PR #176, opened 2021-06-25
Documentation refactor to one method per page

akaszynski:
This PR implements a class template to list class methods as a table rather than outputting them on a single page.  This approach has the advantage of making it easier to navigate methods by viewing class summaries and reduce the total page size for each class.  This matches `numpy`, `scipy`, and `pandas` documentation styles.

The caveat to this is increased build time as one page needs to be generated for each method.


akaszynski:
@maxcapodi78 this is ready for review, and the development documentation can be seen at https://dev.aedtdocs.pyansys.com/API/ (or at least until it's overwritten with the next PR merge).

I worked around the long build time by creating a CRON job that runs nightly at 0400 UTC.  This way we can get fresh daily doc builds, but we don't have to wait an hour on every PR.

---

BTW, I think we're looking at 93 public classes and over 3600 public methods and attributes.  No wonder a single page per method takes forever to build!

akaszynski:
@maxcapodi78, reminder to uncomment L63 on `conf.py`

pyaedt/modeler/Primitives.py
```diff
@@ -2288,15 +2288,14 @@ def get_edges_for_circuit_port(self, face_id, XY_plane=True, YZ_plane=True, XZ_p
...
-        accepts one or more sheet objects.
+        This method accepts a face ID in the input, while the `get_edges_for_circuit_port_from_port`
+        method accepts one or more sheet objects.
                 
         Parameters
         ----------
         face_id :
             ID of the face.
-        XY_plane : bool, optional
+         XY_plane : bool, optional
```
akaszynski: Fix indentation here.

pyaedt/q3d.py
```diff
@@ -114,7 +114,7 @@ class Q3d(QExtractor, object):
         Name of the setup to use as the nominal. The default is
         ``None``. If ``None``, the active setup is used or nothing is
         used.
-    specified_version: str, optional
+        specified_version: str, optional
```
akaszynski: Fix indentation here.

pyaedt/modeler/Primitives.py
```diff
@@ -2153,7 +2153,7 @@ def get_edges_for_circuit_port_from_sheet(self, sheet, XY_plane=True, YZ_plane=T
         the IDs of these lines.
         
         This method accepts a one or more sheet objects as input, while 
-         method :func:`Primitives.get_edges_for_circuit_port` accepts a face ID.
+        the method `get_edges_for_circuit_port` accepts a face ID.
```
akaszynski: Would prefer using the `:func:` directive for linking methods.

akaszynski: Fixed in b0cd37663872337d80d574b514dfdec680490d40.

pyaedt/q3d.py
akaszynski: Fixed in b0cd37663872337d80d574b514dfdec680490d40.

pyaedt/modeler/Primitives.py
akaszynski: Fixed in b0cd37663872337d80d574b514dfdec680490d40.

maxcapodi78 (approved):

PR #189, opened 2021-06-30
Material lib refactoring

maxcapodi78:


_unittest/test_Materials.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def test_duplicate_material(self):
+        assert self.aedtapp.materials.duplicate_material("new_copper2", "copper3")
+        assert not self.aedtapp.materials.duplicate_material("new_copper3", "copper3")
+
+    def test_delete_material(self):
+        assert self.aedtapp.materials.remove_material("copper3")
+        assert not self.aedtapp.materials.remove_material("copper4")
+
+    def test_surface_material(self):
```
akaszynski, line 87: Thanks for moving away from numbered tests.

pyaedt/application/Analysis3D.py
```diff
@@ -309,7 +309,7 @@ def get_all_conductors_names(self):
         -------
 
         """
-        cond = self.materials.GetConductors()
+        cond = self.materials.conductors
```
akaszynski, line 312: Looks much cleaner to have properties.

pyaedt/examples/downloads.py
```diff
@@ -15,7 +15,7 @@ def delete_downloads():
 
 def _get_file_url(directory, filename):
     if not filename:
-        return EXAMPLE_REPO + '/'.join([directory])
+        return EXAMPcoordinateLE_REPO + '/'.join([directory])
```
akaszynski, line 18: We should consider using `os.path.join` in the future.  This works, but it tends to be brittle.

pyaedt/modules/Material.py
```diff
@@ -26,143 +27,111 @@ class MatProperties(object):
...
-    fullname =     ['Permittivity', 'Permeability', 'Conductivity', 'Dielectric Loss Tangent', 'Magnetic Loss Tangent', 'Thermal Conductivity', 'Mass Density', 'Specific Heat', 'Thermal Expansion Coefficient', 'Young\'s Modulus', 'Poisson\'s Ratio', 'Emissivity', 'Diffusivity', 'Molecular Mmass', 'Viscosity', 'Hysteresis application Loss', 'Eddy-Current application Loss', 'Excess application Loss']
-    catname =      ['Permittivity', 'Permeability', 'Conductivity', 'Dielectric Loss Tangent', 'Magnetic Loss Tangent', 'Thermal Conductivity', 'Mass Density', 'Specific Heat', 'Thermal Expansion Coefficient', 'Elasticity',       'Elasticity',       'Emissivity', 'Fluid Properties', 'Fluid Properties', 'Fluid Properties', 'Hysteresis application Loss', 'Eddy-Current application Loss', 'Excess application Loss']
+    aedtname =     ['permittivity', 'permeability', 'conductivity', 'dielectric_loss_tangent', 'magnetic_loss_tangent', 'thermal_conductivity', 'mass_density', 'specific_heat', 'thermal_expansion_coefficient', 'youngs_modulus',   'poissons_ratio',   'diffusivity', 'molecular_mass', 'viscosity', 'core_loss_kh', 'core_loss_kc', 'core_loss_ke']
```
akaszynski, line 30: Consider placing keys like these at the top of the module.  Also, these should be a dictionary with `aedtname` being the key name and each key containing its `defaultvalue`, `defaultunit`, `diel_order`, etc. as a tuple.

pyaedt/modules/Material.py
```diff
@@ -26,143 +27,111 @@ class MatProperties(object):
...
 
         """
-        if fullname:
-            return cls.defaultunit[cls.fullname.index(fullname)]
-        elif catname:
-            return cls.defaultunit[cls.catname.index(catname)]
-        elif aedtname:
+        if aedtname:
             return cls.defaultunit[cls.aedtname.index(aedtname)]
         else:
```
akaszynski, line 53: Unnecessary `else` here.

pyaedt/modules/Material.py
```diff
@@ -26,143 +27,111 @@ class MatProperties(object):
...
+            defaultunit
 
         """
-        if fullname:
-            return cls.aedtname[cls.fullname.index(fullname)]
-        elif catname:
-            return cls.aedtname[cls.catname.index(catname)]
+        if aedtname:
+            return cls.defaultvalue[cls.aedtname.index(aedtname)]
         else:
```
akaszynski, line 74: Unnecessary else here.

pyaedt/modules/Material.py
```diff
@@ -26,143 +27,111 @@ class MatProperties(object):
...
-            raise TypeError("get_aedtname: either fullname or catname MUST be defined")
+            raise TypeError("get_defaultunit: either fullname or catname MUST be defined")
```
akaszynski, line 75: Let's work on not capitalizing words to tell the user something.  None of the other python libraries out there (at least not the big ones) do this, and it seems bad form.  Generally the blood red text is enough to get the user's attention.

pyaedt/modules/Material.py
```diff
@@ -26,143 +27,111 @@ class MatProperties(object):
...
-        elif aedtname:
-            return cls.catname[cls.aedtname.index(aedtname)]
-        else:
-            raise TypeError("get_aedtname: either fullname or aedtname MUST be defined")
+    """
+    aedtname =     ['surface_emissivity', 'surface_roughness', 'surface_diffuse_absorptance', 'surface_incident_absorptance']
+    defaultvalue = [1.0,             0,            0.4,              0.4]
+    defaultunit  = [None,            '[m]',          None, None]
 
     @classmethod
```
akaszynski, line 93: We've got to get in the habit of having `__init__` be the first method.

pyaedt/modules/Material.py
```diff
@@ -26,143 +27,111 @@ class MatProperties(object):
...
+    internalame=named used in script
+    xmlname=named used in XML syntax
 
-        Parameters
-        ----------
-        fullname :
-            optiona full name of the property (Default value = None)
-        aedtname :
-            optional aedtname (Default value = None)
+    Parameters
```
akaszynski, line 82: We need to document parameters and show an example in class docstrings.

pyaedt/modules/Material.py
```diff
@@ -213,52 +211,274 @@ def set_type(self, type):
...
+                if i >= len(self._property_value):
+                    self._property_value.append(BasicValue())
+
+                self._property_value[i].value = el
+                i += 1
+        else:
+            self._property_value[0].value = val
 
+    @property
+    def unit(self):
```
akaszynski, line 245: We've got to get in the habit of providing docstrings for every method (and property).  Doc team has a hard time as is, and it's generally unclear (even for me) what these methods are doing.

pyaedt/modules/Material.py
```diff
@@ -213,52 +211,274 @@ def set_type(self, type):
...
         -------
+        bool
+
+        Examples
+        --------
+
+        >>> from pyaedt import Hfss
+        >>> hfss = Hfss(specified_version="2021.1")
+        >>> mat1 = hfss.materials.add_material("new_copper2")
+        >>> mat1.aadd_thermal_modifier_free_form("if(Temp > 1000cel, 1, if(Temp < -273.15cel, 1, 1))")
```
akaszynski, line 330: Thanks for adding the example here.  If this method returns anything, even a simple `True`, `False`, we need to add the return on the next line (without the `>>> `).

pyaedt/modules/Material.py
```diff
@@ -213,52 +211,274 @@ def set_type(self, type):
...
+    def add_thermal_modifier_closed_form(self, tref=22, c1=0.0001, c2=1e-6, tl=-273.15, tu=1000, units="cel",
+                                         auto_calc=True, tml=1000, tmu=1000, index=0):
         """
-        (Permittivity, Permeability, Conductivity, DielectricLossTangent, MagneticLossTangent,
-         ThermalConductivity, MassDensity, SpecificHeat, ThermalExpCoeff,
-         YoungsModulus, PoissonsRatio, Emissivity, Diffusivity, MolecularMass, Viscosity) = (
-            'permittivity', 'permeability', 'conductivity', 'dielectric_loss_tangent', 'magnetic_loss_tangent',
-            'thermal_conductivity', 'mass_density', 'specific_heat', 'thermal_expansion_coefficient',
-            'youngs_modulus', 'poissons_ratio', 'emissivity', 'diffusivity', 'molecular_mass', 'viscosity')
+        Add a Thermal modifier to a material Property using Closed Form
```
akaszynski, line 366: Short description should be on the line with `"""`.  Should read:
```
    """Add a Thermal modifier to a material Property using Closed Form.
```

pyaedt/modules/Material.py
```diff
@@ -213,52 +211,274 @@ def set_type(self, type):
...
+
+        Parameters
+        ----------
+        tref : float
+            Reference Temperature
+        c1 : float
```
akaszynski, line 372: We've got to have descriptions for these.

pyaedt/modules/Material.py
```diff
@@ -272,579 +492,507 @@ def desktop(self):
...
-            
+    @aedt_exception_handler
+    def _get_args(self, props=None):
+        if not props:
+            props = self._props
+        arg = ["NAME:" + self.name]
+        dict2arg(props, arg)
+        return arg
+
+    def _update_props(self, propname, provpavlue, update_aedt=True):
```
akaszynski, line 532: Protected methods don't need well written docstrings, but we need at least a one liner.

pyaedt/modules/Material.py
```diff
@@ -272,579 +492,507 @@ def desktop(self):
...
+                                                            MatProperties.get_defaultvalue(aedtname=property), None)
+        pass
 
-        Parameters
-        ----------
-        fullname :
-            
+    @property
+    def permittivity(self):
+        """
```
akaszynski, line 589: Docstrings for properties must have a one line short description.  Copy/paste what we have in `Returns` should be sufficient.  Reasoning is when we finally merge #176, we don't want to have empty descriptions in our tables.  It's not too obvious that our descriptions are empty, but it will be quite obvious when each property/method stands alone.

PR #191, opened 2021-06-30
Update EDB_Data.py

PipKat:


pyaedt/edb_core/EDB_Data.py
```diff
@@ -726,8 +726,7 @@ def __init__(self, edb_padstack, layer_name, pad_type, parent):
 
     @property
     def geometry_type(self):
-        """Retrieve the geometry type.
-    
+        """
```
akaszynski, line 729: We have to have shot descriptions for these methods.  ``Geometry type.`` is sufficient.

akaszynski: Just saw you added this back in.

PipKat: @akaszynski  So do this for all with the @property? I assume that I will continue to see syntax errors for subsequent entries if I don't. My notes say that @properties indicates that there are no parameters for this, but yet some of the @properties do have them?!

pyaedt/edb_core/EDB_Data.py
```diff
@@ -671,36 +726,41 @@ def __init__(self, edb_padstack, layer_name, pad_type, parent):
 
     @property
     def geometry_type(self):
-        """
-
+        """Retrieve the geometry type.
```
PipKat, line 729:
```suggestion
        """Geometry type.
```

pyaedt/edb_core/EDB_Data.py
PipKat: @akaszynski I shorted to Geometry type--but it still failed at the same location. Other earlier @property items have short descriptions that are just as long as this one was with the "Retrieve" portion. I can't figure out what is wrong!

pyaedt/edb_core/EDB_Data.py
```diff
@@ -635,13 +688,15 @@ def remove_layer(self, layername):
...
         if not lcNew.AddLayers(newLayers) or not self._parent.active_layout.SetLayerCollection(lcNew):
-            self.messenger.add_error_message('Failed to set new layers when updating stackup info')
+            self.messenger.add_error_message('Failed to set new layers when updating the stackup information.')
             return False
         self._update_edb_objects()
         return True
 
 
 class EDBPadProperties(object):
+    """EDBPadProperties class.""
```
akaszynski, line 698:
```suggestion
    """EDBPadProperties class."""
```

akaszynski: @PipKat, this was causing the syntax error.  Funny how a single character can ruin your whole day.

One takeaway from this is to look above for what might be causing the issue.  It might be a change you made several lines up rather than where the compiler reports the syntax error.

pyaedt/edb_core/EDB_Data.py
```diff
@@ -61,68 +62,68 @@ def messenger(self):
...
         int
-            Layer Type
+            Type of the layer.
         """
         if not self._layer_type:
             self._layer_type = self._layer.GetLayerType()
         return self._layer_type
 
     @layer_type.setter
     def layer_type(self, value):
```
akaszynski, line 103: Setters don't have to be documented.  All documentation for a setter should be placed where it's defined in `@property`.  Reason for this is when our documentation is built, it's only going to print out what's listed in `@property`, so anything we write in the setter is effectively invisible in our generated docs.

`@layer_type.setter` is pretty much self documenting, so we can just skip the docstrings for all setters.

One note.  When documenting properties that also have setters, you'd document it with:

```
Retrieve or update the layer type.
```
Basically, we're exposing a `get/set` to the user with these methods, rather than just the `get` method.

akaszynski (changes requested):
Please see my notes regarding docstrings for setter methods.

pyaedt/edb_core/EDB_Data.py
akaszynski: Figured it out, just a missing `"` in the class docstring above this.  I've committed and pushed the change to get the CI rolling.

As for `@properties` with parameters, could you please point out where that occurs?  I'd like to take a look at that.

pyaedt/edb_core/EDB_Data.py
```diff
@@ -710,74 +770,113 @@ def parameters(self, propertylist):
...
         """
         self._update_pad_parameters_parameters(offsety = offset_value)
-
+        
+        Returns
+        -------
+        bool
+            ``True`` when successful, ``False`` when failed.
+        """
     @property
```
akaszynski, line 826:
```suggestion
        """
        self._update_pad_parameters_parameters(offsety=offset_value)

    @property
```

pyaedt/edb_core/EDB_Data.py
```diff
@@ -710,74 +770,115 @@ def parameters(self, propertylist):
...
-        """
+        """Update the offset for the Y axis.
 
         Parameters
         ----------
         offset_value : float, str
-            New Offset Y Value
+           Offset value for the Y axis.
         """
         self._update_pad_parameters_parameters(offsety = offset_value)
```
akaszynski, line 819:
```suggestion
```

pyaedt/edb_core/EDB_Data.py
```diff
@@ -942,111 +1065,141 @@ def hole_properties(self, propertylist):
...
+        Parameters
+        ----------
+        ratio : float
+            Ratio value for the hole plating.
+        
+        Returns
+        -------
+        bool
+            ``True`` when successful, ``False`` when failed.
+        """"
```
akaszynski, line 1189:
```suggestion
        """
```

pyaedt/edb_core/EDB_Data.py
```diff
@@ -1093,6 +1252,7 @@ def material(self, materialname):
 
 
 class EDBPinInstances(object):
+    """EDBPinInstances class.
```
akaszynski, line 1255:
```suggestion
    """EDBPinInstances class."""
```

pyaedt/edb_core/EDB_Data.py
```diff
@@ -1115,23 +1275,23 @@ def net(self):
...
     @property
     def position(self):
-        """
+        """Retrieve the pin position.
 
         Returns
         -------
         list
-            [X, Y] position
+            List of the pin position in the format [X, Y].
```
akaszynski, line 1294:
```suggestion
            List of the pin position in the format ``[x, y]``.
```

pyaedt/edb_core/EDB_Data.py
```diff
@@ -1157,52 +1318,56 @@ def rotation(self):
...
 
         Returns
         -------
         int
+            Top/bottom association of the placement layer.
             0 Top associated.
             1 No association.
             2 Bottom associated.
             4 Number of top/bottom association type.
             -1 Undefined.
```
akaszynski, line 1364:
```suggestion

            * 0 - Top associated.
            * 1 - No association.
            * 2 - Bottom associated.
            * 4 - Number of top/bottom association type.
            * -1 - Undefined.
```

When creating bulleted lists in rst, make sure there's a space between the line above it or else it won't render correctly.  Additionally, each line needs to have a `*` before it.

pyaedt/edb_core/EDB_Data.py
```diff
@@ -1069,22 +1222,28 @@ def hole_finished_size(self):
 
     @property
     def material(self):
-        """
+        """Retrieve the hole material
```
PipKat, line 1225:
```suggestion
        """Retrieve or update the hole material.
```

pyaedt/edb_core/EDB_Data.py
```diff
@@ -1069,22 +1222,28 @@ def hole_finished_size(self):
...
         str
-            Hole Material
+            Material of the hole.
         """
         return self.edb_padstack.GetData().GetMaterial()
 
     @material.setter
     def material(self, materialname):
-        """
+        """Update the material.
```
PipKat, line 1236:
```suggestion
        """ """
```

pyaedt/edb_core/EDB_Data.py
```diff
@@ -132,25 +122,20 @@ def material_name(self):
 
     @material_name.setter
     def material_name(self, value):
-        """Set the material name
+        """ """
```
akaszynski, line 125: We can completely remove the docstring here:
```suggestion
```

pyaedt/edb_core/EDB_Data.py
```diff
@@ -132,25 +122,20 @@ def material_name(self):
...
-            Material name
-        """
         #self.stackup_methods.EditLayerName(self.builder, self._name, value)
         self._material_name = value
         self.update_layers()
 
     @property
     def thickness_value(self):
-        """
+        """Retrieveor update the thickness value.
```
akaszynski, line 133:
```suggestion
        """Retrieve or update the thickness value.
```

pyaedt/edb_core/EDB_Data.py
```diff
@@ -160,25 +145,20 @@ def thickness_value(self):
 
     @thickness_value.setter
     def thickness_value(self, value):
-        """
+        """ """
```
akaszynski, line 149:
```suggestion
```

pyaedt/edb_core/EDB_Data.py
```diff
@@ -190,25 +170,26 @@ def filling_material_name(self):
...
     def filling_material_name(self, value):
-        """Set the filling material
-
-        Parameters
-        ----------
-        value : str
-            Filling material
-        """
+        """ """
+
```
akaszynski, line 174: Same with the rest of these, just remove the docstrings on all setters.

akaszynski (approved):
Thanks for making the changes.  Merging as status checks are green.

PR #193, opened 2021-06-30
Update general.py

PipKat:


pyaedt/edb_core/general.py
```diff
@@ -55,32 +53,35 @@ def convert_netdict_to_pydict(dict):
 
 @aedt_exception_handler
 def convert_pydict_to_netdict(dict):
-    """
+    """Convert a Python dictionarty to a Net dictionary.
 
     Parameters
     ----------
-    dict :
+    dict : dic
```
akaszynski, line 60:
```suggestion
    dict : dict
```
Problem with these methods is that we're using the built in `dict` as a parameter name.  This should be fixed in a separate PR.  For now, we're just going to have to deal with `dict : dict`.

pyaedt/edb_core/general.py
```diff
@@ -55,32 +53,35 @@ def convert_netdict_to_pydict(dict):
...
     ----------
-    pylist :
+    pylist : list
+        Python list to convert.
         
-
     Returns
     -------
-
+    List
```
akaszynski, line 83:
```suggestion
    list
```

akaszynski (approved):
Approved with suggestions.

PR #196, opened 2021-06-30
Update nets.py

PipKat:


pyaedt/edb_core/nets.py
```diff
@@ -92,14 +110,13 @@ def is_power_gound_net(self, netname_list):
...
-            ``True`` if one of the net names is ``"power"`` or ``"ground"``.
-
+            ``True`` when one of the net names is ``"power"`` or ``"ground"``, ``False`` otherwise.
         """
         if self.builder:
             return self.nets_methods.IsPowerGroundNetInList(self.builder, netname_list)
 
     def get_dcconnected_net_list(self, ground_nets=["GND"]):
-        """List the nets connected to DC through inductors.
+        """Retrive the nets connected to DC through inductors.
```
PipKat, line 119:
```suggestion
        """Retrieve the nets connected to DC through inductors.
```

akaszynski (approved):
Tentatively approved, but please have those methods with the @property decorator changed to nouns rather than verbs.

PR #211, opened 2021-07-02
Update rmxprt.py

PipKat:
Add missing parameter descriptions and change back ticks.


akaszynski (approved):
Approved with no suggestions.

PR #217, opened 2021-07-05
Edb layout

maxcapodi78:
Make hidden all properties in Edb Core that have not to be exposed.
Bug fix on Edb Layers which could probably affect IronPython


pyaedt/edb_core/siwave.py
```diff
@@ -109,7 +111,7 @@ def do_pin_grouping(self,value):
 
 class CircuitPort(Source):
     def __init(self):
-        super().__init__()
+        Source.__init__(self)
```
Alberto-DM, line 114:
```suggestion
        super(CircuitPort, self).__init__()
```
This is compatible with Python 2.x and more indicated if Source class inherits from ``object`` like  ``class Source(object):``.
Use ``Source.__init__(self)`` only when Source is defined as classic class ``class Source():``

pyaedt/edb_core/siwave.py
```diff
@@ -125,9 +127,10 @@ def impedance(self,value):
     def get_type(self):
         return self.type
 
+
 class VoltageSource(Source):
     def __init__(self):
-        super().__init__()
+        Source.__init__(self)
```
Alberto-DM, line 133: See comment before.

pyaedt/edb_core/siwave.py
```diff
@@ -161,9 +164,10 @@ def impedance(self,value):
     def source_type(self):
         return self.source_type
 
+
 class CurrentSource(Source):
     def __init__(self):
-        super().__init__()
+        Source.__init__(self)
```
Alberto-DM, line 170: See comment before.

pyaedt/edb_core/siwave.py
```diff
@@ -197,9 +201,10 @@ def impedance(self,value):
     def source_type(self):
         return self.source_type
 
+
 class ResistorSource(Source):
     def __init__(self):
-        super().__init__()
+        Source.__init__(self)
```
Alberto-DM, line 207: See comment before.

Alberto-DM (approved):
Just one comment on object initialization. Code is ok.

PR #219, opened 2021-07-06
Added Student Version support

maxcapodi78:
Bug fix in Edb.py Database Opening


akaszynski:
We need to document the `student_version` parameter before merging this in the applicable classes.

pyaedt/generic/general_methods.py
```diff
@@ -201,6 +201,55 @@ def env_value(input_version):
...
+            version -= 1
+        else:
+            release -= 2
+    v_key = "ANSYSEMSV_ROOT{0}{1}".format(version, release)
+    return os.getenv(v_key)
+
+
+@aedt_exception_handler
+def env_value_student(input_version):
+    """
```
akaszynski, line 231: Would be nice to have this documented (at least a provide a summary statement).

pyaedt/generic/general_methods.py
```diff
@@ -201,6 +201,55 @@ def env_value(input_version):
     v_key = "ANSYSEM_ROOT{0}{1}".format(version, release)
     return v_key
 
+
+@aedt_exception_handler
+def env_path_student(input_version):
+    """
```
akaszynski, line 207: Would be nice to have this documented (at least the short summary).

pyaedt/edb.py
```diff
@@ -177,31 +181,23 @@ def open_edb(self, init_dlls=False):
...
-                self.builder = self.layout_methods.GetBuilder(self._db, self._active_cell)
-            else:
-                print("opening from EDBLib")
-                self.builder = self.layout_methods.OpenEdbStandAlone(self.edbpath, self.edbversion)
-                self._db = self.builder.EdbHandler.dB
-                self._active_cell = self.builder.EdbHandler.cell
+        print(self.edbpath)
+        print(self.edbversion)
+        print(_ironpython)
+        print(dir())
```
akaszynski, line 187: We should avoid using print statements and instead use the `logging` module.  Many users will want to suppress output to stdout or will want to have a different level of logging depending on their workflow.  Using the `logging` module lets them do that.

akaszynski (approved):
Approved with minor suggestions.  `print` --> `logging` can be saved for a different PR since it's out of scope of this PR, but we really should implement that.

pyaedt/generic/general_methods.py
maxcapodi78: Done

pyaedt/edb.py
maxcapodi78: Done. I've replaced with messenger calls

PR #222, opened 2021-07-06
Update maxwell.py

PipKat:


pyaedt/maxwell.py
```diff
@@ -889,7 +887,7 @@ def get_model_depth(self):
                         a = self.variable_manager[value_str].value
                     return a
                 else:
-                    raise Exception('Design data not found by get_model_depth function - inconsistency !!!')
+                    raise Exception('Design data is not found by the get_model_depth function. Find and fix the inconsistency.')
```
akaszynski, line 890:
```suggestion
                    raise RuntimeError('Design data is not found by the get_model_depth function. Find and fix the inconsistency.')
```

akaszynski (approved):
One minor non-doc suggestion.

PR #228, opened 2021-07-07
Added optional argument props to method add_material

Alberto-DM:
- Added optional argument props to method add_material.
- Made Material __getitem__ case insensitive.
- Renamed import_materials_to_file into import_materials_from_file


pyaedt/modules/MaterialLib.py
```diff
@@ -156,6 +157,9 @@ def add_material(self, materialname):
         materialname : str
             material name
 
+        props : dict
+            Optional material property dictionary. Default = `None`.
```
akaszynski, line 161:
```suggestion
            Optional material property dictionary. Default is `None`.
```

akaszynski (approved):
Trivial suggestion, otherwise LGTM.

pyaedt/modules/MaterialLib.py
Alberto-DM: Done.

maxcapodi78 (approved):

PR #229, opened 2021-07-07
Sherlock example fix

maxcapodi78:


akaszynski:
@maxcapodi78, can you create a new branch from `main`, could you run the following?

```
git checkout main
git pull
git checkout -b <mynewbranch>
```

I'm seeing a ton of commits prior to your new commits, and it looks like the branch you "branch" from isn't fresh.

maxcapodi78:
@akaszynski  could it be that, instead of creating the new branch from main, I created from release branch right after the merge from main?

Alberto-DM (approved):
Looks all good to me.

akaszynski:
> @akaszynski could it be that, instead of creating the new branch from main, I created from release branch right after the merge from main?

Perhaps,  I'd generate feature branches always from main rather than from the release branch.

_unittest/test_01EDB.py
```diff
@@ -410,6 +414,7 @@ def test_export_to_hfss(self):
         out= edb.export_hfss(scratch_path)
         assert os.path.exists(out)
 
+    @pytest.mark.skipif(config["build_machine"]==True, reason="Not running in non-graphical mode")
```
akaszynski, line 417:
```suggestion
    @pytest.mark.skipif(config["build_machine"], reason="Not running in non-graphical mode")
```

_unittest/test_01EDB.py
```diff
@@ -402,6 +404,8 @@ def test_create_edb(self):
         assert edb
         assert edb.active_layout
         edb.close_edb()
+
+    @pytest.mark.skipif(config["build_machine"]==True, reason="Not running in non-graphical mode")
```
akaszynski, line 408:
```suggestion
    @pytest.mark.skipif(config["build_machine"], reason="Not running in non-graphical mode")
```

examples/00-basic-examples/EDB_Create_VIA.py
```diff
@@ -15,7 +15,10 @@
 import os
 from pyaedt import Edb
 from pyaedt.generic.general_methods import generate_unique_name
-aedb_path=os.path.join(r"C:\Temp", generate_unique_name("Edb_custom")+".aedb")
+if os.name == "posix":
+    aedb_path = os.path.join("/tmp", generate_unique_name("Edb_custom") + ".aedb")
+else:
+    aedb_path=os.path.join(r"C:\Temp", generate_unique_name("Edb_custom")+".aedb")
```
akaszynski, line 21: Not necessary to change it now, but in the future you can use `/` instead of `\` to avoid having to use raw strings.

akaszynski (approved):
LGTM.

PR #234, opened 2021-07-07
Update Design.py

PipKat:
Minor updates--typos--missing Parameters section heading, and so on.

### DIFF
--- a/pyaedt/application/Design.py
+++ b/pyaedt/application/Design.py
@@ -427,7 +427,7 @@ class Design(object):
         or AEDTZ archive to open. The default is ``None``, in which
         case an attempt is made to get an active project. If no 
         projects are present, an empty project is created.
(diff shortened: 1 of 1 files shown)


pyaedt/application/Design.py
```diff
@@ -439,13 +439,16 @@ class Design(object):
         the active version or latest installed version is used.
     NG : bool, optional
         Whether to run AEDT in the non-graphical mode. The default 
-        is ``False``, which launches AEDT in the graphical mode.  
+        is ``False``, in which case AEDT launces in the graphical mode.
```
akaszynski, line 442:
```suggestion
        is ``False``, in which case AEDT launches in the graphical mode.  
```

akaszynski (approved):
One spelling fix, otherwise LGTM.

PR #239, opened 2021-07-08
Update AnalysisCore.rst

PipKat:
All RST files have  now been updated.

In this PR, I'm going to update all the PRs to use a single back tick since the double one isn't displaying correctly in the rendered doc. I'll add a comment when they are all done.


akaszynski (approved):
Good catch regarding the single vs. double tick in the caption.

PR #243, opened 2021-07-09
Updated pyaedt import to avoid error in IPY

maxcapodi78:


akaszynski:
This seems like a bad practice to me.  We should identify which module is causing the circular import and patch that (or change the order of imports) rather than importing everything twice.

maxcapodi78:
@akaszynski it's a strange error. it only fails the first time you import a module, randomly when you run the script within AEDT. if you use Ironpython from outside AEDT it works fine.

maxcapodi78:
@akaszynski in addition to this, I'm not able to reproduce it on my machine. Only happens randomly on some colleague machine

pyaedt/__init__.py
```diff
@@ -1,16 +1,31 @@
...
-from .icepak import Icepak
-from .edb import Edb
-from .maxwell import Maxwell2d, Maxwell3d
-from .mechanical import Mechanical
-from .rmxprt import Rmxprt
-from .simplorer import Simplorer
-from .desktop import Desktop
-from .emit import Emit
-
+try:
```
akaszynski, line 2:
```suggestion
# Import exception handling here due to:
# https://github.com/pyansys/PyAEDT/pull/243
try:
```

akaszynski: Added a note here so we know why there's this witchcraft here.

akaszynski (approved):
Added a comment so we could track the addition of the exception handling to this PR.

PR #249, opened 2021-07-09
distutils-compatible setup script

isaacansys:
Add a new distutils-compatible setup script to facilitate installation of PyAEDT with the AEDT-provided IronPython installation.  Move common setup data into a shared file so it has a single source location.  In local testing I confirmed that CPython 3.8 and IronPython from AEDT can run the install command of the setup.py and setup-distutils.py scripts, respectively.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+include _setup_common.py
--- a/_setup_common.py
+++ b/_setup_common.py
@@ -1,0 +1,64 @@
+import fnmatch
+import os
+import sys
+
+# required since Python 2.7's glob.iglob does not support recursive keyword
+def recursive_glob(startpath, filepattern):
+    """Return a list of files matching a pattern, searching recursively from a start path.
+
+    Keyword Arguments:
+    startpath -- starting path (directory)
+    filepattern -- fnmatch-style filename pattern
+    """
+    return [
+        os.path.join(dirpath, filename)
+            for dirpath, _, filenames in os.walk(startpath)
+            for filename in filenames if fnmatch.fnmatch(filename, filepattern)
+    ]
+
+open3 = open
+if sys.version_info < (3, 0):
+    import io
+    open3 = io.open
+
+# loosely from https://packaging.python.org/guides/single-sourcing-package-version/
+HERE = os.path.abspath(os.path.dirname(__file__))
+
+name='pyaedt'
+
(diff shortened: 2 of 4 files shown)


akaszynski:
Can't confirm this works on IronPython, but CI for CPython reports the following error:
```
    ERROR: Command errored out with exit status 1:
     command: /usr/bin/python -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-req-build-bz16r2y4/setup.py'"'"'; __file__='"'"'/tmp/pip-req-build-bz16r2y4/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-e34u70re
         cwd: /tmp/pip-req-build-bz16r2y4/
    Complete output (5 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-req-build-bz16r2y4/setup.py", line 4, in <module>
        from _setup_common import name, version, author, maintainer, maintainer_email, description, long_description, packages, data_files, license, classifiers
    ModuleNotFoundError: No module named '_setup_common'
```

Reproduce this error within CPython with:
```
python setup.py sdist
pip install dist/pyaedt-0.3.dev0.tar.gz
```

akaszynski (changes requested):
Please fix the CPython install from the source distribution.

akaszynski:
Turns out the trick is to include `_setup_common.py` in `MANIFEST.in`.  Normally, only *.py source files within package directories are picked by setup.py when building `sdist`.  This way, we force it to be included with the source distribution.

akaszynski (approved):
@maxcapodi78, I'd like you to validate this prior to merge or to review the PR as well, but it works correctly now for CPython with the `manifest.in`.

isaacansys:
@akaszynski thanks for fixing this up.  @maxcapodi78 and I had discussed the need to add setup-distutils to the data_files but I missed adding the base _setup_common.py

maxcapodi78:
I agree. This is an important point to get Linux users a smart way to install and use pyaedt.

PR #254, opened 2021-07-09
Update MessageManager.py

PipKat:


pyaedt/application/MessageManager.py
```diff
@@ -275,24 +304,28 @@ def clear_messages(self, proj_name=None, des_name=None, level=2):
...
 
-            * ``0`` : Clear all info messages
-            * ``1`` : Clear all info and warning messages
-            * ``2`` : Clear all info, warning and error messages
-            * ``3`` : Clear all messages included info, warning,
-              error, and fatal-error.
+            * ``0`` : Clear all info messages.
+            * ``1`` : Clear all info and warning messages.
+            * ``2`` : Clear all info, warning, and error messages.
+            * ``3`` : Clear all messages, which includ info, warning,
```
PipKat, line 321:
```suggestion
            * ``3`` : Clear all messages, which include info, warning,
```

pyaedt/application/MessageManager.py
```diff
@@ -33,18 +19,19 @@ class Msg():
 
 
 class MessageList:
-    """Data structure to collect and return message information from the AEDT Message Manager for a specified
-       project name and design name
+    """ MessageList class.
```
akaszynski, line 22:
```suggestion
    """MessageList class.
```

pyaedt/application/MessageManager.py
```diff
@@ -117,23 +127,33 @@ def __init__(self, parent=None, loadondesktop=True, loadonlog=True):
...
     @aedt_exception_handler
     def get_messages(self, project_name, design_name):
-        """Return the AEDT Message Manager Contents for a specified project and design
-
-        If the design- or project names are invalid, they will be ignored
+        """Retrieve the Message Manager content for a specified project and design.
+        
+        Parameters
+        ----------
+        project_name: str,
```
akaszynski, line 145:
```suggestion
        project_name : str
```

pyaedt/application/MessageManager.py
```diff
@@ -117,23 +127,33 @@ def __init__(self, parent=None, loadondesktop=True, loadonlog=True):
...
+            Name of the project.
+        design_name, str
```
akaszynski, line 147:
```suggestion
        design_name : str
```

pyaedt/application/MessageManager.py
```diff
@@ -117,23 +127,33 @@ def __init__(self, parent=None, loadondesktop=True, loadonlog=True):
...
+            Name of the design within the specified project.
+        
+        ..note::
+          If the design or project names are invalid, they are ignored.
```
akaszynski, line 151:
```suggestion
        .. note::
           If the design or project names are invalid, they are ignored.
```

akaszynski: Always a space after `..` for note and warning directives.  The standard indentation here is three spaces (rst standard).

pyaedt/application/MessageManager.py
```diff
@@ -208,31 +235,33 @@ def add_info_message(self, message_text, level=None):
     @aedt_exception_handler
     def add_debug_message(self, message_text, level=None):
         """
```
akaszynski, line 237: Deprecated in favor of `add_info_message`.

pyaedt/application/MessageManager.py
```diff
@@ -208,31 +235,33 @@ def add_info_message(self, message_text, level=None):
...
-        .. deprecated use add_info_message
+        .. Deprecated. Use the method `add_info_message`.
```
akaszynski, line 238:
```suggestion
        .. deprecated:: 0.2.0
           Use the method :func:`MessageList.add_info_message`.
```
Here we use two sphinx directives.  The first is the `.. depreciated` directive (see [directive-deprecated](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated)).  It shows up well in the docs.  Secondly, we provide a link to the actual method with the `:func:` directive.  Note how we have to provide the class and then the method name when providing this link.

pyaedt/application/MessageManager.py
```diff
@@ -325,7 +358,15 @@ def _project_name(self):
...
 
 class EDBMessageManager(object):
-    """Class that manage EDB Messaging to the logger file"""
+    """EDBMessageManager class.
+    
+    This class provides all functionalities for managing EDB messaging to the log file.
+    
+    Parameters
+    ----------
+    project_dir :
```
akaszynski, line 367:
```suggestion
    project_dir : str
```

pyaedt/application/MessageManager.py
```diff
@@ -343,44 +384,46 @@ def __init__(self, project_dir=None):
...
+        str
+            Message text added to the log file.
 
         """
 
         self.logger.error(message_text)
 
     def add_warning_message(self, message_text):
-        """Add a "Warning" message to the logger
+        """Add a "Warning" message to the log file.
```
akaszynski, line 404: For this and all other mentions, it should actually be "logger".  The logger in this case could be a file, stream handler (standard output), or both.

pyaedt/application/MessageManager.py
```diff
@@ -343,44 +384,46 @@ def __init__(self, project_dir=None):
...
-        Parameters
-        ----------
-        message_text :
-            message to show
+        message_text : str
+            Text to display as the message.
 
         Returns
         -------
+        str
```
akaszynski, line 411: No return here.  See how there's no decorator for the method?  That means that we're using the vanilla implementation of Python methods, meaning if there's no `return`, it's returning nothing (well, it's returning `None`, so that's something but you don't document it).  Please go through and update the other methods to account for this.

akaszynski (approved):
LGTM.

PR #260, opened 2021-07-12
Update components.py

PipKat:
Change to one back tick for classes, put line after each last >>> example line (though this apparently isn't necessary).

@akaszynski 
So many classes have a "parent" parameter that isn't yet defined. Is there a generic definition that I can use? Also, main pages under "Application Objects" are not showing.  Only those with subpages have the subpage content. Since I can't see a lot of the rendered output, I'll continue editing files that haven't been changed for quite a while.

### DIFF
--- a/pyaedt/edb_core/components.py
+++ b/pyaedt/edb_core/components.py
@@ -1,4 +1,10 @@
-"""This module contain the ``Components`` class."""
+"""This module contain the `Components` class.
+
+Parameters
+----------
+parent : str
+
+"""
 import re
 import random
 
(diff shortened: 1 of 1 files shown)


akaszynski:
I'll update my docs and finally get the one page method per method I was talking about.  Tough to get that out today with hospital appointments, but this is critical since our docs are effectively broken.

akaszynski (approved):
Looks good to me.  One note, Jes and Maxime will be working on the examples section and we've come to the agreement that one line break prior to the ending `"""` was a good idea.  Saw that you added them here, so thanks!

PR #262, opened 2021-07-12
Fixed bug in Edb methods that use GetProductProperty.

Alberto-DM:
Fixed bug in Edb methods that use GetProductProperty.
GetProductProperty returns a response structured differently in CPython.
Improved unittest.


akaszynski (approved):
LGTM.  Thanks!

maxcapodi78 (approved):

PR #268, opened 2021-07-13
Update PostProcessor.py

PipKat:
Mostly minor edits but this topic uses the term function and call. I don't know what the difference is between a method and a function, so I left the terminology as is.

### DIFF
--- a/pyaedt/modules/PostProcessor.py
+++ b/pyaedt/modules/PostProcessor.py
@@ -1,7 +1,7 @@
 """
 This module contains these classes: `FieldPlot`, `PostProcessor`, and `SolutionData`.
 
-These classes provide all functionalities for creating and editing plots in the 3D tools.
+This module provides all functionalities for creating and editing plots in the 3D tools.
 
 .. note::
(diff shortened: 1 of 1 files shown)


pyaedt/modules/PostProcessor.py
```diff
@@ -476,17 +492,17 @@ def modify_folder(self):
...
     @property
     def plotFolder(self):
-        """ """
+        """Field plot folder."""
         return self.name
 
     @property
     def plotGeomInfo(self):
-        """ """
+        """Plot gemoetry information."""
```
PipKat, line 505:
```suggestion
        """Plot geometry information."""
```

akaszynski:
@PipKat, CI keeps failing on our new public CI/CD on `Machine name: 'pyaedt-ci-1'`.  We're investigating that.  Restarting the CI/CD shifts it to a different machine, so try restarting should you think it's due to the machine and not source changes.

akaszynski (approved):
Good to merge.

PR #273, opened 2021-07-14
Update JobManager.py

PipKat:
I don't see this module anywhere in the API doc. I added an issue about this and line 144:

def update_config_name(self, name, machinename):
"""Update the name of the machine.

I believe that line 144 is supposed to be for updating the name of the configuration file, yet it asks for the name of the machine. There is already a method for updating the machine name.

### DIFF
--- a/pyaedt/application/JobManager.py
+++ b/pyaedt/application/JobManager.py
@@ -1,17 +1,16 @@
 
 def get_hpc_info(filename):
-    """
+    """Retrieve HPC information.
 
     Parameters
     ----------
-    filename :
-        
+    filename : str
+        Name of the flie.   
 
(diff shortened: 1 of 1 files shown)


pyaedt/application/JobManager.py
```diff
@@ -1,17 +1,16 @@
...
 def get_hpc_info(filename):
-    """
+    """Retrieve HPC information.
 
     Parameters
     ----------
-    filename :
-        
+    filename : str
+        Name of the flie.
```
PipKat, line 8:
```suggestion
        Name of the file.   
```

akaszynski (approved):
Approved with no suggestions.

PR #274, opened 2021-07-14
Update SolutionType.py

PipKat:


akaszynski (approved):
Approved with no suggestions.

PR #276, opened 2021-07-14
Update Boundary.py

PipKat:
Minor edits


akaszynski (approved):
Approved with no suggestions.

PR #277, opened 2021-07-14
Update Mesh.py

PipKat:
Minor edits.

### DIFF
--- a/pyaedt/modules/Mesh.py
+++ b/pyaedt/modules/Mesh.py
@@ -29,11 +29,6 @@ class MeshOperation(object):
     
     meshoptpe :
     
-    Returns
-    -------
-    type
-    
-    
     """
     def __init__(self, parent, name, props, meshoptype):
         self._parent = parent
@@ -444,7 +439,7 @@ class Mesh(object):
 
     @aedt_exception_handler
     def delete_mesh_operations(self, mesh_type=None):
-        """Remove a mesh operations from a design. 
+        """Remove mesh operations from a design. 
 
         Parameters
         ----------
@@ -810,7 +805,7 @@ class Mesh(object):
 
         Returns
         -------
-        type
+        :class:`MeshOperation`
             Mesh operation object.
             
         """
--- a/pyaedt/modules/Mesh.py
+++ b/pyaedt/modules/Mesh.py
@@ -618,7 +618,7 @@ class Mesh(object):
         names : list
             List of objects or faces.
         enable : bool, optional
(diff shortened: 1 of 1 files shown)


pyaedt/modules/Mesh.py
```diff
@@ -390,14 +436,15 @@ def generate_mesh(self, name):
...
+            ``True`` when successful, ``False`` when failed.
+            
         """
         return self.odesign.GenerateMesh(name) == 0
 
 
     @aedt_exception_handler
     def delete_mesh_operations(self, mesh_type=None):
-        """Remove mesh operations from a design. 
+        """Remove a mesh operations from a design.
```
akaszynski, line 447:
```suggestion
        """Remove mesh operations from a design. 
```

Seems like it should be plural.

pyaedt/modules/Mesh.py
```diff
@@ -760,6 +812,7 @@ def assign_density_control(self, names, refine_inside=True, maxelementlength=Non
         -------
         type
```
akaszynski, line 813:
```suggestion
        :class:`MeshOperation`
```
We can use the class directive here, and we don't need the full class name spelled out since the class is defined in this module.  Otherwise we'd have to use 
```
:class:`pyaedt.modules.Mesh.MeshOperation`
```

Please fix this in the other docstrings.  If you're unclear what the return type is, please ping me.

pyaedt/modules/Mesh.py
```diff
@@ -17,7 +17,24 @@
...
+    
+    props :
+    
+    meshoptpe :
+    
+    Returns
+    -------
+    type
+    
+
```
akaszynski, line 36:
```suggestion
```

akaszynski: Class docstrings have no returns.

akaszynski (changes requested):
See comment regarding adding in the `:class:` directive.

pyaedt/modules/Mesh.py
```diff
@@ -575,14 +623,15 @@ def assign_curvilinear_elements(self, names, enable=True, meshop_name=None):
         names : list
             List of objects or faces.
         enable : bool, optional
-            The default is ``True``.
+            Whether  to apply curvilinear elements. The default is ``True``.
```
PipKat, line 621:
```suggestion
            Whether to apply curvilinear elements. The default is ``True``.
```

akaszynski (approved):
Thanks for adding the directives.

PR #283, opened 2021-07-14
Update modeler_constants.py

PipKat:
Just added the class name to the class docstring. Not useful but I'm assuming something there is better than nothing.


akaszynski (approved):
Agreed, better than nothing.  These classes will likely be private, so we may not have to worry about them.

Issue #269, opened 2021-07-13
Let's avoid adding print statements and instead use the standard python logging module.  Users won't want their stdout polluted.


voss70:
Let's avoid adding print statements and instead use the standard python logging module.  Users won't want their stdout polluted.

_Originally posted by @akaszynski in https://github.com/pyansys/PyAEDT/pull/264#discussion_r668847453_

akaszynski:
What you'll want to do is define a module level logger at the top of each module and then log using that.  For example:

```py
import logging
log = logging.getLogger(__name__)
log.setLevel('CRITICAL')
```

Then have the messenger print to that logging module.  You can change the logging level of this module via
```
logging.getLogger('pyaedt.<module>').setLevel(logging.INFO)
```

Alternatively, you could create a logger just for the messaging module and set its level there.  Regardless, we need to convert all calls to print to use a logging module and then have the level of that logger set based on the logging level requested by the user.

akaszynski:
Pinging @MaxJPRey to take a look at this as well since it's forcing us to skip certain doctests.

PR #286, opened 2021-07-14
Initial set up of doctest for desktop.py docstrings

jleonatti:
This PR includes initial setup for running doctest on docstring examples within the pyaedt code and refactoring of the desktop.py examples. 

#### pytest.ini
The pytest.ini file adds options to run pytest with doctest integration. The `doctest_optionflags = ELLIPSIS` allows ellipses in the docstring example expected output to match any substring in the actual output. This is needed in some cases, since the output of some functions can be random. 

The `-p no:faulthandler` option was also added. Without this option, some Windows exceptions are thrown after the AEDT application is forced to shutdown in the pytest fixture. It seems like something related to how [new versions of pytest handle these exceptions](https://stackoverflow.com/questions/57523762/pytest-windows-fatal-exception-code-0x8001010d).

#### pyaedt/conftest.py
The conftest.py file defines a pytest fixture that is used during the start up of the doctest run. It creates a single instance of AEDT that is used throughout the pytest session and is closed at the end. This file will be modified to include other setup, specific to different solvers. For example, there will be a set up method to create an Icepak object that can be used throughout the icepak.py docstring examples. 

#### Running doctest on the examples
Currently, running doctest on all examples (using `pytest -v pyaedt`) will fail, since all the current docstring examples are not runnable. Since the docstring examples are being addressed in stages, only certain files will be able to run doctest. After this PR, the desktop.py file can run doctest with `pytest -v pyaedt/desktop.py`.

There are several docstring examples in this PR that include `# doctest: +SKIP` at the end of an example command. doctest will not run those lines. It is needed in some cases because the output is extremely random or because running that command will put AEDT in an irreversible state.


pyaedt/conftest.py
```diff
@@ -0,0 +1,12 @@
+import pytest
```
akaszynski, line 1: Cool!  Had no idea you could define fixtures in a module level `conftest.py`.

pyaedt/desktop.py
```diff
@@ -298,12 +289,17 @@ class Desktop:
 
     >>> import pyaedt
     >>> desktop = pyaedt.Desktop("2021.1", NG=True)
-    >>> hfss = pyaedt.Hfss()
+    pyaedt Info: pyaedt v...
+    pyaedt Info: Python version ...
```
akaszynski, line 293: This will eventually be removed (or appear as a logger output) as we should avoid any printing to stdout.  See #269.

akaszynski:
Give me a bit to add in the CI/CD step to incorporate `pytest -v pyaedt/desktop.py` prior to merging this.  I think the PR is actually good to go as-is, but extra CI/CD would be nice to have, especially as we have 3 machines.

akaszynski:
> There are several docstring examples in this PR that include # doctest: +SKIP at the end of an example command. doctest will not run those lines. It is needed in some cases because the output is extremely random or because running that command will put AEDT in an irreversible state.

I think this is unavoidable considering we're testing a stateful desktop application.

One way we might be able to get around this when we expand doctesting is by specifying which python modules we wish to test should it not be possible to reset the state of AEDT when testing different modules (i.e. Hfss vs. Edb).  We'll have to see how much of an issue this becomes.

akaszynski (approved):
Added CI/CD and it ran successfully.  I've also modified the existing unit testing to remove the fault handler for a cleaner output.

Thanks for adding this and teaching me something.

pyaedt/desktop.py
jleonatti: Definitely agree.

jleonatti:
> One way we might be able to get around this when we expand doctesting is by specifying which python modules we wish to test should it not be possible to reset the state of AEDT when testing different modules (i.e. Hfss vs. Edb). We'll have to see how much of an issue this becomes.

@akaszynski, I suspect that this will have the biggest impact on the desktop level. Here is where there are methods for closing AEDT and updating the AEDT registry, etc, which are irreversible or cannot be accessed from the doctest for some reason. For docstrings in the solver level, like for Icepak, I think most examples will be able to execute without issue.

pyaedt/desktop.py
```diff
@@ -103,30 +101,24 @@ def update_aedt_registry(key, value, desktop_version="193"):
     desktop_version : str, optional
         Version of AEDT to use. The default is ``"193"``
```
MaxJPRey, line 102: @jleonatti the default version is 193 while in the conftest.py file, we use the 2021.1 as parameter of the Desktop class
`desktop = Desktop("2021.1", NG=True)`.

Is it expected?

jleonatti: @MaxJPRey that default version of 193 is only for the `update_aedt_registry` method. I'm not sure why that specific version was chosen in that case. But in the case of starting AEDT through the Desktop object, you can specify the version you want to start in the Desktop constructor or it will select the newest version, I believe. I think that most users will be using newer version of AEDT here, which is why I selected 2021.1.

akaszynski: @jleonatti, I think this should be raised as an issue.

jleonatti: @akaszynski is the issue that 193 is the default version or that the format is expected to be "123" instead of "1234.5"? Or both?

akaszynski: Issue is default version.  Do we even "officially" support 19.3 as of 2021R2?

PR #291, opened 2021-07-15
Doc/hfss create frequency sweep

MaxJPRey:
Add an example for the `create_frequency_sweep()` in the hfss module.

### DIFF
--- a/pyaedt/hfss.py
+++ b/pyaedt/hfss.py
@@ -389,15 +389,18 @@ class Hfss(FieldAnalysis3D, object):
         Examples
         --------
 
-        Create a frequency sweep named ``'Sweep'`` for the analysis setup named ``'Setup1'``.
+        Create a setup named ``'MySetup'``.
+        Create a fast frequency sweep named ``'MySweepFast'`` for the analysis setup created above.
 
-        >>> setup = hfss.create_setup("MySetup41")
+        >>> setup = hfss.create_setup("MySetup")
         >>> setup.props["Frequency"] = "1GHz"
         >>> setup.props["BasisOrder"] = 2
         >>> setup.props["MaximumPasses"] = 1
-        >>> frequency_sweep = hfss.create_frequency_sweep(setupname="MySetup41", sweepname="MySweepFast", unit="MHz",
-        ...                             freqstart=1.1e3, freqstop=1200.1, num_of_freq_points=1234,
-        ...                             sweeptype="Fast")
+        >>> frequency_sweep = hfss.create_frequency_sweep(setupname="MySetup", sweepname="MySweepFast", unit="MHz",
+        ...                                               freqstart=1.1e3, freqstop=1200.1, num_of_freq_points=1234,
+        ...                                               sweeptype="Fast")
+        >>> type(frequency_sweep)
+        <class 'pyaedt.modules.SetupTemplates.SweepHFSS'>
 
         """
 
--- a/pyaedt/hfss.py
+++ b/pyaedt/hfss.py
@@ -376,6 +376,13 @@ class Hfss(FieldAnalysis3D, object):
         -------
         SweepHFSS
             Sweep object.
+
+        Examples
+        --------
+
+        Create a frequency sweep named 'Sweep' for the analysis setup named 'Setup1'.
+
+        >>> frequency_sweep = aedtapp.create_frequency_sweep("Setup1", sweepname ="Sweep")
         """
 
         if sweepname is None:
--- a/pyaedt/hfss.py
+++ b/pyaedt/hfss.py
@@ -383,6 +383,7 @@ class Hfss(FieldAnalysis3D, object):
         Create a frequency sweep named 'Sweep' for the analysis setup named 'Setup1'.
 
         >>> frequency_sweep = aedtapp.create_frequency_sweep("Setup1", sweepname ="Sweep")
+
         """
 
         if sweepname is None:
--- a/pyaedt/hfss.py
+++ b/pyaedt/hfss.py
@@ -391,7 +391,13 @@ class Hfss(FieldAnalysis3D, object):
 
         Create a frequency sweep named ``'Sweep'`` for the analysis setup named ``'Setup1'``.
 
-        >>> frequency_sweep = hfss.create_frequency_sweep("Setup1", sweepname ="FrequencySweep")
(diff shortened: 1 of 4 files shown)


pyaedt/hfss.py
```diff
@@ -373,6 +373,14 @@ def create_frequency_sweep(self, setupname, unit="GHz", freqstart=1e-3, freqstop
         -------
         SweepHFSS
             Sweep object.
+
+        Examples
+        --------
+
+        Create a frequency sweep named 'Sweep' for the analysis setup named 'Setup1'.
```
akaszynski, line 380:
```suggestion
        Create a frequency sweep named ``'Sweep'`` for the analysis setup named ``'Setup1'``.
```

pyaedt/hfss.py
```diff
@@ -373,6 +373,14 @@ def create_frequency_sweep(self, setupname, unit="GHz", freqstart=1e-3, freqstop
         -------
         SweepHFSS
```
akaszynski, line 374:
```suggestion
        :class:`pyaedt.modules.SetupTemplates.SweepHFSS` or :class:`pyaedt.modules.SetupTemplates.SweepQ3D`
```
New documentation style (in next release and dev.aedtdocs.pyansys.com) will be able to directly link classes

pyaedt/hfss.py
```diff
@@ -373,6 +373,14 @@ def create_frequency_sweep(self, setupname, unit="GHz", freqstart=1e-3, freqstop
...
             Sweep object.
+
+        Examples
+        --------
+
+        Create a frequency sweep named 'Sweep' for the analysis setup named 'Setup1'.
+
+        >>> frequency_sweep = aedtapp.create_frequency_sweep("Setup1", sweepname ="FrequencySweep")
```
akaszynski, line 382:
```suggestion
        >>> frequency_sweep = aedtapp.create_frequency_sweep("Setup1", sweepname="FrequencySweep")
```

akaszynski: This needs the application setup beforehand as in #286, and we need the `__repr__` for `frequency_sweep`.

pyaedt/hfss.py
```diff
@@ -373,6 +373,14 @@ def create_frequency_sweep(self, setupname, unit="GHz", freqstart=1e-3, freqstop
         -------
         SweepHFSS
             Sweep object.
```
akaszynski, line 375:
```suggestion
            Sweep object if successful.  ``False`` if unsuccessful.
```

pyaedt/hfss.py
MaxJPRey: OK, that 's great. That's avoid any wrong spelling and in case we change the name of the class, I am assuming the link will be updated as in c#.

pyaedt/hfss.py
jleonatti: @akaszynski I plan to get a working example of how doctest setup will work with Icepak and then show @MaxJPRey how it works. He can then update this example for HFSS using the same approach.

pyaedt/hfss.py
akaszynski: > I am assuming the link will be updated as in c#

We'll have a link within the documentation, not sure what doc generator you're using for C# (doxygen?).

pyaedt/hfss.py
MaxJPRey: @akaszynski I am not sure what you are expecting regarding the __repr__ for frequency_sweep.
Do you expect me to directly write something such as:
`>>> aedtapp.create_frequency_sweep("Setup1", sweepname="FrequencySweep")`
?
Or do you want me to print frequency_sweep after declaring/assigning it?

akaszynski: > Or do you want me to print frequency_sweep after declaring/assigning it?

Yes, that's ideal.  It's always helpful to see the return of a method, even if it's something trivial.  The `__repr__` for `frequency_sweep` isn't likely to be great, so you could even run `type(frequency_sweep)`, but regardless, showing something is best.

pyaedt/hfss.py
```diff
@@ -300,8 +303,13 @@ def assigncoating(self, obj, mat=None,
 
         Examples
         --------
-        >>> id1 = aedtapp.modeler.primitives.get_obj_id("inner")
-        >>> coat = aedtapp.assigncoating([id1], "copper",usethickness=True, thickness="0.2mm")
+
```
akaszynski, line 306: Good example!

Since this is a fairly long example, explain using plain english what you're doing.  For example:

Create a cylinder at the XY working plane and assign a copper coating of 0.2 mm to it.

Also, we only need one line of white space prior to the end of the docstring.

pyaedt/hfss.py
```diff
@@ -55,37 +55,40 @@ class Hfss(FieldAnalysis3D, object):
...
 
-    >>> hfss = Hfss(specified_version="2021.1", projectname="myfile.aedt")
+    >>> hfss2 = Hfss(specified_version="2021.1", projectname="myfile.aedt")
 
-    Create an instance of HFSS using the 2021 R2 student version and open
+    Create an instance of `HFSS` using the 2021 R2 student version and open
     the specified project, which is named ``"myfile.aedt"``.
 
-    >>> hfss = Hfss(specified_version="2021.2", projectname="myfile.aedt", student_version=True)
+    >>> hfss2 = Hfss(specified_version="2021.2", projectname="myfile.aedt", student_version=True)
```
akaszynski, line 91: Out of curiosity, why is this `hfss2` instead of `hfss`?

pyaedt/hfss.py
```diff
@@ -984,7 +1000,7 @@ def create_wave_port_microstrip_between_objects(self, startobj, endobject, axisd
 
         Examples
         --------
-        >>>  from pyaedt import Hfss
```
akaszynski, line 987: This also needs a short description for this example.

akaszynski (approved):
Nice work on your first PR.  I have some non-blocking suggestions/comments, but once you address those feel free to merge.

pyaedt/hfss.py
MaxJPRey: Great catch, it was just a temporary to fix a former hfss test/example issue (I fixed it.). It should definitely be ``hfss``.

pyaedt/hfss.py
MaxJPRey: Done.

PR #293, opened 2021-07-15
split PR doc build and full doc build

akaszynski:
Nightly docs aren't getting built, so I've rewritten the docbuild job to be split into a nightly cronjob and a PR light doc build.

### DIFF
--- a/.github/workflows/nightly-docs.yml
+++ b/.github/workflows/nightly-docs.yml
@@ -1,12 +1,8 @@
-name: Documentation Build
+name: Nightly Documentation Build
 
 on:
-  push:
-    branches:
-      - '*'
-  pull_request:
-    branches:
-      - '*'
+  schedule:  # UTC at 0400
+    - cron:  '0 4 * * *'
 
 jobs:
   docs_build:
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-docs.yml
```diff
@@ -0,0 +1,47 @@
...
+        run: |
+          pip install .
+
+      - name: Install doc build requirements
+        run: |
+          pip install -r requirements_docs.txt
+
+      - name: Full Documentation Build
+        run: |
+          make -C doc phtml
```
MaxJPRey, line 29: @akaszynski, is it supposed to be:
`make -C doc phtml`
or
`make -C doc html`
?

akaszynski: See the makefile.  Effectively, `phtml` is running a parallel build of sphinx using as many cores available.  Github agents run https://docs.microsoft.com/azure/virtual-machines/dv2-dsv2-series#dsv2-series, meaning we get once genuine core and one hyperthreaded.  There's some small improvement, so its better than nothing.

PR #295, opened 2021-07-15
Update Design.py

PipKat:
@akaszynski 
I don't understand the long description for this module. One reason is because of all of the "it" pronouns!
Also, search for "Design properties." You'll find the property "design_properties". It shows a parameter of "design_name" in the code. In the rendering, there is no list of parameters following the property name (as there is for methods). Should properties have parameters?

### DIFF
--- a/pyaedt/application/Design.py
+++ b/pyaedt/application/Design.py
@@ -2,8 +2,7 @@
 This module contains these classes: `Design` and `DesignCache`.
 
 This module provides all functionalities for basic project information and objects. 
-Because this module is inherited in the Main Tool class, 
-it will be a simple call from it.
+These classes is are inherited in the main tool class.
 
 """
 from __future__ import absolute_import
@@ -210,7 +209,7 @@ def list_difference(list1, list2):
     return list(set(list1) - set(list2))
 
 class DesignCache(object):
-    """DesignCache class.
+    """Provides for analyzing the differences in the state of a design between two points in time.
     
     This class provides for analyzing the differences in the state of a design between two points in time.
 
@@ -626,7 +625,27 @@ class Design(object):
 
     @property
     def design_type(self):
-        """Design type.    
+        """Design type.
+        
+        One of the following:
+        - ``'Maxwell 2D'``
+        - ``'Maxwell 3D'``
+        - ``'Twin Builder'``
+        - ``'Circuit Design'``
+        - ``'2D Extractor'``
+        - ``'Q3D Extractor'``
+        - ``'HFSS'``
+        - ``'Icepak'``
+        - ``'RMxprtSolution'``
+        - ``'ModelCreation'``
+        - ``'HFSS 3D Layout Design'``
+        - ``'Mechanical'``
+        - ``'EMIT'
+
+        Returns
+        --------
+        str
+            Design type.  See above for one of the possible return values.
 
         """
         #return self._odesign.GetDesignType()
@@ -875,7 +894,7 @@ class Design(object):
 
         Returns
         -------
-         str
+        str
              Full absolute path for the project's working directory. 
              If this directory does not exist, it is created.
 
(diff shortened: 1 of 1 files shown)


pyaedt/application/Design.py
```diff
@@ -1,7 +1,7 @@
 """
 This module contains these classes: `Design` and `DesignCache`.
 
-It contains all basic project information and objects. 
+This module provides all functionalities for basic project information and objects. 
 Because this module is inherited in the Main Tool class, 
 it will be a simple call from it.
```
akaszynski, line 6:
```suggestion
These classes is are inherited in the main tool class.
```

Effectively, this means that

pyaedt/application/Design.py
```diff
@@ -212,7 +212,7 @@ def list_difference(list1, list2):
 class DesignCache(object):
     """DesignCache class.
```
akaszynski, line 213:
```suggestion
    """Provides for analyzing the differences in the state of a design between two points in time.
```

I'm also ok with "Used for..."

pyaedt/application/Design.py
```diff
@@ -631,15 +626,7 @@ def design_list(self):
...
-        """Design type.
-
-        Parameters
-        ----------
-        type
-            Type of the design.
-
-        Returns
-        -------
+        """Design type.
```
akaszynski, line 629:
```suggestion
        """Design type.
        
        One of the following:
        - ``'Maxwell 2D'``
        - ``'Maxwell 3D'``
        - ``'Twin Builder'``
        - ``'Circuit Design'``
        - ``'2D Extractor'``
        - ``'Q3D Extractor'``
        - ``'HFSS'``
        - ``'Icepak'``
        - ``'RMxprtSolution'``
        - ``'ModelCreation'``
        - ``'HFSS 3D Layout Design'``
        - ``'Mechanical'``
        - ``'EMIT'

        Returns
        --------
        str
            Design type.  See above for one of the possible return values.
```

pyaedt/application/Design.py
```diff
@@ -889,7 +876,7 @@ def working_directory(self):
         Returns
         -------
          str
```
akaszynski, line 878:
```suggestion
        str
```

pyaedt/application/Design.py
```diff
@@ -2431,15 +2419,15 @@ def validate_simple(self, logfile=None):
 
     @aedt_exception_handler
     def get_evaluated_value(self, variable_name, variation=None):
-        """Return the evaluated value of a design property or project variable in SI units.
+        """Retrieve the evaluated value of a design property or project variable in SI units.
```
akaszynski, line 2424:
```suggestion
```

pyaedt/application/Design.py
```diff
@@ -2499,11 +2487,11 @@ def evaluate_expression(self, expression_string):
     def design_variation(self, variation_string=None):
         """Generate a string to specify a desired variation.
 
-        This method converts a user input string defining a desired solution variation into a valid
+        This method converts an input string defining a desired solution variation into a valid
         string taking into account all existing design properties and project variables, including 
         dependent (non-sweep) properties.
 
-        This is needed because the standard function `GetVariationVariableValue` does not work for obtaining
+        This is needed because the standard method :funct:`GetVariationVariableValue` does not work for obtaining
```
akaszynski, line 2494:
```suggestion
        This is needed because the standard method COM function ``GetVariationVariableValue`` does not work for obtaining
```
This is a special one because the COM object (the underlying API) doesn't have any python documentation, so we can't link to this.

akaszynski (approved):
Approved with various suggestions.

pyaedt/application/Design.py
PipKat: Fix "is are" in Alex's suggestion.

PR #299, opened 2021-07-16
Initial setup for icepak doctest

jleonatti:
- Add doctest_fixtures directory under pyaedt to hold module related fixtures for doctest.
- Fixtures defined in other files are imported into the conftest.py and the module path added to the `pytest_plugins ` list.
- Add icepak_fixture.py to setup a project with an Icepak design for testing throughout icepak.py
- One icepak docstring example was updated to show how this works with the fixture setup. My next task will be to update the rest of the icepak docstrings.


jleonatti:
Currently working to fix failing unittests.

akaszynski:
It's an insane traceback, but issue boils down to:
```
doctest_namespace["icepak"] = Icepak(project_name=project_path, design_name="IcepakDesign1")
E       TypeError: __init__() got an unexpected keyword argument 'project_name'
```

akaszynski:
Should be fixed in 357dadf0c51336c7e5921820849c883f41227f44.  Doesn't help that the kwarg names aren't snake case...

That's a refactor for another day.

pyaedt/icepak.py
```diff
@@ -124,7 +128,16 @@ def assign_openings(self, air_faces):
         Returns
         -------
         type
```
akaszynski, line 130: As you're adding these examples, if you see incorrect or generic returns, please modify them to the correct return type using the sphinx `:class:` directive.

```suggestion
        :class:`pyaedt.modules.Boundary.BoundaryObject`
```

.github/workflows/unit_tests.yml
```diff
@@ -62,7 +62,7 @@ jobs:
         run: |
           testenv\Scripts\Activate.ps1
           pip install pytest
-          pytest -v pyaedt/desktop.py -p no:faulthandler
+          pytest -v pyaedt/desktop.py pyaedt/icepak.py -p no:faulthandler
```
akaszynski, line 65:
```suggestion
          pytest -v pyaedt/desktop.py pyaedt/icepak.py
```
We can remove the option `no:faulthandler` here as it's in `pytest.ini`, correct?

pyaedt/conftest.py
```diff
@@ -10,3 +20,14 @@ def start_aedt():
     # Wait to run doctest on docstrings
     yield desktop
     desktop.force_close_desktop()
+
+@pytest.fixture(autouse=True, scope="session")
+def get_root_directory(doctest_namespace):
+    """Setup paths to the root and project directories.
```
akaszynski, line 26:
```suggestion
    """Set up paths to the root and project directories.
```
I had to look this up myself. Please see [Setup vs set up](https://www.enhancemywriting.com/setup-vs-set-up).

akaszynski (approved):
Approved with minor suggestions.

.github/workflows/unit_tests.yml
jleonatti: @akaszynski yes, that's correct.

pyaedt/icepak.py
jleonatti: @akaszynski I will do that moving forward. I didn't know about referencing classes like that.

jleonatti:
@akaszynski thanks for the fix for the failing tests!

MaxJPRey (commented):
Approved.

PR #301, opened 2021-07-17
Reintroduced assign_material in Analysis3D, plus several bug fixes!

Alberto-DM:
- Fixed `get_group_bounding_box`.
- Reintroduced `assignmaterial` in Analysis3D (still useful -and faster- in case of material assignment to a list of objects). 
- Renamed `assignmaterial` in `assign_material`.
- Refactored `add_workbench_link` --> `set_object_deformation`.
- Fixed bug in `delete_design` not working with 3DLayout. 
- Fixed bug in `copy_design_from` not working with 3DLayout.
- Fixed bug in `copy_design_from` not correctly setting back the active design.

### DIFF
--- a/_unittest/test_01_Design.py
+++ b/_unittest/test_01_Design.py
@@ -94,8 +94,8 @@ class TestClass():
         print(self.aedtapp.variable_manager)
         print(self.aedtapp.materials)
 
-    def test_09_add_workbench_link(self):
-        assert self.aedtapp.modeler.add_workbench_link(["inner"], "25")
+    def test_09_set_objects_deformation(self):
+        assert self.aedtapp.modeler.set_objects_deformation(["inner"])
 
     def test_09_set_objects_temperature(self):
         ambient_temp = 22
--- a/pyaedt/application/Analysis2D.py
+++ b/pyaedt/application/Analysis2D.py
@@ -74,7 +74,7 @@ class FieldAnalysis2D(Analysis):
     #     return self._post
 
     @aedt_exception_handler
-    def assignmaterial(self, obj, mat):
+    def assign_material(self, obj, mat):
         """Assign a material to one or more objects. 
 
         Parameters
--- a/pyaedt/application/Analysis2D.py
+++ b/pyaedt/application/Analysis2D.py
@@ -1,3 +1,5 @@
+import warnings
+
 from .Analysis import Analysis
 from ..modeler.Model2D import Modeler2D
 from ..modules.Mesh import Mesh
@@ -72,6 +74,19 @@ class FieldAnalysis2D(Analysis):
     # @property
     # def post(self):
     #     return self._post
+
+    @aedt_exception_handler
+    def assignmaterial(self, obj, mat):
+        """Assign a material to one or more objects.
+
+        .. deprecated:: 0.3.1
+           Use :func:`FieldAnalysis2D.assign_material` instead.
+
+        """
+        # raise a DeprecationWarning.  User won't have to change anything
+        warnings.warn('assignmaterial is deprecated.  Please use assign_material instead.',
+                      DeprecationWarning)
+        self.assign_material(obj, mat)
 
     @aedt_exception_handler
     def assign_material(self, obj, mat):
--- a/pyaedt/application/Analysis3D.py
+++ b/pyaedt/application/Analysis3D.py
@@ -233,9 +233,9 @@ class FieldAnalysis3D(Analysis, object):
             ``True`` when successful, ``False`` when failed.
             
         """
-        body_list = design.modeler.solid_bodies
+        body_list = design.modeler.primitives.solid_names
         if include_sheets:
-            body_list += design.modeler.primitives.get_all_sheets_names()
+            body_list += design.modeler.primitives.sheet_names
         selection_list = []
         material_properties = design.modeler.primitives.objects
         for body in body_list:
--- a/pyaedt/application/Analysis3D.py
+++ b/pyaedt/application/Analysis3D.py
@@ -1,6 +1,7 @@
-import glob
 import os
 import ntpath
+import warnings
+
 from ..generic.general_methods import aedt_exception_handler, generate_unique_name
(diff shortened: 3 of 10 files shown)


_unittest/test_01_Design.py
```diff
@@ -94,8 +94,8 @@ def test_08_objects(self):
         print(self.aedtapp.variable_manager)
         print(self.aedtapp.materials)
 
-    def test_09_add_workbench_link(self):
-        assert self.aedtapp.modeler.add_workbench_link(["inner"], "25")
+    def test_09_set_objects_deformation(self):
+        assert self.aedtapp.modeler.set_objects_deformation(["inner"])
```
akaszynski, line 98: I'd like for us to validate more than just the method ran.  Perhaps if we also has (and this code probably isn't right):
```
        assert self.aedtapp.modeler.get_objects_deformation(["inner"]) == 25
```

pyaedt/application/Analysis2D.py
```diff
@@ -74,7 +74,7 @@ def mesh(self):
     #     return self._post
 
     @aedt_exception_handler
-    def assignmaterial(self, obj, mat):
+    def assign_material(self, obj, mat):
```
akaszynski, line 77: Whenever we change the API like this, we __must__ provide an old method that calls the new method and raise a warning (or deprecation error) for the user.  Also, it's best to provide a [Deprecated Directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated) in the docstring of the older method.

Pinging @jleonatti and @MaxJPRey as well.  Should we change the API, give the user a chance to change theirs too or at least be notified when it changes.  Otherwise, they'll simply stop upgrading, or worse, using our API due to stability concerns.  For this reason, I'm in favor of using the warning approach first and then using an error after a minor release or two.

Add the following above this:
```py
         def assignmaterial(self, obj, mat):
            """Assign a material to one or more objects.

            .. deprecated:: 0.3.1
               Use :func:`FieldAnalysis2D.assign_material` instead.

            """
            # one of the following:

            # raise a DeprecationWarning.  User won't have to change anything
            warnings.warn('assignmaterial is deprecated.  Please use assign_material instead',
                          warnings.DeprecationWarning)
            self.assign_material(obj, mat)

            # or raise an AttributeError (could also make a custom DeprecationError)
            raise AttributeError('assignmaterial is deprecated.  Please use assign_material instead')
```

pyaedt/application/Analysis3D.py
```diff
@@ -291,6 +291,65 @@ def set_source_context(self, sources, number_of_modes=1):
...
+            One or more objects to assign materials to.
+        mat : str
+            Material to assign. If this material is not present, it will be
+            created.
+
+        Returns
+        -------
+        bool
+            ``True`` when successful, ``False`` when failed.
+
```
akaszynski, line 310: Let's get in the habit of adding an `Example` section for all our methods as we update the API.

pyaedt/application/Design.py
```diff
@@ -2035,18 +2035,16 @@ def delete_design(self, name=None, fallback_design=None):
...
+                self.odesign = None
         except:
             pass
-
+        if fallback_design and (fallback_design in
+                                [x for i in list(self._oproject.GetDesigns()) for x in (i.GetName(), i.GetName()[2:])]):
+            try:
+                self.set_active_design(fallback_design)
+            except:
+                pass
```
akaszynski, line 2047: Should we not be returning ``False`` here?

pyaedt/modeler/Modeler.py
```diff
@@ -833,58 +833,27 @@ def set_working_coordinate_system(self, name):
         return True
 
     @aedt_exception_handler
-    def add_workbench_link(self, objects, ambient_temp=22, create_project_var=False, enable_deformation=True):
-        """Assign Temperature and Deformation Objects for WorkBench Link. From 2020R2 Material needs to have Thermal Modifier
+    def set_objects_deformation(self, objects):
```
akaszynski, line 836: Add deprecation error to the old `add_workbench_link`.  Looks like we're completely removing it.

akaszynski (changes requested):
See the comments regarding depreciation warnings and errors.

pyaedt/application/Analysis2D.py
akaszynski: Added a "fun" write up regarding this here:
https://sphinxdocs.pyansys.com/coding_style.html

_unittest/test_01_Design.py
Alberto-DM: I agree with your suggestion, but this method just set a flag option and AFAIK there is no API call to check its status.

pyaedt/application/Analysis2D.py
Alberto-DM: I am being victim of my own suggestion 😄 
I'll do the change.

pyaedt/application/Design.py
Alberto-DM: I was thinking the same.
I have refactored the method.

pyaedt/modeler/Modeler.py
Alberto-DM: Done.

pyaedt/application/Analysis3D.py
Alberto-DM: Done

maxcapodi78 (approved):

maxcapodi78 (approved):

pyaedt/application/Analysis2D.py
```diff
@@ -75,6 +77,19 @@ def mesh(self):
 
     @aedt_exception_handler
     def assignmaterial(self, obj, mat):
+        """Assign a material to one or more objects.
+
+        .. deprecated:: 0.3.1
```
akaszynski, line 82: Perfect.  Thanks!

akaszynski (approved):
Thanks for making the changes.  LGTM.

PR #302, opened 2021-07-18
Shorten method names in the API Reference

akaszynski:
The method links in the left sidebar are often so long that the method name itself isn't visible without scrolling; this is undesired behavior.  This PR shortens the displayed links to just the method by overriding the `autosummary/base.rst` template.

### Before
broken

### After
fixed


maxcapodi78 (approved):

maxcapodi78:
I merge it because I'll create a new doc version

Issue #259, opened 2021-07-12
Refactor Desktop handling 


voss70:
If the Desktop option AlwaysNew=False is set, then an existing Desktop will be used even if the non-graphical setting is not as defined by the NG setting. We need to identify when a session is running in graphical/non-graphical mode and open a new session if the NG of the existign session is not as desired

voss70:
Desktop will be called exclusively from within the application classes. A Desktop Object will be stored within the Design Object, and created as needed from oDesktop if AlwaysNew=False.  In this PR we will also rename the variables AlwaysNew  and NG to use_existing_desktop and non_graphical. Re-using the desktop is only allowed if the non-graphical status is consistent

voss70:
Also in this context address the issue that aedtapp.release_desktop(close_projects=False, close_desktop=False) does not release the desktop: it is not possible to close the desktop after this command at the design level. Also condider here that the default arguments for this function should probably be aedtapp.release_desktop(close_projects=**False**, close_desktop=**False**) or maybe even remove the close_desktop input to this function. tbd

voss70:
It would simplify the API if we could eliminate the student_version parameter, and just let the user specify specified_version="2021.1.SV" or something like that.

PR #304, opened 2021-07-19
Refactor desktop handling #259

voss70:
Rename NG and AlwaysNew for all applications to obtain

For Desktion:
__init__(self, specified_version=None, non_graphical=False, launch_new_desktop=False, release_on_exit=True, student_version=False)

For Applications:
def __init__(self, projectname=None, designname=None, solution_type=None, setup_name=None,
                 specified_version=None, non_graphical=False, launch_new_desktop=True, release_on_exit=False,student_version=False)

Minor code refactoring


voss70:
Working on merge conflicts in the docstrings ... please wait to review

_unittest/conftest.py
```diff
@@ -85,12 +85,8 @@ def setup_class(self, project_name=None, design_name=None, solution_type=None, a
                 application = Hfss
 
             self.aedtapp = application(projectname=self.test_project, designname=design_name,
```
akaszynski, line 87: During this refactor, can we also change the kwargs to `project_name` and `design_name`?  Would be great to be consistent with snake case.  Pinging @jleonatti as well.

PR #306, opened 2021-07-19
updated docstrings returns

maxcapodi78:


Alberto-DM (approved):
Looks good to me.

pyaedt/application/Analysis2D.py
```diff
@@ -61,12 +61,22 @@ def __init__(self, application, projectname, designname, solution_type, setup_na
...
+        Returns
+        -------
+        :class:`pyaedt.modeler.Model2D.Modeler2D`
+        """
         return self._modeler
 
     @property
     def mesh(self):
-        """Mesh object"""
+        """
```
akaszynski, line 74: @maxcapodi78, didn't quite get a change to respond to this.

By removing the "short summary" (first sentence, immediately after the start of `"""`, these attributes will have an empty description in the autogenerated attributes table in our docs.  We should have something there (we could even have the returns type so the the link is in the table itself.  For example:

```
    @property
    def mesh(self):
        """Mesh object :class:`pyaedt.modules.Mesh.Mesh`

        Returns
        -------
        :class:`pyaedt.modules.Mesh.Mesh`
        """
```

PR #308, opened 2021-07-19
Fix reference to messenger

jleonatti:
Test failing locally due to incorrect reference to messenger
image


akaszynski (approved):
Thanks for fixing this.

PR #315, opened 2021-07-21
Doc/hfss sweep examples

MaxJPRey:
Rename the setup for the test `create_frequency_sweep`.
This will avoid confusion with the other tests as every test will have its own _Setup_ under _Analysis_.

Add examples for the following Hfss methods:
`create_linear_count_sweep()`
`create_linear_step_sweep()`

### DIFF
--- a/pyaedt/hfss.py
+++ b/pyaedt/hfss.py
@@ -389,17 +389,17 @@ class Hfss(FieldAnalysis3D, object):
         Examples
         --------
 
-        Create a setup named ``'MySetup'``.
+        Create a setup named ``'FrequencySweepSetup'``.
         Create a fast frequency sweep named ``'MySweepFast'`` for the analysis
         setup created above.
 
-        >>> setup = hfss.create_setup("MySetup")
+        >>> setup = hfss.create_setup("FrequencySweepSetup")
         >>> setup.props["Frequency"] = "1GHz"
         >>> setup.props["BasisOrder"] = 2
         >>> setup.props["MaximumPasses"] = 1
-        >>> frequency_sweep = hfss.create_frequency_sweep(setupname="MySetup", sweepname="MySweepFast", unit="MHz",
-        ...                                               freqstart=1.1e3, freqstop=1200.1, num_of_freq_points=1234,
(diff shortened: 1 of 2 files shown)


pyaedt/hfss.py
```diff
@@ -459,8 +460,24 @@ def create_linear_count_sweep(self, setupname, unit, freqstart, freqstop, num_of
...
+            Sweep object if successful. ``False`` if unsuccessful.
+
+        Examples
+        --------
+
+        Create a setup named ``'LinearCountSetup'``.
+        Create a linear count sweep named ``'LinearCountSweep'`` for the analysis setup
+        created above.
+
+        >>> setup = hfss.create_setup("LinearCountSetup")
```
akaszynski, line 473: Good example.

pyaedt/hfss.py
```diff
@@ -459,8 +460,24 @@ def create_linear_count_sweep(self, setupname, unit, freqstart, freqstop, num_of
...
-            Sweep object.
+        :class:`pyaedt.modules.SetupTemplates.SweepHFSS` or bool
+            Sweep object if successful. ``False`` if unsuccessful.
+
+        Examples
+        --------
+
+        Create a setup named ``'LinearCountSetup'``.
+        Create a linear count sweep named ``'LinearCountSweep'`` for the analysis setup
+        created above.
```
akaszynski, line 471:
```suggestion
        Create a setup named ``'LinearCountSetup'`` and use it in a linear count sweep
        named ``'LinearCountSweep'``.
```
I think this reads better.  Please use a similar description in `create_frequency_sweep`

MaxJPRey: Sure, I will change all the similar comments in my other examples.

akaszynski (approved):
Perfect.  Good examples.  Let's plan to get an entire module's examples documented and then evaluate where we are at that point.

There's 3000+ methods to document (though not all need it), and pyansys team's goal won't be to do them all on our own.  Regardless, this is a great start and will serve as the foundation of examples moving forward.

jleonatti (approved):

PR #316, opened 2021-07-21
Add example for discrete sweep.

MaxJPRey:
Add example for `discrete_sweep()`


akaszynski (approved):
Great job on this!  Even basic users will be able to figure out how to instantiate hfss, so I think we're good with not initializing it on each example.

jleonatti (approved):

PR #318, opened 2021-07-22
Testrunner ironpython

voss70:
Created launch stubs for all IronPython versions of the pytest tests.
test_01EDB.py is renamed to ignore_01EdB.py because desktop crashes when this test is run inside desktop

Run the unittest using run_unittests.py in Pycharm

Batch version under run_unittests_batchmode.cmd

results stored in runner_unittest.log

The individual unitttests can be run individually within Pycharm, although the desktop still needs to be closed manually after that run (tbd)


maxcapodi78 (approved):
I've reviewed the code, corrected most of the remaining errors in UnitTest and added the pipeline to Ironpython that now works ok. Open Items:
- Few failing UT to debug on IPY
- improve pipeline to exit promptly if AEDT crashes  (it happens sometimes)
- add parametric Mark to unittest

pyaedt/modeler/Circuit.py
```diff
@@ -112,17 +112,21 @@ def __init__(self, parent):
...
+        #     if "isoutsideDesktop" in dir(_main) and not _main.isoutsideDesktop and self._parent.oproject.GetEDBHandle():
+        #         try:
+        #             self._edb = Edb(edb_folder, self._parent.design_name, True, self._parent._aedt_version, isaedtowned=True,
+        #                             oproject=self._parent.oproject)
+        #         except:
+        #             self._edb = None
+        #     else:
+        #         self._edb = None
+        # else:
+        #     self._mttime = 0
```
akaszynski, line 129: May consider just removing rather than commenting this.  If we do comment it, we might consider adding a note why we commented this and kept it around.

maxcapodi78: I added a todo into def edb just below. I'll check if this code is still needed and eventually remove. right now it doesn't work in Ironpython

pyaedt/edb_core/EDB_Data.py
```diff
@@ -944,10 +944,16 @@ def _update_hole_parameters(self, hole_type=None, params=None, offsetx=None, off
             offsety = self.hole_offset_y
         if not rotation:
             rotation = self.hole_rotation
-        newPadstackDefinitionData.SetHoleParameters(hole_type, convert_py_list_to_net_list(params),
-                                                      self._edb_value(offsetx),
-                                                      self._edb_value(offsety),
-                                                      self._edb_value(rotation))
+        if "IronPython" in sys.version or ".NETFramework" in sys.version:
```
akaszynski, line 947: We can consider having a global attribute within pyaedt that checks if we're within IronPython, so we don't have to keep writing this check everwhere.

within `pyaedt/__init__.py`

```py
USING_IRONPYTHON = "IronPython" in sys.version or ".NETFramework" in sys.version
```
Then used with:
```py
pyaedt.USING_IRONPYTHON
# or
from pyaedt import USING_IRONPYTHON
```

pyaedt/application/Design.py
```diff
@@ -2105,7 +2106,10 @@ def insert_design(self, design_name=None):
 
         """
         if self.design_type == "Circuit Design" or self.design_type == "HFSS 3D Layout Design":
-            self.modeler.edb.close_edb()
+            try:
+                self.modeler.edb.close_edb()
+            except:
```
akaszynski, line 2111: It's a bit of an antipattern to have naked `except:`.  Let's not do this in the future.

pyaedt/application/Design.py
```diff
@@ -1347,6 +1347,7 @@ def activate_variable_sensitivity(self, variable_name, min_val=None, max_val=Non
         """
         arg = ["NAME:AllTabs"]
         self._optimetrics_variable_args(arg, "Sensitivity",variable_name, min_val, max_val)
+        print(arg)
```
akaszynski, line 1350: We really should be using the logger here.

_unittest_ironpython/test_Simplorer.py
```diff
@@ -0,0 +1,23 @@
...
+test_obj = mymodule.TestClass()
+
+class TestSequenceFunctionsGenerate(PytestMockup):
+    @classmethod
+    def setUpClass(cls):
+        test_obj.setup_class()
+
+    @classmethod
+    def tearDownClass(cls):
+        test_obj.teardown_class()
```
akaszynski, line 18: This appears to be shared across several of these unit tests.  Can we consider a refactor to have it within a fixture/shared module?

_unittest_ironpython/run_unittests.py
```diff
@@ -0,0 +1,36 @@
+import unittest
+import os
+import sys
+from datetime import datetime
+import time
+
+sys.path.append(r"C:\Program Files\AnsysEM\AnsysEM21.1\Win64\PythonFiles\DesktopPlugin")
```
akaszynski, line 7: Avoid hardcoding paths.

_unittest_ironpython/run_unittests.py
```diff
@@ -0,0 +1,36 @@
...
+path_dir = os.path.join(os.path.abspath(os.path.dirname(__file__)), "..")
+sys.path.append(path_dir)
+
+os.environ["UNITTEST_CURRENT_TEST"] = "1"
+run_dir = os.path.abspath(os.path.dirname(__file__))
+print(run_dir)
```
akaszynski, line 13: If we're going to use print statements within here (which I'm fine with here within a testing framework), let's at least make sure that we're printing out what we're printing out (i.e. `print('Working directory', run_dir)`).

pyaedt/modeler/Circuit.py
akaszynski: If it adds functionality, we can have a branch that checks if in IronPython and then doesn't execute.

pyaedt/edb_core/EDB_Data.py
maxcapodi78: Good points. I'll create another PR on that suggestions and remove all the checks

pyaedt/application/Design.py
maxcapodi78: this is needed to bypass the aedt_exception_handler. Anyway I've replaced with if statement

pyaedt/application/Design.py
maxcapodi78: removed. it was in my debug phase

_unittest_ironpython/test_Simplorer.py
maxcapodi78: @vossleon next PR?

PR #336, opened 2021-07-26
Add examples for method that assign boundary conditions from sheet.

MaxJPRey:
Add examples for HFSS methods that assign boundary conditions from sheet.


pyaedt/hfss.py
```diff
@@ -1664,7 +1664,18 @@ def assig_voltage_source_to_sheet(self, sheet_name, axisdir=0, sourcename=None):
...
+
+        Examples
+        --------
+
+        Create a sheet and assign to it some voltage.
+
+        >>> sheet = hfss.modeler.primitives.create_rectangle(hfss.CoordinateSystemPlane.XYPlane,
+        ...                                                  [0, 0, -70], [10, 2], name="VoltageSheet",
+        ...                                                  matname="copper")
+        >>> hfss.assig_voltage_source_to_sheet(sheet.name, hfss.AxisDir.XNeg, "VoltageSheetExample")
```
jleonatti, line 1676: @MaxJPRey I don't know how many places this method is used, but it might be worth it to fix the typo in the method name here in this PR, so that it is `assign_voltage_source_to_sheet`.

jleonatti (approved):

pyaedt/hfss.py
MaxJPRey: @jleonatti Agreed, this need to be fixed. Today I created #342 to fix it in a dedicated PR.

PR #339, opened 2021-07-26
Update Application.rst

PipKat:
Remove duplicated Icepak entry from Applications list


jleonatti (approved):
Thanks for fixing this Kathy!

PR #347, opened 2021-07-27
Replace 'analyse' by 'analyze'.

### LINKED ISSUE #333: analyse versus analyze
In hfss.py (and likely other files), some methods use analyse_ and others use analyze_. We should always use analyze as we are to use American spellings. Also, in hfss.py, we are missing an "in" in assign in this method: assig_voltage_source_to_sheet

MaxJPRey:
Fix #333.
Replace 'analyse' by 'analyze' in pyAEDT.

### DIFF
--- a/pyaedt/modeler/PrimitivesNexxim.py
+++ b/pyaedt/modeler/PrimitivesNexxim.py
@@ -1058,7 +1058,7 @@ class NexximComponents(CircuitComponents):
 
         for prt in ports:
             arg_temp = []
-            arg_temp = ["NAME:Enabledanalyzes", ["NAME:"+str(prt), str(prt)+":=", [] ], ["NAME:VoltageSinusoidal"+str(id), str(prt)+":=", [] ] ]
+            arg_temp = ["NAME:EnabledAnalyses", ["NAME:"+str(prt), str(prt)+":=", [] ], ["NAME:VoltageSinusoidal"+str(id), str(prt)+":=", [] ] ]
             arg3.append(arg_temp)
 
         arg2.append(arg3)
--- a/pyaedt/modeler/PrimitivesNexxim.py
+++ b/pyaedt/modeler/PrimitivesNexxim.py
@@ -1058,7 +1058,7 @@ class NexximComponents(CircuitComponents):
 
         for prt in ports:
             arg_temp = []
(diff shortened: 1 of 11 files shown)


pyaedt/modeler/PrimitivesNexxim.py
```diff
@@ -1058,7 +1058,7 @@ def assign_sin_excitation2ports(self, ports, settings):
 
         for prt in ports:
             arg_temp = []
-            arg_temp = ["NAME:EnabledAnalyses", ["NAME:"+str(prt), str(prt)+":=", [] ], ["NAME:VoltageSinusoidal"+str(id), str(prt)+":=", [] ] ]
+            arg_temp = ["NAME:Enabledanalyzes", ["NAME:"+str(prt), str(prt)+":=", [] ], ["NAME:VoltageSinusoidal"+str(id), str(prt)+":=", [] ] ]
```
jleonatti, line 1061:
```suggestion
            arg_temp = ["NAME:EnabledAnalyses", ["NAME:"+str(prt), str(prt)+":=", [] ], ["NAME:VoltageSinusoidal"+str(id), str(prt)+":=", [] ] ]
```
I think this one should stay the same, since it is using the plural of analysis, which is analyses.

jleonatti (approved):

pyaedt/modeler/PrimitivesNexxim.py
MaxJPRey: Agreed. Thanks @jleonatti !

maxcapodi78, line 1061: @jleonatti @MaxJPRey  I agree with you. This is a native AEDT command and we cannot modify syntax.

maxcapodi78 (changes requested):

maxcapodi78 (commented):
@MaxJPRey analyze_nominal method is used in many functions and examples already by customer. I would leave the deprectated method and remove it in 0.4.0 @akaszynski @jleonatti  what do you think?

jleonatti:
> @MaxJPRey analyze_nominal method is used in many functions and examples already by customer. I would leave the deprectated method and remove it in 0.4.0 @akaszynski @jleonatti what do you think?

@maxcapodi78 yes, this is a good point. @MaxJPRey can you follow the deprecation guidelines [here](https://sphinxdocs.pyansys.com/coding_style.html#deprecation-best-practice)? You can leave the analyse_nominal API, and have it call the analyze_nominal method with a deprecation warning.

akaszynski (approved):

PR #348, opened 2021-07-27
changed way to check if a str is a number

maxcapodi78:


voss70:
Typically I would use exception handling for this

try : 
    float(test_string)
    res = True
except :
    res = False

Is there a reason for inventing a new way to do this ?

If we do accept this approach, then are we sure that the implementation is complete. What about the "+" sign, so that "3e+6" is valid ?

voss70:
For the case where we expect a float value, such as the pad_percent of the region, it does not really make sense to convert the input value to a string in order to judge it's validity. 

e.g. pad_percent is a number, not a string, so if isinstance(pad_percent, numbers.Number) would be the logical approach. Is there a fundamental issue of using numbers in IronPython ?

maxcapodi78:
@voss70 concerning 1st comment, you are right the + is missed. the reason I used that approach is that from literature this seems faster than try except. Anyway it should be small impact so I can revert to it

voss70 (approved):

pyaedt/modeler/Primitives.py
```diff
@@ -895,7 +895,7 @@ def _change_geometry_property(self, vPropChange, names_list):
             vPropServers.append(el)
         vGeo3d = ["NAME:Geometry3DAttributeTab", vPropServers, vChangedProps]
         vOut = ["NAME:AllTabs", vGeo3d]
-        self.oeditor.ChangeProperty(vOut)
+        retry_ntimes(10,self.oeditor.ChangeProperty, vOut)
```
akaszynski, line 898:
```suggestion
        retry_ntimes(10, self.oeditor.ChangeProperty, vOut)
```

pyaedt/generic/general_methods.py
```diff
@@ -343,4 +343,8 @@ def time_fn( fn, *args, **kwargs ):
     return results
 
 def isclose(a, b, rel_tol=1e-9, abs_tol=0.0):
-    return abs(a-b) <= max( rel_tol * max(abs(a), abs(b)), abs_tol )
\ No newline at end of file
+    return abs(a-b) <= max( rel_tol * max(abs(a), abs(b)), abs_tol )
+
+
+def is_number(a):
+    return str(a).replace(".", "").replace("+", "").replace("-", "").replace("e","").replace("E","").isnumeric()
```
akaszynski, line 350:
```suggestion
    return str(a).replace(".", "").replace("+", "").replace("-", "").replace("e", "").replace("E", "").isnumeric()
```

akaszynski: Is there any reason why `eval` wouldn't work here?  Something like:
```py
try:
    float(eval(a))
    return True
except:
    return False

voss70: I would tend to agree with Alex here.

voss70:
I believe that this function would fail on the input float("3f") which returns an error

voss70 (changes requested):
I believe that this function would fail on the input float("3f") which returns an error

I also still maintain that using a try/except on this would be safer and easier

pyaedt/generic/general_methods.py
maxcapodi78: changed

PR #361, opened 2021-07-30
Doc/hfss example validate design

MaxJPRey:
Add an example for the HFSS method named validate_full_design.
Add an example for the HFSS method named create_scattering.

### DIFF
--- a/pyaedt/hfss.py
+++ b/pyaedt/hfss.py
@@ -2326,8 +2326,9 @@ class Hfss(FieldAnalysis3D, object):
 
         Returns
         -------
-        list
-            List of all the validation information for later use.
+        tuple
+            Tuple containing a list of all the validation information for later use
+            and a bool indicating if the validation was succesful or not.
 
         Examples
         --------
@@ -2335,8 +2336,10 @@ class Hfss(FieldAnalysis3D, object):
         Validate the current design and save the log file into
         the current project directory.
 
-        >>> wave_port = hfss.validate_full_design()
+        >>> validation = hfss.validate_full_design()
         pyaedt Info: Design Validation Checks
+        >>> validation[1]
+        False
 
         """
 
@@ -2375,7 +2378,7 @@ class Hfss(FieldAnalysis3D, object):
                 val_list.extend(temp)
             os.remove(temp_val_file)
         else:
-            msg = "** No design validation file isfound. **"
+            msg = "** No design validation file is found. **"
             self._messenger.add_debug_message(msg)
             val_list.append(msg)
         msg = "** End of design validation messages. **"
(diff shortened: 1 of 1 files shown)


pyaedt/hfss.py
```diff
@@ -2365,7 +2378,7 @@ def validate_full_design(self, dname=None, outputdir=None, ports=None):
                 val_list.extend(temp)
             os.remove(temp_val_file)
         else:
-            msg = "** No design validation file isfound. **"
+            msg = "** No design validation file is found. **"
```
MaxJPRey, line 2381: @maxcapodi78 I am assuming it won't cause any deprecation issue here as this error message is appended to the list of validation message returned.

pyaedt/hfss.py
```diff
@@ -2447,8 +2460,17 @@ def create_scattering(self, plot_name="S Parameter Plot Nominal", sweep_name=Non
...
         bool
             ``True`` when successful, ``False`` when failed.
-            
+
+        Examples
+        --------
+
+        Create a scattering named ``"S Parameter Plot Nominal"``.
+
+        >>> hfss.create_scattering()
```
MaxJPRey, line 2469: Here the example is very short but if we decide to assign every parameter a value different from the default ones, we might have to create a new sweep, new ports... This could lead to a confusing example for the customer.

pyaedt/hfss.py
```diff
@@ -2326,10 +2326,23 @@ def validate_full_design(self, dname=None, outputdir=None, ports=None):
 
         Returns
         -------
-        list
-            List of all the validation information for later use.
+        tuple
+            Tuple containing a list of all the validation information for later use
+            and a bool indicating if the validation was successful or not.
```
akaszynski, line 2331:
```suggestion
        list
            List of all the validation information for later use.
        bool
            Indicates if the validation was successful or not.
```

Would prefer if detail the items we return separately.

pyaedt/hfss.py
akaszynski: Correct, same would go with changes within any methods that do not affect the public API.

pyaedt/hfss.py
```diff
@@ -2447,8 +2460,17 @@ def create_scattering(self, plot_name="S Parameter Plot Nominal", sweep_name=Non
         -------
         bool
             ``True`` when successful, ``False`` when failed.
-            
+
+        Examples
+        --------
+
+        Create a scattering named ``"S Parameter Plot Nominal"``.
```
akaszynski, line 2467:
```suggestion
        Create a scattering named ``"S Parameter Plot Nominal"`` using
        the default parameters.
```

pyaedt/hfss.py
akaszynski: Added a note to explain that we're just using the default parameters.

akaszynski (approved):
Approved with minor suggestions.

pyaedt/hfss.py
MaxJPRey: Ok. I will add this information in the example for the next ones.

PR #368, opened 2021-08-04
Doc/doctest coverage

MaxJPRey:
Add a python module to be able to evaluate the doctest example coverage in each pyaedt modules and submodules.
Here is an example of the output provided by this module:
Coverage


pyaedt/misc/doctest_coverage.py
```diff
@@ -0,0 +1,143 @@
...
+        total = len(doctests[module_name])
+        missing = len(methods_without_example)
+        covered = total - missing
+        if total!=0:
+            percentage_covered = covered/total*100
+        else:
+            # If no docstring is available in the module, coverage is considered to be 100%.
+            percentage_covered = 100
+
+        print(f'{module_name : <37}{total : ^19}{missing : ^4}{percentage_covered:8.2f}')
```
jleonatti, line 124: @MaxJPRey just a small suggestion, but you may want to display the percentage similarly to the unit test coverage report:
image

One decimal place may be sufficient and adding the percent sign would help be more clear.

akaszynski: Agreed.

akaszynski:
Thought about this a bit more, I think we might want to make this a standalone module `ansys-docs-example-coverage`.  This way we can share this across several modules, ensure we have the correct requirements to test, etc.  Add this to:
https://github.com/pyansys/example-coverage

---

Also, found a great [library](https://pypi.org/project/docstr-coverage/) that reports on documentation coverage (and this would have been really helpful to have at the start of this project!).  @jleonatti, could you please investigate adding this to our CI/CD?  Getting a coverage report that we can reference would be helpful to track total doc coverage and if we're missing anything.

akaszynski:
Just added this to https://github.com/pyvista/pyvista/pull/1562.  The community there (when not on vacation) is generally great about reviewing PRs.

Had to make a series of modifications to get the test to display correctly.  You may want to run a diff on my changes there.

Should we publish a library reporting coverage, we can replace the code used in `pyvista` by the "official" package.

pyaedt/misc/doctest_coverage.py
MaxJPRey: @jleonatti , Done.

jleonatti:
> Also, found a great library that reports on documentation coverage (and this would have been really helpful to have at the start of this project!). @jleonatti, could you please investigate adding this to our CI/CD? Getting a coverage report that we can reference would be helpful to track total doc coverage and if we're missing anything.
Yes I can look into this!

MaxJPRey:
> Thought about this a bit more, I think we might want to make this a standalone module `ansys-docs-example-coverage`. This way we can share this across several modules, ensure we have the correct requirements to test, etc. Add this to:
> https://github.com/pyansys/example-coverage
> 
> Also, found a great [library](https://pypi.org/project/docstr-coverage/) that reports on documentation coverage (and this would have been really helpful to have at the start of this project!). @jleonatti, could you please investigate adding this to our CI/CD? Getting a coverage report that we can reference would be helpful to track total doc coverage and if we're missing anything.

@akaszynski Shall I put this python module under https://github.com/pyansys/example-coverage?
Last week, we mentioned the fact that we could postpone the creation this package.
Is it acceptable to put this python file under this repo? Shall I had tests accordingly?

akaszynski:
> @akaszynski Shall I put this python module under https://github.com/pyansys/example-coverage?

I'd prefer this as a standalone package.  Reason being is that we need this to be usable for multiple packages.

MaxJPRey:
@akaszynski thanks, I am migrating the file to the appropriate location.

PR #375, opened 2021-08-09
Doc/circuit example

MaxJPRey:
Rename some variables to respect the PEP8.


akaszynski (approved):
Looks good; this reads much better.

PR #378, opened 2021-08-11
Doc/edits to start classes for app interfaces with verbs

PipKat:
Super simple PR....
Just hated the inconsistencies!


akaszynski (approved):
Agree with the changes.  Thanks!

PR #379, opened 2021-08-12
Doc/minor doc edits

PipKat:
Minor edits, such as initial capital letters in titles and rendering fixes.

### DIFF
--- a/doc/source/Resources/User_Guide.rst
+++ b/doc/source/Resources/User_Guide.rst
@@ -33,7 +33,7 @@ The same result can be obtained with the following code:
 
 .. code:: python
 
-    Launch the latest installed version of AEDT in graphical mode
+    # Launch the latest installed version of AEDT in graphical mode.
 
     from pyaedt import Circuit
     with Circuit(specified_version="2021.1", NG=False) as circuit:
--- a/README.rst
+++ b/README.rst
@@ -80,7 +80,7 @@ The main advantages of PyAEDT are:
 
 Example Workflow
 -----------------
-1. Initialize the ``Desktop`` class with the version of AEDT to use.
+1. Initialize the `Desktop` class with the version of AEDT to use.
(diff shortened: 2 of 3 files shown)


doc/source/Resources/User_Guide.rst
```diff
@@ -33,7 +33,7 @@ The same result can be obtained with the following code:
 
 .. code:: python
 
-    Launch the latest installed version of AEDT in graphical mode
+    # Launch the latest installed version of AEDT in graphical mode.
```
akaszynski, line 36: We should standardize this, and I'm find with either approach:

#### Use `>>>`

```py
Let's create an array of numbers.

>>> import numpy as np
>>> array = np.arange(10)
array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])

```

#### don't use `>>>`
```py

# Let's create an array of numbers.
import numpy as np
array = np.arange(10)

```

I like using `>>>` because we we can explain and show the output.

PipKat: @akaszynski I will gladly do the first--but there are a few code examples in this file, and I'm not confident on how to change them. Would you please edit these examples so I can learn from your edits? Thanks!

PipKat:
@akaszynski  
How do we delete this PR now that we know that the PY files edited herein are not the source files?

doc/source/Resources/User_Guide.rst
akaszynski: Sorry @PipKat, been bogged down with other work.  I'll try to get to this today.

akaszynski: Actually, let's add this in a later PR.  We can add this to the sphinx documentation in a later PR.

akaszynski (approved):
Looks good.  We may end up changing all of the single tick (which comes up as italic) to double tick (which renders as literal), because code should be literal.

PR #387, opened 2021-08-18
Fix: sets self._active_cell to default cell if cell name is not specified.

Alberto-DM:
In `open_edb` correctly sets `self._active_cell` to default cell if `cellname` is not specified.
Minor code clean-up.


akaszynski (approved):
Looks good to me.  Thanks!

PR #390, opened 2021-08-20
Added material appearance property to materials.

Alberto-DM:
Added material appearance property to materials. 
Created unit tests.

### DIFF
--- a/pyaedt/modules/Material.py
+++ b/pyaedt/modules/Material.py
@@ -361,7 +361,7 @@ class MatProperty(object):
         >>> from pyaedt import Hfss
         >>> hfss = Hfss(specified_version="2021.1")
         >>> mat1 = hfss.materials.add_material("new_copper2")
-        >>> mat1.aadd_thermal_modifier_free_form("if(Temp > 1000cel, 1, if(Temp < -273.15cel, 1, 1))")
+        >>> mat1.add_thermal_modifier_free_form("if(Temp > 1000cel, 1, if(Temp < -273.15cel, 1, 1))")
         """
         self._property_value[index].thermalmodifier = formula
         return self._add_thermal_modifier(formula, index)
@@ -691,31 +691,38 @@ class Material(CommonMaterial, object):
 
         Returns
         -------
-        type
+        list
             Appearance of the material.
+
+        Examples
+        --------
+
+        >>> from pyaedt import Hfss
+        >>> hfss = Hfss(specified_version="2021.1")
+        >>> mat1 = hfss.materials.add_material("new_material")
+        >>> rgbcolor = mat1.material_appearance
+        >>> mat1.material_appearance = [0, 153, 153]
         """
         return self._material_appearance
 
     @material_appearance.setter
-    def material_appearance(self, value):
-        if type(value) is list and len(value) == 3:
-            value_int = []
-            for i in value:
-                try:
-                    tmp = int(i)
-                except ValueError:
-                    raise
-                if tmp < 0 or tmp > 255:
-                    raise ValueError
-                value_int.append(tmp)
-            self._material_appearance = value_int
-            self._props["AttachedData"] = OrderedDict({"MatAppearanceData":
-                                                       OrderedDict({'property_data': 'appearance_data',
-                                                                    'Red': value_int[0],
-                                                                    'Green': value_int[1],
-                                                                    'Blue': value_int[2]})})
-        else:
-            raise ValueError
+    def material_appearance(self, rgb):
+        if not isinstance(rgb, (list, tuple)):
+            raise TypeError('`material_apperance` must be a list or tuple')
+        if len(rgb) != 3:
+            raise ValueError('`material_appearance` must be three items (RGB)')
+        value_int = []
+        for rgb_item in rgb:
+            rgb_int = int(rgb_item)
+            if rgb_int < 0 or rgb_int > 255:
+                raise ValueError('RGB value must be between 0 and 255')
+            value_int.append(rgb_int)
+        self._material_appearance = value_int
+        self._props["AttachedData"] = OrderedDict({"MatAppearanceData":
+                                                   OrderedDict({'property_data': 'appearance_data',
+                                                                'Red': value_int[0],
+                                                                'Green': value_int[1],
+                                                                'Blue': value_int[2]})})
 
     @property
     def permittivity(self):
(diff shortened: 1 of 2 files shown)


pyaedt/modules/Material.py
```diff
@@ -648,7 +648,17 @@ def __init__(self, parent, name, props=None):
...
-            self._props["PhysicsTypes"] = OrderedDict({"set":['Electromagnetic', 'Thermal', 'Structural']})
+            self._props["PhysicsTypes"] = OrderedDict({"set": ['Electromagnetic', 'Thermal', 'Structural']})
+        if "AttachedData" in self._props:
+            self._material_appearance = []
+            self._material_appearance.append(self._props["AttachedData"]["MatAppearanceData"]["Red"])
+            self._material_appearance.append(self._props["AttachedData"]["MatAppearanceData"]["Green"])
+            self._material_appearance.append(self._props["AttachedData"]["MatAppearanceData"]["Blue"])
+        else:
+            self._material_appearance = [128, 128, 128]
+            self._props["AttachedData"] = OrderedDict({"MatAppearanceData":
```
akaszynski, line 659: Any reason why we use an ordered dictionary here?

pyaedt/modules/Material.py
```diff
@@ -675,6 +685,38 @@ def __init__(self, parent, name, props=None):
...
 
+    @property
+    def material_appearance(self):
+        """Material Appearance specified as an RGB list.
+
+        Returns
+        -------
+        type
+            Appearance of the material.
+        """
```
akaszynski, line 696: Please include a basic example here along with the correct return type (looks to be a list).

pyaedt/modules/Material.py
```diff
@@ -675,6 +685,38 @@ def __init__(self, parent, name, props=None):
...
+        return self._material_appearance
+
+    @material_appearance.setter
+    def material_appearance(self, value):
+        if type(value) is list and len(value) == 3:
```
akaszynski, line 701: Raising a `ValueError` won't be helpful here.  Let's try this

```suggestion
        if not isinstance(value, (list, tuple)):
            raise TypeError('`material_apperance` must be a list or tuple)
        if len(value) != 3:
            raise ValueError('`material_appearance` must be three items (RGB)')
```

pyaedt/modules/Material.py
```diff
@@ -675,6 +685,38 @@ def __init__(self, parent, name, props=None):
...
+            value_int = []
+            for i in value:
+                try:
+                    tmp = int(i)
+                except ValueError:
+                    raise
```
akaszynski, line 707: We gain nothing here with a `try`/`except` by re raising the error.  We can either provide a helpful error:
```suggestion
                try:
                    tmp = int(i)
                except ValueError:
                    raise ValueError('List of RGB must be able to be converted to int')
```
Or simply:
tmp = int(i)

Also, `i` is not a good variable name.  Perhaps `rgb_item` works better.  Also, use `rgb` instead of `value` in the function signature.

pyaedt/modules/Material.py
```diff
@@ -675,6 +685,38 @@ def __init__(self, parent, name, props=None):
...
+                if tmp < 0 or tmp > 255:
+                    raise ValueError
```
akaszynski, line 709:
```suggestion
                    raise ValueError('RGB value must be between 0 and 255')
```

pyaedt/modules/Material.py
```diff
@@ -675,6 +685,38 @@ def __init__(self, parent, name, props=None):
...
+                value_int.append(tmp)
+            self._material_appearance = value_int
+            self._props["AttachedData"] = OrderedDict({"MatAppearanceData":
+                                                       OrderedDict({'property_data': 'appearance_data',
+                                                                    'Red': value_int[0],
+                                                                    'Green': value_int[1],
+                                                                    'Blue': value_int[2]})})
+        else:
+            raise ValueError
```
akaszynski, line 718: You will have to deindent this section.  With setters, it's best to do type checking and raise early rather than branching.

pyaedt/modules/Material.py
Alberto-DM: The entire `self._props` need to be a nested ordered dictionary (see line 564), otherwise AEDT will fail if the arguments are not given in the expected order. It's crazy I know.....

pyaedt/modules/Material.py
Alberto-DM: Done.

pyaedt/modules/Material.py
Alberto-DM: You are fully right. Much better.

pyaedt/modules/Material.py
Alberto-DM: Already done it.

pyaedt/modules/Material.py
Alberto-DM: Done. Chosen the simple way. Renamed also `tmp` to `rgb_int`.

pyaedt/modules/Material.py
Alberto-DM: Done.

pyaedt/modules/Material.py
akaszynski: Thanks for the explanation!

pyaedt/modules/Material.py
```diff
@@ -675,6 +685,45 @@ def __init__(self, parent, name, props=None):
...
+        """Material Appearance specified as an RGB list.
+
+        Returns
+        -------
+        list
+            Appearance of the material.
+
+        Examples
+        --------
+
```
akaszynski, line 699:
```suggestion
        Create a new material with color ``[0, 153, 153]`` (darker cyan).

```

pyaedt/modules/Material.py
```diff
@@ -675,6 +685,45 @@ def __init__(self, parent, name, props=None):
...
         pass
 
+    @property
+    def material_appearance(self):
+        """Material Appearance specified as an RGB list.
+
+        Returns
+        -------
+        list
+            Appearance of the material.
```
akaszynski, line 695:
```suggestion
            Color of the material in RGB.  Values are in the range ``[0, 255]``.
```

akaszynski (approved):
Thanks for the changes.  Made two more, but otherwise LGTM.

PR #396, opened 2021-08-23
Added 3.9 support

maxcapodi78:
Corrected edb example


akaszynski (approved):
Good idea for adding the wheel to the release.

Issue #349, opened 2021-07-27
The returned string is "10WWatt". It should be "10W" or "10Watts"


MaxJPRey:
Hi @maxcapodi78 and @PipKat 
I saw a small issue in the following code.
https://github.com/pyansys/PyAEDT/blob/7db32e4ad317a7ed94b363bf7ab454484b01c105/pyaedt/hfss.py#L2099-L2100

We can expect to get either:
`Setting up power to Eigenmode 10W`
or
`Setting up power to Eigenmode 10Watts`

To fix it, either we update the keyword argument `powerin `or we update the line:
`self._messenger.add_info_message("Setting up power to Eigenmode " + powerin + "Watt")`
to be` self._messenger.add_info_message("Setting up power to Eigenmode " + powerin)`

Please tell me what is the more appropriate and I will fix it accordingly.

PR #411, opened 2021-08-30
Do not repeat the name of power unit.

MaxJPRey:
Fix issue #349 .
The string returned when editing the source was repeating the power unit.

### DIFF
--- a/pyaedt/hfss.py
+++ b/pyaedt/hfss.py
@@ -2373,12 +2373,12 @@ class Hfss(FieldAnalysis3D, object):
         >>> wave_port = hfss.create_wave_port_from_sheet(sheet, 5, hfss.AxisDir.XNeg, 40,
         ...                                              2, "SheetWavePort", True)
         >>> hfss.edit_source("SheetWavePort" + ":1", "10W")
-        pyaedt Info: Setting up power to Eigenmode 10WWatt
+        pyaedt Info: Setting up power to Eigenmode 10W
         True
 
         """
 
-        self._messenger.add_info_message("Setting up power to Eigenmode " + powerin + "Watt")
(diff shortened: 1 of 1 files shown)


maxcapodi78 (approved):

pyaedt/hfss.py
```diff
@@ -2373,12 +2373,12 @@ def edit_source(self, portandmode, powerin, phase="0deg"):
...
         ...                                              2, "SheetWavePort", True)
         >>> hfss.edit_source("SheetWavePort" + ":1", "10W")
-        pyaedt Info: Setting up power to Eigenmode 10WWatt
+        pyaedt Info: Setting up power to Eigenmode 10W
         True
 
         """
 
-        self._messenger.add_info_message("Setting up power to Eigenmode " + powerin + "Watt")
+        self._messenger.add_info_message("Setting up power to Eigenmode " + powerin)
```
PipKat, line 2381:
```suggestion
        self._messenger.add_info_message("Setting up power to Eigenmode " + power in.)
```

PipKat, line 2381:
```suggestion
        self._messenger.add_info_message("Setting up power to Eigenmode " + ``powerin``.)
```

PipKat: @MaxJPRey I'm not sure if the literal tagging is acceptable within a message, but if so, it seems more readable, especially when concluding the sentence with a period.

PipKat (approved):

pyaedt/hfss.py
maxcapodi78: as part of a string passed to an argument we should not use ``. Original command from @MaxJPRey was correct

Issue #382, opened 2021-08-16
Add syntax to remove variables


dcrawforAtAnsys:
Please make it possible to remove a design or project variable using the following syntax:
```py
>>> from pyaedt import Hfss
>>> app = Hfss()
...
>>> app['var'] = None
```
This would remove 'var1' if it is not used.

akaszynski:
Would it be better to use:
```py
>>> from pyaedt import Hfss
>>> app = Hfss()
...
>>> del app['var']
```

This way it's more like a `dict`.

dcrawforAtAnsys:
Yes. This is better.

akaszynski:
@maxcapodi78, @MaxJPRey 

Let's implement the following `__delitem__`:
```py
def __del__(self, key):
    """Implement del with array name or index."""
    del self.variable_manager(key)
```

We should also consider implementing a nested interface via a `variables` property:

```
    @property
    def variables(self):
        """Variables.."""
        self.variable_manager
```

While "flat is better than nested", in this case I think this is too flat.  Application should act as an "application", not as a variables dictionary as well.  This is something we'd deprecated in the next minor release.

PR #417, opened 2021-08-31
Add destructors.

MaxJPRey:
Fix issue #382.
Add a destructor for the application/design's variables.


pyaedt/application/Design.py
```diff
@@ -516,6 +516,11 @@ def __init__(self, design_type, project_name=None, design_name=None, solution_ty
         self.project_datasets = self._get_project_datasets()
         self.design_datasets = self._get_design_datasets()
 
+        def __del__(self, key):
+            """Implement destructor with array name or index."""
+            object_to_delete = self._variable_manager(key)
+            del object_to_delete
```
akaszynski, line 522: This should be simplified with
```suggestion
            del self._variable_manager[key]
```

akaszynski:
This also needs a test to verify it works and to ensure we maintain code coverage.

pyaedt/application/Design.py
maxcapodi78: I agree. we should follow @akaszynski suggestion

akaszynski (approved):
Good work.  This is ready to merge.

PR #419, opened 2021-08-31
Format top level modules with black

jleonatti:
Begin to format code using black, to enforce line length of 120.

This PR formats the top-level pyaedt modules.


akaszynski:
Since this is automated, I'd prefer getting all existing PRs cleaned out (including #420) so we don't end up with merge conflicts.

akaszynski:
Will have to merge with main and then redo:

Also, there's this error:
```
    @property
    def composed_name(self):
        """Composed names."""
        if self.id:
            return self.name + ";" +str(self.id) + ";" + str(self.schematic_id)
        else:
>           return self.name + ";" + str(self.schematic_id)
E           TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```
We need to deal with the None return for `self.name`.  I feel like we should consider changing the default value for name to be an empty string to avoid this.

jleonatti:
I have no idea where the unittest errors are coming from and I cannot run the tests locally, because AEDT on my machine is corrupt somehow. I am talking about EBU support about it. I am still investigating these unittest failures.

jleonatti:
@maxcapodi78 any ideas why this unit test is failing https://github.com/pyansys/PyAEDT/runs/3499206386?check_suite_focus=true?

```
   def test_09_import_netlist(self):
        self.aedtapp.insert_design("SchematicImport")
>       assert self.aedtapp.create_schematic_from_netlist(
            os.path.join(self.local_scratch.path, netlist1))

_unittest\test_Circuit.py:105:
```
```

    @aedt_exception_handler
    def get_pins(self, partid):
        """Retrieve one or more pins.
    
        Parameters
        ----------
        partid : int or str
            One or more IDs or names for the pins to retrieve.
    
        Returns
        -------
        type
            Pin with properties.
    
        """
        if type(partid) is str:
            pins = retry_ntimes(10, self.oeditor.GetComponentPins, partid)
            #pins = self.oeditor.GetComponentPins(partid)
        else:
            pins = retry_ntimes(10, self.oeditor.GetComponentPins, self.components[partid].composed_name)
            #pins = self.oeditor.GetComponentPins(self.components[partid].composed_name)
>       return list(pins)
E       TypeError: 'bool' object is not iterable

pyaedt\modeler\PrimitivesCircuit.py:755: TypeError
```

maxcapodi78:
not sure where the error comes from. my PRs done today work well. I noticed also that Ironphyton UT produces some additional failures (fails 8 UT Vs expected 1-3)

maxcapodi78:
please don't merge this PR until we figure out what is wrong

PR #420, opened 2021-08-31
Use numpydoc

akaszynski:
### Use numpydoc

This PR switches our docstring build to use `numpydoc` rather than `sphinx.ext.napoleon`.  This decision was made based on a variety of factors:

- Eventual inclusion of the `pyvista-plot` directive into `numpydoc`.  It's easier to do with `numpydoc` and it's more likely that the PR will be merged.
- Better parameters hyperlinks with `numpydoc_xref_param_type`.
- Includes `matplotlib` plots in examples.  We can add those in now.
- Can enable `numpydoc_validation_checks`.  This will be really useful for encouraging good documentation beyond the basic checks we have right now.

#### Comparison (`numpydoc` (new) left,  `sphinx.ext.napoleon` (old) right)
sc

As you can see, `numpydoc` is doing a much better job picking up on class links, even though we didn't specify the `:class:` directive for these.  Furthermore, the docs are more spread out (which I personally prefer) versus the less compressed in `sphinx.ext.napoleon`.

#### Validation

The following validation checks have been enabled to help us automate the PR approval process:

- GL06: Found unknown section
- GL07: Sections are in the wrong order.
- GL08: The object does not have a docstring
- GL09: Deprecation warning should precede extended summary
- GL10: reST directives {directives} must be followed by two colons

- SS01: No summary found
- SS02: Summary does not start with a capital letter
- SS03: Summary does not end with a period
- SS04: Summary contains heading whitespaces
- SS05: Summary must start with infinitive verb, not third person

- RT02: The first line of the Returns section should contain only the type, unless multiple values are being returned"

There are several more that should be on the roadmap for enabling.  See [numpydoc validation](https://numpydoc.readthedocs.io/en/latest/validation.html) for the full list, but the next ones should probably be the parameters section including PR01- PR10.


akaszynski:
@jleonatti, I'm getting failures here and on main.  Any idea what happened in regards to flake8?

jleonatti (approved):

PR #433, opened 2021-09-07
Format pyaedt files with black

jleonatti:
New PR because of strange CI behavior. Same work as #419.


jleonatti:
Marking PR as draft until I can debug the failing unittest locally.

akaszynski (approved):

PR #431, opened 2021-09-07
Multi part component

maxcapodi78:
Added Multipart Component Concept. User can add multiple 3d Components linked together and with motion associated.

PR #435, opened 2021-09-07
Format edb core files

jleonatti:
Run black on all files under edb_core folder to address line length issues.


akaszynski:
Wait until #431 is merged before proceeding with this.

akaszynski (approved):

PR #436, opened 2021-09-07
Format modules files

jleonatti:
Run black on all files under modules and generic folders to address line length issues.


akaszynski (approved):

PR #438, opened 2021-09-08
Update Contributing.rst

PipKat:
I know that this info will be part of the About project--but Alina asked me a formatting question and in my investigation, I found two format errors n this file that I fixed.


jleonatti (approved):

akaszynski (approved):
LGTM.

PR #447, opened 2021-09-10
Organize imports for unittests

jleonatti:
The last directive to add for flake8 configuration is to catch unused module imports.

This is one of several PRs to organize module imports before adding this check to flake8. This PR addresses imports in the unittest files.


_unittest/test_02_2D_modeler.py
```diff
@@ -1,15 +1,11 @@
 # standard imports
 import math
 
-try:
-    import pytest
-except ImportError:
-    import _unittest_ironpython.conf_unittest as pytest
```
akaszynski, line 7: I think we need this for ironpython compatibility.

jleonatti: Good point. Because of this, checking unused module imports with flake8 can't be applied to the unit test files, so I will probably just abandon this PR.

PR #448, opened 2021-09-10
Organize module imports

jleonatti:
Organize module imports and remove unused imports with the hopes of adding this check to the flake8 config once all refactoring is complete.


akaszynski:
These seem to be sorted to. Is that the case?

jleonatti:
> These seem to be sorted to. Is that the case?

Yes, I organized them too while I was at it, but this isn't enforced by flake8.

akaszynski (approved):
Thanks for going though this.

akaszynski:
Merging to give europe and east coast a fresh start for Monday.

PR #449, opened 2021-09-10
Apply flake8 policy for unused imports

jleonatti:
Final refactor to unit test files to conform to unused module import policy. In the case when modules absolutely need to be imported but unused, the comment ` # noqa: F401 ` can be added at the end of that line to make it exempt from the policy. This is used in several places due to the need to import pytest in some cases for Iron Python.

The flake8 configuration is also updated to have this policy checked in the CI.

The style check CI for this PR will not pass until #448 is merged.

### DIFF
--- a/examples/01-advanced-postprocessing/Touchstone_Management.py
+++ b/examples/01-advanced-postprocessing/Touchstone_Management.py
@@ -9,11 +9,9 @@ must be installed on the machine.
 This example runs only on Windows using CPython.
 """
 
-import sys
 import os
 import pathlib
-import shutil
-import time
+import sys
 
 local_path = os.path.abspath("")
 module_path = pathlib.Path(local_path)
@@ -30,9 +28,9 @@ example_path = examples.download_touchstone()
 
 ###############################################################################
 
-from pyaedt.generic.TouchstoneParser import *
-import numpy as np
 import matplotlib.pyplot as plt
+import numpy as np
+from pyaedt.generic.TouchstoneParser import *
 
 ###############################################################################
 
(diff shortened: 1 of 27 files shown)


akaszynski:
I think the CPython failure was due to instability, but I can't tell why IronPython failed...

examples/01-advanced-postprocessing/Touchstone_Management.py
```diff
@@ -30,9 +28,9 @@
 
 ###############################################################################
 
-from pyaedt.generic.TouchstoneParser import *
-import numpy as np
 import matplotlib.pyplot as plt
+import numpy as np
+from pyaedt.generic.TouchstoneParser import *
```
akaszynski, line 33: Would prefer to avoid `*` imports.  @jleonatti, is there a flake8 standard for that?  I'd like to enforce it.

akaszynski: > I'd like to enforce it.

Just not in this PR.  We can do it in a follow-up PR.

akaszynski (approved):
Approved with no suggestions.  See my comment regarding "star imports":
https://www.geeksforgeeks.org/why-import-star-in-python-is-a-bad-idea/

Consider follow-up PR.

examples/01-advanced-postprocessing/Touchstone_Management.py
jleonatti: @akaszynski yes it would directive F403. I can add then in a follow-up PR.

Issue #334, opened 2021-07-23
Possibly redundant methods in Q2d.py


PipKat:
The Q2d.py file has two methods with the same descriptions:
analyse_from_initial_mesh
analyse_nominal

Only analyse_nominal appears in Q3d.py. 
Perhaps the descriptions are wrong in Q2d.py?

akaszynski:
Pinging @maxcapodi78

maxcapodi78:
The two methods has different actions. 
I noticed that the analyse_nominal as wrong description. 
It has not to be "Revert the solution to the initial mesh and re-run the solve." 
but it has to be "Solves nominal design."

PR #454, opened 2021-09-13
Update Analysis.py

PipKat:
Back in July, Massimo indicated in #334 that the description for analyse nominal/analyze_nominal was wrong and asked me to correct it to "Solve the nominal design". This method is described in Analysis.py. Somehow, I missed following up on his request.

Also changing tagging to double ticks for class name since I'm editing this file anyway.


akaszynski (approved):
LGTM.

PR #453, opened 2021-09-13
Desktop Refactoring and Multipart Improvements

maxcapodi78:

PR #456, opened 2021-09-13
Move doctest to nightly build

jleonatti:
Remove doctest run from unit test CI to the nightly doc build.

### DIFF
--- a/.github/workflows/nightly-docs.yml
+++ b/.github/workflows/nightly-docs.yml
@@ -1,6 +1,7 @@
 name: Nightly Documentation Build
 
 on:
+  workflow_dispatch:
   schedule:  # UTC at 0400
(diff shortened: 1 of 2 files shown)


akaszynski:
I moved the doc testing into its own job and briefly ran by triggering it.
https://github.com/pyansys/PyAEDT/actions/runs/1232269764

In the future, we can run these manually as I've added the [Manual Trigger](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/) using `workflow_dispatch`.

akaszynski (approved):
I'll merge this as it's needed for #453.  I'm in favor of adding this back in as a non-required action when we fix logging, but as its own job.

PR #459, opened 2021-09-14
Change the name of some variables.

MaxJPRey:
Change the name of some variables.
Set the value for the "i" in the while loop of the `force_close_desktop()` function.

### DIFF
--- a/pyaedt/desktop.py
+++ b/pyaedt/desktop.py
@@ -363,9 +363,9 @@ class Desktop:
     @property
     def current_version_student(self):
         """Current student version of AEDT."""
-        for el in self.version_keys:
-            if "SV" in el:
-                return el
+        for version_key in self.version_keys:
+            if "SV" in version_key:
+                return version_key
         return None
 
     def _init_desktop(self):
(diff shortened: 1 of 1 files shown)


pyaedt/desktop.py
```diff
@@ -589,16 +591,16 @@ def _exception(self, ex_value, tb_data):
...
+                design_name = oproject.GetActiveDesign().GetName()
+                if ";" in design_name:
+                    design_name = design_name.split(";")[1]
             except:
-                des_name = ""
+                design_name = ""
         except:
-            proj_name = ""
-            des_name = ""
+            project_name = ""
```
MaxJPRey, line 602: @maxcapodi78 This variable is assigned but never used. We could get rid of it.

pyaedt/desktop.py
```diff
@@ -589,16 +591,16 @@ def _exception(self, ex_value, tb_data):
         """
         try:
             oproject = self._main.oDesktop.GetActiveProject()
-            proj_name = oproject.GetName()
+            project_name = oproject.GetName()
```
MaxJPRey, line 594: @maxcapodi78 This variable is assigned but never used. We could get rid of it.

maxcapodi78: I agree. that method can be refactored

pyaedt/desktop.py
maxcapodi78: I agree

akaszynski (approved):
LGTM.

PR #465, opened 2021-09-16
Added Registry Editor Methods to Design and Desktop

maxcapodi78:


pyaedt/desktop.py
```diff
@@ -691,6 +624,113 @@ def disable_autosave(self):
         """
         self._main.oDesktop.EnableAutoSave(False)
 
+    def change_license_type(self, license_type="Pool"):
+        """Change the License Type between ``Pack`` and ``Pool``.
+
+        ..note: The command returns True even if the Key is wrong due to API limitation.
```
akaszynski, line 630: This won't get rendered unless it follows this format
```suggestion
        .. note::
           The command returns True even if the Key is wrong due to API limitation.
```

pyaedt/desktop.py
```diff
@@ -691,6 +624,113 @@ def disable_autosave(self):
...
+
+        Parameters
+        ----------
+        license_type : str, optional
+            Set License Type between ``Pack`` and ``Pool``.
```
akaszynski, line 635:
```suggestion
            Set license type.  Must be either ``"Pack"`` or ``"Pool"``.
```

pyaedt/desktop.py
```diff
@@ -691,6 +624,113 @@ def disable_autosave(self):
...
+
+        Returns
+        -------
+        bool
```
akaszynski, line 639:
```suggestion
        bool
           ``True`` when successful, ``False`` when failed.
```

pyaedt/desktop.py
```diff
@@ -691,6 +624,113 @@ def disable_autosave(self):
...
+        """
+        try:
+            self._main.oDesktop.SetRegistryString("Desktop/Settings/ProjectOptions/HPCLicenseType", license_type)
+            return True
+        except:
```
akaszynski, line 644: We should log failure here.

maxcapodi78: the API doesn't return always error unfortunately

PR #471, opened 2021-09-20
Stackup material name and dataset fixes

Alberto-DM:
- Set the stackup material names to lowercase. It is necessary to match the same materials set in HFSS.
- Fixed sorting of dataset points. It was sorted as string rather than as float.
- Docstrings updated.


maxcapodi78 (approved):
LGTM

PR #476, opened 2021-09-20
Q2d methods

ring630:
create_rectangle()
asisgn_single_signal_line

PR #483, opened 2021-09-22
Ipc 2581

maxcapodi78:


akaszynski:
Failures here are from #476.  Branch protections for our CPython unit testing are now enabled so this doesn't happen again.

PR #484, opened 2021-09-22
Added method delete_setup

Alberto-DM:
Added method delete_setup for hfss and hfss3dlayout.


akaszynski:
I'm seeing this failure on other builds:
```
================================== FAILURES ===================================
_____________________ TestClass.test_02a_create_rectangle _____________________

self = <_unittest.test_30_Q2D.TestClass object at 0x00000277054B9C70>

    def test_02a_create_rectangle(self):
>       o = self.aedtapp.create_rectangle((0, 0), [5, 3], name="Rectangle1")

_unittest\test_30_Q2D.py:36: 
```

This can be traced to #476, which was merged despite failure.  I'm turning on branch protection for our build step so this doesn't happen again.

pyaedt/application/Analysis.py
```diff
@@ -748,6 +748,36 @@ def create_setup(self, setupname="MySetupAuto", setuptype=None, props={}):
...
+
+        >>> import pyaedt
+        >>> hfss = pyaedt.Hfss()
+        >>> setup1 = hfss.create_setup(setupname='Setup1')
+        >>> hfss.delete_setup(setupname='Setup1')
+        ...
+        pyaedt Info: Sweep was deleted correctly.
+        """
+        if setupname in self.existing_analysis_setups:
+            self.oanalysis.DeleteSetups([setupname])
```
maxcapodi78, line 777: I think you need to remove the setup from the list of self.setups

maxcapodi78 (commented):
Please check if we need to update the list of setups after deleting it

maxcapodi78 (approved):
Please check the app.setups list. you should probably delete the object once the boundary is deleted.

pyaedt/application/Analysis.py
Alberto-DM: done

PR #487, opened 2021-09-23
Improve the comment to be accurate regarding the code behavior.

MaxJPRey:
The comments of the 3 logging methods suggest that the messages can only be logged at the design level, whereas it can also be done at either the 'global' or 'project' level.
'Design' is the default level and not the only valid level.


akaszynski (approved):
LGTM.

PR #499, opened 2021-09-30
q2d enhancement

ring630:


pyaedt/q3d.py
```diff
@@ -581,16 +581,19 @@ def create_rectangle(self, position, dimension_list, name="", matname=""):
         return self.modeler.primitives.create_rectangle(position, dimension_list=dimension_list, name=name,
                                                         matname=matname)
 
-    def assign_single_signal_line(self, target_objects, name="", solve_option="SolveInside", thickness=None, unit="um"):
+    def assign_single_conductor(self, target_objects, name="", conductor_type="SignalLine", solve_option="SolveInside",
```
akaszynski, line 584: Since we're changing the name, we need to add in a deprecation warning and forward the call over.

See: https://sphinxdocs.pyansys.com/coding_style.html?highlight=deprication

ring630:
@akaszynski if everything is ok, could you approve?

PR #502, opened 2021-10-01
export design image

koubaa:
Add API to export design image as jpg.  This could be used by a jupyter notebook


maxcapodi78 (approved):
LGTM. thanks

PR #508, opened 2021-10-07
improve ironpython unit testing

koubaa:
add command line arg for test filter
use env var in case AEDT is installed in a non-default location

### DIFF
--- a/_unittest_ironpython/run_unittests.py
+++ b/_unittest_ironpython/run_unittests.py
@@ -47,4 +47,3 @@ if is_ironpython and "oDesktop" in dir(sys.modules["__main__"]):
             os.kill(pid, 9)
         except:
             successfully_closed = False
-


_unittest_ironpython/run_unittests.py
```diff
@@ -24,20 +29,22 @@ def discover_and_run(start_dir, pattern=None):
...
     f.write("ok")
+
 if is_ironpython and "oDesktop" in dir(sys.modules["__main__"]):
     pid = sys.modules["__main__"].oDesktop.GetProcessID()
     if pid > 0:
         try:
             os.kill(pid, 9)
         except:
             successfully_closed = False
+
```
akaszynski, line 50:
```suggestion
```

_unittest_ironpython/run_unittests.py
```diff
@@ -10,8 +11,12 @@
 
 os.environ["UNITTEST_CURRENT_TEST"] = "1"
 run_dir = os.path.abspath(os.path.dirname(__file__))
-print(run_dir)
 
+args_env = os.environ.get("RUN_UNITTESTS_ARGS", "")
```
MaxJPRey, line 15: @koubaa We might not need to create a variable here if it is only used one time in the code below. You can use it directly line 18.

PR #512, opened 2021-10-18
Doc/logger

MaxJPRey:
Add the logger and the handler classes to the documentation.

### DIFF
--- a/.github/workflows/ironpython_linux.yml
+++ b/.github/workflows/ironpython_linux.yml
@@ -29,6 +29,7 @@ jobs:
 
       - name: 'Run Unit Tests in Ironpython'
         run: |
+          export DISPLAY="127.0.0.1:1"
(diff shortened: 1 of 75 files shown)


doc/source/API/LogHandler.rst
```diff
@@ -0,0 +1,17 @@
...
+
+.. currentmodule:: pyaedt
+
+.. autosummary::
+   :toctree: _autosummary
+   :template: custom-class-template.rst
+   :nosignatures:
+
+.. autoclass:: pyaedt.log_handler.LogHandler
+   :inherited-members: Handler
```
MaxJPRey, line 15: @akaszynski I am not sure we can have both `autoclass `and `autosummary `in the same rst. At some point, I got some conflicts for the class and I was expected it, but then I did not get any errors all of a sudden.

PR #519, opened 2021-10-26
IronPython job shows passing regardless of if it actually passes or fails in pytest

### LINKED ISSUE #513: IronPython CI fails but still shows passing
IronPython job shows passing regardless of if it actually passes or fails in pytest.

MaxJPRey:
Fix #513 
IronPython job shows passing regardless of if it actually passes or fails in pytest.


maxcapodi78 (approved):
LGTM

PR #528, opened 2021-11-01
Antenna example

svandenb-dev:


pyaedt/edb_core/components.py
```diff
@@ -386,7 +386,7 @@ def get_components_from_nets(self, netlist=None):
         return cmp_list
 
     @aedt_exception_handler
-    def create_component_from_pins(self, pins, component_name):
+    def create_component_from_pins(self, pins, component_name, partname, placement_layer=None):
```
akaszynski, line 389: Please document these new parameters.

svandenb-dev: fixed. Partname removed as not used anymore. Added placement layer parameter doc.

PR #533, opened 2021-11-03
added get variation strings

beliaev-maksim:


pyaedt/application/Analysis.py
```diff
@@ -480,6 +480,28 @@ def variations(self, setup_sweep=None):
...
+            -------
+            list of str
+                List of variation families.
+
+            """
+            if not setup_sweep:
+                setup_sweep = self._app.existing_analysis_sweeps[0]
+            vs = self._app.osolution.GetAvailableVariations(setup_sweep)
+
+            return vs
```
akaszynski, line 503:
```suggestion
            return self._app.osolution.GetAvailableVariations(setup_sweep)
```

pyaedt/application/Analysis.py
```diff
@@ -480,6 +480,28 @@ def variations(self, setup_sweep=None):
                 families.append(family)
             return families
 
+        @aedt_exception_handler
+        def get_variation_strings(self, setup_sweep=None):
+            """
+            Variations.
```
akaszynski, line 486:
```suggestion
            """Return variation strings.
```

pyaedt/application/Analysis.py
beliaev-maksim: @akaszynski 
thanks!

maxcapodi78 (approved):
LGTM. I'll add UT to it

beliaev-maksim:
@maxcapodi78 
thanks!

Issue #535, opened 2021-11-04
Warning because no title is part of the example 03_5G_antenna_example


MaxJPRey:
When we generate the documentation locally a warning appears for the example: 'examples/00-basic-examples/03_5G_antenna_example'

_`D:\GitHub\pyaedt\doc\source\examples\index.rst:107: WARNING: toctree contains reference to document 'examples/00-basic-examples/03_5G_antenna_example' that doesn't have a title: no link will be generated`_

PR #536, opened 2021-11-04
Add a title '5G linear array antenna'.

MaxJPRey:
Fix issue #535.
When the doc is generated, a warning appears for the example 03_5G_antenna_example.
It is because the example does not have a title in its docstring.


maxcapodi78 (approved):
LGTM

PR #540, opened 2021-11-04
Update the disconnect on linux. the release_desktop is not needed.

maxcapodi78:
Updated the max chunk size to 256MB

### DIFF
--- a/pyaedt/common_rpc.py
+++ b/pyaedt/common_rpc.py
@@ -13,6 +13,7 @@ from rpyc.utils.server import ThreadedServer
 from pyaedt.rpc.rpyc_services import GlobalService
 import rpyc.core.consts
 
+# Maximum Stream message size. Set to 256MB
 rpyc.core.consts.STREAM_CHUNK = 256000000
 
 def launch_server(port=18000, ansysem_path=None, non_graphical=False):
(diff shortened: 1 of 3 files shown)


pyaedt/rpc/rpyc_services.py
```diff
@@ -42,10 +42,11 @@ def on_disconnect(self, connection):
...
         if self.app:
-            try:
-                self.app[0].release_desktop()
-            except:
-                pass
+            if not os.name == "posix":
+                try:
+                    self.app[0].release_desktop()
+                except:
+                    pass
```
MaxJPRey, line 49: Maybe we can raise an exception here to say that the desktop was not released properly.
`raise Exception("The desktop was not released properly.") `

pyaedt/common_rpc.py
```diff
@@ -11,7 +11,9 @@
 import rpyc
 from rpyc.utils.server import ThreadedServer
 from pyaedt.rpc.rpyc_services import GlobalService
+import rpyc.core.consts
 
+rpyc.core.consts.STREAM_CHUNK = 256000000
```
MaxJPRey, line 16: Maybe just add a small comment to explain why we must set STREAM_CHUNK to 256000000.

maxcapodi78: Done. Thanks

pyaedt/rpc/rpyc_services.py
maxcapodi78: I've removed the try except since should be handled by the release destkop

PR #542, opened 2021-11-04
Logger removal legacy messenger

maxcapodi78:


MaxJPRey (approved):

pyaedt/common_rpc.py
```diff
@@ -11,7 +11,10 @@
 import rpyc
 from rpyc.utils.server import ThreadedServer
 from pyaedt.rpc.rpyc_services import GlobalService
+import rpyc.core.consts
 
+# Maximum Stream message size. Set to 256MB
+rpyc.core.consts.STREAM_CHUNK = 256000000
```
MaxJPRey, line 17: @maxcapodi78 I thought this was already modified into #540 . and should be into main already.
Why is it modified in this PR?

maxcapodi78: I think because I've merged from main

PR #545, opened 2021-11-05
added method to export other report types

beliaev-maksim:


pyaedt/modules/PostProcessor.py
```diff
@@ -1201,8 +1201,8 @@ def steal_focus_oneditor(self):
         return True
 
     @aedt_exception_handler
-    def export_report_to_csv(self, project_dir, plot_name):
```
maxcapodi78, line 1204: Please add a DeprecationWarning and leave the old method. We cannot change a name without notifying it to customers

beliaev-maksim: @maxcapodi78 
old method is not removed

maxcapodi78 (approved):
LGTM

Issue #111, opened 2021-06-17
Methods not snake case


jleonatti:
https://github.com/pyansys/PyAEDT/blob/15aacb0c2becf53e04b833f78422dc5964d945eb/pyaedt/hfss.py#L217

There are several methods (assigncoating, assignmaterial, etc) throughout that don't follow the python conventions of snake case. I think it would be best to rename these methods to make it cleaner for the user.

PR #547, opened 2021-11-05
Fix/issue 111

MaxJPRey:
Fix issue #111 .

Deprecate some remaining methods that are not snake case.


maxcapodi78 (approved):
LGTM

akaszynski:
One note:
When deprecating methods/functions, best practice is to pass all args and kwargs with:

```python

def old_method(*args, **kwargs):
    new_method(*args, **kwargs)

```

MaxJPRey:
> One note: When deprecating methods/functions, best practice is to pass all args and kwargs with:
> 
> ```python
> def old_method(*args, **kwargs):
>     new_method(*args, **kwargs)
> ```

@akaszynski Makes sense. I will do it in a small PR. Thanks for reviewing the PR.

Issue #552, opened 2021-11-08
create_via() fails on hole_diam


fwlfGH:
Could it be that in Primitives3DLayout.py method create_via (line 423)
 `arg.append("hole diameter:="), arg.append([self.arg_with_dim("hole_diam")])`
should be:
 `arg.append("hole diameter:="), arg.append([self.arg_with_dim(hole_diam)])`
without quotes. I receive an error on unknown variable hole_diam.

PR #556, opened 2021-11-09
hole_diam cannot be a string to be passed to arg_with_dim.

MaxJPRey:
Fix the issue #552 .
The `hole_diam` variable must be passed to the `arg_with_dim()` and not a string "hole_diam".


maxcapodi78 (approved):
LGTM

PR #559, opened 2021-11-09
fix maxwell example

beliaev-maksim:


pyaedt/modeler/Primitives2D.py
```diff
@@ -148,23 +148,23 @@ def create_rectangle(self, position, dimension_list, is_covered=True, name=None,
...
         if self.plane2d == "Z":
-            Height = self._arg_with_dim(dimension_list[1])
-            Width = self._arg_with_dim(dimension_list[0])
+            height = self._arg_with_dim(dimension_list[0])
+            width = self._arg_with_dim(dimension_list[1])
         else:
-            Height = self._arg_with_dim(dimension_list[0])
-            Width = self._arg_with_dim(dimension_list[1])
+            height = self._arg_with_dim(dimension_list[0])
+            width = self._arg_with_dim(dimension_list[1])
```
MaxJPRey, line 158: @beliaev-maksim The code in this if/else condition seems the same. Is the if/else condition really necessary or shall we go with:
```
        if self.plane2d == "Z":
            height = self._arg_with_dim(dimension_list[1])
            width = self._arg_with_dim(dimension_list[0])
        else:
            height = self._arg_with_dim(dimension_list[0])
            width = self._arg_with_dim(dimension_list[1])
```

beliaev-maksim: good catch. For  XY planes my fix is definitely correct. I need to check RZ, may be in RZ it should be swapped

beliaev-maksim:
@MaxJPRey @maxcapodi78 
solution is following. Remove this swapping of `height` and `width`. Explicitly name the dimensions in doc strings, otherwise this is not obvious and leads to mistakes.

I added tests to cover both XY and RZ scenarios

_unittest/test_02_2D_modeler.py
```diff
@@ -14,8 +14,21 @@
...
 class TestClass(BasisTest):
-    def setup_class(self):
-        BasisTest.setup_class(self, project_name="test_primitives", design_name="2D_Primitives", application=Maxwell2d)
+    def setup_class(self, project_name=None, design_name=None, solution_type=None, application=None):
+        """
+        redefine not to mock empty call
+        """
+
+    def setup(self):
+        super().setup_class(project_name="test_primitives", design_name="2D_Primitives", application=Maxwell2d)
```
MaxJPRey: @maxcapodi78  Will the call to `super()` work with IronPython. FRom what I can remember in the past it used to break some ironpython tests. If yes, @beliaev-maksim you can use directly the base class name directly.

beliaev-maksim: @MaxJPRey 
change is outdated :)

MaxJPRey: Just saw that. Thanks @beliaev-maksim

beliaev-maksim: just FYI, you should not call `BaseClass.method`

in IPY you need to run `super(MyClass, self).method()`

MaxJPRey (approved):

PR #567, opened 2021-11-11
Set solder on component added

svandenb-dev:
I added this command as it was missing. Creating solder balls directly from component prior coax port definition.


maxcapodi78 (approved):
LGTM

PR #569, opened 2021-11-11
bullet

ring630:
Enhanced crreat_padstack to support bullet shape antipad

### DIFF
--- a/pyaedt/edb_core/padstack.py
+++ b/pyaedt/edb_core/padstack.py
@@ -262,8 +262,8 @@ class EdbPadstacks(object):
     @aedt_exception_handler
     def create_padstack(
             self, padstackname=None, holediam="300um", paddiam="400um", antipaddiam="600um",
-            startlayer=None, endlayer=None, antipadshape="Circle", xsize="600um", ysize="600um", cornerradius="300um",
-            offsetx="0.0", offsety="0.0", rotation="0.0"
+            startlayer=None, endlayer=None, antipad_shape="Circle", x_size="600um", y_size="600um", corner_radius="300um",
+            offset_x="0.0", offset_y="0.0", rotation="0.0"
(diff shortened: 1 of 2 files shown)


maxcapodi78:
@ring630 the PR looks good to me. Two comments:
1- it would be better to add parameters at the end of the list of existing parameters to maintain the back compatibility with existing customer code.
2- maybe should be helpful to add a unit test for bullet

MaxJPRey:
@ring630 I would just suggest to rename the parameters in order to follow the snake_case convention.

ring630:
> @ring630 the PR looks good to me. Two comments: 1- it would be better to add parameters at the end of the list of existing parameters to maintain the back compatibility with existing customer code. 2- maybe should be helpful to add a unit test for bullet

Done.

ring630:
> @ring630 I would just suggest to rename the new parameters (the one added in the signature of `create_padstack()`) in order to follow the snake_case convention.

Done

pyaedt/edb_core/padstack.py
```diff
@@ -281,7 +283,20 @@ def create_padstack(
...
+        y_size : str, optional
+            Only applicable to bullet shape. The default is ``"600um"``.
+        corner_radius :
+            Only applicable to bullet shape. The default is ``"300um"``.
+        offset_x : str, optional
+            X offset of antipad. The default is ``"0.0"``.
+        offset_y : str, optional
+            Y offset of antipad. The default is ``"0.0"``.
+        rotation : str, optional
+            rotation of antipad. The default is ``"0.0"``.
```
MaxJPRey, line 299:
```suggestion
            Rotation of antipad. The default is ``"0.0"``.
```

svandenb-dev (approved):
good for me

PR #573, opened 2021-11-12
Fix #428 .

### LINKED ISSUE #428: Duplicate GND pin groups are created when there are two positive pin groups on the same device
https://github.com/pyansys/PyAEDT/blob/8e7bcff6bc66a7da026957439d771aa2ef5da798/pyaedt/edb_core/siwave.py#L583

MaxJPRey:
Duplicate GND pin groups were created when there were two positive pin groups on the same device.
It was because the positive pins were assigned to the negative node.
Fix #428 .


maxcapodi78 (approved):
LGTM

PR #575, opened 2021-11-15
Bug fix clip design

svandenb-dev:
@maxcapodi78 bug fixed at edb cutout method to prevent crash when output aedb path was empty.


maxcapodi78 (approved):
LGTM

PR #576, opened 2021-11-15
ANSYSEM path and version.

MaxJPRey:
Refactor the way we get the ANSYSEM path and version.
It will avoid code duplication which is error prone.


maxcapodi78 (approved):
LGTM

PR #578, opened 2021-11-15
Remove the PDSHole class duplicate and use the outter one.

### LINKED ISSUE #574: modeler/Object3d.py: Padstack - PDSHole Duplicate
Hi I'm not sure if that's desired, but:
PDSHole class is defined twice, looks like copy/paste to me:

- Once in line 1531 
- and again inside of PDSLayer in line 1615

please check if that is correct, thanks.

MaxJPRey:
The inner class `PDSHole` in the outer class `PDSLayer `is not necessary anymore. We can use the `PDSHole `class in `Padstack `directly.

Fix #574.

### DIFF
--- a/pyaedt/modeler/Object3d.py
+++ b/pyaedt/modeler/Object3d.py
@@ -1586,7 +1586,7 @@ class Padstack(object):
             if value:
                 self._pad = value
             else:
-                self._pad = self.PDSHole(holetype="None", sizes=[])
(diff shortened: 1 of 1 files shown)


maxcapodi78 (approved):
LGTM. Maybe in the future we can make PDSHole indipendent from any Class

MaxJPRey:
> LGTM. Maybe in the future we can make PDSHole indipendent from any Class

@maxcapodi78 Agreed. I think that is something we could do.

PR #598, opened 2021-11-21
Disable the log handlers for the stdout and output file.

### LINKED ISSUE #587: Logger Improvements
Here a list of Logger improvements:
- if another logger exists, no new handlers have to be created and the existing logger has to be used
- there should be methods for enabling/disabling stdout logging. Actually Env Variable PYAEDT_SCREEN_LOGS is used
- there should be methods for enabling/disabling logging. Actually Env Variable PYAEDT_FILE_LOGS is used
- there should be methods for enabling/disabling Desktop logging. Actually Env Variable PYAEDT_DESKTOP_LOGS is used

MaxJPRey:
Disable the log handlers for the stdout and output file.
Add some unit tests to make sure the different log handlers are disable/re-enable properly.

Fix #587.


akaszynski:
FYI, since we can now mark [draft requests](https://github.blog/2019-02-14-introducing-draft-pull-requests/) for private and public PRs due to our upgrade to the enterprise account, we can now stop using `[WIP]`.

MaxJPRey:
> FYI, since we can now mark [draft requests](https://github.blog/2019-02-14-introducing-draft-pull-requests/) for private and public PRs due to our upgrade to the enterprise account, we can now stop using `[WIP]`.

@akaszynski Ok,  was tagging with both "draft" and "WIP". I will just use the the draft feature from now on.

_unittest/test_14_AedtLogger.py
```diff
@@ -157,12 +158,94 @@ def test_03_stdout_with_app_filter(self):
...
-        # for handler in logger.handlers:
-        #     handler.close()
-        # for handler in project_logger.handlers:
-        #     handler.close()
-        # for handler in design_logger.handlers:
-        #     handler.close()
+    def test_04_disable_output_file_handler(self):
+        content = None
+        temp_dir = tempfile.gettempdir()
+        path = os.path.join(temp_dir, "test04.txt")
```
beliaev-maksim, line 164: where is this file removed?

have you considered [NamedTemporaryFile](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) ?

MaxJPRey: I have to use `gettempdir` because otherwise with `NamedTemporaryFile`, I cannot read the file latter on because it is deleted as soon as I closed it.
The temp dir and all its content will be deleted at the end of the test to be sure that no handlers is preventing from this operation. Thanks for the reminder.

maxcapodi78 (approved):
LGTM

PR #602, opened 2021-11-23
flip stackup layers

svandenb-dev:


pyaedt/edb_core/stackup.py
```diff
@@ -218,6 +219,47 @@ def create_debye_material(
...
+            if not 'RadBox' in lyr.GetName():
+                upper_elevation = lyr.GetUpperElevation() * 1.0e6
+                updated_lower_el = max_elevation - upper_elevation
+                val = self._edb_value("{}um".format(updated_lower_el))
+                cloned_layer.SetLowerElevation(val)
+                new_lc.AddStackupLayerAtElevation(cloned_layer)
+
+        layer_list = general.convert_py_list_to_net_list(non_stackup_layers)
+        new_lc.AddLayers(layer_list)
+        if self._active_layout.SetLayerCollection(new_lc):
```
Alberto-DM, line 257: I have a question. 
You have excluded the RadBox from the flipping algorithm, but then it is not added back to the ``new_lc`` layer.
Is that correct?

_unittest/test_00_EDB.py
```diff
@@ -520,3 +520,7 @@ def test_67_add_void(self):
         void = self.edbapp.core_primitives.create_path(path, layer_name="TOP",
                                                        width="0.1mm")
         assert self.edbapp.core_primitives.add_void(plane, void)
+
+
+    def test_68_flip_layer_stackup(self):
+        assert self.edbapp.core_stackup.flip_stackup()
```
Alberto-DM, line 526:
```suggestion
        assert self.edbapp.core_stackup.flip_stackup()
        
```
New line at the end of the file is missing. Style check will complain :)

Alberto-DM (commented):
Question and minor style change required. 
Otherwise ok.

pyaedt/edb_core/stackup.py
```diff
@@ -218,6 +219,47 @@ def create_debye_material(
...
 
+    @aedt_exception_handler
+    def flip_stackup(self):
+        """Flip the current layer stackup of a layout.
+        Parameters
+        ----------
+        Returns
+        -------
+        type bool
+            True when layer stackup update success False if not.
```
MaxJPRey, line 230:
```suggestion
            ``True`` when layer stackup update success ``False`` if not.
```

MaxJPRey (approved):

pyaedt/edb_core/stackup.py
svandenb-dev: that's correct since we build a new stackup and override the original one

_unittest/test_00_EDB.py
svandenb-dev: thanks

PR #603, opened 2021-11-24
Circuit Component Refactoring

maxcapodi78:


pyaedt/modeler/Circuit.py
```diff
@@ -137,6 +160,20 @@ def primitives(self):
             self._primitivesDes = self._app.project_name + self._app.design_name
         return self._primitives
 
+    @property
+    def primitives(self):
+        """Primitives.
+
+        .. deprecated:: 0.4.10
+           Use :func:`Circuit.modeler.layout` instead.
```
MaxJPRey, line 168: @maxcapodi78 Was it deprecated in 0.4.10 or will it be done in the coming 0.4.11 release?

pyaedt/modeler/Object3d.py
```diff
@@ -1853,18 +1854,91 @@ def remove(self):
...
+        """
 
+        page_name = "{}_{}".format(self._circuit_comp.composed_name, self.name)
+        try:
+            x_loc = AEDT_UNITS["Length"][decompose_variable_value(self._circuit_comp.location[0])[1]] * float(
+                decompose_variable_value(self._circuit_comp.location[1])[0])
+        except:
+            x_loc = float(self._circuit_comp.location[0])
+        if self.position[0] < x_loc:
+            angle = 6.28318530717959
```
Alberto-DM, line 1894: Is it not the same as angle = 0? i.e. no rotation?

pyaedt/generic/constants.py
```diff
@@ -11,6 +11,7 @@
 METER2IN = 0.0254
 METER2MILES = 1609.344051499
 
+MILS2METER = 39370.078740157
```
MaxJPRey, line 14: Are you sure that MILS2METER = 39370.078740157
I thought that 1 mils = 2.54e-5 meters.

pyaedt/modeler/Object3d.py
```diff
@@ -17,7 +17,8 @@
 
 from pyaedt import aedt_exception_handler, _retry_ntimes
 from pyaedt.modeler.GeometryOperators import GeometryOperators
-
+from pyaedt.application.Variables import decompose_variable_value
+from pyaedt.generic.constants import AEDT_UNITS, MILS2METER
 clamp = lambda n, minn, maxn: max(min(maxn, n), minn)
```
MaxJPRey, line 22:
```suggestion
from pyaedt.generic.constants import AEDT_UNITS, MILS2METER

clamp = lambda n, minn, maxn: max(min(maxn, n), minn)
```

pyaedt/modeler/Object3d.py
```diff
@@ -1875,64 +1949,126 @@ def composed_name(self):
...
+        return self._location
+
+    @location.setter
+    def location(self, location_xy):
+        """Set the part location.
+
+        Parameters
+        ----------
+        location_xy : list
+            List of x and y coordinates. If float is provided, ``mils`` will be used
```
MaxJPRey, line 2028:
```suggestion
            List of x and y coordinates. If float is provided, ``mils`` will be used.
```

pyaedt/generic/constants.py
maxcapodi78: it's the inverse of it

MaxJPRey (approved):

pyaedt/modeler/Circuit.py
maxcapodi78: you are right. I'll fix it

pyaedt/modeler/PrimitivesNexxim.py
```diff
@@ -346,7 +346,7 @@ def create_field_model(self, design_name, solution_name, pin_names, model_type="
         return False
 
     @aedt_exception_handler
-    def create_resistor(self, compname=None, value=50, xpos=0, ypos=0, angle=0, use_instance_id_netlist=False):
+    def create_resistor(self, compname=None, value=50, location=None, angle=0, use_instance_id_netlist=False):
```
Alberto-DM, line 349:
```suggestion
    def create_resistor(self, compname=None, value=50, location=[], angle=0, use_instance_id_netlist=False):
```
If location is a list, it would be better so set to an empty list. It still returns ``False`` when checked.

pyaedt/modeler/Object3d.py
```diff
@@ -1853,18 +1854,91 @@ def remove(self):
...
+                                                                       posy=self.location[1], angle=angle)
+        try:
+            x_loc = AEDT_UNITS["Length"][decompose_variable_value(component_pin._circuit_comp.location[0])[1]] * float(
+                decompose_variable_value(component_pin._circuit_comp.location[0])[0])
+        except:
+            x_loc = float(self._circuit_comp.location[0])
+        if component_pin.position[0] < x_loc:
+            angle = 6.28318530717959
+        else:
+            angle = 3.14159265358979
```
Alberto-DM:
```suggestion
            angle = math.pi
```
Use of math.pi is better like in other cases.

Alberto-DM (changes requested):

examples/03-Circuit/EMIT_Example.py
```diff
@@ -37,15 +36,18 @@
 # Create 3 radios and connect an antenna to each.
 rad1 = aedtapp.modeler.components.create_component("UE - Handheld")
 ant1 = aedtapp.modeler.components.create_component("Antenna")
-ant1.move_and_connect_to(rad1)
+if rad1 and ant1:
```
myoung301, line 39: Under what circumstances will this be false? In that situation does it make sense to proceed with the example?

I think it would only happen when there is an error with create_component. If there are errors, I think it may be better to fail sooner rather than later.

Alberto-DM (approved):
LGTM. 
Great Refactoring 🥇 ! Much needed!

examples/03-Circuit/EMIT_Example.py
maxcapodi78: Hi @myoung301 , it is false when running in non-graphical. If you notice the latest version of the documentation has not been updated because the doc build runs in non-graphical. With this trick the documentation should be generated and the example should run on customer machine

PR #606, opened 2021-11-26
Fix/issue 496

### LINKED ISSUE #496: docs
it is not a bool, I would expect str with the path to interpreter. 
on top of this I would expect this to fail and be ignored in 2021R2+

https://github.com/pyansys/PyAEDT/blob/c1bf3bf6f3b68b9664ac4c65804ef431b55f8886/pyaedt/maxwell.py#L110

MaxJPRey:
Update the example, because other part of the code were refactored.
Maxwell2D template was renamed Maxwell2d.

Fix #496 .

### DIFF
--- a/pyaedt/generic/DataHandlers.py
+++ b/pyaedt/generic/DataHandlers.py
@@ -499,7 +499,7 @@ setup = os.getenv('setup')
 with Desktop() as d:
     maxwell_2d = Maxwell2d(designname=design_name, setup_name=setup)
     maxwell_2d.setup_ctrlprog(keep_modifications=True )
-    oDesktop.AddMessage( maxwell_2d.project_name, maxwell_2d.design_name, 0, "Successfully updated project definitions")
+    d.logger.info("Successfully updated project definitions")
     maxwell_2d.save_project()
 """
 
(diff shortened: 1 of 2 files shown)


pyaedt/generic/DataHandlers.py
```diff
@@ -491,16 +491,16 @@ def from_rkm_to_aedt(code):
...
 setup = os.getenv('setup')
 
 with Desktop() as d:
-    mxwl = Maxwell2D(designname=design_name, setup_name=setup)
-    mxwl.setup_ctrlprog(keep_modifications=True )
-    oDesktop.AddMessage( mxwl.project_name, mxwl.design_name, 0, "Successfully updated project definitions")
-    mxwl.save_project()
+    maxwell_2d = Maxwell2d(designname=design_name, setup_name=setup)
+    maxwell_2d.setup_ctrlprog(keep_modifications=True )
+    oDesktop.AddMessage( maxwell_2d.project_name, maxwell_2d.design_name, 0, "Successfully updated project definitions")
```
MaxJPRey, line 502: @maxcapodi78 Do you think that we should use the logger: `desktop.logger.info` here?

beliaev-maksim (approved):

pyaedt/generic/DataHandlers.py
maxcapodi78: yes

maxcapodi78 (approved):
LGTM

Issue #609, opened 2021-11-29
decompose_variable_value cannot decompose m2


boyang2022:
trying to decompose "m2" (meter square)

```python
from pyaedt.application.Variables import decompose_variable_value
values = ["3.123456m2"]
for value in values:
    print("*"*50)
    print("pyaedt:", decompose_variable_value(value))

```

output
pyaedt ('3.123456m2', 'm')

PR #616, opened 2021-12-02
Fix/issue 609

MaxJPRey:
Fix the issue #609 .
The variable decomposition was not working for units including both letters and digits such as 'm2'.

With the new modification we get the index of the first letter in the `variable `string and then we split it to get the unit and the value.

### DIFF
--- a/pyaedt/application/Variables.py
+++ b/pyaedt/application/Variables.py
@@ -269,22 +269,17 @@ def decompose_variable_value(variable_value, full_variables={}):
             float_value = float(variable_value)
         except ValueError:
             # search for a valid units string at the end of the variable_value
-            if full_variables=={}:
-                value = re.findall(r"[0-9.]+", variable_value)[0]
-                float_value = float(value)
-                units = variable_value.replace(value,"")
-            else:
-                units = _find_units_in_dependent_variables(variable_value, full_variables)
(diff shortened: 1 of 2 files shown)


pyaedt/application/Variables.py
```diff
@@ -267,18 +269,17 @@ def decompose_variable_value(variable_value, full_variables={}):
...
         except ValueError:
             # search for a valid units string at the end of the variable_value
-            loc = re.search("[a-z_A-Z]+$", variable_value)
+            loc = re.search("[a-z_A-Z]+", variable_value)
             units = _find_units_in_dependent_variables(variable_value, full_variables)
 
             if loc:
                 loc_units = loc.span()[0]
                 extract_units = variable_value[loc_units:]
-                if unit_system(extract_units):
```
MaxJPRey, line 276: @maxcapodi78  This `unit_system()` method is using the dictionary `AEDT_UNITS`. This dictionary does not contain units such as area m², volume m^3. Consequently it was not working for "3.123456m2" as described in issue #609.

maxcapodi78: got it. Do you think we should add it or your fix handles all conditions?

maxcapodi78 (approved):
Great job. Thanks also for detailed UT

Issue #620, opened 2021-12-03
Desktop initialization fails in german language machines


maxcapodi78:
Failure occur in _get_tasks_list_windows method in desktop.py file.

Error is:
"UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 75: character maps to <undefined>

PR #621, opened 2021-12-03
Fix/issue 620

MaxJPRey:
The output provided when we list the sub-processes running is facing some issues with encoding.
We must specify the encoding and ignore the errors as the only string that we are looking for is "ansysedtsv.exe" or "ansysedt.exe".

Fix issue #620 .


maxcapodi78 (approved):
Looks great

Issue #623, opened 2021-12-03
Encoding is not working on build agent.


MaxJPRey:
When building the documentation on the build agent, an exception is raised when we try to get the ansys aedt already running on  the machine.

![image](https://user-images.githubusercontent.com/87315832/144639879-34422c26-2508-4c3f-8b3c-3c9a9ebb18c5.png)

PR #624, opened 2021-12-03
Set the encoding to 'utf-8'.

MaxJPRey:
Encoding must be set to utf-8 because the full documentation will fail on the build agent otherwise.

Fix issue #623 .


maxcapodi78 (approved):
LGTM

PR #635, opened 2021-12-09
Enhanchement/bh curves

maxcapodi78:


_unittest/test_07_Object3D.py
```diff
@@ -224,7 +224,7 @@ def test_09_to_boolean(self):
         assert not _to_boolean("no")
 
     def test_10_chamfer(self):
-        initial_object = self.prim.create_box([0, 0, 0], [10, 10, 5], "ChamferTest", "Copper")
+        initial_object = self.aedtapp.modeler.create_box([0, 0, 0], [10, 10, 5], "ChamferTest", "Copper")
```
beliaev-maksim, line 227: why do we need to intermix modeler with BH curve PR?

I think 2 separate PRs would be better

maxcapodi78: you are right. In fact there are 2 PR. but since I've created the new branch from the refactoring you are seeing all the modifications here. Sorry about that

beliaev-maksim: @maxcapodi78 
first, you should not do it and branch only from master
second, you will have merge conflicts due to squashing

I would reset branch and just cherry-pick your commits

beliaev-maksim: @maxcapodi78 
done

maxcapodi78: @beliaev-maksim you are right. I did by mistake and discovered only when I finished. I fixed it. As you can see now only files related to this PR are available and there are no conlicts

beliaev-maksim: @maxcapodi78 
I reverted your commits, that is why only these changes are here :)

PR #637, opened 2021-12-09
Added First Implementation of BH Curves. At this time Temperature dep…

maxcapodi78:
…endance is not implemented

### DIFF
--- a/pyaedt/generic/DataHandlers.py
+++ b/pyaedt/generic/DataHandlers.py
@@ -70,7 +70,15 @@ def _dict2arg(d, arg_out):
 
     """
     for k, v in d.items():
-        if isinstance(v, (OrderedDict, dict)):
+        if k == "Point":
+            if isinstance(v[0], (list, tuple)):
+                for e in v:
+                    arg = ["NAME:" + k, e[0], e[1]]
+                    arg_out.append(arg)
+            else:
+                arg = ["NAME:" + k, v[0], v[1]]
+                arg_out.append(arg)
+        elif isinstance(v, (OrderedDict, dict)):
             arg = ["NAME:" + k]
             _dict2arg(v, arg)
             arg_out.append(arg)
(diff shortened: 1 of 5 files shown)


maxcapodi78:
@beliaev-maksim  I've created a new branch and created a new PR

pyaedt/generic/DataHandlers.py
```diff
@@ -92,41 +100,46 @@ def _dict2arg(d, arg_out):
...
     elif arg[0][:5] == "NAME:":
         top_key = arg[0][5:]
         dict_in = OrderedDict()
         i = 1
         while i < len(arg):
-            if (type(arg[i]) is list or type(arg[i]) is tuple or str(type(arg[i])) == r"<type 'List'>") and arg[i][0][
-                :5
-            ] == "NAME:":
+            if (isinstance(arg[i], (list, tuple)) or str(type(arg[i])) == r"<type 'List'>") and arg[i][0][
+                                                                                                :5] == "NAME:":
```
beliaev-maksim, line 118:
```suggestion
            if arg[i][0][:5] == "NAME:" and (
                isinstance(arg[i], (list, tuple)) or str(type(arg[i])) == r"<type 'List'>"
            ):
```

pyaedt/generic/DataHandlers.py
```diff
@@ -92,41 +100,46 @@ def _dict2arg(d, arg_out):
...
-    dict_out :
-
-
-    Returns
-    -------
-
-    """
     if arg[0] == "NAME:DimUnits" or "NAME:Point" in arg[0]:
-        dict_out[arg[0][5:]] = list(arg[1:])
+        if arg[0][5:] in list(dict_out.keys()):
```
beliaev-maksim, line 104:
```suggestion
        if arg[0][5:] in dict_out:
```

beliaev-maksim (changes requested):

MaxJPRey (approved):

pyaedt/generic/DataHandlers.py
```diff
@@ -92,41 +100,46 @@ def _dict2arg(d, arg_out):
...
-                :5
-            ] == "NAME:":
+            if (isinstance(arg[i], (list, tuple)) or str(type(arg[i])) == r"<type 'List'>") and arg[i][0][
+                                                                                                :5] == "NAME:":
                 _arg2dict(list(arg[i]), dict_in)
                 i += 1
             elif arg[i][-2:] == ":=":
                 if str(type(arg[i+1])) == r"<type 'List'>":
-                    dict_in[arg[i][:-2]] = list(arg[i + 1])
+                    if arg[i][:-2] in list(dict_in.keys()):
```
MaxJPRey, line 123: @maxcapodi78 @beliaev-maksim Could it be modified to
```suggestion
                    if arg[i][:-2] in dict_in:
```

pyaedt/generic/DataHandlers.py
```diff
@@ -92,41 +100,46 @@ def _dict2arg(d, arg_out):
...
+                        dict_in[arg[i][:-2]].append(list(arg[i + 1]))
+                    else:
+                        dict_in[arg[i][:-2]] = list(arg[i + 1])
                 else:
-                    dict_in[arg[i][:-2]] = arg[i + 1]
+                    if arg[i][:-2] in list(dict_in.keys()):
```
MaxJPRey, line 128: @maxcapodi78 @beliaev-maksim Could it be modified to
```suggestion
                    if arg[i][:-2] in list(dict_in.keys()):
```

pyaedt/modules/Material.py
```diff
@@ -232,6 +232,12 @@ def __init__(self, material, name, val=None, thermalmodifier=None):
         elif val is not None and val["property_type"] == "AnisoProperty":
             self.type = "anisotropic"
             self.value = [val["component1"], val["component2"], val["component3"]]
+        elif val is not None and val["property_type"] == "nonlinear":
```
MaxJPRey, line 235:
```suggestion
        elif val["property_type"] == "nonlinear":
```

MaxJPRey:
Approved with suggestion..

pyaedt/modules/Material.py
maxcapodi78: @MaxJPRey I cannot commit this modification. If val is None then it would fail in looking for "property_type". So I need first to check that val is filled and then search for the type

MaxJPRey: Yes you are write, it would require a try/except block to prevent from  that and raise and exception.

pyaedt/generic/DataHandlers.py
beliaev-maksim: not clear suggestion

pyaedt/modules/Material.py
beliaev-maksim: try/except is more expensive when it raises compared to type check

pyaedt/generic/DataHandlers.py
beliaev-maksim: yes, could be

PR #645, opened 2021-12-14
Maxwell TEAM3 Example

DaveTwyman:
This is a new Maxwell example for adding to the Maxwell Examples.
It should consist of a script and an image for uploading.
I've not managed to get the documentation up and running locally so hopefully it views well in the browser


beliaev-maksim:
I applied `black` to the code

beliaev-maksim (approved):

examples/00-basic-examples/Maxwell3DTeam3.py
```diff
@@ -0,0 +1,188 @@
+"""
+Maxwell 3D TEAM3 Bath Plate
+-------------------
```
maxcapodi78, line 3:
```suggestion
---------------------------
```

DaveTwyman:
> I applied `black` to the code

Ok, not seen that before, impressively brutal but can't argue with the end result!

examples/00-basic-examples/Maxwell3DTeam3.py
```diff
@@ -0,0 +1,188 @@
...
+                "OffsetF1:=",
+                False,
+                "Synchronize:=",
+                0,
+            ],
+        ],
+        ["NAME:Sweep Operations"],
+        ["NAME:Goals"],
+    ],
+)
```
maxcapodi78, line 138:
```suggestion
param = M3D.opti_parametric.add_parametric_setup("Coil_Position", 
                                                                                      "LIN -20mm 0mm 20mm", parametricname=sweepname)
param.props["ProdOptiSetupDataV2"]["SaveFields"] = True
param.props["ProdOptiSetupDataV2"]["CopyMesh"] = False
param.props["ProdOptiSetupDataV2"]["SolveWithCopiedMeshOnly"] = True
param.update()
```

maxcapodi78: @DaveTwyman once you create an optimetrics setup in Pyaedt you get the object with properties so you can simply edit one of the props and update it once finished.

examples/00-basic-examples/Maxwell3DTeam3.py
```diff
@@ -0,0 +1,188 @@
...
+
+Plot.CreateReport(
+    "mag(Bz) Along 'Line_AB' Centred Coil",
+    "Fields",
+    "Rectangular Plot",
+    "Setup1 : LastAdaptive",
+    ["Context:=", "Line_AB", "PointCount:=", 1001],
+    ["Distance:=", ["All"], "Freq:=", ["All"], "Phase:=", ["0deg"], "Coil_Position:=", ["0mm"]],
+    ["X Component:=", "Distance-60mm", "Y Component:=", ["mag(Bz)"]],
+)
```
maxcapodi78, line 178: good this custom report. we need to wrap it better in the future in M3D.post.create_rectangular_plot method

maxcapodi78 (changes requested):
You can replace the optimetrics editing method with the one proposed. 
LGTM

examples/00-basic-examples/Maxwell3DTeam3.py
maxcapodi78: @DaveTwyman of course in the example I've put all the three properties set, but if you need to change only SaveFields you can get rid of line 109 and 110

DaveTwyman: > @DaveTwyman once you create an optimetrics setup in Pyaedt you get the object with properties so you can simply edit one of the props and update it once finished.

That's great, I do now recall seeing similar comment in the documentation. Let's leave the three lines in they are clear and adds detail to the example. Thanks

PR #657, opened 2021-12-16
Enhancement/wpf builder

maxcapodi78:
First version of wpf automatic builder


Alberto-DM (approved):
LGTM

pyaedt/application/MessageManager.py
```diff
@@ -309,7 +309,10 @@ def add_message(self, type, message_text, level=None, proj_name=None, des_name=N
                 des_name = self._design_name
             if des_name and ";" in des_name:
                 des_name = des_name[des_name.find(";") + 1 :]
-            self._desktop.AddMessage(proj_name, des_name, type, message_text)
+            try:
+                self._desktop.AddMessage(proj_name, des_name, type, message_text)
+            except:
+                print("pyaedt info: Failed in Adding Desktop Message")
```
akaszynski, line 315: Is there a way we can raise exceptions instead of printing?

@MaxJPRey, can you make sure these get either logged or raised?

MaxJPRey: @akaszynski Sure Alex. I will.

MaxJPRey: Actually that is the MessageManager itself that we are wrapping with the standard python logger.

maxcapodi78: Exactly. We don't have print anywhere else.

Issue #660, opened 2021-12-17
The name of the net for the polygon creation must be improved.


MaxJPRey:
https://github.com/pyansys/pyaedt/blob/ddab51735aa50a41996d515657aecddfb776bc56/pyaedt/edb.py#L1039

maxcapodi78:
@MaxJPRey  The name of the net has to be DUMMY_XXX. this is managed differently by the solver from a different naming. you cannot change it in with POLYGON_XXX

PR #661, opened 2021-12-17
Fix/issue 660

MaxJPRey:
Fix issue #660 .
Rename the net for the polygon creation.
Delete some variables that are not required anymore.

### DIFF
--- a/pyaedt/edb.py
+++ b/pyaedt/edb.py
@@ -1036,7 +1036,7 @@ class Edb(object):
         plane = self.core_primitives.Shape("polygon", points=point_list)
         polygonData = self.core_primitives.shape_to_polygon_data(plane)
         self.core_primitives.create_polygon(
-            plane, list(self.core_stackup.signal_layers.keys())[0], net_name="DUMMY_CUTOUT"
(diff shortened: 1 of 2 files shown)


pyaedt/edb.py
```diff
@@ -1036,9 +1036,8 @@ def create_cutout_on_point_list(
         plane = self.core_primitives.Shape("polygon", points=point_list)
         polygonData = self.core_primitives.shape_to_polygon_data(plane)
         self.core_primitives.create_polygon(
-            plane, list(self.core_stackup.signal_layers.keys())[0], net_name="DUMMY_CUTOUT"
```
maxcapodi78, line 1039: you cannot rename DUMMY_CUTOUT. the DUMMY word into the net name has to be preserved otherwise the solver will mesh it. We don't want to have it meshed since it is an added polygon.

maxcapodi78 (changes requested):
Please revert the DUMMY_CUTOUT renaming

pyaedt/edb.py
MaxJPRey: @maxcapodi78 Sure I will revert it.

MaxJPRey: Done

maxcapodi78 (approved):
LGTM

Issue #662, opened 2021-12-19
Misspelling argument in create_ellipse() in Primitives3D.


MaxJPRey:
Hi,
Sorry for commenting on a closed issue, but it seems that the typo is still in main:

https://github.com/pyansys/pyaedt/blob/146680f6b2d9fc73d5f70c09cf0556cda8cd2645/pyaedt/modeler/Primitives3D.py#L565

_Originally posted by @Piwakk in https://github.com/pyansys/pyaedt/issues/172#issuecomment-995775418_

PR #667, opened 2021-12-20
Rename argument 'major_raidus' as 'major_radius'.

MaxJPRey:
Fix issue #662 .
Rename argument 'major_raidus' as 'major_radius' in the method create_ellipse() in the Primitives3D module.

### DIFF
--- a/pyaedt/modeler/Primitives3D.py
+++ b/pyaedt/modeler/Primitives3D.py
@@ -562,7 +562,7 @@ class Primitives3D(Primitives, object):
         return self._create_object(new_object_name)
 
     @aedt_exception_handler
-    def create_ellipse(self, cs_plane, position, major_raidus, ratio, is_covered=True, name=None, matname=None):
+    def create_ellipse(self, cs_plane, position, major_radius, ratio, is_covered=True, name=None, matname=None):
         """Create an ellipse.
 
         Parameters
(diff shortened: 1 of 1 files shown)


pyaedt/modeler/Primitives3D.py
```diff
@@ -562,7 +562,7 @@ def create_circle(self, cs_plane, position, radius, numSides=0, is_covered=True,
         return self._create_object(new_object_name)
 
     @aedt_exception_handler
-    def create_ellipse(self, cs_plane, position, major_raidus, ratio, is_covered=True, name=None, matname=None):
+    def create_ellipse(self, cs_plane, position, major_radius, ratio, is_covered=True, name=None, matname=None):
```
MaxJPRey, line 565: @maxcapodi78 I don't know if we should add a deprecation warning as it is an argument name and not the method name that was changed. Let say that one customer uses named arguments in his script with something like:
`create_ellipse(, "XY", [0., 0.,0.], major_raidus = 3.2, ratio = 0.8, is_covered=True, name=None, matname=None)`
Then his script will fail.
What do you think?

maxcapodi78: it's a clear spelling error. I don't think it's needed deprecation warning as soon as it's documented

maxcapodi78 (approved):
LGTM

PR #671, opened 2021-12-22
Implemented arc segmentation for edb plot

Alberto-DM:
Arcs in the layout are now fully represented represented. The arc is segmented.


maxcapodi78 (approved):
LGTM

PR #673, opened 2021-12-23
Plot voids in EDB plots

Alberto-DM:
- EDB plots is now plotting voids correctly.
- It is possible to specify the figure size in pixel.
- The colors remain the same between runs.


maxcapodi78 (approved):
Great! LGTM

Issue #77, opened 2021-06-11
Parameter questions/issues


PipKat:
Hffs.py file contains parameter descriptions that are a bit confusing or different from other files. We'd  like to strive for consistency.

Lumped--The HFSS has a few function functions and parameter descriptions with lumped in them. Some of these seem like they might need to be rewritten.

For the parameter "sweeptype," choices are listed here have initial captial letters. I know in another file, I saw them as lower case. Can they standardized. If so, which form is preferred. Alex said that there may be an API variation between the different products.

For one occurrence, the description for the parameter "startobj" has this additional sentence: This is typically the reference plane.
Should this sentence be added to the description of all occurrences of "startobj"?

MaxJPRey:
@maxcapodi78 Do you want me to rename the sweeptype values in lower case. 
If we proceed so, "Interpolating" would become "interpolating".

Let me know what you think. Otherwise, I will close the issue as the second part of this issue is the expected behavior for the "startobj" parameter when dealing with microstrip.

PR #674, opened 2021-12-24
All sweeptype default parameter value now starts with lower case.

MaxJPRey:
Fix issue #77 .

It is a minor typo modification. It is without impact because it only concerns a default value string in a method's signature that has been deprecated.
All other issues (upper case uniformity across the entire package) have been resolved along the way.

@maxcapodi78 I checked that the `sweeptype` argument default values were upper case in different files.


maxcapodi78 (approved):
LGTM

Issue #659, opened 2021-12-17
Update to latest pyansys-sphinx-theme


akaszynski:
You no longer need to include the logo file or the templates. See:
https://github.com/pyansys/pyansys-sphinx-theme/blob/0a94c8c2edfb7b67ab4829f9e1722aa477c58a74/doc/source/conf.py#L4-L14

PR #675, opened 2021-12-24
Fix/issue 659

MaxJPRey:
Fix issue #659 .
Logo is now provided through the pyansys_sphinx_theme.
In addition, the templates are no longer required.

Notify that PyAEDT is part of the larger PyANSYS effort in the README.rst file.


doc/source/conf.py
```diff
@@ -20,7 +22,7 @@
 
 sys.path.append(os.path.join(root_path))
 project = "PyAEDT"
-copyright = "(c) 2021 ANSYS, Inc. All rights reserved"
+copyright = f"(c) {datetime.datetime.now().year} ANSYS, Inc. All rights reserved"
 author = "Ansys Inc."
 
 # Check for the local config file, otherwise use default desktop configuration
```
MaxJPRey, line 28: @akaszynski Few lines below in this file, to get the semantic version, we are doing:
```
with open(os.path.join(root_path, "pyaedt", "version.txt"), "r") as f:
    release = version = f.readline()
```
Do you think that we should expose the version in the `_version.py` with an attribute like this `__version__` and then do in the `conf.py` file:
`release = version = __version__`

akaszynski: Both approaches work. Since we already need to import `pyaedt` for the auto documentation to be built, I think it’s easier to:
```py
from pyaedt import __version__
```

akaszynski (approved):
LGTM. Thanks for fixing this. Please validate that the docs at dev.docs.pyaedt.com are valid once this is merged.

Issue #622, opened 2021-12-03
Ibis Parses


maxcapodi78:
In order to import ibis models in AEDT we need to parse Ibis model before.

kbhagat-ansys:
I wonder if there is a way to take advantage of Nexxim (and or Circuit UI) as they use an enhanced version of the committee IBIS parser. 
Please DM me and I can put you in touch with the developer(s) concerned.

PR #676, opened 2021-12-24
Add Ibis reader capability to insert pins and buffers into the schematic.

MaxJPRey:
_Draft will be removed as long as the doc is added._
Fix issue #622 :
Add a reader for *.ibs file in order to extract all data of the IBIS model included in this file.
Add features to add, insert and locate both BUFFER and PIN Component into the AEDT graphical window.
Add some unit tests.
The IBIS-AMI model is not supported yet.

**Note:** 
This is a first version. The reader was mainly created to extract data required by AEDT. 
Some packages are available on PyPi and will be analyzed to see if we can benefit from them. 
(https://pypi.org/project/PyIBIS-AMI/)
Indeed, the replacement of the current reader exposed in this PR by an existing reader available on Pypi is not excluded in case we find one from which we can extract easily the data expected by AEDT product.


MaxJPRey:
Documentation files still need to be modified to include these new features.

pyaedt/circuit.py
```diff
@@ -360,6 +361,26 @@ def create_schematic_from_netlist(self, file_to_import):
         self.logger.info("Netlist correctly imported into %s", self.design_name)
         return True
 
+    @aedt_exception_handler
+    def read_ibis(self, path):
+
```
akaszynski, line 366:
```suggestion
```

pyaedt/generic/ibis_reader.py
```diff
@@ -0,0 +1,748 @@
...
+                "Data:=",
+                [],
+                "GRef:=",
+                [],
+            ],
+        )
+
+    def insert(self, x, y, angle=0.0):
+        """
+        Insert a pin at a defined location inside the graphical window.
```
akaszynski, line 144:
```suggestion
        """Insert a pin at a defined location inside the graphical window.
```

pyaedt/generic/ibis_reader.py
```diff
@@ -0,0 +1,748 @@
...
+
+        Parameters
+        ----------
+        x: float
+            X position of the pin.
+        y: float
+            Y position of the pin.
+        angle: float, optional
```
akaszynski, line 152:
```suggestion
        x : float
            X position of the pin.
        y : float
            Y position of the pin.
        angle : float, optional
```

pyaedt/generic/ibis_reader.py
```diff
@@ -0,0 +1,748 @@
...
+                True,
+                "Data:=",
+                [],
+                "GRef:=",
+                [],
+            ],
+        )
+
+    def insert(self, x, y, angle=0.0):
+        """
```
akaszynski, line 202: Short description should always be on the first line, even on multi-line docstrings:

pyaedt/generic/ibis_reader.py
```diff
@@ -0,0 +1,748 @@
...
+        Insert a buffer at a defined location inside the graphical window.
+
+        Parameters
+        ----------
+        x: float
```
akaszynski, line 207: Leave a space after the parameter name and the colon `:`, otherwise sphinx won’t render the parameter correctly.

pyaedt/generic/ibis_reader.py
```diff
@@ -0,0 +1,748 @@
...
+        circuit : class:`pyaedt.circuit.Circuit`
+            Circuit in which the ibis components will be used.
+
+        Returns
+        ----------
+        :class:`pyaedt.generic.ibis_reader.Ibis`
+            Ibis object exposing all data from the ibis file.
+        """
+
+        if os.path.exists(filename) == False:
```
akaszynski, line 392:
```suggestion
        if not os.path.exists(filename):
```

pyaedt/generic/ibis_reader.py
```diff
@@ -0,0 +1,748 @@
...
+            error_message = filename + " does not exist."
+            raise FileExistsError(error_message)
```
akaszynski, line 394: There’s no reason to assign and use a single use variable (unless for readability). Also, `FileExistsError` would be the wrong exception in Python 3, you’d use `FileNotFoundError`, however to maintain compatibility with IronPython 2.7, we would either have to create a custom exception or just use `OSError` or just a plain `Exception`.

pyaedt/generic/ibis_reader.py
```diff
@@ -0,0 +1,748 @@
...
+
+        ibis_name = pyaedt.generic.general_methods.get_filename_without_extension(filename)
+        ibis = Ibis(ibis_name, circuit)
+
+        # Read *.ibis file.
+        with open(filename, "r") as file:
```
akaszynski, line 400: Nitpick: `file` is a keyword for Python 2.7 and you may want to use `fid` instead if your IDE complains about it. Since we may drop IronPython at some point, it’s up to you if you want to change this.

pyaedt/generic/ibis_reader.py
```diff
@@ -0,0 +1,748 @@
...
+            while True:
+                current_line = file.readline()
+                if not current_line:
+                    break
+
+                if self.IsStartedWith(current_line, "[Component] ") == True:
```
akaszynski, line 406: There’s no reason to compare with `True`, the `if` statement will still work. `if True == True:` is the same as `if True:`, though even if you were comparing a Boolean with a Boolean, you would use `is` rather than `==`.
```suggestion
                if self.IsStartedWith(current_line, "[Component] "):
```

pyaedt/generic/ibis_reader.py
```diff
@@ -0,0 +1,748 @@
...
+
+        """
+
+        if line == "":
+            return ""
+
+        data = line.split(" ")
+        return data[0].strip()
+
+    def IsStartedWith(self, src, find, ignore_case=True):
```
akaszynski, line 720: This should be pascal case and not camel case as it’s a method.

akaszynski: This should not be a class method, just a function since it does not depend on any class data.

pyaedt/generic/ibis_reader.py
```diff
@@ -0,0 +1,748 @@
...
+        if ignore_case == True:
+            if src[: len(find)].lower() == find.lower():
+                return True
+            else:
+                return False
+        else:
+            if src[: len(find)] == find:
+                return True
+            else:
+                return False
```
akaszynski, line 748: This could be greatly simplified. Also, it appears this method only extends the built-in `startswith` by adding the ability to ignore case.

```suggestion
        if ignore_case:
            return src.lower().startswith(find.lower())
        return src.startswith(find)
```

pyaedt/generic/ibis_reader.py
```diff
@@ -0,0 +1,748 @@
...
+
+        current_string = current_string[len(pin.r_value) + 1 :].strip()
+        pin.l_value = self.get_first_parameter(current_string)
+
+        current_string = current_string[len(pin.l_value) + 1 :].strip()
+        pin.c_value = self.get_first_parameter(current_string)
+
+        return pin
+
+    def get_first_parameter(self, line):
```
akaszynski, line 699: This class method may not belong in this class as it does not depend on the state of the class. If you choose to have it in this class for style/organization, consider using `@classmethod`.

pyaedt/generic/ibis_reader.py
```diff
@@ -0,0 +1,748 @@
...
+
+        Returns
+        -------
+        str
+            Name of the component.
+
+        """
+        name = ""
+        name = line.replace("[Component]", "")
+        return name.strip()
```
akaszynski, line 654:
```suggestion
        return line.replace("[Component]", "").strip()
```

pyaedt/generic/ibis_reader.py
```diff
@@ -0,0 +1,748 @@
...
+        if self.IsStartedWith(current_line, "R_pkg") == True:
+            component.R_pkg = current_line.strip()
+            current_line = file.readline()
+        elif self.IsStartedWith(current_line, "L_pkg") == True:
+            component.L_pkg = current_line.strip()
+            current_line = file.readline()
+        elif self.IsStartedWith(current_line, "C_pkg") == True:
+            component.C_pkg = current_line.strip()
+
+    def get_component_name(self, line):
```
akaszynski, line 638: This method is a candidate for `@classmethod` or to be moved out of the class as a standard function.

akaszynski:
Change the PR title so it’s more descriptive. These PR titles are used in the autogenerated release notes summary and it’s helpful to have a clear title. If applicable, include the issue being resolved in the body of the PR.

pyaedt/generic/ibis_reader.py
MaxJPRey: Make sense.

pyaedt/generic/ibis_reader.py
MaxJPRey: @akaszynski ,You are right, this function was based on a Visual Basic code. Some modifications must be made to be more pythonic.

pyaedt/generic/ibis_reader.py
MaxJPRey: I fixed this issue in the entire file.

pyaedt/generic/ibis_reader.py
MaxJPRey: I was looking for a name on my side too.
I think I will just call it ibis_file.

pyaedt/generic/ibis_reader.py
MaxJPRey: It is now a ``@classmethod``.

pyaedt/generic/ibis_reader.py
MaxJPRey: Great suggestion @akaszynski .It avoids to use a list.

_unittest_ironpython/test_15_ibs_reader.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+class TestSequenceFunctionsGenerate(PytestMockup):
+    @classmethod
+    def setUpClass(cls):
+        test_obj.setup_class()
+
+    @classmethod
+    def tearDownClass(cls):
+        test_obj.teardown_class()
```
akaszynski, line 18: Method names, even within our unit tests, should conform to PEP8. These should be pascal case.

pyaedt/circuit.py
```diff
@@ -360,6 +361,25 @@ def create_schematic_from_netlist(self, file_to_import):
...
+        Returns
+        ----------
+        :class:`pyaedt.generic.ibis_reader.Ibis`
+            Ibis object exposing all data from the ibis file.
+        """
+
+        reader = ibis_reader.IbisReader()
+        ibis = reader.read_project(path, self)
+        return ibis
```
akaszynski, line 381: There's no reason to keep around a single use parameter.
```suggestion
        return reader.read_project(path, self)
```

pyaedt/generic/ibis_reader.py
```diff
@@ -0,0 +1,739 @@
...
+    """Component extracted from ibis model."""
+
+    def __init__(self):
+        self._name = None
+        self._manufacturer = None
+        self._pins = {}
+
+    @property
+    def name(self):
+        """Name of the component."""
```
akaszynski, line 16: Consider adding examples to these properties.

pyaedt/generic/ibis_reader.py
```diff
@@ -0,0 +1,739 @@
...
+    def signal(self):
+        return self._signal
+
+    @signal.setter
+    def signal(self, value):
+        """Signal of the pin."""
+        self._signal = value
+
+    @property
+    def model(self):
```
akaszynski, line 87: Missing docstring.

pyaedt/generic/ibis_reader.py
```diff
@@ -0,0 +1,739 @@
...
+
+        Parameters
+        ----------
+        ibis : :class:`pyaedt.generic.ibis_reader.Ibis`
+            ibis object containing all info.
+        current_line : str
+            Current line content.
+        ibis_file : TextIO
+            File's stream.
+
```
akaszynski, line 465: This should have a basic example.

pyaedt/generic/ibis_reader.py
```diff
@@ -0,0 +1,739 @@
...
+    def buffers(self):
+        """Buffers included into the ibis model."""
+        return self._buffers
+
+    @buffers.setter
+    def buffers(self, value):
+        self._buffers = value
+
+
+class IbisReader:
```
akaszynski, line 373: There's no `__init__` defined for this class. It's unclear to me, but if `read_project` must be called for this class to be useful, I would just convert `read_project` --> `__init__`.

pyaedt/circuit.py
```diff
@@ -360,6 +361,25 @@ def create_schematic_from_netlist(self, file_to_import):
...
+        Parameters
+        ----------
+        path : str
+            Path of the ibis file.
+
+        Returns
+        ----------
+        :class:`pyaedt.generic.ibis_reader.Ibis`
+            Ibis object exposing all data from the ibis file.
+        """
```
akaszynski, line 377: You may not choose to provide an example within `IbisReader`, but one should be provided here.

pyaedt/generic/ibis_reader.py
```diff
@@ -0,0 +1,739 @@
...
+            ibis object containing all info.
+        current_line : str
+            Current line content.
+        ibis_file : TextIO
+            File's stream.
+
+        """
+
+        if is_started_with(current_line, "[Component] ") != True:
+            return
```
akaszynski, line 574:
```suggestion
        if not is_started_with(current_line, "[Component] "):
            return
```

pyaedt/generic/ibis_reader.py
```diff
@@ -0,0 +1,739 @@
...
+
+        while True:
+            current_line = ibis_file.readline()
+            if is_started_with(current_line, "[Package]"):
+                break
+
+        self.fill_package_info(component, current_line, ibis_file)
+
+        # [pin]
+        while is_started_with(current_line, "[Pin] ") != True:
```
akaszynski, line 591:
```suggestion
        while not is_started_with(current_line, "[Pin] "):
```

pyaedt/generic/ibis_reader.py
```diff
@@ -0,0 +1,739 @@
...
+            current_line = ibis_file.readline()
+
+        while True:
+            current_line = ibis_file.readline()
+            if is_started_with(current_line, "|"):
+                break
+
+        current_line = ibis_file.readline()
+
+        while is_started_with(current_line, "|") == False:
```
akaszynski, line 601:
```suggestion
        while not is_started_with(current_line, "|"):
```

pyaedt/generic/ibis_reader.py
```diff
@@ -0,0 +1,739 @@
...
+        if line == "":
+            return ""
+
+        data = line.split(" ")
+        return data[0].strip()
+
+
+def is_started_with(src, find, ignore_case=True):
+    """Verifies if a string content starts with a specific string or not.
+
```
akaszynski, line 720:
```suggestion

    This is identical to `str.startswith`, except that it includes
    the `ignore_case` parameter.

```

akaszynski (commented):
Provided that we can include more examples and add this class to the overall API documentation, I'm happy with this PR. I'll leave it up to @maxcapodi78 to approve.

pyaedt/generic/ibis_reader.py
MaxJPRey: Done

pyaedt/generic/ibis_reader.py
MaxJPRey: The issue is that in this method `read_project` I returned an IBIS object. I would not be able to do that within the initializer.

pyaedt/circuit.py
MaxJPRey: Excellent idea. Done.

_unittest_ironpython/test_15_ibs_reader.py
MaxJPRey: @akaszynski The camelCase naming of the class methods comes from the `PytestMockup` class inherited from `TestCase` in the `unittest` module.
Those methods are overridden.

akaszynski:
Responding to https://github.com/pyansys/pyaedt/pull/676/files#r778147185

For `IbisReader`, try overriding the constructor. For example.

```py
class MyTotallyDifferentClass:

    def __init__(self, init_value):
        self._myattr = init_value

    @property
    def myattr(self):
        return self._myattr


class MyClass:

    def __new__(cls, init_value=None):
        if init_value is None:
            return super(MyClass, cls).__new__(MyClass)
        return MyTotallyDifferentClass(init_value)

    @property
    def diff_myattr(self):
        return 'hello world'


my_inst = MyClass('whatever')
print(type(my_inst))
my_inst.myattr

my_inst = MyClass()
print(type(my_inst))
my_inst.diff_myattr
```

MaxJPRey:
> Responding to https://github.com/pyansys/pyaedt/pull/676/files#r778147185
> 
> For `IbisReader`, try overriding the constructor. For example.
> 
> ```python
> class MyTotallyDifferentClass:
> 
>     def __init__(self, init_value):
>         self._myattr = init_value
> 
>     @property
>     def myattr(self):
>         return self._myattr
> 
> 
> class MyClass:
> 
>     def __new__(cls, init_value=None):
>         if init_value is None:
>             return super(MyClass, cls).__new__(MyClass)
>         return MyTotallyDifferentClass(init_value)
> 
>     @property
>     def diff_myattr(self):
>         return 'hello world'
> 
> 
> my_inst = MyClass('whatever')
> print(type(my_inst))
> my_inst.myattr
> 
> my_inst = MyClass()
> print(type(my_inst))
> my_inst.diff_myattr
> ```

This has been implemented. A constructor (`__new__`) has been added. It will return an `Ibis` object.

pyaedt/generic/ibis_reader.py
MaxJPRey: I did and it highlighted the fact that some tests were missing.

akaszynski (approved):
LGTM. Nice work implementing the constructor.

PR #682, opened 2021-12-29
Added UnitTest to new EdbNetsData and EDBPrimitives Classes. Added Do…

maxcapodi78:
…cumentation

Added Plot method to layers.

### DIFF
--- a/pyaedt/edb_core/EDB_Data.py
+++ b/pyaedt/edb_core/EDB_Data.py
@@ -314,24 +314,22 @@ class EDBPrimitives(object):
         return voids
 
     @aedt_exception_handler
-    def points_with_arcs(self):
-        """Return a list of points in [x,y] coordinate.
-        For arc, y is infinite and the arc height is represented with x coordinate.
+    def points_raw(self):
+        """Return a list of Edb points.
 
         Returns
         -------
-        list of list
-            list of [x,y] couple of data.
+        list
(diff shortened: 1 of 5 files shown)


pyaedt/edb_core/EDB_Data.py
```diff
@@ -314,24 +314,22 @@ def voids(self):
...
         -------
-        list of list
-            list of [x,y] couple of data.
+        list
+            Edb Points.
         """
         points = []
         try:
             my_net_points = list(self.primitive_object.GetPolygonData().Points)
             for i, point in enumerate(my_net_points):
```
MaxJPRey, line 328: @maxcapodi78 We don't need the enumerate anymore because ``i`` is not used.
```suggestion
            my_net_points = list(self.primitive_object.GetPolygonData().Points)
            for point in my_net_points:
```

MaxJPRey (approved):

pyaedt/edb_core/EDB_Data.py
maxcapodi78: @Alberto-DM  you implemented the enumerate in this method. Is there any specific issue why you didin't iterate directly my_net_points or can I apply suggestion from @MaxJPRey

Alberto-DM (approved):
Looks good to me!

Issue #684, opened 2021-12-31
plot() in Object3d is not working


MaxJPRey:
'Modeler3D' object has no attribute '_appp'

_app attribute is misspelled.

https://github.com/pyansys/pyaedt/blob/97cfb19f44d7a99fa682f19e36b8e8c397da91dc/pyaedt/modeler/Object3d.py#L779

PR #685, opened 2021-12-31
Correct the Object3d.plot() implementation.

MaxJPRey:
`Modeler3D` object has no attribute `_appp`
So when calling the `plot()` method in `Object3d`, an error was raised.
Attribute `_appp` is renamed as `_app`.

Fix issue #684 .


maxcapodi78 (approved):
Great catch

Issue #689, opened 2022-01-04
Rename class methods following snake_case convention.


MaxJPRey:
The IronPython unit tests are based on some class methods.
The signature of those methods does not follow the snake_case convention.

MaxJPRey:
The camelCase naming of the class methods comes from the PytestMockup class inherited from TestCase in the unittest module.
Those methods are overridden.

PR #690, opened 2022-01-04
Rename 'setUpClass' as 'setup_class' and 'tearDownClass' as 'teardown_class'.

MaxJPRey:
In the _unittest_ironpython folder I renamed 'setUpClass' as 'setup_class' and 'tearDownClass' as 'teardown_class'.
All methods, class methods or instance methods must follow the snake_case convention.
Fix issue #689 .


MaxJPRey:
The camelCase naming of the class methods comes from the `PytestMockup` class inherited from `TestCase` in the `unittest` module.
Those methods are overridden.

PR #695, opened 2022-01-06
Expose odesktop property from Design class.

### LINKED ISSUE #694: self.odesktop is not exposed as a property
In the `Design` class, `self.odesktop` is not exposed as a property.
It could be otherwise the attribute `self.odesktop` is useless.
https://github.com/pyansys/pyaedt/blob/1eb337eff5008556078cb5b6e1e37a4bf529db90/pyaedt/application/Design.py#L591

MaxJPRey:
Expose `odesktop` through a read-only property from the `Design`.
Use the `odesktop` accessor instead of the private member. `_desktop`.
Fix #694

### DIFF
--- a/_unittest/test_01_Design.py
+++ b/_unittest/test_01_Design.py
@@ -14,6 +14,8 @@ try:
 except ImportError:
     import _unittest_ironpython.conf_unittest as pytest  # noqa: F401
 
+from pyaedt.generic.general_methods import is_ironpython
+
 test_project_name = "Coax_HFSS"
 example_project = os.path.join(local_path, "example_models", test_project_name + ".aedt")
 
@@ -236,5 +238,7 @@ class TestClass:
         assert self.aedtapp.omaterial_manager
 
     def test_27_odesktop(self):
-        print(type(self.aedtapp.odesktop))
-        assert str(type(self.aedtapp.odesktop)) == "<class 'win32com.client.CDispatch'>"
+        if is_ironpython:
+            assert str(type(self.aedtapp.odesktop)) == "<type 'ADesktopWrapper'>"
+        else:
+            assert str(type(self.aedtapp.odesktop)) == "<class 'win32com.client.CDispatch'>"
--- a/_unittest/test_01_Design.py
+++ b/_unittest/test_01_Design.py
@@ -236,4 +236,5 @@ class TestClass:
         assert self.aedtapp.omaterial_manager
 
     def test_27_odesktop(self):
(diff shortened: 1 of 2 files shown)


pyaedt/application/Design.py
```diff
@@ -588,12 +588,16 @@ def __init__(
...
-        self.odesktop = self._desktop
         self._variable_manager = VariableManager(self)
         self.solution_type = self._solution_type
         self.project_datasets = self._get_project_datasets()
         self.design_datasets = self._get_design_datasets()
 
+    @property
+    def odesktop(self):
+        """Desktop instance containing all projects and designs."""
+        return self._desktop
```
akaszynski, line 599: I'm of the opinion that "odesktop" isn't a python friendly attribute. Recommending renaming to `desktop`. The leading "o" won't make sense for new users of PyAEDT.

We should also provide a trivial example and return value for this.

MaxJPRey: @akaszynski  I named it "odesktop" because it returns a COM object and that have been the naming convention applied to the entire project. Should I get rid of the convention for this property or get the name more pythonic?
I will add example and return value.

akaszynski: Good question. Whatever we do should be consistent, and if we have a standard to use "o" for COM objects, let's use it.

MaxJPRey: Ok, so I will use `odesktop` then.

akaszynski: Still needs an example I think, unless we don't intend for the user to access this attribute.

_unittest/test_01_Design.py
```diff
@@ -230,3 +230,6 @@ def test_24_change_type(self):
 
     def test_25_change_registry_from_file(self):
         assert self.aedtapp.set_registry_from_file(os.path.join(local_path, "example_models", "Test.acf"))
+
+    def test_27_odesktop(self):
+        assert self.aedtapp.odesktop
```
akaszynski, line 235: Should be some sort of `isinstance` here. If you can't get a type, this works as well:
```py
assert 'COM' in `str(type(self.aedtapp.odesktop))`
```
Note: I "COM" probably isn't the string to test.

MaxJPRey: @akaszynski Excellent idea. Done. I will try to improve other existing tests too in an other PR.

akaszynski (approved):
LGTM

maxcapodi78 (approved):
LGTM

PR #704, opened 2022-01-11
Now Cutout includes also original Setups

maxcapodi78:


MaxJPRey (approved):
@maxcapodi78 Can you add a short unit test to also test the content of the loop ``if "GetSimSetupInfo" in dir(_setup):``

PR #712, opened 2022-01-12
Replace CI bot token by an application.

### LINKED ISSUE #696: Nightly documentation build is broken
Nightly build is broken https://github.com/pyansys/pyaedt/actions/workflows/nightly-docs.yml

Reason being is our PAT from https://github.com/pyansys-ci-bot no longer works due to SSO. We could renew the tokens, but IMO we should be using github applications. Reference: https://michaelheap.com/ultimate-guide-github-actions-authentication/

To fix this, follow the approach used in https://github.com/pyansys/pymapdl/pull/810. I've given the bot auth to this repo and the dependent documentation repos.

MaxJPRey:
Replace the current token generated with the PyAnsys CI bot by a GitHub application.
For the deployment, we will now use `JamesIves/github-pages-deploy-action` because it is maintained and tested.
We added a step to retry the deployment in case it failed the first time because of some network issue. It already happened in the past.

Fix #696


.github/workflows/main.yml
```diff
@@ -51,44 +60,44 @@ jobs:
...
+      #     git checkout -b master
+      #     git config --local user.name "maxcapodi78"
+      #     git config --local user.email "massimo.capodiferro@ansys.com"
+      #     New-Item -ItemType file .nojekyll
+      #     New-Item -ItemType file CNAME
+      #     Add-Content CNAME "aedtdocs.pyansys.com"
+      #     git add .
+      #     git commit -m "Documentation generated by PyAedt Build"
+      #     git remote add origin https://${{ secrets.PYAEDT_DOCS }}@github.com/pyansys/PyAEDT-docs
+      #     git push -u origin master --force
```
MaxJPRey, line 85: @maxcapodi78 @akaszynski I thought that we could use gh-pages branching as in PyMAPDL. If you both agree, I will remove this commented section.

.github/workflows/main.yml
```diff
@@ -51,44 +60,44 @@ jobs:
...
-      #   branch: master
-      #    folder: doc/_build/html/
-      #    token: ${{ secrets.PYANSYS_CI_BOT_TOKEN }}
-      #    clean: true
+      - name: Deploy
+        if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
+        uses: JamesIves/github-pages-deploy-action@4.1.4
+        with:
+          repository-name: pyansys/pyaedt-docs
+          branch: gh-pages
```
akaszynski, line 100: Note that changing the branch we push to will mean we also have to change the "pages" branch used by gh-pages in https://github.com/pyansys/PyAEDT-docs/settings/pages

akaszynski (approved):
LGTM. Take note that we change the pages branch in https://github.com/pyansys/PyAEDT-docs/settings/pages. Be sure @maxcapodi78 is aware.

maxcapodi78 (approved):
Better to keep master now and move to gh-pages later

.github/workflows/main.yml
```diff
@@ -51,44 +60,41 @@ jobs:
...
-      #  uses: JamesIves/github-pages-deploy-action@4.1.4
-      #  with:
-      #  repository-name: pyansys/pyaedt-docs
-      #   branch: master
-      #    folder: doc/_build/html/
-      #    token: ${{ secrets.PYANSYS_CI_BOT_TOKEN }}
-      #    clean: true
+        # Retry the deployment in case some network issue prevented git to push
+        # the modifications properly. It happened in the past.
+      - name: Deploy Second try
```
akaszynski, line 92: Forgot:
```
        continue-on-error: true          # make the step always success
```

.github/workflows/main.yml
```diff
@@ -78,6 +78,7 @@ jobs:
           retention-days: 7
 
       - name: Deploy
+        continue-on-error: true
```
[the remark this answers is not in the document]
MaxJPRey: @akaszynski I thought it was already the case because ligne 94:
`if: steps.deploy.outcome!='success' && github.event_name == 'push' && startsWith(github.ref, 'refs/tags')`
we check the outcome of the first deployment before to execute the second try?
Am I missing a point?

PR #715, opened 2022-01-13
Correct the add_range() method for sweep "LogScale" subrange.

### LINKED ISSUE #707: sweep.add_subrange() Bug
line 1283 of pyaedt/pyaedt/modules/SetupTemplates.py is wrong, it should be:
            range["RangeEnd"] = str(end) + unit
Maybe you can put the line out of if-else.

image

MaxJPRey:
In the `range` dictionary, correct the `RangeEnd` value to be a string and not a float.

Return `True` at the end of the method `add_subrange`.

Fix #707 .


akaszynski (approved):
LGTM.

PR #718, opened 2022-01-13
Enhancement/plot class

maxcapodi78:


pyaedt/modules/AdvancedPostProcessing.py
```diff
@@ -1327,22 +1118,20 @@ def export_mesh_obj(self, setup_name=None, intrinsic_dict={}):
 
     @aedt_exception_handler
     def plot_model_obj(
```
akaszynski, line 1120: The Return on this method is incorrect.

akaszynski:
We need to add the ability to return the `pyvista.Plotter` instance when running:
https://github.com/pyansys/pyaedt/blob/1fde412f5591d2528a8d514da8b1e8d67fc95901/pyaedt/modules/AdvancedPostProcessing.py#L622

Recommend changing the function signature to:

```py
 def plot(self, export_image_path=None, return_plotter=False, show=True): 
```

That way, downstream users can add to the existing pyvista plotter and make their own animations or scenes. Then, in `plot_model_obj`, provide the ability to `return_plotter` and then:

```py

pl = hfss.plot.plot_model_obj(return_plotter=True)

##############################
# Add Radar Data to Plotter

radar_sgrid = pyvista.StructuredGrid(x, y, z)

# add individual radar frames
for i in range(100):
    radar_sgrid['radar_%03d' % i] = mydata[i]

pl.add_mesh(radar_sgrid)

# then animate by following this example
# https://docs.pyvista.org/examples/02-plot/gif.html
# 
# Differences:
# - don't plotter.write_frame()
# - don't update_scalars, rather radar_sgrid.active_scalars_name = 'radar_%03d' % frame
```

Pinging @ushechip

---

Main point is to provide an interface so we can customize the output from:
https://github.com/pyansys/pyaedt/blob/be9cccd51b8bcc29a487457a5e22d5d329d4db25/pyaedt/modules/AdvancedPostProcessing.py#L1329

pyaedt/modules/AdvancedPostProcessing.py
maxcapodi78: I was doing exactly the same :-).
I've also added the possibility to show or not the plot.

pyaedt/modeler/Object3d.py
```diff
@@ -770,24 +771,28 @@ def _odesign(self):
...
+    def plot(self, show=True):
         """Plot model with PyVista.
 
         .. note::
         Works from AEDT 2021.2 in CPython only. PyVista has to be installed.
+
+        Parameters
+        ----------
+        show : bool, optional
+            Show the plot after generation.
```
MaxJPRey, line 783:
```suggestion
            Show the plot after generation.  The default value is ``True``.
```

pyaedt/modules/AdvancedPostProcessing.py
```diff
@@ -81,179 +89,296 @@ def is_float(istring):
...
+class ObjClass(object):
+    """Class that manages mesh files to be plotted in pyvista."""
 
-    Examples
-    --------
-    Basic usage demonstrated with an HFSS, Maxwell, or any other design:
+    def __init__(self, path, color, opacity, units):
+        self.path = path
+        self._color = (0, 0, 0)
+        self.color = color
```
MaxJPRey, line 98: I think that you can only set the protected attribute. Then the color property will return the correct value latter on.
```suggestion
        self._color = color
```

pyaedt/modules/AdvancedPostProcessing.py
```diff
@@ -81,179 +89,296 @@ def is_float(istring):
...
 
-class PostProcessor(Post):
-    """Contains advanced postprocessing functionalities that require Python 3.x packages like NumPy and Matplotlib.
-
-    Parameters
-    ----------
-    app :
-        Inherited parent object.
+class ObjClass(object):
+    """Class that manages mesh files to be plotted in pyvista."""
```
MaxJPRey, line 93:
```suggestion
    """Class that manages mesh files to be plotted in pyvista.

    Parameters
    ----------
    path : str
        Full path to the file.
    color : str or tuple
        Can be a string with color name or a tuple with (r,g,b) values.
    opacity : float
        Value between 0 to 1 of opacity.
    units : str
        Model units.

    """
```

pyaedt/modules/AdvancedPostProcessing.py
```diff
@@ -81,179 +89,296 @@ def is_float(istring):
...
+            self._color = tuple(int(h[i : i + 2], 16) for i in (0, 2, 4))
 
-    @aedt_exception_handler
-    def nb_display(self, show_axis=True, show_grid=True, show_ruler=True):
-        """Show the Jupyter Notebook display.
 
-          .. note::
-              .assign_curvature_extraction Jupyter Notebook is not supported by IronPython.
+class FieldClass(object):
+    """Class to manage Field data to be plotted in pyvista."""
```
MaxJPRey, line 119:
```suggestion
    """Class to manage Field data to be plotted in pyvista.

    Parameters
    ----------
    path : str
        Full path to the file.
    log_scale : bool, optional
        Either if the field has to be plotted log or not. The default value is ``True``.
    coordinate_units : str, optional
        Fields coordinates units. The default value is ``"meter"``.
    opacity : float, optional
        Value between 0 to 1 of opacity. The default value is ``1``.
    color_map : str, optional
        Color map of field plot. The default value is ``"rainbow"``.
    label : str, optional
        Name of the field. The default value is ``"Field"``.
    tolerance : float, optional
        Delauny tolerance value used for interpolating points. The default value is ``1e-3``.
    headers : int, optional
        Number of lines to of the file containing header info that has to be removed.
        The default value is ``2``.
    """
```

pyaedt/modules/AdvancedPostProcessing.py
```diff
@@ -81,179 +89,296 @@ def is_float(istring):
...
-             Whether to show the ruler. The default is ``True``.
+    def __init__(
+        self,
+        path,
+        log_scale=True,
+        coordinate_units="meter",
+        opacity=1,
+        color_map="rainbow",
+        label="Field",
+        tol=1e-3,
```
MaxJPRey, line 129:
```suggestion
        tolerance=1e-3,
```

pyaedt/modules/AdvancedPostProcessing.py
```diff
@@ -81,179 +89,296 @@ def is_float(istring):
...
+        self.log_scale = log_scale
+        self.units = coordinate_units
+        self.opacity = opacity
+        self.color_map = color_map
+        self._cached_mesh = None
+        self._cached_polydata = None
+        self.label = label
+        self.name = os.path.splitext(self.path)[0]
+        self.color = (255, 0, 0)
+        self.surface_mapping_tolerance = tol
```
MaxJPRey, line 142:
```suggestion
        self.surface_mapping_tolerance = tolerance
```

pyaedt/modules/AdvancedPostProcessing.py
```diff
@@ -81,179 +89,296 @@ def is_float(istring):
...
+    def background_color(self, value):
+        if isinstance(value, (tuple, list)):
+            self._background_color = value
+        elif value in CSS4_COLORS:
+            h = CSS4_COLORS[value].lstrip("#")
+            self._background_color = tuple(int(h[i : i + 2], 16) for i in (0, 2, 4))
+
+    @property
+    def fields(self):
+        """Field Object.
```
MaxJPRey, line 185:
```suggestion
        """List of fields object.
```

pyaedt/modules/AdvancedPostProcessing.py
```diff
@@ -81,179 +89,296 @@ def is_float(istring):
...
-        file_name = self.export_model_picture(show_axis=show_axis, show_grid=show_grid, show_ruler=show_ruler)
-        return Image(file_name, width=500)
+        return self._frames
 
-    @aedt_exception_handler
-    def get_efields_data(self, setup_sweep_name="", ff_setup="Infinite Sphere1", freq="All"):
-        """Compute Etheta and EPhi.
+    @property
+    def objects(self):
+        """Field Object.
```
MaxJPRey, line 205: I agree that my description/suggestion is not of great help for the user neither.
```suggestion
        """List of class objects.
```

pyaedt/modules/AdvancedPostProcessing.py
```diff
@@ -81,179 +89,296 @@ def is_float(istring):
...
-            Name of the setup for computing the report. The default is ``""``, in
-            which case the nominal adaptive is applied.
-        ff_setup : str, optional
-            Far field setup. The default is ``"Infinite Sphere1"``.
-        freq : str, optional
-            The default is ``"All"``.
+        cad_path : str
+            Full path to the file.
+        cad_color : str or tuple
+            Can be a string with color name or a tuple with (r,g,b) values.
```
MaxJPRey, line 222:
```suggestion
            Can be a string with color name or a tuple with (r,g,b) values.
            The default value is ``"dodgerblue"``.
```

pyaedt/modules/AdvancedPostProcessing.py
```diff
@@ -81,179 +89,296 @@ def is_float(istring):
...
+        opacity : float
+            Value between 0 to 1 of opacity.
```
MaxJPRey, line 224:
```suggestion
            Value between 0 to 1 of opacity. The default value is ``1``.
```

pyaedt/modules/AdvancedPostProcessing.py
```diff
@@ -81,179 +89,296 @@ def is_float(istring):
...
+        units : str
+            Model units.
```
MaxJPRey, line 226:
```suggestion
            Model units. The default value is ``"mm"``.
```

pyaedt/modules/AdvancedPostProcessing.py
```diff
@@ -81,179 +89,296 @@ def is_float(istring):
...
         ----------
-        ff_data :
-
-        xphase : float, optional
-            Phase in the X-axis direction. The default is ``0``.
-        yphase : float, optional
-            Phase in the Y-axis direction. The default is ``0``.
-
+        field_files : list
+            list of full path to frame file.
```
MaxJPRey, line 302:
```suggestion
            List of full path to frame file.
```

pyaedt/modules/AdvancedPostProcessing.py
```diff
@@ -81,179 +89,296 @@ def is_float(istring):
...
+        coordinate_units : str
+            Fields coordinates units.
+        opacity : float
+            Value between 0 to 1 of opacity.
+        color_map : str
+            Color map of field plot. Default rainbow.
+        label_name : str, optional
+            Name of the field.
+        surface_mapping_tolerance : float, optional
+            Delauny tolerance value used for interpolating points.
```
MaxJPRey, line 366:
```suggestion
            Delauny tolerance value used for interpolating points.

```

pyaedt/modules/AdvancedPostProcessing.py
maxcapodi78: the color setter is handling both a tuple or a string "white". That's why you have to use the setter and not instantiate directly the internal variable

MaxJPRey: ok, got it.

pyaedt/modules/AdvancedPostProcessing.py
```diff
@@ -81,179 +89,331 @@ def is_float(istring):
...
-             Whether to show the axes. The default is ``True``.
-         show_grid : bool, optional
-             Whether to show the grid. The default is ``True``.
-         show_ruler : bool, optional
-             Whether to show the ruler. The default is ``True``.
+    Examples
+    --------
+    This Class can be instantiated within Pyaedt (with plot_model_object or different field plots
+    and standalone.
+    Hera an example of standalone project
```
MaxJPRey, line 156:
```suggestion
    Here an example of standalone project
```

MaxJPRey (approved):

PR #727, opened 2022-01-18
The name of the udp can now be different from the dll's one.

### LINKED ISSUE #647: M2D.modeler.primitives.create_udp function does not assign name to the created object
Dear All,

while using the function:

M2D.modeler.primitives.create_udp

the udp creates the desired object correctly but assigns automatically the udp default name. The optional parameter name is not used further in the function implementation (I saw obj_name is used and maybe this is the problem).

Also the function to rename objects is missing, analogous to the API:

oEditor.ChangeProperty(
	[
		"NAME:AllTabs",
		[
			"NAME:Geometry3DAttributeTab",
			[
				"NAME:PropServers", 
				"VentSlotCore"
			],
			[
				"NAME:ChangedProps",
				[
					"NAME:Name",
					"Value:="		, "Stator"
				]
			]
		]
	])

Thank you for your help,

Tiziana

MaxJPRey:
The name of the udp can now be different from the dll's one.
Previously, the name provided by the user was never used. Only thename of the udp from the dll was used.

Fix #647.


_unittest/test_28_Maxwell3D.py
```diff
@@ -154,6 +154,7 @@ def test_26_create_udp(self):
         mypair = ["InfoCore", "0"]
         my_udpPairs.append(mypair)
 
+        # Test udp with default with a custom name.
```
MaxJPRey, line 157:
```suggestion
        # Test udp with a custom name.
```

maxcapodi78 (approved):
LGTM

PR #735, opened 2022-01-19
Improved Terminal Name assignment.

maxcapodi78:
Improved Terminal Naming in HFSS Port Creation

### DIFF
--- a/pyaedt/application/Analysis.py
+++ b/pyaedt/application/Analysis.py
@@ -480,6 +480,28 @@ class Analysis(Design, object):
         return SOLUTIONS()
 
     @aedt_exception_handler
+    def get_excitations_name(self):
+        """Get all excitation names.
+
+        Returns
+        -------
+        list
+            List of excitation names. Excitations with multiple modes will return one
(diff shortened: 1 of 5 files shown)


pyaedt/application/Analysis.py
```diff
@@ -479,6 +479,28 @@ def SolutionTypes(self):
...
+            excitation for each mode.
+
+        References
+        ----------
+
+        >>> oModule.GetExcitations
+        """
+        try:
+            list_names = list(self.oboundary.GetExcitations())
+            del list_names[1::2]
```
MaxJPRey, line 499: @maxcapodi78 
Out of curiosity those items of the list?

MaxJPRey (approved):

MaxJPRey:
LGTM.

pyaedt/application/Analysis.py
maxcapodi78: because each excitations return a terminal and and excitation. This is not properly correct. We should consider refactor since on multiple terminals ports that could be wrong

PR #739, opened 2022-01-19
Primitive functions

ring630:


pyaedt/edb_core/layout.py
```diff
@@ -536,6 +536,43 @@ def create_polygon(self, main_shape, layer_name, voids=[], net_name=""):
                     self._primitives_by_layer[layer_name] = [polygon]
             return polygon
 
+    @aedt_exception_handler
+    def get_primitive(self, net_name=None, layer_name=None, prim_type=None, is_void=False):
```
maxcapodi78, line 540: maybe we can rename the method get_primitives since it is returning a list of primitives.

maxcapodi78 (approved):
LTGM. Maybe I would only change the method name

pyaedt/edb_core/layout.py
ring630: Done.

MaxJPRey (approved):

PR #748, opened 2022-01-20
JamesYves action does not suit the need for the Full Documentation workflow.

MaxJPRey:
JamesYves action is primarily developed for Linux.
To use it on Windows you must first upload artifacts with a Windows platform and then downloads those artifacts with Linux.
Switching platform during the workflow is not optimal for our requirements.


maxcapodi78 (approved):
LGTM

PR #754, opened 2022-01-21
Enhancement/remote command

maxcapodi78:
Added 2 methods:
-common_rpc.launch_ironpython_server to initialize and start ironpython server and, at the same time, launch and return a Cpython client.
- the client has now the method convert_remote_object to convert list and dictionaries returned by IPY to Cpython


Alberto-DM (approved):
LGTM

PR #760, opened 2022-01-24
Fix, port renorm option is respected also in Driven Terminal

Alberto-DM:
_create_port_terminal had no "renorm" parameter. Consequently the renorm option was ignored if the solution type was Driven Terminal.


MaxJPRey:
@Alberto-DM, can you just add a unit test for the new argument `renorm`?
Otherwise LGTM.

MaxJPRey (approved):

PR #762, opened 2022-01-24
Model depth for Maxwell 2D transient analysis is now a property with getter/setter.

### LINKED ISSUE #753: Method missing: set model depth in Maxwell 2D
Dear All,

I could find the method: Maxwell2d.get_model_depth()
 but not the way to set the model depth for 2D transient analysis.
In the GUI it is under:

Maxwell2D-> Design Settings->Model Settings
Thank you very much,

Tiziana

MaxJPRey:
Model depth for Maxwell 2D transient analysis is now a property with getter/setter.
Add unit test for model depth property.

Fix #753 .

### DIFF
--- a/pyaedt/maxwell.py
+++ b/pyaedt/maxwell.py
@@ -1159,9 +1159,22 @@ class Maxwell2d(Maxwell, FieldAnalysis2D, object):
     def xy_plane(self, value=True):
         self.design_solutions.xy_plane = value
 
-    @aedt_exception_handler
-    def get_model_depth(self):
-        """Get model depth."""
+    @property
+    def model_depth(self):
+        """Get model depth.
+
+        Example
+        -------
+
+        Set and get the model depth for a 2D Maxwell analysis.
+
+        >>> from pyaedt import Maxwell2d
+        >>> maxwell_2d = Maxwell2d()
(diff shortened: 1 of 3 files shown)


pyaedt/maxwell.py
```diff
@@ -1172,6 +1185,29 @@ def get_model_depth(self):
...
+        >>> maxwell_2d.model_depth
+        0.09
+        """
+
+        try:
+            depth_value = float_units(value)
+        except:
+            depth_value = self.variable_manager[value].value
+
+        self.design_properties["ModelDepth"] = value
```
maxcapodi78, line 1209: @MaxJPRey this way is not gonna work. design_properties contains the .aedt file parsing information. It's a readonly information. You cannot update it from it. You need to check in AEDT what is the relative command to assign model depth.

maxcapodi78 (changes requested):
The setter is not gonna work in this way

pyaedt/maxwell.py
MaxJPRey: OK. The get was previously just extracting info from the aedt file. It seems that this property should be on the design. I will modify that.
Thanks @maxcapodi78

PR #785, opened 2022-01-27
Fixed Hfss Default Solution Type

maxcapodi78:


pyaedt/application/design_solutions.py
```diff
@@ -542,17 +534,88 @@ def __init__(self, odesign, design_type, aedt_version):
...
+        if self._odesign:
+            try:
+                self._solution_type = self._odesign.GetSolutionType()
+                if "Modal" in self._solution_type:
+                    self._solution_type = "Modal"
+                elif "Terminal" in self._solution_type:
+                    self._solution_type = "Terminal"
+            except:
+                self._solution_type = solutions_defaults[self._design_type]
+        elif self._solution_type is None:
```
MaxJPRey, line 549: @maxcapodi78 I am not sure if it is a `if` or `elif` here because we are not testing the same value. But I think what you did also works.
```suggestion
        if self._solution_type is None:
```

pyaedt/application/design_solutions.py
```diff
@@ -542,17 +534,88 @@ def __init__(self, odesign, design_type, aedt_version):
...
+                except:
+                    self._solution_type = solutions_defaults[self._design_type]
+            else:
+                self._solution_type = solutions_defaults[self._design_type]
+        elif value and value in self._solution_options and self._solution_options[value]["name"]:
+            if value == "Transient":
+                value = "Transient Network"
+                self._solution_type = "Transient Network"
+            elif value == "DrivenModal":
+                value = "Modal"
```
MaxJPRey, line 588: @maxcapodi78 I think it is not used after.
```suggestion
```

pyaedt/application/design_solutions.py
```diff
@@ -542,17 +534,88 @@ def __init__(self, odesign, design_type, aedt_version):
...
+                        self._solution_type = "Modal"
+                    elif "Terminal" in self._solution_type:
+                        self._solution_type = "Terminal"
+                except:
+                    self._solution_type = solutions_defaults[self._design_type]
+            else:
+                self._solution_type = solutions_defaults[self._design_type]
+        elif value and value in self._solution_options and self._solution_options[value]["name"]:
+            if value == "Transient":
+                value = "Transient Network"
```
MaxJPRey, line 585: @maxcapodi78 I think it is not used after.
```suggestion
```

pyaedt/application/design_solutions.py
```diff
@@ -542,17 +534,88 @@ def __init__(self, odesign, design_type, aedt_version):
...
+                self._solution_type = "Transient Network"
+            elif value == "DrivenModal":
+                value = "Modal"
+                self._solution_type = "Modal"
+            elif value == "DrivenTerminal":
+                value = "Terminal"
```
MaxJPRey, line 591: @maxcapodi78 I think it is not used after.
```suggestion
```

MaxJPRey (approved):

MaxJPRey:
Well done @maxcapodi78 , especially with the solution type string that depends on the AEDT version.

pyaedt/application/design_solutions.py
maxcapodi78: it is intended.

PR #788, opened 2022-01-27
Improve the test of create_component_from_pins()

### LINKED ISSUE #783: create_component_from_pins doesn't work on my machine
Below code is from test_54_create_component_from_pins in test_EDB.py. The unitest passes, but it returns false on my machine. 

pins = self.edbapp.core_components.get_pin_from_component("R13")
print(self.edbapp.core_components.create_component_from_pins(pins, "newcomp"))

result
-----------------------------
(False, None)

MaxJPRey:
Improve the test and its assertions for the component creation.
Make sure that the tuple returned by `create_component_from_pins` is correct.
Fix #783


maxcapodi78 (approved):
LGTM

PR #790, opened 2022-01-28
Add Points to the primitives in the modeler

### LINKED ISSUE #747: CreatePoint not available in Pyaedt
I believe createpoint is not yet available in Pyaedt
image

The below is a workaround but could this feature be added?
oDesign=M3D.oproject.GetActiveDesign()
oEditor=oDesign.SetActiveEditor("3D Modeler")
oEditor.CreatePoint (["NAME:PointParameters","PointX:=","-165mm","PointY:=","-245mm","PointZ:=","0mm"],
    ["NAME:Attributes","Name:=","Point1","Color:=","(143 175 143)"])

MaxJPRey:
It is now possible to add point primitives.
`Point` is different from `Object3d` because it doesn't have any ids.
Points are identified by their name.
The color is handled differently because they have no material (material are used for `Object3d` to set the color if none is provided when creating the object).
It is not possible to change the color of the point because of the "Edit" button.

Points can be accessed from primitives either by a list of a dictionary (where the point's name is the key).

Fix #747 .


pyaedt/modeler/Object3d.py
```diff
@@ -3197,3 +3197,232 @@ def set_net_name(self, netname=""):
...
+                changed_property = ["NAME:ChangedProps", name_property]
+                property_servers = ["NAME:PropServers"]
+                property_servers.append(self._name)
+                point_tab = ["NAME:Geometry3DPointTab", property_servers, changed_property]
+                all_tabs = ["NAME:AllTabs", point_tab]
+                _retry_ntimes(10, self._primitives._oeditor.ChangeProperty, all_tabs)
+                self._name = point_name
+                self._primitives.cleanup_objects()
+        else:
+            # TODO check for name conflict
```
maxcapodi78, line 3287: Here I would suggest, at the moment, a logger.warning that name exists

pyaedt/modeler/Primitives3D.py
```diff
@@ -45,6 +47,56 @@ def is3d(self):
...
+            List of ``[x, y, z]`` coordinates. Note, The list can be empty or contain less than 3 elements.
+        name : str, optional
+            Name of the point. The default is ``None``, in which case the
+            default name is assigned.
+        color : str, optional
+            String exposing 3 int values such as "(value1 value2 value3)". Default value is ``"(143 175 143)"``.
+
+        Returns
+        -------
+        :class:`pyaedt.modeler.Object3d.Object3d`
```
maxcapodi78, line 66: Are you sure it returns Object3d and not Point Object?

pyaedt/modeler/Primitives3D.py
```diff
@@ -45,6 +47,56 @@ def is3d(self):
...
+        parameters = ["NAME:PointParameters"]
+        parameters.append("PointX:="), parameters.append(x_position)
+        parameters.append("PointY:="), parameters.append(y_position)
+        parameters.append("PointZ:="), parameters.append(z_position)
+
+        attributes = ["NAME:Attributes"]
+        attributes.append("Name:="), attributes.append(name)
+        attributes.append("Color:="), attributes.append(color)
+
+        point = _retry_ntimes(10, self._oeditor.CreatePoint, parameters, attributes)
```
maxcapodi78, line 97: is the CreatePoint method returning something (like a boolean)? If so we could use it to detect if method executed correctly or not

maxcapodi78 (approved):
LGTM

pyaedt/modeler/Primitives3D.py
MaxJPRey: @maxcapodi78 Great catch!

pyaedt/modeler/Object3d.py
MaxJPRey: Done.

pyaedt/modeler/Primitives3D.py
MaxJPRey: After verification, the method CreatePoint() returns None.

PR #791, opened 2022-01-28
Added method to plot Edb in Linux with Client Server mode

maxcapodi78:


pyaedt/edb_core/nets.py
```diff
@@ -356,24 +334,50 @@ def plot(
...
-        end_time = time.time() - start_time
-        self._logger.info("Plot Generation time %s seconds", round(end_time, 3))
-        if save_plot:
-            plt.savefig(save_plot)
-        else:
-            plt.show()
+        Parameters
+        ----------
+        nets : str, list
+            Name of the net or list of nets to plot. If `None` all nets will be plotted.
```
MaxJPRey, line 354: @maxcapodi78 Maybe the default value should be a string: "all" instead of None to indicate that by default all layers or nets are plotted.

pyaedt/edb_core/nets.py
```diff
@@ -356,24 +334,50 @@ def plot(
...
+        show_legend : bool, optional
+            If `True` the legend is shown in the plot. (default)
+            If `False` the legend is not shown.
+        save_plot : str, optional
+            If `None` the plot will be shown.
+            If a file path is specified the plot will be saved to such file.
+        outline : list, optional
+            List of points of the outline to plot.
+        size : tuple, optional
+            Image size in pixel (width, height).
```
MaxJPRey, line 369:
```suggestion
            Image size in pixel (width, height). Default value is ``(2000, 1000)``
```

pyaedt/edb_core/nets.py
maxcapodi78: This is a good idea but maybe we should think to a biggest refactoring since this approach (with None) is widely used in pyaedt and users expect the same behaviour.

PR #802, opened 2022-02-01
Added new method for bounding box calculation

Alberto-DM:
A new method to evaluate the object bounding box has been implemented. 
It exports the object as .sat file and reads the bounding box values from the .sat file. 
It is faster by a factor of 9 and safer because it does not change the object properties. 
The new implementation rolls back to the old method for the student version (no export available) and in case of errors.


maxcapodi78 (approved):
Great Refactoring

PR #822, opened 2022-02-04
Added create_fan method to icepak.py

maxcapodi78:


ansys-icepak:
@maxcapodi78 

Tested on my end and works great, thanks!

PR #832, opened 2022-02-07
Added method to create dynamic link subcircuit valid for Q3d, Q2d and HFSS

maxcapodi78:


pyaedt/modeler/PrimitivesNexxim.py
```diff
@@ -1117,6 +1117,155 @@ def add_subcircuit_hfss_link(
...
+    ):
+        """Add a subcircuit 2D Extractor link.
+
+        Parameters
+        ----------
+        pyaedt_app : :class:`pyaedt.q3d.Q3d` or :class:`pyaedt.q3d.Q2d` or :class:`pyaedt.q3d.Hfss`
+            pyaedt application object to include. It could be an Hfss object, a Q3d object or a Q2d.
+        solution_name : str, optional
+            Name of the solution and sweep. The
+            default is ``"Setup1 : Sweep"``.
```
MaxJPRey, line 1189:
```suggestion
            Name of the solution and sweep. The default is ``"Setup1 : Sweep"``.
```

pyaedt/modeler/PrimitivesNexxim.py
```diff
@@ -1117,6 +1117,155 @@ def add_subcircuit_hfss_link(
...
+            comp_name=comp_name,
+            pin_names=pin_names,
+            source_project_path=source_project_path,
+            source_design_name=source_design_name,
+            solution_name=solution_name,
+            image_subcircuit_path="",
+            model_type=pyaedt_app.design_type,
+            variables=variables,
+            extrusion_length_q2d=extrusion_length,
+            matrix=matrix,
```
MaxJPRey, line 1253: @maxcapodi78 I might be wrong but is seems that matrix will always be `None`, as it is set to `None` line 1210 and never changed after that. Is it expected?

MaxJPRey (approved):

pyaedt/modeler/PrimitivesNexxim.py
Samuelopez-ansys: @maxcapodi78 , I agree with @MaxJPRey, matrix should be obtained before the call of _add_subcircuit_link(...), I think with the new PR (#834) this should be accessible

PR #833, opened 2022-02-07
Enhancement/nfd

maxcapodi78:


Samuelopez-ansys (approved):

pyaedt/generic/near_field_import.py
```diff
@@ -0,0 +1,164 @@
...
+    commented_header_line += 'Hx(real, imag), Hy(real, imag), Hz(real, imag)\n'
+
+    with open(nfd_full_file, 'w') as file:
+        writer = csv.writer(file, delimiter=',', lineterminator='\n')
+        file.write(commented_header_line)
+        file.write('Frequencies 1\n')
+        file.write('Frequency ' + str(frequency) + 'GHz\n')
+        writer.writerows(full_data)
+
+    print('.nfd file written to %s' % nfd_full_file)  # Prints if running ipy64 through external editor
```
akaszynski, line 135: Should log here.

pyaedt/generic/near_field_import.py
```diff
@@ -0,0 +1,164 @@
+import os
+import re
+import csv
+import glob
+
+
+class BoxFacePointsAndFields(object):  # pragma: no cover
```
akaszynski, line 7: Both this class and `convert_nearfield_data` the next should have coverage.

maxcapodi78: good catch. Thanks I've forgot to remove once I've added UT.

PR #837, opened 2022-02-08
Removed modeler.primitives from unit tests and examples as it is deprecated.

maxcapodi78:
…ecated and now any methods from primitives can be called directly from modeler

eg. modeler.primitives.create_box becomes modeler.create_box

### DIFF
--- a/_unittest/test_02_2D_modeler.py
+++ b/_unittest/test_02_2D_modeler.py
@@ -45,8 +45,8 @@ class TestClass(BasisTest):
         assert self.aedtapp.modeler._omaterial_manager
 
     def test_create_rectangle(self):
-        rect1 = self.aedtapp.modeler.primitives.create_rectangle([0, -2, -2], [3, 8])
-        rect2 = self.aedtapp.modeler.primitives.create_rectangle(
+        rect1 = self.aedtapp.modeler.create_rectangle([0, -2, -2], [3, 8])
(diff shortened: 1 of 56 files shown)


_unittest/test_02_2D_modeler.py
```diff
@@ -124,14 +124,14 @@ def test_create_regular_polygon(self):
 
     @pytest.mark.skipif(config["build_machine"] or is_ironpython, reason="Not running in ironpython")
     def test_plot(self):
-        self.aedtapp.modeler.primitives.create_regular_polygon([0, 0, 0], [0, 2, 0])
-        self.aedtapp.modeler.primitives.create_regular_polygon(
+        self.aedtapp.modeler.create_regular_polygon([0, 0, 0], [0, 2, 0])
```
Alberto-DM, line 127: There are some lines, like this one, without an assert. Any reason why there is no assert? 
They are notified by codecov.
Maybe we can exclude all unit tests from code coverage.

Alberto-DM (approved):
I made one comment. Other than that it looks good to me.

_unittest/test_02_2D_modeler.py
maxcapodi78: because those one are needed to the test assertion which is a couple of lines below

MaxJPRey (approved):

MaxJPRey:
LGTM. Well Done @maxcapodi78 .

PR #839, opened 2022-02-08
Deprecate `_convert_list_to_ids`

Alberto-DM:
`_convert_list_to_ids` is being deprecated. Using `convert_to_selections` instead.


maxcapodi78 (approved):
LGTM

PR #848, opened 2022-02-10
Added short_component_pins to edb and parameters_values  and polygon_data to Padstack Instances

maxcapodi78:
short_component_pins method
parameters_values property for pin padstack definitition


Alberto-DM (approved):
LGTM

PR #850, opened 2022-02-14
Update an existing user-defined primitive (UDP).

### LINKED ISSUE #847: Improve handling of UDPs
Some suggestions:

1) udp_parameters_list is a required parameter. If none is specified, there is an error. I suggest that this is not required and in case none is specified, the default parameters are used. Currently, if at least one is specified, it works OK.  

2) Remove "name" from the input parameters, as it anyway does not work. The name can be changed later by using UDP.name.

3) Correct the spelling for udp_type parameter. It is written as udptye.

4) Currently, once a UDP is created, there is no way (at least I could not find) to change the UDP parameters. The only way would be use the original API, that is:
```
oEditor=UDP.m_Editor
oEditor.ChangeProperty(
	[
		"NAME:AllTabs",
		[
			"NAME:Geometry3DCmdTab",
			[
				"NAME:PropServers",
				"rotor:CreateUserDefinedPart:1"
			],
			[
				"NAME:ChangedProps",
				[
					"NAME:InfoCore",
					"Value:="		, "1"
				]
			]
		]
	])
```
So there is a workaround, but if there is a way to improve this and access the parameters directly, like UDP.InfoCore, that would be great.

MaxJPRey:
It is now possible to update an existing geometry object that was originally created using UDP.
Add unit tests for the UDP updates.
Fix #847 .

### DIFF
--- a/_unittest/test_28_Maxwell3D.py
+++ b/_unittest/test_28_Maxwell3D.py
@@ -186,6 +186,7 @@ class TestClass:
             upd_library="syslib",
             udptye="Solid",
         )
+
         assert second_udp
         assert second_udp.name == "ClawPoleCore"
         assert "ClawPoleCore" in udp._primitives.object_names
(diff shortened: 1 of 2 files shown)


maxcapodi78 (approved):
LGTM

PR #877, opened 2022-02-18
Fixed export_field_file and export_field_file_on_grid methods

maxcapodi78:


_unittest/test_12_PostProcessing.py
```diff
@@ -147,6 +147,30 @@ def test_07_export_fields_from_Calculator(self):
             intrinsics="5GHz",
         )
         assert os.path.exists(os.path.join(self.local_scratch.path, "Efield.fld"))
+        self.aedtapp.post.export_field_file_on_grid(
```
MaxJPRey, line 150:
```suggestion
        assert os.path.exists(os.path.join(self.local_scratch.path, "Efield.fld"))

        self.aedtapp.post.export_field_file_on_grid(
```

MaxJPRey (approved):

PR #884, opened 2022-02-21
Change parameter name from 'udptye' to 'udp_type' in create_udp() method.

MaxJPRey:
Change parameter name from 'udptye' to 'udp_type' in create_udp() method.


akaszynski (approved):
LGTM

PR #893, opened 2022-02-25
Handle the scenario if the Design name has or doesn't have a ";" in it

kbhagat-ansys:


kbhagat-ansys:
@MaxJPRey  could you check why one of the checkss didn't pass?
@maxcapodi78

MaxJPRey:
> @MaxJPRey could you check why one of the checkss didn't pass? @maxcapodi78

Hi @kbhagat-ansys ,
The issue is that you are modifying a line of code that is not tested by any of our unit tests. We recently added the code coverage (with a threshold of 90%) for every new patch. That means that at least 90% of the code you are modifying must be covered by unit test. It is true that it should already have been covered by tests in the past to avoid that.
I will work on your PR if you want to add this test and get all the check passing.

PR #894, opened 2022-02-25
3DComponent creation refactoring. 

Samuelopez-ansys:
Now users could define specific Objets, CS, boundaries, and excitations from the HFSS Design. By default, it will export all the objects/boundaries available in the Design


Samuelopez-ansys:
@maxcapodi78 , @MaxJPRey do you know why is Style Check failing? Thanks

PR #898, opened 2022-02-25
Clean up related to older class names for Twin Builder

SushobhitGupt:
Clean up related to older class names for Twin Builder
Renaming earlier classes referring to Simplorer


pyaedt/modeler/Circuit.py
```diff
@@ -314,25 +314,25 @@ def rotate(self, selections, degrees=90):
         return True
 
 
-class ModelerSimplorer(ModelerCircuit):
-    """ModelerSimplorer class.
+class ModelerTwinBuilder(ModelerCircuit):
```
MaxJPRey, line 317: @maxcapodi78 I am not sure if we need a deprecation warning here.

maxcapodi78: No, since it is an internal class and cannot be called by users.

maxcapodi78 (approved):
LGTM

PR #901, opened 2022-02-28
Add unit tests for HFSS and EDB.

MaxJPRey:
Add unit test for spiral lumped port with non-collinear objects.
Add test for edb with dxf.
It is impossible to create a wave port without faces. An exception is raised prior the creation.


pyaedt/hfss.py
```diff
@@ -3085,7 +3085,7 @@ def create_wave_port_from_sheet(
                 faces = sheet
             else:
                 faces = self.modeler.get_object_faces(sheet)[0]
-            if not faces:
+            if not faces:  # pragma: no cover
                 self.logger.error("Wrong Input object. it has to be a face id or a sheet.")
                 return False
```
akaszynski, line 3090: This should be a `TypeError`.

akaszynski (approved):
Approved with suggestions.

pyaedt/hfss.py
MaxJPRey: Agreed. It will be part of a global refactoring work.

PipKat (approved):
@MaxJPRey 
Hope I didn't hold you up.

MaxJPRey:
> @MaxJPRey Hope I didn't hold you up.

Not at all @PipKat . It even allowed @akaszynski and I to talk about a refactoring.

PR #902, opened 2022-02-28
Add Non Conservative Periodic Waveform Source for Twin Builder [SINE,…

SushobhitGupt:
… PULSE, TRIANG, SAWTOOTH]

### DIFF
--- a/pyaedt/modeler/PrimitivesTwinBuilder.py
+++ b/pyaedt/modeler/PrimitivesTwinBuilder.py
@@ -416,7 +416,7 @@ class TwinBuilderComponents(CircuitComponents):
             use_instance_id_netlist=use_instance_id_netlist,
         )
 
-        if type == "SINE" or type == "PULSE" or type == "TRIANG" or type == "SAWTOOTH":
+        if type in ["SINE", "PULSE", "TRIANG", "SAWTOOTH"]:
             id.set_property("AMPL", amplitude)
             id.set_property("FREQ", freq)
             id.set_property("PHASE", phase)
--- a/pyaedt/modeler/PrimitivesTwinBuilder.py
+++ b/pyaedt/modeler/PrimitivesTwinBuilder.py
@@ -356,3 +356,73 @@ class TwinBuilderComponents(CircuitComponents):
         )
 
         return id
+
+    @aedt_exception_handler
+    def create_periodic_waveform_source(
+        self,
+        compname=None,
+        type="SINE",
+        amplitude=100,
+        freq=50,
+        phase=0,
+        offset=0,
+        delay=0,
+        location=[],
+        angle=0,
+        use_instance_id_netlist=False,
+    ):
+        """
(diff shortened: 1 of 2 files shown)


pyaedt/modeler/PrimitivesTwinBuilder.py
```diff
@@ -356,3 +356,73 @@ def create_pnp(self, compname=None, location=[], angle=0, use_instance_id_netlis
...
+        location=[],
+        angle=0,
+        use_instance_id_netlist=False,
+    ):
+        """
+        Create a periodic waveform source (non conservative real output).
+
+        Parameters
+        ----------
+        compname : str, optional
```
MaxJPRey, line 379: @SushobhitGupt Should we rename this argument ``source_name`` as it is the name of the voltage source. I understand that we are creating a component behind the scene. It is just to be sure.

MaxJPRey (approved):

pyaedt/modeler/PrimitivesTwinBuilder.py
SushobhitGupt: @MaxJPRey compname is a known Twin Builder terminology, we probably would not want to change args based on what type of component is being placed (like sources in this case). keeping it same would keep consistency with other components defined in the file

MaxJPRey: @SushobhitGupt It makes sense.

pyaedt/modeler/PrimitivesTwinBuilder.py
```diff
@@ -356,3 +356,73 @@ def create_pnp(self, compname=None, location=[], angle=0, use_instance_id_netlis
...
+        id = self.create_component(
+            compname,
+            component_library="Basic Elements\\Tools\\Time Functions",
+            component_name=type,
+            location=location,
+            angle=angle,
+            use_instance_id_netlist=use_instance_id_netlist,
+        )
+
+        if type == "SINE" or type == "PULSE" or type == "TRIANG" or type == "SAWTOOTH":
```
akaszynski, line 419:
```suggestion
        if type in ["SINE", "PULSE", "TRIANG", "SAWTOOTH"]:
```

maxcapodi78 (approved):
@SushobhitGupt maybe you could have a look also to components_catalog object that is now in schematic and which provides you the full list of components in Twinbuilder syslib and as well the possibility to place it. Anyway I think that most common used components should have dedicated method. I like it

SushobhitGupt:
> @SushobhitGupt maybe you could have a look also to components_catalog object that is now in schematic and which provides you the full list of components in Twinbuilder syslib and as well the possibility to place it. Anyway I think that most common used components should have dedicated method. I like it

Thanks, I will check it out. As you mentioned, some commonly used components can be exposed so that we can create some good examples and tutorials. That is what my intent is here.

PR #907, opened 2022-03-02
Do not enforce global threshold

akaszynski:
Recommend only enforcing patch coverage. There will be times when you remove source and come out with more efficient code by "lose" coverage. We should only worry about patch.

BTW, great job getting above 80%!


MaxJPRey:
Make sense @akaszynski . LGTM.

MaxJPRey (approved):

PR #912, opened 2022-03-03
Added place_3d_component method to Hfss3dLayout

maxcapodi78:
Refactored Components3DLayout and Pins3dLayout classes replacing methods with properties where possible


maxcapodi78:
closing because the method added has been tested in Linux because it doesn't run in non-graphical on CICD.

PR #917, opened 2022-03-04
Add group property and setter to object3d

maxcapodi78:
group_name property was there but was bugged. I've fixed and added also a group setter


pyaedt/modeler/Object3d.py
```diff
@@ -1237,6 +1238,61 @@ def group_name(self):
             )
             return self._m_groupName
 
+    @group_name.setter
+    def group_name(self, name):
+        """Assign Object to a specific group. it creates a new group if group doesn't exists.
```
MaxJPRey, line 1243:
```suggestion
        """Assign Object to a specific group. it creates a new group if the group doesn't exist.
```

pyaedt/modeler/Object3d.py
```diff
@@ -1237,6 +1238,61 @@ def group_name(self):
...
+
+        Parameters
+        ----------
+        name : str
+            Name of the group to assign. Group will be created if not exists.
```
MaxJPRey, line 1248:
```suggestion
            Name of the group to assign. Group will be created if it does not exist.
```

MaxJPRey (approved):

MaxJPRey:
LGTM.

PR #918, opened 2022-03-04
Catch all exceptions raised when exporting ipc.

### LINKED ISSUE #871: export_to_ipc2581() method in edb always return True
`export_to_ipc2581()` method in edb always return `True`.
Instead of checking the return type, we will check for any exception raised during the export.

MaxJPRey:
Catch all exceptions raised when exporting ipc.
Previously we were not catching raised exception in the edlib assembly.

Fix #871.


_unittest/test_00_EDB.py
```diff
@@ -15,6 +15,7 @@
 
 try:
     import pytest
+    import unittest.mock
 except ImportError:
```
maxcapodi78, line 19:
```suggestion
except ImportError:  # pragma: no cover
```

maxcapodi78 (approved):
LGTM

PR #926, opened 2022-03-08
Fix Voids replacement after cutout is performed.

maxcapodi78:
Now also polygon voids intersecting cutout region are added.


Alberto-DM (approved):
LGTM

MaxJPRey (approved):

PR #927, opened 2022-03-08
Create helix from a polyline.

### LINKED ISSUE #914: create_helix() in Primitives3D does not work.
In the module named `Primitives3D`, the method named `create_helix()` is not working.

`vArg1.append("Selections:="), vArg1.append(o.name)` command line does not work as 'o' is not define.

MaxJPRey:
The method named `create_helix` was not tested and did not work.
It is now working properly.
Tests have been added accordingly.
Fix #914

### DIFF
--- a/_unittest/test_08_Primitives3D.py
+++ b/_unittest/test_08_Primitives3D.py
@@ -1211,8 +1211,6 @@ class TestClass(BasisTest, object):
                 z_start_dir=1.0,
             )
         except ValueError as exc_info:
-            import pdb
-            pdb.set_trace()
             assert "The name of the polyline cannot be an empty string." in str(exc_info.args[0])
         else:
             assert False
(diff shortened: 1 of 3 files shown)


_unittest/test_08_Primitives3D.py
```diff
@@ -1156,3 +1156,58 @@ def test_76_check_value_type(self):
...
+            polyline_name="helix_polyline_left",
+            position=[0, 0, 0],
+            x_start_dir=1.0,
+            y_start_dir=1.0,
+            z_start_dir=1.0,
+            right_hand=False,
+        )
+
+        # Test that an exception is raised if the name of the polyline is not provided.
+        with pytest.raises(ValueError) as exc_info:
```
Alberto-DM, line 1205: You can use a workaround like this to be compatible with IronPython:
```
try:
   function_that_will_crash
except:
   assert True
else:
   assert False
```
So you don't need to skip the test in Ironpython

MaxJPRey: @Alberto-DM Done. Thanks.

maxcapodi78 (approved):
LGTM

PR #935, opened 2022-03-09
Support touchstone file with more than 9 ports.

### LINKED ISSUE #928: create_touchsthone_component() is not working properly
Hello @maxcapodi78 

Said method is not working properly in case of touchstone component having more than 9 ports.

It is due to the following line (line no. 338 in modeler/PrimitivesCircuit.py).

`num_terminal = int(touchstone_full_path[-2:-1])`

A possible fix could be the following line.

`num_terminal = int(os.path.splitext(touchstone_full_path)[1].strip('.sp'))`

This worked in properly loading a 45 port touchstone model into Circuit.

MaxJPRey:
Support touchstone file with more than 9 ports.
Previously we were only selecting the last digit to get the number of ports contained in the touchstone file.

Fix #928

### DIFF
--- a/pyaedt/modeler/PrimitivesCircuit.py
+++ b/pyaedt/modeler/PrimitivesCircuit.py
@@ -335,7 +335,7 @@ class CircuitComponents(object):
             model_name = os.path.splitext(os.path.basename(touchstone_full_path))[0]
         if model_name in list(self.o_model_manager.GetNames()):
             model_name = generate_unique_name(model_name, n=2)
-        num_terminal = int(touchstone_full_path[-2:-1])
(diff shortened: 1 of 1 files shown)


jhillairet:
If not done before in the code, I would suggest to `lower()` the `touchstone_full_path` somewhere, as one can often sees Touchstone files with using upper cases extentions such as : `'.S2P'`

MaxJPRey:
> If not done before in the code, I would suggest to `lower()` the `touchstone_full_path` somewhere, as one can often sees Touchstone files with using upper cases extentions such as : `'.S2P'`

Great suggestion @jhillairet . It is done. Thanks.

PR #936, opened 2022-03-09
Place on target's outermost signal layer

svandenb-dev:
In place_in_layout_3d_placement use the signal layers rather than the
stackup layers of the target to determine the top/bottom placement layer
to handle cases where dielectric layers are outside the metal layers.
For example, this occurs when a soldermask or mold compound is present.


maxcapodi78 (approved):
LGTM

PR #938, opened 2022-03-10
Access edb_value from components using a method instead of a property.

### LINKED ISSUE #911: self._edb_value is not callable in component module
Currently `_edb_value` is a read-only property. A setter must be added to be able to set a new value.
Currently the code: `self._edb_value(0.0)` cannot work.

MaxJPRey:
Replace `_edb_value` protected property by a protected method `_get_edb_value()`.
Codacy tool was raising mentioning that `_edb_value` was not callable. The property was returning a function object and using it to call an edb method.
Even though it was valid, it could be confusing when maintaining code.

Fix #911


maxcapodi78:
@MaxJPRey I like the approach but we should standardize in edb_core classes. can you replace it in any classes in edb_core the property with the method?

MaxJPRey:
> @MaxJPRey I like the approach but we should standardize in edb_core classes. can you replace it in any classes in edb_core the property with the method?

Sure @maxcapodi78 I will do it.

maxcapodi78 (approved):
LGTM

PR #939, opened 2022-03-10
Enhancement props setters and getters

maxcapodi78:
All objects created and returned by pyaedt (setup, mesh, boundaries) return an object containing a dictionary named props.
Till now user that want to change an object in AEDT needs to do
object.props["propertyname"] = property_value
object.update()

From now on the props will be an automatic getter and setter for the property


pyaedt/edb_core/EDB_Data.py
```diff
@@ -2022,6 +2022,7 @@ def id(self):
 
     @property
     def name(self):
+        """Padstack Instance Name. If is a pin the syntax will be like in AEDT ComponentName-PinName."""
```
MaxJPRey, line 2025:
```suggestion
        """Padstack Instance Name. If it is a pin, the syntax will be like in AEDT ComponentName-PinName."""
```

MaxJPRey (approved):

PR #942, opened 2022-03-11
Fix 3D placement with solder height

isaacansys:
Use the source design's signal layers to calculate alignment to properly
account for solder height without separately calculating the height of
exterior dielectrics on the source design.  Also fix
addition/subtraction for the initial elevation calculation:

- when not flipping, subtract the bottom/top source elevation from the
  top/bottom target elevation when placing on top/bottom
- when flipping, add the top/bottom source elevation to the top/bottom
  target elevation when placing on top/bottom

Add two new test cases on a new model with a solder height defined.


maxcapodi78 (approved):
I've checked on a couple of cases and works. Thanks

PR #953, opened 2022-03-14
Pre-commit applied to unit test files.

### LINKED ISSUE #948: pre-commit is not activated for the unit tests.
The test files located under the repository "unittests" is not used when pre-commit is run.
They should be part of this process when committing.

MaxJPRey:
The test files (except the example models) are now part of the pre-commit control.
Fix #948 .


maxcapodi78 (approved):
LGTM. Good cleaning of import calls too

PR #954, opened 2022-03-14
added create_report method to replace create_rectangular plot method …

maxcapodi78:
…and extend funtionalities to 3d Plot (polar), smith charts and contour plots.


pyaedt/circuit.py
```diff
@@ -980,9 +981,10 @@ def create_touchstone_report(
             Name of the solution. The default value is ``None``.
         variation_dict : dict, optional
             Dictionary of variation names. The default value is ``None``.
+        differential_pairs : bool, optional
+            Specify if the plot is on differential pairs tracs. The default value is ``False``.
```
MaxJPRey, line 985:
```suggestion
            Specify if the plot is on differential pairs traces. The default value is ``False``.
```

maxcapodi78:
UT will be increased by @Samuelopez-ansys

PR #956, opened 2022-03-15
Add Unittest rmxprt and Circuit Create_Report method

Samuelopez-ansys:


maxcapodi78 (approved):
LGTM

MaxJPRey (approved):

_unittest/test_21_Circuit.py
```diff
@@ -80,9 +80,26 @@ def test_05c_create_component(self):
             "Test1", [1, 2], parameter_list=["Author:=", "NumTerminals:="], parameter_value=["pyaedt", 2]
         )
 
-    def test_06_add_3dlayout_component(self):
+    def test_06a_create_setup(self):
+        setup_name = "LNA"
+        LNA_setup = self.aedtapp.create_setup(setup_name)
+        assert LNA_setup
```
MaxJPRey, line 86: @Samuelopez-ansys I think that the variable is not mandatory here.
```suggestion
        assert self.aedtapp.create_setup(setup_name)
```

_unittest/test_21_Circuit.py
```diff
@@ -80,9 +80,26 @@ def test_05c_create_component(self):
...
+
+    def test_06b_add_3dlayout_component(self):
         myedb = self.aedtapp.modeler.schematic.add_subcircuit_3dlayout("Galileo_G87173_204")
         assert type(myedb.id) is int
+        ports = myedb.pins
+        tx = ports
+        rx = ports
```
MaxJPRey, line 93: @Samuelopez-ansys 
Is it expected to have `tx` and `rx` variables having the same value?

PR #960, opened 2022-03-16
Improve the polygon parametrization in the edb layout module.

### LINKED ISSUE #957: _edb_value() expect only one argument.
The number of arguments provided to the private method `_edb_value()` is not correct. 
The variable name `var_server` is not appropriate.

image

MaxJPRey:
`_get_edb_value()` only take one parameter.
`var_server` variable is not required and creating a crash if not removed.
The method named `calc_slope()` in the `edb_core\layout.py` module could not work properly.
Fix #957


maxcapodi78 (approved):
LGTM

PR #968, opened 2022-03-18
Improved all Examples by adding reports

maxcapodi78:


MaxJPRey (approved):

MaxJPRey:
LGTM.

MaxJPRey (changes requested):
@maxcapodi78 Some files in the IronPython package are modified. They should not. It might come from black tool when pre-commit is performed. I thought those files were excluded. I will check.

MaxJPRey:
> @maxcapodi78 Some files in the IronPython package are modified. They should not. It might come from black tool when pre-commit is performed. I thought those files were excluded. I will double check.

@maxcapodi78 I doubled check and the python files located under third-party are not affected by the pre-commit . We should not have any issue with that in the future. Here it might be due to our black formatting tool version conflict that we fixed couple days ago.

MaxJPRey (approved):

PR #970, opened 2022-03-18
S588850: Linked coupling pyaedt support

jianwei425:


examples/03-Circuit/EMIT_HFSS_Example.py
```diff
@@ -0,0 +1,101 @@
...
+
+temp_folder = os.path.join(tmpfold, ("EmitHFSSExample"))
+if not os.path.exists(temp_folder):
+    os.mkdir(temp_folder)
+
+example_name = "Cell Phone RFI Desense"
+example_aedt = example_name + ".aedt"
+example_lock = example_aedt + ".lock"
+
+example_project = os.path.join(local_path, "example_models", example_aedt)
```
maxcapodi78, line 51: @jianwei425 the examples files of published examples have not to be in that folder. On monday I can explain where to put it and to download from Examples.

pyaedt/emit_core/Couplings.py
```diff
@@ -0,0 +1,74 @@
+"""
```
maxcapodi78, line 1: I would suggest to not create another folder. I think this file could stay in modules or modeler folders.

maxcapodi78 (changes requested):
overall looks good. A couple of modifications are needed. Especially the example

pyaedt/emit_core/Couplings.py
jianwei425: @maxcapodi78 @myoung301, we thought we would group all EMIT only files under this emit_core foulder, this Coupling.py is the first one, there will be files for result nodes, and more

examples/03-Circuit/EMIT_HFSS_Example.py
jianwei425: decided to copy it from the EDT install and not saving a copy of the example file in pyaedt

PR #973, opened 2022-03-21
Remove the mutable from the defaults argument.

MaxJPRey:
Severe side effects can arose when using mutable objects as default argument.
Indeed the default values of the argument will be evaluated the first time the method is called.
In the subsequent method calls, the same value is used.


akaszynski (approved):
LGTM.

Issue #946, opened 2022-03-14
Add edb.core_components.set_component_type() in edb application


linmingchih:
Allow users to change component type between IO, Other, and IC.

ring630:
@linmingchih implemented. Please try

```
comp = self.edbapp.core_components.components["R2L18"]
comp.type = "Capacitor"
```

PR #977, opened 2022-03-22
Set component type

ring630:
#946


MaxJPRey (approved):

ring630:
@maxcapodi78 it is failing ironpython. I should I skip ironpython for the test?

PR #980, opened 2022-03-22
Fix get_component_placement_vector

isaacansys:
Remove the usage of the merged component's center location in
get_component_placement_vector for the offset calculation.  Its return
values are expected to be passed into place_in_layout_3d_placement, and
this function applies the rotation around the origin.

This change corrects the offset calculation when the merged component
is not centered about the origin, for example when the component has
been cut out from a larger design.


maxcapodi78 (approved):
Lgtm

PR #983, opened 2022-03-23
made more robust the launch_ironpython_server. now it is checking the aedt_path

maxcapodi78:
…  and it is checking if connection to the client is successful before initializing it.


pyaedt/common_rpc.py
```diff
@@ -244,6 +245,9 @@ def client(server_name, server_port=18000, beta_options=None):
         except:
             t -= 1
             time.sleep(1)
+    if not c:
+        print("Failing to connect to {} on port {}. Check the settings".format(server_name, server_port))
```
akaszynski, line 249: These should raise errors.

akaszynski, line 249: These should raise errors.

pyaedt/common_rpc.py
```diff
@@ -403,6 +407,9 @@ def launch_ironpython_server(aedt_path, non_graphical=False, port=18000, launch_
         str(val),
         str(port1),
     ]
+    if not os.path.exists(os.path.join(aedt_path, "common", "IronPython", "ipy64.exe")):
+        print("Check the aedt_path and retry.")
+        return False
     proc = subprocess.Popen(" ".join(command), shell=True)
     print("Process {} started on {}".format(proc.pid, socket.getfqdn()))
     print("Using port {}".format(port1))
```
akaszynski, line 415: This should be logged.

akaszynski, line 415: This should be logged.

maxcapodi78:
@akaszynski That's is the only part of pyaedt not logget because it is the linux client_server. this cannot be logged because nothing is started yet. the logs starts after the client is up and running

pyaedt/common_rpc.py
```diff
@@ -1,22 +1,27 @@
...
+
 if is_ironpython:
     pyaedt_path = os.path.normpath(os.path.abspath(os.path.dirname(__file__)))
-    sys.path.insert(0, os.path.join(pyaedt_path, "third_party", "ironpython"))
-
-import rpyc
-from rpyc.utils.server import ThreadedServer
-from pyaedt.rpc.rpyc_services import GlobalService, check_port
-import rpyc.core.consts
+    # sys.path.insert(0, os.path.join(pyaedt_path, "third_party", "ironpython"))
```
MaxJPRey, line 12:
```suggestion
```

MaxJPRey (approved):

maxcapodi78:
> @maxcapodi78 If we decide to go with the log option instead of the print. We cannot use the AEDT message logger because it won't be available. We could use the root logger instead. Something like:
> 
> ```
> import logging
> logger = logging.getLogger(__name__)
> logger.setLevel(logging.CRITICAL)
> ```

@MaxJPRey If it is ok for you we can take in another PR. We need to test the log and see if no issues are in both Windows and Linux, CPython and IPY

PR #987, opened 2022-03-23
Fixed assign surface mesh manual flag when surface deviation is disabled

maxcapodi78:


pyaedt/modules/Mesh.py
```diff
@@ -350,9 +350,9 @@ def assign_surface_mesh_manual(self, names, surf_dev=None, normal_dev=None, aspe
         ----------
         names : list
             List of faces to apply the surface mesh to.
```
jeremich41, line 352: If we put a list of object it is working too.

jeremich41: So the doc can be "List of faces or directly objects to apply the surface mesh to."?

MaxJPRey (approved):

pyaedt/modules/Mesh.py
```diff
@@ -382,9 +382,10 @@ def assign_surface_mesh_manual(self, names, surf_dev=None, normal_dev=None, aspe
         aspect_ratio_enable = 2
 
         if not surf_dev:
+            surf_dev_enable = 0
```
MaxJPRey, line 385: @maxcapodi78 I added some unit tests for the default value while "playing" with this feature and others.

maxcapodi78: great thanks

MaxJPRey (approved):

PR #988, opened 2022-03-24
The default value of mutable arguments is None.

MaxJPRey:
Mention in the methods docstring that the default value of mutable arguments is `None`.


maxcapodi78 (approved):
LGTM

PR #992, opened 2022-03-24
Fix emit example

maxcapodi78:


examples/03-Circuit/EMIT_HFSS_Example.py
```diff
@@ -12,23 +12,21 @@
...
 from pyaedt import Desktop
 
+
 ###############################################################################
 # Initialization Settings
 # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 # Change NonGraphical Boolean to False to open AEDT in graphical mode
 # With NewThread = False, an existing instance of AEDT will be used, if
-# available. This example will use AEDT 2022.2
+# available. This example will use AEDT 2022.1
```
myoung301, line 24: @maxcapodi78 - I think we should add the same note you made in the other example about this working on 2022.2.
```suggestion
# available. This example will use AEDT 2022.1 However this example is supposed to work
# on AEDT 2022R2 and on.
```

PR #1002, opened 2022-03-27
Read IBIS file from circuit class.

### LINKED ISSUE #966: Ibis reader is not working from circuit.
Ibis reader is not working from circuit.
The `filename` is a required parameter and is currently missing.
The method `read_project` in the reader has been renamed `parse_ibis_file`

MaxJPRey:
It is now possible to extract all ibis file model info directly from `Circuit` class.

Fix #966 .

### DIFF
--- a/pyaedt/circuit.py
+++ b/pyaedt/circuit.py
@@ -372,7 +372,7 @@ class Circuit(FieldAnalysisCircuit, object):
         return True
 
     @pyaedt_function_handler()
-    def read_ibis(self, path):
+    def get_ibis_model_from_file(self, path):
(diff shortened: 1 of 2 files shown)


pyaedt/circuit.py
```diff
@@ -372,7 +372,7 @@ def create_schematic_from_netlist(self, file_to_import):
         return True
 
     @pyaedt_function_handler()
-    def read_ibis(self, path):
+    def get_ibis_model_from_file(self, path):
```
MaxJPRey, line 375: @maxcapodi78 I don't think we need any deprecation warning here as the method was not working properly.

akaszynski (approved):

PR #1005, opened 2022-03-28
remove coverage of unit tests

akaszynski:
Coverage of unit tests should not be reported.


Alberto-DM (approved):
I agree with the change.

MaxJPRey (approved):

PR #1009, opened 2022-03-28
Fix bug for flipped mount when rotation zero

isaacansys:
get_component_placement_vector would return an incorrect Y-offset when
flipped was True and the rotation angle was calculated to be zero.  Add
a test case for and fix this by modifying the merged component pin
locations to account for the flip around the X-axis (Y-value changes
sign) so that the correct flipped values propagate through the rest of
the calculations.


pyaedt/edb_core/components.py
```diff
@@ -467,21 +467,21 @@ def get_component_placement_vector(
...
         rotation = GeometryOperators.v_angle_sign_2D(vector1, vector2, False)
         if rotation != 0.0:
-            x_v2 = m_pin1_pos[0] * math.cos(rotation) + multiplier * m_pin1_pos[1] * math.sin(rotation)
-            y_v2 = -1 * m_pin1_pos[0] * math.sin(rotation) + multiplier * m_pin1_pos[1] * math.cos(rotation)
-            new_vector = [x_v2, y_v2]
-            vector = [h_pin1_pos[0] - new_vector[0], h_pin1_pos[1] - new_vector[1]]
+            x_v2 = m_pin1_pos[0] * math.cos(rotation) + m_pin1_pos[1] * math.sin(rotation)
+            y_v2 = -1 * m_pin1_pos[0] * math.sin(rotation) + m_pin1_pos[1] * math.cos(rotation)
+            offset_from = [x_v2, y_v2]
+        vector = [h_pin1_pos[0] - offset_from[0], h_pin1_pos[1] - offset_from[1]]
```
Alberto-DM, line 484: Now that the flipped condition is fixed and the `multiplier` has been removed from the rotation equation, I think that the `if rotation != 0.0:` condition is not necessary anymore. 
In fact, if `rotation=0.0`,  `x_v2 = m_pin1_pos[0]` and  `y_v2 = m_pin1_pos[1]` which are the same as line 478. 

It seems that everything could be reduced to:
```
if flipped:
    m_pin1_pos[1] *= -1.0
    m_pin2_pos[1] *= -1.0

vector1 = GeometryOperators.v_points(m_pin1_pos, m_pin2_pos)
vector2 = GeometryOperators.v_points(h_pin1_pos, h_pin2_pos)

rotation = GeometryOperators.v_angle_sign_2D(vector1, vector2, False)

x_v2 = m_pin1_pos[0] * math.cos(rotation) + m_pin1_pos[1] * math.sin(rotation)
y_v2 = -1 * m_pin1_pos[0] * math.sin(rotation) + m_pin1_pos[1] * math.cos(rotation)

vector = [h_pin1_pos[0] - x_v2, h_pin1_pos[1] - y_v2]
```
To be tested of course.

Alberto-DM (commented):
Maybe the code can be further simplified.

pyaedt/edb_core/components.py
isaacansys: Good catch, will implement this.

maxcapodi78 (approved):
LGTM

PR #1010, opened 2022-03-28
Add number_aware_string_key

isaacansys:
Add a method to calculate keys for list.sort and sorted from strings
that treats embedded strings of digits as integers.  This is useful to
produce an "expected" ordering of component names, pin names, etc.  For
example, we expect "U2" to be sorted before "U10" in such a list but the
default sort behavior would have the opposite ordering.


maxcapodi78 (approved):
LGTM

akaszynski:
Style check failure can be ignored, will be fixed in #1012.

PR #1033, opened 2022-04-12
RMxprt Setup templates added for multiple machine types

pblarsen-ansys:
Added setup templates for standard RMxprt machine types.  Additional unittests within test_32_RMxprt provide some examples of updates for different machine types that can successfully create_setup with predefined templates.


_unittest/test_32_RMxprt.py
```diff
@@ -28,7 +28,28 @@ def test_02_changeesolution(self):
...
+        # second test ASSM setup
+        assert self.aedtapp.disable_modelcreation("ASSM")
+        mysetup = self.aedtapp.create_setup()
+        mysetup.props["RatedSpeed"] = "3600rpm"
+        assert mysetup.update()  # update only needed for assertion
+        # third test TPSM/SYNM setup
+        assert self.aedtapp.disable_modelcreation("TPSM")
+        mysetup = self.aedtapp.create_setup()
+        mysetup.props["RatedVoltage"] = "208V"
+        assert mysetup.update()  # update only needed for assertion
```
MaxJPRey, line 55: @pblarsen-ansys Is it possible to check the setup properties value after the setup has been updated?

pblarsen-ansys: Yes, it's possible and works as expected.  I pushed a quick change that checks that the property values are set.  I can return later and improve this by checking that the value is identical to the template value for each machine type.

pblarsen-ansys: @MaxJPRey I made this change, and all the verification checks passed.  Let me know if there is anything else that you suggest.

svandenb-dev (approved):

_unittest/test_32_RMxprt.py
MaxJPRey: Thanks @pblarsen-ansys . It sounds good to me.

PR #1046, opened 2022-04-15
Handling Maxwell Circuit Variables

maxcapodi78:


pyaedt/application/Design.py
```diff
@@ -3152,6 +3153,17 @@ def get_evaluated_value(self, variable_name, variation=None, units=None):
...
         >>> eval_p3 = M3D.get_evaluated_value("p3")
         """
+        if self.design_type == "Maxwell Circuit":
+            if "$" in variable_name:
+                val_units = self._oproject.GetVariableValue(variable_name)
+            else:
+                val_units = self._odesign.GetVariableValue(variable_name)
+            val, units = decompose_variable_value(val_units)
+            try:
+                float(val)
```
MaxJPRey, line 3163: @maxcapodi78 I am not sure it is mandatory as we are returning the expression the line after.
```suggestion
```

MaxJPRey (approved):

pyaedt/application/Design.py
maxcapodi78: not sure either.... I'll check better

PR #1048, opened 2022-04-15
Set python packages version for dependabot.

### LINKED ISSUE #1021: Use dependabot
Implement [dependabot](https://github.com/dependabot) and:
- move requirements into a `requirements` directory
- follow similar implementation as https://github.com/pyansys/pymapdl/blob/main/.github/dependabot.yml

This will help solve issues related to upstream dependencies breaking doc generation and unit testing.

MaxJPRey:
Add dependabot.
Set python packages version for dependabot.
Fix #1021 .


requirements.txt
```diff
@@ -1,31 +1,31 @@
...
 # https://github.com/mhammond/pywin32
 # License: PSF
 
-pythonnet >= 2.4.0; python_version > '2.7'
+pythonnet==2.5.2; python_version > '2.7'
 # https://github.com/pythonnet/pythonnet
 # License: MIT
 
-rpyc >= 5.0.1; python_version > '2.7'
+rpyc==5.1.0; python_version > '2.7'
```
maxcapodi78, line 12: I would need to check if this version is compatible with the one embedded in pyaedt for ironpython usage. Should be ok but it good to have a try on linux machine next week before closing it or simply stay with 5.0.1 for now.

akaszynski (approved):
LGTM

maxcapodi78 (commented):
LGTM apart the rpyc comment

requirements.txt
akaszynski:
```suggestion
rpyc==5.0.1; python_version > '2.7'
```

akaszynski:
I downgraded pyrc, we can wait for dependabot to suggest.

PR #1056, opened 2022-04-15
Fix solder height calculation

### LINKED ISSUE #1055: Solder height calculation in place_in_layout_3d_placement is incorrect for certain cases
In certain cases there may be more than one layer at the bottom- or
top-most elevation in a design.  Checking only one of these layers (the
first as returned by the signal_layers collection) may result in a zero
solder height even when other layers at the same elevation have a
non-zero solder height.  In these cases we need to check all these
layers for components to calculate the solder height from.

isaacansys:
In certain cases there may be more than one layer at the bottom- or
top-most elevation in a design.  Checking only one of these layers (the
first as returned by the signal_layers collection) may result in a zero
solder height even when other layers at the same elevation have a
non-zero solder height.  In these cases we need to check all these
layers for components to calculate the solder height from.

To accomplish this, modify the solder height calculation in
place_in_layout_3d_placement to loop over all signal layers at either
the bottom-most or top-most signal layer elevation.  Take the maximum
solder height of any of these layers as the solder height for the
component.

Fixes #1055


MaxJPRey (approved):

MaxJPRey:
LGTM.

PR #1065, opened 2022-04-20
when get_solution_data method is used the user can pass a list of variations with values or a dict

maxcapodi78:
There was an issue when list was passed that is now fixed


Samuelopez-ansys (approved):
LGTM

PR #1069, opened 2022-04-21
posx and posy variable name replacement to standardization xpos and ypos

Samuelopez-ansys:
- Replace posx,posy,posz in all methods less deprecated ones


maxcapodi78 (approved):
LGTM

PR #1094, opened 2022-04-26
Read material .amat files

Alberto-DM:
LoadAEDTFile now supports .amat files.
It is possible to add materials directly from an .amat file.


maxcapodi78 (approved):
Great. LGTM

PR #1098, opened 2022-04-26
boolean cutout bug fixed

svandenb-dev:


pyaedt/edb_core/hfss.py
```diff
@@ -1192,12 +1192,7 @@ def get_ports_number(self):
            Number of ports.
 
         """
-        port_list = []
-        for term in self._active_layout.Terminals:
-            if str(term.GetBoundaryType()) == "PortBoundary":
-                if "ref" not in term.GetName():
-                    port_list.append(term)
-        return len(port_list)
+        return len([term for term in list(self._active_layout.Terminals) if term.GetBoundaryType() == 0])
```
maxcapodi78, line 1195:
```suggestion
        return len([term for term in list(self._active_layout.Terminals) if int(term.GetBoundaryType()) == 0])
```

maxcapodi78 (commented):
LGTM. Maybe just a simple modification to make it more robust for Ironpython too

maxcapodi78 (approved):

PR #1103, opened 2022-04-27
Impedance support for Lumped ports and Wave ports in Terminal mode.

Alberto-DM:
Now it is possible to set impedance for Lumped ports and Wave ports in Terminal mode.


maxcapodi78 (approved):
LGTM

PR #1104, opened 2022-04-27
Added 2D Report Edit methods

maxcapodi78:


maxcapodi78:
I'll force the merge because the coverage on new methods cannot be achieved with 22R1 but it's achieved with 22R2

MaxJPRey (approved):

_unittest/test_12_PostProcessing.py
```diff
@@ -317,6 +317,59 @@ def test_09e_add_line_from_equation(self):
         assert new_report.create()
         assert new_report.add_limit_line_from_equation(1, 20, 0.5, "GHz")
 
+    @pytest.mark.skipif(
+        config["desktopVersion"] < "2022.2", reason="Not working in non-graphical mode in version earlier than 2022.2."
```
PipKat, line 321: Would the reason be better stated as "X does not work in non-graphical mode in a version earlier than 2022.2?
I'm not exactly sure what doesn't work--the report creation?

pyaedt/modules/report_templates.py
```diff
@@ -462,6 +462,488 @@ def _context(self):
...
+        style_minor="Solid",
+        style_major="Solid",
+        minor_color=(0, 0, 0),
+        major_color=(0, 0, 0),
+    ):
+        """Edit Plot Grid Settings.
+
+        Parameters
+        ----------
+        minor_x : bool, optional
```
PipKat, line 493: I would edit this parameter descriptions but seem to no longer have the button to do so. I guess I'll edit them later. Descriptions of Boolean parameters should be written like this: "Whether to enable the minor X grid. The default is ``True``. Descriptions should also use only an initial capital letter. For example, Minor grid style.

PR #1117, opened 2022-04-29
Improve solution data

maxcapodi78:
Improved the way solutions data are obtained in order to manage well also complex variations data


Alberto-DM (approved):
Great improvement of the data object. LGTM.

PR #1123, opened 2022-05-02
Improve the behavior of the `ChangePageCallback `called when adding a button.

MaxJPRey:
- Improve the behavior of the `ChangePageCallback `called when adding a button.

- Deprecated method create_iport was not working.


maxcapodi78 (approved):
LGTM

PR #1124, opened 2022-05-02
Added Clip Plane creation method. it works only on 22R2

maxcapodi78:


maxcapodi78:
Coverage looks low because this method works on 22R2 only

PR #1130, opened 2022-05-03
add in

jeremich41:
>>> sphere1.props["ThetaStart"] = 45 , does not work, the unit degree must be specified.


maxcapodi78 (approved):
LGTM

PR #1134, opened 2022-05-04
Added new methods to Object3d FacePrimitive and EdgePrimitive classes

maxcapodi78:
To handle object creation from edge, object creation from face, edge normal movement and find top/bottom edges.


Samuelopez-ansys (approved):
LGTM

PR #1140, opened 2022-05-04
Improved Ibis Import. When the ibis is now imported, only the pins the associated models will bi imported.

maxcapodi78:
This will make the log much simple


MaxJPRey (approved):

MaxJPRey:
LGTM.

PR #1145, opened 2022-05-05
Thermal property getter was defined two times for padstack PDS layer.

MaxJPRey:
Thermal property getter was defined two times for padstack PDS layer.


maxcapodi78 (approved):
LGTM

PR #1147, opened 2022-05-05
fix assign_winding and assign_coil + update tests

gmalinve:
In this PR:
- Fix in assign_coil() -> polarity is now case insensitive.
- Fix in assign_winding() -> fix in coil_terminals if its None and voltage, current, res, ind fixes
- new 2D/3D tests implementation


maxcapodi78 (approved):
LGTM

PR #1149, opened 2022-05-05
Report note

cmesibov:


maxcapodi78 (approved):
LGTM

maxcapodi78:
I'll approve and force the merge because the coverage is not possible on this class because it runs only in graphical mode

PR #1151, opened 2022-05-06
add displacement current in eddy effect

gmalinve:
In this PR:
- Displacement current selection has been added to eddy_effects_on() method.
- Update tests


maxcapodi78 (approved):
LGTM

PR #1158, opened 2022-05-06
Introduced an environment variable for example build 

maxcapodi78:
All the examples now will run in non-graphical on sphinx build and in graphical when customer executes it

### DIFF
--- a/examples/01-Modeling-Setup/Configurations.py
+++ b/examples/01-Modeling-Setup/Configurations.py
@@ -36,6 +36,13 @@ from pyaedt import Icepak
 from pyaedt import examples
 from pyaedt import generate_unique_name
 
+
+##########################################################
+# Set Non Graphical Mode.
+# Default is False
+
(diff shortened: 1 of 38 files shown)


examples/01-Modeling-Setup/Configurations.py
```diff
@@ -108,7 +115,7 @@
 # ~~~~~~~~~~~~~~~~~~~~
 # This section we create a new Icepak project and iport the step.
 
-app = Icepak(projectname="new_proj_Ipk")
+app = Icepak(projectname="new_proj_Ipk", non_graphical=False)
```
MaxJPRey, line 118: Is it expected to have Icepak always run in non-graphical  even when not in the documentation process/

MaxJPRey (approved):

examples/01-Modeling-Setup/Configurations.py
maxcapodi78: effectively that bool is ignored 2nd time because Desktop is already opened. Anyway good catch I've removed it

PR #1174, opened 2022-05-11
Fix User List Init

maxcapodi78:
@Samuelopez-ansys we found a bug in user_list. Please give a look at the fix


Samuelopez-ansys (approved):
LGTM

MaxJPRey (approved):

PR #1175, opened 2022-05-11
Setup Y connection for Maxwell2D transient design.

MaxJPRey:
It is now possible to set an Y connection from windings in Maxwell2D.


maxcapodi78 (approved):
LGTM

PR #1181, opened 2022-05-13
Fix issue when reading Lists in AEDT File.

Samuelopez-ansys:
Bug found. If no Lists are in the AEDT project it was returning a Log error ("Lists not found"). It is not an error because typically, the AEDT file will not have Lists.


MaxJPRey (approved):

MaxJPRey:
Excellent fix @Samuelopez-ansys .

PR #1188, opened 2022-05-17
Remove the releases workflow file as it is not used anymore.

MaxJPRey:
Remove the `releases `workflow file as it is not used anymore.


maxcapodi78 (approved):
LGTM

PR #1216, opened 2022-05-24
Added unit test to Variable manage to project variables

maxcapodi78:


Alberto-DM (approved):
LGTM. Regression is fixed and the UT will catch it in the future.

PR #1231, opened 2022-05-30
Export mesh exception for Q3D API

Samuelopez-ansys:


pyaedt/application/Analysis3D.py
```diff
@@ -250,7 +250,7 @@ def plot(
             )
 
     @pyaedt_function_handler()
-    def export_mesh_stats(self, setup_name, variation_string="", mesh_path=None):
+    def export_mesh_stats(self, setup_name, variation_string="", mesh_path=None, setup_type="CG"):
```
beliaev-maksim, line 253: is it good to put Q3D specific parts into general function?

@maxcapodi78 
your thoughts?

Samuelopez-ansys: Another option could be to create a different method only for Q3D...

beliaev-maksim: don't know. However, changing signature just because of Q3D.
from other side over-complicating things is also not good

beliaev-maksim (approved):

pyaedt/application/Analysis3D.py
Samuelopez-ansys: I have created an overridden method in q3d.py, to keep the original one. But the  Codacy is complaining because I added one parameter that is not available in the original method, any suggestion? should I rename the method in q3d.py to avoid this issue? Thanks! @maxcapodi78 @beliaev-maksim

PR #1257, opened 2022-06-08
fixed signal_layers and diel_layers properties in Hfss3dLayout modeler

maxcapodi78:
Improved assign_material for huge number of assignments.


Samuelopez-ansys (approved):
LGTM

PR #1261, opened 2022-06-09
Refresh pyaedt object when design duplication

Samuelopez-ansys:
- Duplicate design refresh pyaedt object

- Fix small bug in End Connection boundary


maxcapodi78 (approved):
Lgtm

PR #1284, opened 2022-06-14
Evaluation of expression must use abstract tree.

MaxJPRey:
Evaluation of expression must use abstract tree.


maxcapodi78 (approved):
LGTM I've tested in linux ironpython and works ok.

PR #1288, opened 2022-06-15
Verify design object has the attribute 'GetDesignType'

Samuelopez-ansys:
Maxwell Circuit has not the attribute 'GetDesignType'. Check before call it.


MaxJPRey:
LGTM. I am assuming it could raised some exception without this check.

PR #1306, opened 2022-06-20
fixed return value of an edge length, for non line edges such as arcs

Marc-Vieweger:
routine now supports arcs and splines

old version was only valid for lines. For an arc or spline, the length of the edge is not the distance between to vertecies.

### DIFF
--- a/pyaedt/modeler/Object3d.py
+++ b/pyaedt/modeler/Object3d.py
@@ -468,19 +468,15 @@ class EdgePrimitive(EdgeTypePrimitive, object):
         Returns
         -------
         float or bool
-            Edge length in model units when edge has two vertices, ``False`` othwerise.
+            Edge length in model units.
 
         References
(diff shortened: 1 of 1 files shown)


maxcapodi78 (approved):
LGTM

maxcapodi78:
@Marc-Vieweger can you contact me on Teams? I'd like to know you as you are first time runner

maxcapodi78 (changes requested):
I think a merge from main is needed before going on

maxcapodi78:
@Marc-Vieweger please merge from main and resolve conflict and push again. it should then pass the tests

MaxJPRey:
@Marc-Vieweger This is now part of the code (encapsulated in a `try/catch`)
See:

```
try:
    return float(self.oeditor.GetEdgeLength(self.id))
except:
    return False
```

PR #1307, opened 2022-06-20
Add dictionary context in get_solution_data

Samuelopez-ansys:
- Fixed Bug in report_templates for Spectrum report
- Add the option to define a dictionary in the context for solution_data. Useful to define the setup values in reports.
- Get units sweep
- Unittest


maxcapodi78 (approved):
Great catch. LGTM

PR #1312, opened 2022-06-20
added extra dependcies, eg pip install pyaedt[all]

### LINKED ISSUE #1308: Bug located in ...Iron python import pyadet error due to numpy

### Description of the bug

Iron python import pyadet error due to numpy

pyadet version: main branch
numpy version: 1.21.6

error message:

```
  File "c:\users\byang\desktop\pyaedt_numpy\venv\lib\site-packages\numpy\__init__.py", line 132
            raise ImportError(msg) from e

                                   ^
SyntaxError: unexpected token 'from'
```

Both cpython and ironpython are used in aedt test framework. And ironpython is not compatible with numpy version 1.21.6.

We want to install some minimal packages like this: 
`pip install pyaedt[minimal] `

and this will install only pyaedt + COM interfaces

### Steps To Reproduce

- create venv (python 3.7)
- pip install git+https://github.com/pyansys/pyaedt.git@main
- launch ironpython
- run:
  ```python
  import sys
  sys.path.append(r"c:\users\byang\desktop\pyaedt_numpy\venv\lib\site-packages")
  import pyaedt
  ```

Windows

3.7

### Installed packages

ironpython

beliaev-maksim:
same as pyvista PR https://github.com/pyvista/pyvista/pull/2826

closes #1308

### DIFF
--- a/doc/source/Resources/Installation.rst
+++ b/doc/source/Resources/Installation.rst
@@ -15,9 +15,11 @@ In addition to the runtime dependencies listed in the installation information, 
 
 Installing on CPython v3.7-v3.9 from pypi
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-Install the prerequisite packages ``pythonnet`` and ``pywin32`` with:
+To install all the additional packages that extend PyAEDT, install using ``pip`` with::
 
-.. code:: python
+    pip install pyaedt[all]
+
+Install only with COM and GRPC interfaces with::
 
     pip install pyaedt
 
--- a/pyaedt/generic/general_methods.py
+++ b/pyaedt/generic/general_methods.py
@@ -30,7 +30,10 @@ except:
 is_remote_server = os.getenv("PYAEDT_IRONPYTHON_SERVER", "False").lower() in ("true", "1", "t")
 
 if not is_ironpython:
-    import psutil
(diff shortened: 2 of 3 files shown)


maxcapodi78:
@beliaev-maksim all the requirements you put in extras we consider now mandatory for pyaedt CPython because are needed for many pre and post processing operations. I'm not sure is a good idea to revert them to optional and exclude them from installation only because of Ironpython. @akaszynski what do you think about it?

beliaev-maksim:
@maxcapodi78 
the problem is that if you do not use post processing installing numpy and matplotlib and other deps is too heavy. And this is not only because of IronPython.
For general AEDT interface none of these libraries is required

akaszynski:
> @akaszynski what do you think about it?

I think the idea behind having optional dependencies is a good one; not everyone is going to want to install additional dependencies. However, I would prefer that we do this the other way around; we have a "light" install option where users can choose to not install the heavy dependencies, but we default to installing them.

maxcapodi78:
> > @akaszynski what do you think about it?
> 
> I think the idea behind having optional dependencies is a good one; not everyone is going to want to install additional dependencies. However, I would prefer that we do this the other way around; we have a "light" install option where users can choose to not install the heavy dependencies, but we default to installing them.

That's is a good idea!

PR #1314, opened 2022-06-20
Fix bounding box extent

### LINKED ISSUE #1313: Cannot set radiation_box and other properties on SimulationConfiguration
The configure_hfss_extents method will never set its radiation_box property because the setter checks that its argument is of type RadiationBoxType which is just an enumeration-style class for several constant integer values.  The setter should instead check that the argument is an integer (and perhaps one of the allowable values).

There appear to be several classes in pyaedt.generic.constants that are used this way:

- CutoutSubdesignType
- RadiationBoxType
- SweepType
- BasisOrder

A failing test for setting the HFSS airbox extent type:

https://github.com/pyansys/pyaedt/blob/32c41385a03efd01eb0c0a8ba28b28bf1655da85/_unittest/test_00_EDB.py#L1616

isaacansys:
A comment and a question up front:

- I welcome comments on the validate_enum_class_value design as I'm not sure if this is particularly pythonic or if there is a standard or better way to do it.
- Should these setters raise a ValueError or other exception if the validation fails?  This is not being done currently and the values not set are silently ignored.

The setters for some SimulationConfiguration properties, including cutout_subdesign_type, radiation_box, sweep_type, and basis_order would never set their property's value because they were checking that the arguments were of a type that would not be passed in.  The arguments passed in are integers taken from attributes on the class.  I call these "enumeration style" classes.

To fix this, add a new method validate_enum_class_value that takes an enumeration style class "cls" that has valid values from zero to N where the value N is assigned to the class attribute called Invalid and an integer value "value" to validate.  The function returns true if "value" is between zero (inclusive) and cls.Invalid (exclusive).  Note that the enumeration class must define this Invalid attribute and its value is not itself valid.

I added a test for setting the boundary box extent type which is the case that was reported to me as not working.

Fixes #1313


pyaedt/generic/constants.py
```diff
@@ -181,6 +181,24 @@ def scale_units(scale_to_unit):
...
+        Enumeration-style class with integer members in range(0, N) where cls.Invalid equals N-1.
+    value : int
+        value to check
+
+    Returns
+    -------
+    bool
+        Whether value is a valid value for enumeration class cls.
+    """
+    return isinstance(value, int) and value >= 0 and value < cls.Invalid
```
maxcapodi78, line 199: are you sure that an enum in ironputhon is classified as an int? I've noticed that pythonnet3.0 return an enum class and the isinstance(value, int) would return false

maxcapodi78 (approved):
@isaacansys I like this approach. Since the number of enum in edb is quite high I would probably create a separate module in edb_core folder and have a method also to retrieve the enum value given an int and an edb class. that would be replaced in all the methods that requires it as input

pyaedt/generic/constants.py
```diff
@@ -528,19 +546,19 @@ class SolverType(object):
...
 
 
 class SweepType(object):
-    (Linear, LogCount) = range(0, 2)
+    (Linear, LogCount, Invalid) = range(0, 3)
 
 
 class BasisOrder(object):
-    (Mixed, Zero, single, Double) = range(0, 4)
+    (Mixed, Zero, single, Double, Invalid) = range(0, 5)
```
MaxJPRey, line 561: @maxcapodi78 I am assuming it is too late to rename `single` as `Single`.
Could this cause deprecation issue?
```suggestion
    (Mixed, Zero, Single, Double, Invalid) = range(0, 5)
```

pyaedt/generic/constants.py
```diff
@@ -181,6 +181,24 @@ def scale_units(scale_to_unit):
     return sunit
 
 
+def validate_enum_class_value(cls, value):
+    """Return bool whether value is a valid value for enumeration-class cls.
```
PipKat, line 185:
```suggestion
    """Check whether the value for the class ``enumeration-class`` is valid.
```

pyaedt/generic/constants.py
```diff
@@ -181,6 +181,24 @@ def scale_units(scale_to_unit):
...
+
+    Parameters
+    ----------
+    cls : class
+        Enumeration-style class with integer members in range(0, N) where cls.Invalid equals N-1.
+    value : int
+        value to check
```
PipKat, line 192:
```suggestion
        Value to check
```

pyaedt/generic/constants.py
```diff
@@ -181,6 +181,24 @@ def scale_units(scale_to_unit):
...
+
+    Returns
+    -------
+    bool
+        Whether value is a valid value for enumeration class cls.
```
PipKat, line 197:
```suggestion
        ``True`` when the value is value for the ``enumeration-class``, ``False`` otherwise.
```

pyaedt/generic/constants.py
maxcapodi78: @maxcapodi78  I think this is a typo and can be fixed since PR is not yet closed. Correct @isaacansys ?

isaacansys: @maxcapodi78 I did not change this so it was in the previous commit as well.  This value with lowercase is already used in EDB_Data.py for example.

MaxJPRey: If customers/users are already using it. We can let it the way it is. I was mentioning "deprecation" because that it could require to update some scripts and examples.

isaacansys: @MaxJPRey after discussion with @maxcapodi78 we think we should make this change now as this was recently added and we can make things consistent with minimum pain now.  Should just make sure we announce this so existing users can adjust.  Thoughts?

isaacansys:
Thanks all for suggestions, have a busy calendar today but will get these incorporated by tomorrow and ready to merge.

pyaedt/generic/constants.py
maxcapodi78: Please add a warning into the class docstrings

MaxJPRey: Agreed.

maxcapodi78:
@isaacansys if you finalize all agreed changes we can close this PR

pyaedt/generic/constants.py
```diff
@@ -528,19 +546,19 @@ class SolverType(object):
...
-    (Conformal, BoundingBox, ConvexHull) = range(0, 3)
+    (Conformal, BoundingBox, ConvexHull, Invalid) = range(0, 4)
 
 
 class SweepType(object):
-    (Linear, LogCount) = range(0, 2)
+    (Linear, LogCount, Invalid) = range(0, 3)
 
 
 class BasisOrder(object):
```
isaacansys, line 560:
```suggestion
class BasisOrder(object):
    """Enumeration-class for HFSS basis order settings.
    
    Warning: the value ``single`` has been renamed to ``Single`` for consistency. Please update references to
    ``single``.
    """
```

pyaedt/generic/constants.py
```diff
@@ -528,19 +546,24 @@ class SolverType(object):
...
-    (Mixed, Zero, single, Double) = range(0, 4)
+    """Enumeration-class for HFSS basis order settings.
+
```
isaacansys, line 562:
```suggestion


```

pyaedt/generic/constants.py
```diff
@@ -528,19 +546,24 @@ class SolverType(object):
...
+    (Linear, LogCount, Invalid) = range(0, 3)
 
 
 class BasisOrder(object):
-    (Mixed, Zero, single, Double) = range(0, 4)
+    """Enumeration-class for HFSS basis order settings.
+    
+    Warning: the value ``single`` has been renamed to ``Single`` for consistency. Please update references to
+    ``single``.
+    """
```
isaacansys, line 565:
```suggestion
    """

```

PR #1319, opened 2022-06-21
Update unit_tests.yml

beliaev-maksim:
currently this pipe will be triggered twice, during PR and right after the merge.

Since in branch protection rules we do not allow pushes to main, this will save CI/CD time, considering that it takes ~30min


maxcapodi78 (approved):
LGTM

MaxJPRey:
Thanks @beliaev-maksim for doing it. I postponed it couple of times.

PR #1338, opened 2022-06-28
Props getitem and setitem

maxcapodi78:
Now user can get or set a property directly from the objects thanks to __getitem__ and setitem


Samuelopez-ansys (approved):
LGTM. It is really great

PR #1345, opened 2022-06-30
Added a check in the create Field plot method on the face ids. 

maxcapodi78:
If the face id belongs to a non model object the syntax has to be changed

### DIFF
--- a/pyaedt/modules/solutions.py
+++ b/pyaedt/modules/solutions.py
@@ -2285,11 +2285,29 @@ class FieldPlot:
             for index in self.volume_indexes:
                 info.append(str(index))
         if self.surfaces_indexes:
-            info.append("Surface")
-            info.append("FacesList")
-            info.append(len(self.surfaces_indexes))
+            model_faces = []
+            nonmodel_faces = []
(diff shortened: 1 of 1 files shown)


pyaedt/modules/solutions.py
```diff
@@ -2285,11 +2285,29 @@ def plotGeomInfo(self):
...
+                except:
+                    pass
+            info.append("Surface")
+            if model_faces:
+                info.append("FacesList")
+                info.append(len(model_faces))
+                for index in model_faces:
+                    info.append(index)
+            if nonmodel_faces:
+                info.append("NonModelFaceList")
```
MaxJPRey, line 2307:
```suggestion
                info.append("NonModelFacesList")
```

maxcapodi78: @MaxJPRey that is AEDT API syntax. We cannot choose

Samuelopez-ansys (approved):
LGTM

pyaedt/modules/solutions.py
MaxJPRey: OK. Got it. Thanks.

PR #1356, opened 2022-07-05
Implement HFSS SetSolveInsideThreshold

gmalinve:
Implement HFSS SetSolveInsideThreshold

### DIFF
--- a/pyaedt/hfss.py
+++ b/pyaedt/hfss.py
@@ -5226,14 +5226,14 @@ class Hfss(FieldAnalysis3D, object):
         )
 
     @pyaedt_function_handler()
-    def set_material_threshold(self, threshold=None):
-        """Set material threshold.
+    def set_material_threshold(self, threshold=100000):
+        """Set material conductivity threshold.
 
         Parameters
         ----------
         threshold : float, optional
             Conductivity threshold.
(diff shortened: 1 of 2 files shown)


pyaedt/hfss.py
```diff
@@ -5224,3 +5224,26 @@ def get_antenna_ffd_solution_data(
...
+        bool
+            ``True`` when successful, ``False`` when failed.
+        """
+        try:
+            if not threshold:
+                threshold = 100000
+            self.odesign.SetSolveInsideThreshold(threshold)
+            return True
+        except:
+            return False
```
MaxJPRey, line 5249: I might be missing a point, but why not using `100000` as the default value instead of `None`.
```suggestion
    def set_material_threshold(self, threshold=100000):
        """Set material conductivity threshold.

        Parameters
        ----------
        threshold : float, optional
            Conductivity threshold.
            The default value is 100000.

        Returns
        -------
        bool
            ``True`` when successful, ``False`` when failed.
        """
        try:
            self.odesign.SetSolveInsideThreshold(threshold)
            return True
        except:
            return False
```

maxcapodi78 (approved):
LGTM

PR #1362, opened 2022-07-07
implement assign symmetry method for maxwell 2D/3D

gmalinve:


maxcapodi78 (changes requested):

pyaedt/maxwell.py
```diff
@@ -1221,6 +1223,71 @@ def set_initial_angle(self, motion_setup, val):
...
+
+        Parameters
+        ----------
+        entity_list : list
+            List IDs or :class:`pyaedt.modeler.Object3d.EdgePrimitive` or
+            :class:`pyaedt.modeler.Object3d.FacePrimitive`.
+        symmetry_name : str, optional
+            Name of the symmetry.
+        is_odd : bool, optional
+            Type of symmetry. Default value is True which means H field is tangential to
```
MaxJPRey, line 1238:
```suggestion
            Type of symmetry. Default value is ``True`` which means H field is tangential to
```

pyaedt/maxwell.py
```diff
@@ -1221,6 +1223,71 @@ def set_initial_angle(self, motion_setup, val):
...
+            the boundary. If False H field is normal to the boundary.
```
MaxJPRey, line 1239:
```suggestion
            the boundary. If ``False`` H field is normal to the boundary.
```

Samuelopez-ansys (approved):
LGTM

PR #1363, opened 2022-07-07
Enhance bom import

ring630:
set RLC to disabled when its refdes is not in BOM.


pyaedt/edb_core/EDB_Data.py
```diff
@@ -2663,6 +2663,23 @@ def refdes(self):
         """
         return self.edbcomponent.GetName()
 
+    @property
+    def is_enabled(self):
+        """Determines whether current object is enabled.
```
PipKat, line 2668:
```suggestion
        """Check if the current object is enabled.
```

pyaedt/edb_core/EDB_Data.py
```diff
@@ -2663,6 +2663,23 @@ def refdes(self):
...
+
+        Returns
+        -------
+        bool
+            True if this component is enabled, False otherwise.
```
PipKat, line 2673:
```suggestion
            ``True`` if this current object is enabled, ``False`` otherwise.
```

pyaedt/edb_core/EDB_Data.py
```diff
@@ -2663,6 +2663,23 @@ def refdes(self):
...
+        """
+        return self.edbcomponent.GetComponentProperty().IsEnabled()
+
+    @is_enabled.setter
+    def is_enabled(self, enabled):
+        """Sets whether current object is enabled."""
```
PipKat, line 2679:
```suggestion
        """Enables the current object."""
```

pyaedt/edb_core/EDB_Data.py
```diff
@@ -2663,6 +2663,23 @@ def refdes(self):
...
+        if self.edbcomponent.GetComponentType() in [1, 2, 3]:
+            self.edbcomponent.GetComponentProperty().SetEnabled(enabled)
+
     @property
     def res_value(self):
         """Resitance Value.
```
PipKat, line 2685:
```suggestion
        """Resitance value.
```

ring630:
@PipKat , do you know why it is failing ironpython?

pyaedt/edb_core/EDB_Data.py
```diff
@@ -2663,9 +2663,26 @@ def refdes(self):
...
+        -------
+        bool
+            ``True`` if current object is enabled, ``False`` otherwise.
+        """
+        return self.edbcomponent.GetComponentProperty().IsEnabled()
+
+    @is_enabled.setter
+    def is_enabled(self, enabled):
+        """Enables the current object."""
+        if self.edbcomponent.GetComponentType() in [1, 2, 3]:
```
svandenb-dev, line 2680: you need to use the right enumerator now not Integer, e.g Edb.Definition.ComponentType.Resistor

svandenb-dev (changes requested):

pyaedt/edb_core/EDB_Data.py
```diff
@@ -2663,9 +2663,32 @@ def refdes(self):
...
+    @property
+    def is_enabled(self):
+        """Check if the current object is enabled.
+
+        Returns
+        -------
+        bool
+            ``True`` if current object is enabled, ``False`` otherwise.
+        """
+        return self.edbcomponent.GetComponentProperty().IsEnabled()
```
MaxJPRey, line 2675: @ring630 .
This is causing the IronPython tests failure.
IronPython is less permissive that Python with some incorrect types provided.
image

You can debug in Python on your local machine if it is more convenient for you and check the type returned by `self.edbcomponent.GetComponentProperty()`.

pyaedt/edb_core/EDB_Data.py
```diff
@@ -2663,9 +2663,32 @@ def refdes(self):
...
+
+    @is_enabled.setter
+    def is_enabled(self, enabled):
+        """Enables the current object."""
+        if self.edbcomponent.GetComponentType() in [
+            self._edb.Definition.ComponentType.Resistor,
+            self._edb.Definition.ComponentType.Capacitor,
+            self._edb.Definition.ComponentType.Inductor,
+        ]:
+            return self.edbcomponent.GetComponentProperty().SetEnabled(enabled)
```
MaxJPRey, line 2685: @ring630 .
This is causing the IronPython tests failure you were asking about.
IronPython is less permissive that Python with some incorrect types provided.
image

You can debug in Python on your local machine if it is more convenient for you and check the type returned by `self.edbcomponent.GetComponentProperty()`.

ring630: @MaxJPRey 
GetComponentProperty() returns RLCComponentProperty on my machine.

svandenb-dev: @ring630 I fixed the issue. Actually with the new version Python behaves exactly the same as the EDB API with C#. I don't understand quite well how python was working before. If you want to access some properties of object (e.g ComponentProperty for instance) you need to clone the object to make it not read only. Then you will have access to all methods. I fixed all getter and setters in the edb components. Now the test is passing

pyaedt/edb_core/EDB_Data.py
```diff
@@ -2637,9 +2637,30 @@ def refdes(self):
...
+        bool
+            ``True`` if current object is enabled, ``False`` otherwise.
+        """
+        if self.type in ["Resistor", "Capacitor", "Inductor"]:
+            return self.component_property.IsEnabled()
+        else:  # pragma: no cover
+            return False
+
+    @is_enabled.setter
+    def is_enabled(self, enabled):
```
maxcapodi78, line 2655: you have implemented the setter but there is no unit test for it.

maxcapodi78 (approved):
Please add Unittest to the setter and then it's ok for me

svandenb-dev (approved):

PR #1365, opened 2022-07-07
Change headings to snake case for source/docs folder

PipKat:
Trying to prepare for implementation of Vale docstyle linter, which requires sentence case headings.

### DIFF
--- a/doc/source/API/Application.rst
+++ b/doc/source/API/Application.rst
@@ -1,10 +1,10 @@
-AEDT Applications
-=================
-The PyAEDT API includes classes for applications and modules. You must initialize the 
-application case. All other classes and methods are inherited into the application class.
-The desktop application is implicitly launched in any of the other applications.
+AEDT apps
+=========
+The PyAEDT API includes classes for apps and modules. You must initialize the 
+app case. All other classes and methods are inherited into the app class.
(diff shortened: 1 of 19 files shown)


doc/source/API/Application.rst
```diff
@@ -1,10 +1,10 @@
...
-application case. All other classes and methods are inherited into the application class.
-The desktop application is implicitly launched in any of the other applications.
+AEDT apps
+=========
+The PyAEDT API includes classes for apps and modules. You must initialize the 
+app case. All other classes and methods are inherited into the app class.
+The desktop app is implicitly launched in any of the other applications.
 
-Example with Desktop:
+Example with desktop:
```
maxcapodi78, line 7: I'm ok with most of the changes. My only doubt is related to Desktop. "Electronics Desktop" is the Ansys name of the product. Maybe we should leave the name with big D.

maxcapodi78 (approved):
I've left only one comment but the documentation looks better now. Thanks

doc/source/API/Application.rst
PipKat: @maxcapodi78 I changed it from "desktop" to AEDT--which is more consistent within the descriptive doc.

PR #1367, opened 2022-07-08
Disable Dependency Check on Linux

maxcapodi78:


MaxJPRey (approved):

MaxJPRey:
That's good to see those simplifications coming for the final users.

PR #1390, opened 2022-07-14
Fixed bug on material assigment

maxcapodi78:


Samuelopez-ansys (approved):

pyaedt/desktop.py
```diff
@@ -181,9 +181,11 @@ def release_desktop(close_projects=True, close_desktop=True):
                 desktop.CloseProject(project)
         pid = _main.oDesktop.GetProcessID()
         if settings.aedt_version >= "2022.2" and settings.use_grpc_api and not is_ironpython:
-            import ScriptEnv
+            import PyDesktopPlugin
 
-            ScriptEnv.Release()
+            PyDesktopPlugin.Release()
+            _delete_objects()
```
MaxJPRey, line 187: @maxcapodi78 Maybe here we could add a comment to explain why we are releasing this way. It might be not trivial for external contributors.

pyaedt/rpc/rpyc_services.py
```diff
@@ -823,6 +825,18 @@ def aedt_grpc(port=None, beta_options=None, use_aedt_relative_path=False, non_gr
...
+        s = socket.socket()
+        machine_name = socket.getfqdn()
+        while timeout > 0:
+            try:
+                s.connect((machine_name, port))
+            except socket.error:
+                timeout -= 2
+                time.sleep(2)
+            else:
+                s.close()
```
MaxJPRey, line 838: @maxcapodi78 I am assuming here we are always closing the socket in the `else` block if no exception is raised because it is just to check the gRPC port.

MaxJPRey (approved):

pyaedt/rpc/rpyc_services.py
MaxJPRey: LGTM.

pyaedt/desktop.py
maxcapodi78: correct. Thanks

pyaedt/rpc/rpyc_services.py
maxcapodi78: correct

PR #1401, opened 2022-07-15
Edits to Modeler and Setup examples

PipKat:
Minor edits were also made to some EDB examples

### DIFF
--- a/examples/01-Modeling-Setup/Configurations.py
+++ b/examples/01-Modeling-Setup/Configurations.py
@@ -5,7 +5,7 @@ This example shows how you can use PyAEDT to export configuration files and reus
 them to import in a new project. A configuration file is supported by these apps:
 
 * HFSS
-* 2d Extractor and Q3D Extractor
+* 2D Extractor and Q3D Extractor
 * Maxwell
 * Icepak
 * Mechanical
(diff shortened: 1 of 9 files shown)


examples/01-Modeling-Setup/Configurations.py
```diff
@@ -1,33 +1,34 @@
...
 ----------------------------
-This example shows how you can use PyAEDT to export config files and reuse to import in a new project.
-Configuration file is actually supported by the following applications:
-* Hfss
-* Q2d and Q3d
+This example shows how you can use PyAEDT to export configuration files and reuse
+them to import in a new project. A configuration file is supported by these apps:
+
+* HFSS
+* 2d Extractor and Q3D Extractor
```
MaxJPRey, line 8:
```suggestion
* Q2D Extractor and Q3D Extractor
```

PipKat: @MaxJPRey I went back and forth on this. When I search the AEDT doc, there are now occurrences of Q2D--only 2D. This is why I used 2D. Similarly, some of the example text used 2D (and some used 3D). This is why I performed the doc search.

PR #1428, opened 2022-07-20
plot_bug_fix

ring630:
exclude void path


pyaedt/edb_core/nets.py
```diff
@@ -244,6 +244,8 @@ def get_plot_data(
             nets = [nets]
 
         for path in self._pedb.core_primitives.paths:
+            if path.is_void:
```
maxcapodi78, line 247: why are you excluding void paths from the plot?

maxcapodi78 (approved):

maxcapodi78:
anyway maybe we have to rethink to plot the void traces correctly in the future

PR #1434, opened 2022-07-20
Add 3DDataset in Icepak and Maxwell at design level

Samuelopez-ansys:
icepak and Maxwell 3D allows 3D Dataset at design level


maxcapodi78 (approved):
LGTM

PR #1507, opened 2022-08-04
Set points coordinates for BP curve coreloss.

### LINKED ISSUE #1490: Import curve and calculate coefficients using 'set_bp_curve_coreloss()' in Maxwell
Hi guys,
 
I’m trying to import an Induction vs. Loss (BP) curve in Maxwell and command to calculate some coefficients using the following script, but it does not work. Any ideas? It returns “true” but the following error happens:
 
mensagemErro_atribuicaoDeMaterial

Code:
self.material.set_bp_curve_coreloss([[0,0],[0.5,1.56],[1.0,3.54],[1.5,19.263],[2.0,40.85]],
                                                    kdc=0.001,
                                                    cut_depth=0.001,
                                                    punit="w/kg",
                                                    bunit="tesla",
                                                    frequency=50,
                                                    thickness="0.5mm")
 
Cheers

MaxJPRey:
The method `set_bp_curve_coreloss()` was not allowing the creation of the the point coordinates properly and consequently, the core loss curve was not updated.

image

Now the curve is created properly.

Fix #1490

### DIFF
--- a/_unittest/test_28_Maxwell3D.py
+++ b/_unittest/test_28_Maxwell3D.py
@@ -430,3 +430,16 @@ class TestClass(BasisTest, object):
             if bound.name == "Symmetry_Test_IsEven":
                 assert bound.type == "Symmetry"
                 assert not bound.props["IsOdd"]
+
+    def test_36_set_bp_curve_loss(self):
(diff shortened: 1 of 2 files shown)


Samuelopez-ansys (approved):
lgtm

PR #1511, opened 2022-08-05
Provide thickness to a new pad stack layer while defining the start layer.

### LINKED ISSUE #1505: No value provided for argument 'thickness' in PadstackLayer() call

### Description of the bug

```
if k == layer:
    found = True
if found and layer not in self._padstacks_by_layer:
    new_stackup[k] = PadstackLayer(self, k, v.elevation)
elif found:
    new_stackup[k] = self._padstacks_by_layer[k]
```

It currently works which means that the parameter named `thickness` could be optional.

### Steps To Reproduce

Launch the test named `test_03_padstackline` in the set `test_16_3d_stackup`.

Windows

3.10

### Installed packages

N/A.

MaxJPRey:
You must provide the `thickness` when instantiating the `PadstackLayer`.

Fix #1505.


Samuelopez-ansys (approved):
LGTM

PR #1517, opened 2022-08-08
Implement export rl matrix

gmalinve:
PR to export RL matrix

### DIFF
--- a/pyaedt/application/Analysis.py
+++ b/pyaedt/application/Analysis.py
@@ -1851,7 +1851,6 @@ class Analysis(Design, object):
             val = "{0}{1}".format(value, units)
         return val
 
-    # Add test and check if Analysis.py is the best place to have the method
     @pyaedt_function_handler()
     def export_rl_matrix(
         self,
--- a/pyaedt/application/Analysis.py
+++ b/pyaedt/application/Analysis.py
@@ -1896,8 +1896,7 @@ class Analysis(Design, object):
         """
         try:
             if not self.solution_type == "EddyCurrent":
-                msg = "RL Matrix can only be exported if solution type is Eddy Current."
-                raise ValueError(msg)
+                raise ValueError("RL Matrix can only be exported if solution type is Eddy Current.")
             matrix_list = [bound for bound in self.boundaries if isinstance(bound, MaxwellParameters)]
             if matrix_name is None:
                 msg = "Matrix name to be exported must be provided."
--- a/pyaedt/application/Analysis.py
+++ b/pyaedt/application/Analysis.py
@@ -1893,44 +1893,45 @@ class Analysis(Design, object):
         bool
             ``True`` when successful, ``False`` when failed.
         """
-        try:
-            if not self.solution_type == "EddyCurrent":
-                raise ValueError("RL Matrix can only be exported if solution type is Eddy Current.")
-            matrix_list = [bound for bound in self.boundaries if isinstance(bound, MaxwellParameters)]
-            if matrix_name is None:
-                msg = "Matrix name to be exported must be provided."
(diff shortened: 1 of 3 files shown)


pyaedt/application/Analysis.py
```diff
@@ -1849,3 +1850,104 @@ def value_with_units(self, value, units=None):
...
+            if matrix_list:
+                if not [
+                    matrix
+                    for matrix in matrix_list
+                    if matrix.name == matrix_name or [x for x in matrix.available_properties if matrix_name in x]
+                ]:
+                    msg = "Matrix name doesn't exist, provide and existing matrix name."
+                    raise ValueError(msg)
+            else:
+                msg = "Matrix list parameters is empty, can't export a valid matrix."
```
Samuelopez-ansys, line 1914: @gmalinve 

You should modify the raise error with:

_self.logger.error(msg)
return False_

If you see any other pyaedt method, it returns False when it fails with a logger error message.

And in the unittest you could also add these tests, I mean, test when it fails (when filepath is None or the extension is not .txt) this will improve even more the coverage.

Samuelopez-ansys, line 1914: This should return False, right?

Samuelopez-ansys (changes requested):

pyaedt/application/Analysis.py
```diff
@@ -1849,3 +1850,104 @@ def value_with_units(self, value, units=None):
                 units = self.modeler.model_units
             val = "{0}{1}".format(value, units)
         return val
+
+    # Add test and check if Analysis.py is the best place to have the method
```
MaxJPRey, line 1854:
```suggestion
```

pyaedt/application/Analysis.py
```diff
@@ -1849,3 +1850,104 @@ def value_with_units(self, value, units=None):
...
+            Matrix name to be exported.
+        file_path : str
+            File path to export R/L matrix file.
+        is_format_default : bool, optional
+            Whether the exported format is default or not.
+            If False the custom format is set (no exponential).
+        width : int, optional
+            Column width in exported .txt file.
+        precision : int, optional
+            Decimal precision number in exported .txt file.
```
MaxJPRey, line 1882:
```suggestion
            Decimal precision number in exported *.txt file.
```

pyaedt/application/Analysis.py
```diff
@@ -1849,3 +1850,104 @@ def value_with_units(self, value, units=None):
...
+        is_exponential : bool, optional
+            Whether the format number is exponential or not.
+        setup_name : str, optional
+            Name of the setup.
+        default_adaptive : str, optional
+            Adaptive type.
+        is_post_processed : bool, optional
+            Boolean to check if it is post processed.
```
MaxJPRey, line 1890:
```suggestion
            Boolean to check if it is post processed. Default value is ``False``.
```

pyaedt/application/Analysis.py
```diff
@@ -1849,3 +1850,104 @@ def value_with_units(self, value, units=None):
...
+
+        Returns
+        -------
+        bool
+            ``True`` when successful, ``False`` when failed.
+        """
+        try:
+            if not self.solution_type == "EddyCurrent":
+                msg = "RL Matrix can only be exported if solution type is Eddy Current."
+                raise ValueError(msg)
```
MaxJPRey, line 1900: Usually, it is not needed to create a new variable when it is used only one time.
```suggestion
                raise ValueError("RL Matrix can only be exported if solution type is Eddy Current.")
```

Samuelopez-ansys (approved):
lgtm

PR #1534, opened 2022-08-11
Fix arguments to pytest; Ignore coverage file when running from PyCharm (since workingDir=_unittest)

kbhagat-ansys:


MaxJPRey:
LGTM.

PR #1538, opened 2022-08-12
Frequency dependent source in circuit excitation

Samuelopez-ansys:


pyaedt/circuit.py
```diff
@@ -1431,6 +1431,90 @@ def assign_power_sinusoidal_excitation_to_ports(self, ports, settings):
...
+                    [
+                        "NAME:VoltageSourceFreq" + str(id),
+                        "DataId:=",
+                        "Source" + str(id),
+                        "Type:=",
+                        17,
+                        "Output:=",
+                        0,
+                        "NumPins:=",
+                        2,
```
MaxJPRey, line 1472: @Samuelopez-ansys Will this value for the Pins always be `2`?

pyaedt/circuit.py
```diff
@@ -1431,6 +1431,90 @@ def assign_power_sinusoidal_excitation_to_ports(self, ports, settings):
...
+        -------
+        bool
+            ``True`` when successful, ``False`` when failed.
+
+        References
+        ----------
+
+        >>> oDesign.UpdateSources
+        """
+        id = self.modeler.schematic.create_unique_id()
```
gmalinve, line 1455: maybe it's worth to add a check if the file path provided exists and then another check that the extension of the file name provided is valid and is .fds format.
You could also add a unit test for that and check that it fails if file doesn't exist or file extension is wrong.

pyaedt/circuit.py
```diff
@@ -1431,6 +1431,90 @@ def assign_power_sinusoidal_excitation_to_ports(self, ports, settings):
...
+
```
gmalinve, line 1456: you could actually do the same for the ports. Check if user has entered valid and existing excitations.

pyaedt/circuit.py
Samuelopez-ansys: @MaxJPRey  this is a good question. If you check the equivalent methods : assign_current_sinusoidal_excitation_to_ports and assign_voltage_sinusoidal_excitation_to_ports, you will see how NumPins is 2 as well. I does not have any 'sense' because we are not defining any component it is just an excitation, so I think it is always 2 hardcoded. Let's keep it, because for now it is working in these other methods.

pyaedt/circuit.py
Samuelopez-ansys: @gmalinve Done

pyaedt/circuit.py
Samuelopez-ansys: @gmalinve Done

pyaedt/circuit.py
MaxJPRey: Make sense. Thanks @Samuelopez-ansys .

MaxJPRey (approved):
Looks great!

PR #1548, opened 2022-08-17
Dependabot for gh-actions wrongly configured

RobPasMue:
It looks like unless we define it at "/" it will not work. Solved it at PyDisco already and it is now working fine.


MaxJPRey (approved):
@RobPasMue Thanks for fixing it.

PR #1579, opened 2022-08-22
Add section to set Desktop Config to Twin Builder for the example

SushobhitGupt:
…  and restore it back at the end so that we find correct license for all regular AEDT users.

### DIFF
--- a/examples/07-TwinBuilder/03-Dynamic_ROM_Creation_And_Visualization.py
+++ b/examples/07-TwinBuilder/03-Dynamic_ROM_Creation_And_Visualization.py
@@ -69,6 +69,14 @@ shutil.copyfile(os.path.join(source_data_folder ,source_build_conf_file), os.pat
 
 tb = TwinBuilder(projectname=generate_unique_project_name(),specified_version=desktop_version, non_graphical=non_graphical, new_desktop_session=new_thread)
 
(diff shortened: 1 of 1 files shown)


examples/07-TwinBuilder/03-Dynamic_ROM_Creation_And_Visualization.py
```diff
@@ -69,6 +69,14 @@
 
 tb = TwinBuilder(projectname=generate_unique_project_name(),specified_version=desktop_version, non_graphical=non_graphical, new_desktop_session=new_thread)
 
+# Switch current desktop configuration and schematic environment to "Twin Builder"
+# dynamic ROM feature is only available with a twin builder license configuration
+# this and the restoring section at the end is not needed if the desktop is already configured as "Twin Builder"
```
MaxJPRey, line 74:
```suggestion
# Switch the current desktop configuration and the schematic environment to "Twin Builder".
# The Dynamic ROM feature is only available with a twin builder license.
# Those actions and the restoring section at the end are not needed if the desktop is already configured as "Twin Builder"
```

MaxJPRey (approved):

MaxJPRey (approved):
LGTM.

PR #1597, opened 2022-08-25
Create a setup in Hybrid solution type with SBR+ region

### LINKED ISSUE #1572: In HFSS Hybrid solution type assign Hybrid regions

It is not implemented the Hybrid region definition (SBR+, IE or PO)

PyAEDT should wrap the AEDT API method Module.AssignHybridRegion
It is very importat for HFSS users.

_No response_

Samuelopez-ansys:
Closes #1572


Samuelopez-ansys:
@maxcapodi78 It was not possible to create a HFSS Setup for a Hybrid solution type with SBR+ regions. In addition I have added Hybrid region definition (SBR+, IE and PO).

Please could check the setup part is correct? In hfss.py I have inserted a loop checking if an SBR+ region exists because then we need to add SBR+ setup options. 

Thanks

maxcapodi78 (approved):
LGTM

PR #1598, opened 2022-08-25
Assign current density and current density terminal implementation - Maxwell2D/3D

gmalinve:


maxcapodi78 (changes requested):
We need to review it. We need to use Boundary Class instead of creating array in this way.

pyaedt/maxwell.py
```diff
@@ -1299,6 +1300,151 @@ def assign_symmetry(self, entity_list, symmetry_name=None, is_odd=True):
...
+            Default value is 0 A/m2.
+        current_density_2d : str, optional
+            Current density 2D value.
+            Default value is 0 A/m2.
+        coordinate_system_name : str, optional
+            Coordinate system name.
+            Default value is 'Global'.
+        coordinate_system_cartesian : str, optional
+            Coordinate system cartesian.
+            Possible values can be 'Cartesian', 'Cylindrical', 'Spherical'.
```
MaxJPRey, line 1346:
```suggestion
            Possible values can be 'Cartesian', 'Cylindrical', and 'Spherical'.
```

pyaedt/maxwell.py
```diff
@@ -1299,6 +1300,151 @@ def assign_symmetry(self, entity_list, symmetry_name=None, is_odd=True):
...
+        current_density_2d : str, optional
+            Current density 2D value.
+            Default value is 0 A/m2.
+        coordinate_system_name : str, optional
+            Coordinate system name.
+            Default value is 'Global'.
+        coordinate_system_cartesian : str, optional
+            Coordinate system cartesian.
+            Possible values can be 'Cartesian', 'Cylindrical', and 'Spherical'.
+            Default value is 'Cartesian'.
```
MaxJPRey, line 1347:
```suggestion
            Possible values can be ``"Cartesian"``, ``"Cylindrical"``, and ``"Spherical"``.
            Default value is ``"Cartesian"``.
```

maxcapodi78 (approved):

PR #1620, opened 2022-08-29
Method to download full example directory or specific file 

### LINKED ISSUE #1599: General method to download files from PyAnsys example folder

Current methods download specific folders inside PyAnsys example folder. We would need a general and public method to download any folder specified by the user

_No response_

_No response_

Samuelopez-ansys:
Close #1599


_unittest/test_00_downloads.py
```diff
@@ -58,3 +58,7 @@ def test_download_3dcomp(self):
     def test_download_twin_builder_data(self):
         example_folder = self.examples.download_twin_builder_data("Ex1_Mechanical_DynamicRom.zip", True)
         assert os.path.exists(example_folder)
+
+    def test_download_folder(self):
+        example_folder = self.examples.download_file("motorcad", "IPM_Vweb_Hairpin.mot")
```
maxcapodi78, line 63: is this method downloading a file only or it can download a directory with all the content?

Samuelopez-ansys: This is using the private method "_download_file", so this method is just downloading a file. Do you think is useful to implement this additional feature?

maxcapodi78: yes, mainly for aedb. I think you can reuse some already available in some of the specific download methods

MaxJPRey: Maybe we can create a download_folder method? (a separate one)

Samuelopez-ansys:
@maxcapodi78 @dcrawforAtAnsys @MaxJPRey 

The method **download_file** method could download a specific file from a directory or if the filename is not specified, then the full directory is downloaded. 
In the second case, it is downloading all the files of this directory, if there are nested directories, they will not be copied, only the .aedb folder will be copied as an exception.

MaxJPRey (approved):
LGTM.

maxcapodi78 (approved):
LGTM

Samuelopez-ansys:
Last modification disable the option in Ironpython to download the full folder because urllib does not work and the os.system(command) does not retrieve the needed data. Then, the Ipython option is disabled for this case.

Issue #1592, opened 2022-08-24
Add more features to HFSS 3D Layout stackup editor module


Marcel0000:
### Description of the feature

It would be nice, if the methods & attributes available in `pyaedt.edb_core.EDB_Data.EDBLayer` would be implemented in `pyaedt.modules.LayerStackup.Layer` as well for usage directly in HFSS 3D Layout.
E.g. change name, thickness, roughness etc.

### Useful links and references

[https://aedt.docs.pyansys.com/API/_autosummary/pyaedt.modules.LayerStackup.Layer.html](https://aedt.docs.pyansys.com/API/_autosummary/pyaedt.modules.LayerStackup.Layer.html)

[https://aedt.docs.pyansys.com/API/_autosummary/pyaedt.edb_core.EDB_Data.EDBLayer.html](https://aedt.docs.pyansys.com/API/_autosummary/pyaedt.edb_core.EDB_Data.EDBLayer.html)

maxcapodi78:
@Marcel0000 I'll check and add this enhancement. 
Anyway you can combine both tools to obtain what you need. 

First open the aedb folder with Edb class and do all modifications
Then open the aedb folder with Hfss3dLayout to run analysis and postprocessing

PR #1621, opened 2022-08-29
Implemented getters and setters on 3d layout layer properties

maxcapodi78:
#1592

### DIFF
--- a/pyaedt/modeler/Modeler.py
+++ b/pyaedt/modeler/Modeler.py
@@ -1374,7 +1374,7 @@ class GeometryModeler(Modeler, object):
                             props["List"] = data["GeometryEntityListParameters"]["EntityList"]
                         design_lists.append(Lists(self, props, name))
             except:
-                self.logger.infi("Lists were not retrieved from AEDT file")
+                self.logger.info("Lists were not retrieved from AEDT file")
         return design_lists
 
     def __get__(self, instance, owner):
--- a/pyaedt/modeler/Modeler.py
+++ b/pyaedt/modeler/Modeler.py
@@ -1374,7 +1374,7 @@ class GeometryModeler(Modeler, object):
                             props["List"] = data["GeometryEntityListParameters"]["EntityList"]
                         design_lists.append(Lists(self, props, name))
             except:
-                self.logger.error("Lists were not retrieved from AEDT file")
+                self.logger.infi("Lists were not retrieved from AEDT file")
         return design_lists
 
     def __get__(self, instance, owner):
--- a/pyaedt/modules/LayerStackup.py
+++ b/pyaedt/modules/LayerStackup.py
@@ -150,7 +150,7 @@ class Layer(object):
 
     @property
     def color(self):
-        """Get/Set the property to the active layer. Color can be an integer or a list of rgb values (0,255).
+        """Get/Set the property of the active layer. Color can be an integer or a list of rgb values (0,255).
 
         Returns
         -------
--- a/pyaedt/modules/LayerStackup.py
+++ b/pyaedt/modules/LayerStackup.py
@@ -335,7 +335,7 @@ class Layer(object):
         self.update_stackup_layer()
 
     @property
-    def topbottom(self):
+    def top_bottom(self):
         """Get/Set the active layer top bottom alignment.
 
         Returns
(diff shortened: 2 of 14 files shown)


Samuelopez-ansys (approved):
LGTM

pyaedt/modules/LayerStackup.py
```diff
@@ -143,10 +143,616 @@ def __init__(self, app, layertype="signal", negative=False):
...
-        self.usp = False
+        self._usp = False
         self.hfssSp = {"si": True, "dt": 0, "dtv": 0.1}
         self.planaremSp = {"ifg": False, "vly": False}
-        self.zones = []
+        self._zones = []
+
+    @property
+    def color(self):
+        """Get/Set the property to the active layer. Color can be an integer or a list of rgb values (0,255).
```
MaxJPRey, line 153:
```suggestion
        """Get/Set the property of the active layer. Color can be an integer or a list of rgb values (0,255).
```

pyaedt/modules/LayerStackup.py
```diff
@@ -143,10 +143,616 @@ def __init__(self, app, layertype="signal", negative=False):
...
+        Returns
+        -------
+        str
+        """
+        return self._toprounghenss
+
+    @toprounghenss.setter
+    def toprounghenss(self, val):
+        self._toprounghenss = val
+        self.update_stackup_layer()
```
MaxJPRey, line 438:
```suggestion
    def top_roughness(self):
        """Get/Set the active layer top roughness (with units).

        Returns
        -------
        str
        """
        return self._top_roughness

    @top_roughness.setter
    def top_roughness(self, val):
        self._top_roughness= val
        self.update_stackup_layer()
```

pyaedt/modules/LayerStackup.py
```diff
@@ -114,25 +114,25 @@ def __init__(self, app, layertype="signal", negative=False):
...
-        self.index = 1
+        self._locked = False
+        self._topbottom = "neither"
+        self._pattern = 1
+        self._drawoverride = 0
+        self._thickness = 0
+        self._lowerelevation = 0
+        self._roughness = 0
+        self._botroughness = 0
+        self._toprounghenss = 0
```
MaxJPRey, line 125:
```suggestion
        self._top_roughness= 0
```

pyaedt/modules/LayerStackup.py
```diff
@@ -143,10 +143,616 @@ def __init__(self, app, layertype="signal", negative=False):
...
+    def toprounghenss(self):
+        """Get/Set the active layer top roughness (with units).
+
+        Returns
+        -------
+        str
+        """
+        return self._toprounghenss
+
+    @toprounghenss.setter
```
MaxJPRey, line 435:
```suggestion
    @top_roughness.setter
```

maxcapodi78: @MaxJPRey the reason I left those names is to keep the back compatibility. their name was this before and If I change I risk to break it. Any suggestion?

pyaedt/modeler/Modeler.py
```diff
@@ -1374,7 +1374,7 @@ def _get_lists_data(self):
                             props["List"] = data["GeometryEntityListParameters"]["EntityList"]
                         design_lists.append(Lists(self, props, name))
             except:
-                self.logger.error("Lists were not retrieved from AEDT file")
+                self.logger.infi("Lists were not retrieved from AEDT file")
```
MaxJPRey, line 1377:
```suggestion
                self.logger.info("Lists were not retrieved from AEDT file")
```

pyaedt/modules/LayerStackup.py
MaxJPRey: Could we have both properties (`top_roughness` and `toprounghenss`)
The old one could return the value of the new one.
For instance for the getter of the old one,
```
@toprounghenss
def toprounghenss (self):
    return self.top_roughness
```

For the setter of the old one:
```
  @top_roughness.setter
  def top_roughness(self, val):
      self._top_roughness= val
      self.update_stackup_layer()
```

And so we don't expose the protected attribute `_toprounghenss` anymore.

maxcapodi78: I've made all exposed properties pythonic

MaxJPRey: @maxcapodi78 Perfect! Thanks

MaxJPRey (approved):

PR #1632, opened 2022-08-31
Remove licenses check.

MaxJPRey:
The license check actions must be updated. It is currently executing some shell commands.
We could have some cmd script performing it.


maxcapodi78 (approved):
LGTM. Better to wait when it's more stable

PR #1634, opened 2022-08-31
1064 add layer

ring630:
Draft


maxcapodi78 (approved):
LGTM. We need to refactor completely those classes to make more user friendly the stackup access for users. @svandenb-dev please discuss it with @ring630 about possible improvements

PR #1636, opened 2022-08-31
disabled Examples that will work in 2023R1 only

maxcapodi78:
Fix for aedtdocs.pyansys.com


maxcapodi78:
> @maxcapodi78 Do you want us to create a branch named 23.1 and reinstate those examples in it? or 23.1 beta? If yes, I can do it.

not needed for now. I guess only examples are affected

Issue #1628, opened 2022-08-30
Consider moving jupyterlab out of install_requires


isaacansys:
### Description of the feature

Jupyter Lab and its dependencies add significant download and install weight.

### Steps for implementing the feature

If Jupyter Lab is not required except for the example ipynb files, consider moving this dependency out of setup.py install_requires and into a new requirements file like requirements-examples.txt.

maxcapodi78:
@isaacansys I agree. I'm implementing it. I leave in pyvista, matplotlib and numpy as they are actively used in different methods

PR #1637, opened 2022-08-31
Remove JupyterLab from pyaedt requirements

maxcapodi78:
#1628


MaxJPRey (approved):
LGTM.

requirements.txt
```diff
@@ -32,16 +32,9 @@ matplotlib==3.5.3; python_version > '3.6'
 # License: https://github.com/matplotlib/matplotlib/blob/master/LICENSE/LICENSE
 # License: PSF
 
-numpy==1.21.6; python_version <= '3.7'
 numpy==1.23.2; python_version > '3.7'
```
isaacansys, line 35: This looks to me like numpy would not be installed for Python 3.7 (e.g. from the Ansys EM shipped version).  This could cause a failure except that its required in setup.py without a version number so maybe this one is redundant?

MaxJPRey: @isaacansys That is an excellent point. Yes, it should be fine thanks to the condition in the requirements.
But as I said, I agree with you that it could be confusing. I will try to clarify our installation.
Thanks again for the feedback.

isaacansys: @MaxJPRey not sure how standard this would be but for my own smaller projects I have been leaving out requirements.txt entirely so that setup.py or setup.cfg is the only source of build and install requirements.  Then I have a requirements-dev, requirements-docs, etc. for specific additions for development.

MaxJPRey: @isaacansys It was mainly for Dependabot, so it can parse the files properly and inform us of the new releases.

Issue #1624, opened 2022-08-30
Add create_dc_terminal(terminal_name) in pingroup Class


linmingchih:
### Description of the feature

Example code:
from pyaedt import Edb

edb = Edb(edbpath=r"D:\demo\Galileo_G87173_20454.aedb", edbversion='2022.1')
pins = edb.core_components.get_pin_from_component('U2A5', 'GND')
pg1 = edb.core_components.create_pingroup_from_pins(pins, 'group1')
x = edb.edb.Cell.Terminal.PinGroupTerminal.Create(pg1.GetLayout(), pg1.GetNet(), 'test', pg1)
x.SetBoundaryType(edb.edb.Cell.Terminal.BoundaryType.kDcTerminal)

edb.save_edb_as('d:/demo/ttt4.aedb')
edb.close_edb()
![image](https://user-images.githubusercontent.com/13580731/187362155-4a907a84-ac53-42d3-b426-99d2b212e1b8.png)

### Steps for implementing the feature

see example code in description

maxcapodi78:
Implemented a new method in core_siwave class

PR #1638, opened 2022-08-31
Implemented create_dc_terminal method in core_siwave class (#1624]

maxcapodi78:
#1624


ring630:
@maxcapodi78 I suggest to rename class DCSource to DCTerminal. A source should consists of 2 terminals.

ring630 (approved):

PR #1649, opened 2022-09-02
User defined component with native properties

Samuelopez-ansys:
With this PR, User defined components will have the native properties (if they are native 3DComponents like Fans, PCBs in Icepak or Antennas in SBR+).

This is the first step to deprecating the NativeComponent class

### DIFF
--- a/pyaedt/modeler/Object3d.py
+++ b/pyaedt/modeler/Object3d.py
@@ -3832,6 +3832,30 @@ class UserDefinedComponentParameters(dict):
         self._component = component
 
 
+class UserDefinedComponentProps(OrderedDict):
+    def __setitem__(self, key, value):
+        OrderedDict.__setitem__(self, key, value)
+        if self._pyaedt_user_defined_component.auto_update:
+            res = self._pyaedt_user_defined_component.update()
+            if not res:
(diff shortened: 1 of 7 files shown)


maxcapodi78 (approved):
Great.

pyaedt/modules/Boundary.py
```diff
@@ -218,17 +218,17 @@ def create(self):
             ``True`` when successful, ``False`` when failed.
 
         """
-        self.name = "InsertNativeComponentData"
+        # self.name = "InsertNativeComponentData"
```
MaxJPRey, line 221:
```suggestion
```

pyaedt/modeler/Object3d.py
```diff
@@ -3832,6 +3833,28 @@ def __init__(self, component, *args, **kw):
         self._component = component
 
 
+class UserDefinedComponentProps(OrderedDict):
```
MaxJPRey, line 3836: @Samuelopez-ansys The description is missing here. If we need to merge before the milestone, we can do it later, no worries.

MaxJPRey (approved):
Looks great.

PR #1653, opened 2022-09-05
Source terms to ground

svandenb-dev:


pyaedt/edb_core/EDB_Data.py
```diff
@@ -5166,6 +5187,11 @@ def _get_list_value(self, value):  # pragma: no cover
                 prop_values = [value.strip()]
             return prop_values
 
+    def add_dc_ground_source_term(self, source_name=None, node_to_ground=1):
```
MaxJPRey, line 5190: @svandenb-dev Please add the docstring of the method here.

PR #1657, opened 2022-09-05
improve change net visibility

gmalinve:


pyaedt/modeler/Primitives3DLayout.py
```diff
@@ -639,17 +639,32 @@ def change_net_visibility(self, netlist=None, visible=False):
         """
         if not netlist:
             netlist = self.nets
-
+        elif [x for x in netlist if x not in self.nets]:
+            self.logger.error("Selected net doesn't exist in current design.")
+            return False
         if type(netlist) is str:
```
MaxJPRey, line 645:
```suggestion
        if isinstance(netlist, str):
```

pyaedt/modeler/Primitives3DLayout.py
```diff
@@ -616,16 +616,16 @@ def padstacks(self):
         return self._padstacks
 
     @pyaedt_function_handler()
-    def change_net_visibility(self, netlist=None, visible=False):
+    def change_net_visibility(self, netlist=None, visible=True):
```
MaxJPRey, line 619: Do we really need to modify the default value.
Indeed, it will modify the method signature and could break some existing users' scripts.

pyaedt/modeler/Primitives3DLayout.py
```diff
@@ -616,16 +616,16 @@ def padstacks(self):
...
         """Change the visibility of one or more nets.
 
         Parameters
         ----------
         netlist : str  or list, optional
             One or more nets to visualize. The default is ``None``.
```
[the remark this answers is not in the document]
gmalinve: yep, I agree!

MaxJPRey (approved):
Thanks for improving the net visibility @gmalinve . Looks very good.

PR #1672, opened 2022-09-08
Array phase offset

Samuelopez-ansys:
New feature for phased array post-processing. Add a phase offset on each element. Needed when one element has more than 1 port, the phase between both ports could not be the same, for example, to achieve circular polarization is needed 90 degrees offset.


gmalinve (approved):

MaxJPRey (commented):
Well Done @Samuelopez-ansys .

PR #1680, opened 2022-09-09
New method to export Circuit log file

Samuelopez-ansys:
This is an interesting capability in Circuit which is not exposed in the API. This method saves the most recent Log file into a new path because sometimes if the project is saved, the log files could dissapear.
image


MaxJPRey (approved):
LGTM.

PR #1681, opened 2022-09-09
Edb set antipad diameter

svandenb-dev:


pyaedt/edb_core/padstack.py
```diff
@@ -470,6 +470,51 @@ def get_pad_parameters(self, pin, layername, pad_type=0):
         rot = padparams[5].ToDouble()
         return geom_type, parameters, offset_x, offset_y, rot
 
+    @pyaedt_function_handler
+    def set_all_antipad_value(self, value):
+        """Set all anti-pads from all pads-tack definition to the given value.
```
MaxJPRey, line 475:
```suggestion
        """Set all anti-pads from all pad-stack definition to the given value.
```

pyaedt/edb_core/padstack.py
```diff
@@ -470,6 +470,51 @@ def get_pad_parameters(self, pin, layername, pad_type=0):
...
+
+        Parameters
+        ----------
+        value float, str
```
MaxJPRey, line 479:
```suggestion
        value : float, str
```

pyaedt/edb_core/padstack.py
```diff
@@ -470,6 +470,51 @@ def get_pad_parameters(self, pin, layername, pad_type=0):
...
+            anti-pad value
+
+        Returns
+        --------
+        bool
+            ``True`` when all succeed ``False`` if one of anti-pads value failed to be assigned.
```
MaxJPRey, line 485:
```suggestion
            ``True`` when all succeed. ``False`` if one of anti-pads value failed to be assigned.
```

pyaedt/edb_core/padstack.py
```diff
@@ -470,6 +470,51 @@ def get_pad_parameters(self, pin, layername, pad_type=0):
...
+                    if geom_type == 1:
+                        params = convert_py_list_to_net_list([self._pedb.edb_value(value)] * len(parameters))
+                        geom = self._edb.Definition.PadGeometryType.Circle
+                        offset_x = self._pedb.edb_value(offset_x)
+                        offset_y = self._pedb.edb_value(offset_y)
+                        rot = self._pedb.edb_value(rot)
+                        antipad = self._edb.Definition.PadType.AntiPad
+                        if cloned_padstack_data.SetPadParameters(lay, antipad, geom, params, offset_x, offset_y, rot):
+                            self._logger.info(
+                                "Pad-stack definition {}, anti-pad on layer {}, has been set to {}".format(
```
MaxJPRey, line 505:
```suggestion
                                "For pad-stack definition {}, anti-pad on layer {} has been set to {}".format(
```

pyaedt/edb_core/padstack.py
```diff
@@ -470,6 +470,51 @@ def get_pad_parameters(self, pin, layername, pad_type=0):
...
+        --------
+        bool
+            ``True`` when all succeed ``False`` if one of anti-pads value failed to be assigned.
+        """
+        if self.padstacks:
+            for padstack in list(self.padstacks.values()):
+                cloned_padstack_data = self._edb.Definition.PadstackDefData(padstack.edb_padstack.GetData())
+                layers_name = cloned_padstack_data.GetLayerNames()
+                all_succeed = True
+                for lay in layers_name:
```
MaxJPRey, line 492: Maybe you can rename `lay` in `layer`.
```suggestion
                for layer in layers_name:
```

MaxJPRey: I you decide to do so, you will have to update it in the loop below also. Your call 😃

pyaedt/edb_core/padstack.py
```diff
@@ -470,6 +470,51 @@ def get_pad_parameters(self, pin, layername, pad_type=0):
...
+                                "Pad-stack definition {}, anti-pad on layer {}, has been set to {}".format(
+                                    padstack.edb_padstack.GetName(), lay, str(value)
+                                )
+                            )
+                        else:
+                            self._logger.error(
+                                "Failed to reassign anti-pad value {} on Pads-stack definition {},"
+                                " layer{}".format(str(value), padstack.edb_padstack.GetName(), lay)
+                            )
+                            all_succeed = False
```
MaxJPRey, line 514: Here, we could return `False`.
```suggestion
                            return False
```

MaxJPRey (approved):
Well done @svandenb-dev

pyaedt/edb_core/padstack.py
svandenb-dev: yes I did thx

pyaedt/edb_core/padstack.py
svandenb-dev: if we return False it will stop the loop although I wanted to keep running the loop. Siince the commandis called set_all_antipad I wanted o notice the user something wrong happened  but not necessarily stop the loop over all padstack definition. So I will keep like that for now

PR #1688, opened 2022-09-12
Added power_in field to create_pcb_from_3dlayout method.

maxcapodi78:
Fixed bug in power_budget method when 3dcomponent is created from pyaedt.


Samuelopez-ansys (approved):
Good spot!

PR #1699, opened 2022-09-14
Add source context option in Far field get solution data

### LINKED ISSUE #1673: "get_far_field_data" function misses a "Source" parameter.

### Description of the bug

In the case of a project containing multiple antennas, it's not possible to get the "far field" data of all antenna using the PostProcessor.get_far_field_data(expression='GainTotal', setup_sweep_name='', domain='Infinite Sphere1', families_dict=None)
function.

### Steps To Reproduce

In the case of a project containing multiple antennas, it's not possible to get the "far field" data of all antenna using the PostProcessor.get_far_field_data(expression='GainTotal', setup_sweep_name='', domain='Infinite Sphere1', families_dict=None)
function.

--> The get_far_field_data should also take the "Source" as parameter.

Explanations:

Right click "Excitations" and select "Edit Sources".

Select several sources : 

image

image

From now there is a new "Sources" menu in the Radiation pattern report : 

image

We need to be able to switch between the different sources (Ports) in order to access the "far field" data of all antennas.

Windows

3.10

### Installed packages

pyaedt 0.5.40
numpy 1.23.1

Samuelopez-ansys:
Close #1673

### DIFF
--- a/_unittest/test_12_PostProcessing.py
+++ b/_unittest/test_12_PostProcessing.py
@@ -316,6 +316,21 @@ class TestClass(BasisTest, object):
             assert data.plot(is_polar=True)
             assert data.plot_3d()
             assert self.field_test.post.create_3d_plot(data)
+
+        self.field_test.set_source_context(["1"])
+        context = {"Context": "3D", "SourceContext": "1:1"}
+        data = self.field_test.post.get_solution_data(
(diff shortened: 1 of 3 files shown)


MaxJPRey (approved):

Maxjrey:
LGTM.

PR #1700, opened 2022-09-14
SolutionData class now is able to support pandas output.

maxcapodi78:
A settings flag enable/disable it (enable_pandas_output). 
This can be set globally (from pyaedt.settings) or in a single solution_data. 
```
from pyaedt import settings
settings.enable_pandas_output = True
```
or 

```
sol = app.post.get_solution_data()
sol .enable_pandas_output = True
```

Added 2 additional properties to solution data full_matrix_real_imag and full_matrix_mag_phase which returns the full Solution, 
Matrix in either dict format or pandas format.


Samuelopez-ansys (approved):
This is a great enhancement!

PR #1705, opened 2022-09-16
minor improvement on get_solution_data

maxcapodi78:
now get_solution_data accept None as argument for expression which will provide all available traces for that solution in report category.
fix few bugs in is_solved flag for time domain analysis


Samuelopez-ansys (approved):
LGTM

Issue #1719, opened 2022-09-20
SiWave explicitly launches 2021 R2


GarrStau:
In the Siwave class initialization, the call to pywin32 uses a string instead of the computed `version`.

```python
self._main.oSiwave = win32com.client.Dispatch("Siwave.Application.2021.2")
```
https://github.com/pyansys/pyaedt/blob/30c5fa787883f0bdf27929abc477fc1442d61f0e/pyaedt/siwave.py#L117

Instead it should be
```python
self._main.oSiwave = win32com.client.Dispatch(version)
```

This is preventing usage with any release other than 2021.2

PR #1740, opened 2022-09-27
Return PyAEDT app object from Desktop object

### LINKED ISSUE #1654: Add method or property to get a list of designs in the currently open project from an instance of the Desktop() class.

We can get a list of project names, but do not have access to the pyaedt designs and type of design (i.e. Hfss(), Maxwell(), ...)

Create a property that returns a list of application object instances.  This could be implemented through an iterator that behaves like a dictionary for the desktop object.  The keys would be project name and design name. For example:
desktop["Project1"]["hfss_design"] would return an instance of Hfss.

_No response_

Samuelopez-ansys:
Close #1654 and #1719

### DIFF
--- a/_unittest/test_01_Design.py
+++ b/_unittest/test_01_Design.py
@@ -326,3 +326,15 @@ class TestClass(BasisTest, object):
             assert e.args[0] == "Project doesn't exists. Check it and retry."
         assert exception_raised
         settings.force_error_on_missing_project = False
+
+    def test_35_get_app(self):
+        d = Desktop(desktop_version, new_desktop_session=False)
+        assert d[[0, 0]]
+        assert d[["Coax_HFSS", "myname"]]
+        assert d[[0, "mydesign"]]
+        assert d[["Coax_HFSS", 2]]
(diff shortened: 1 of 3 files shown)


_unittest/test_01_Design.py
```diff
@@ -326,3 +326,15 @@ def test_34_force_project_path_disable(self):
...
+        d = Desktop(desktop_version, new_desktop_session=False)
+        assert d[[0, 0]]
+        assert d[["Coax_HFSS", "myname"]]
+        assert d[[0, "mydesign"]]
+        assert d[["Coax_HFSS", 2]]
+        assert not d[["Coax_HFSS", 5]]
+        assert not d[[1, 0]]
+        assert not d[[1, 0, 3]]
+        self.aedtapp.create_new_project("Test")
+        assert d[[1, 0]]
```
MaxJPRey, line 340:
```suggestion
        assert d[[1, 0]]
        assert "Test" in d[[1, 0]].project_name
```

maxcapodi78 (approved):
LGTM

PR #1741, opened 2022-09-27
Import bom bug fix

ring630:


ring630:
New class EDBComponentDef.

maxcapodi78 (approved):
I think is great idea. We should also add properties to it in order to get components and properties of the componentdef (like the model).

PR #1742, opened 2022-09-27
Fixed a bug in solution_type and a bug in primitives in edb

maxcapodi78:
Circuit solution type now is correctly setted
edb primitives object used to contain everything. now it contains only polygons, rectangles, circles and path.


Samuelopez-ansys (approved):
LGTM

PR #1747, opened 2022-09-28
Build json documentation for the developer portal.

MaxJPRey:


.github/workflows/ci-build.yml
```diff
@@ -57,6 +57,53 @@ jobs:
...
+          sphinx-build -M json doc/source doc-json -j auto -w build_errors.txt -N;
+
+      - name: Flatten the generated nested files into a single directory
+        run: |
+          echo Flattening a nested directory
+          mkdir doc-flatten-json;
+          echo Move all the JSON file to the flatten directory.
+          find . -name "*.fjson" -exec mv "{}" --backup=numbered ./doc-flatten-json \;
+          echo Make sure all the file has .json extensions instead of the .fjson
+          for f in doc-flatten-json/* ; do mv -- "$f" "$f.json" ; done;
```
germa89, line 72: if ``f`` is the complete file name, you might end up with ``file.json.json`` or ``file.doc.json``.

germa89: image

.github/workflows/ci-build.yml
```diff
@@ -57,6 +57,53 @@ jobs:
...
+          echo Move all static image files to the JSON flatten folder.
+          mv doc-json/json/_images doc-flatten-json/_images;
```
germa89, line 74: Also the images are in a folder:

image

Not sure if that is the idea.

.github/workflows/ci-build.yml
```diff
@@ -57,6 +57,53 @@ jobs:
...
+          application_private_key: ${{ secrets.BOT_APPLICATION_PRIVATE_KEY }}
+
+      - name: Deploy
+        uses: JamesIves/github-pages-deploy-action@v4.4.0
+        with:
+          repository-name: pyansys/devportal-doc-archive
+          branch: drupal
+          folder: pyaedt-doc-flatten-json
+          token: ${{ steps.get_workflow_token.outputs.token }}
+          clean: false
```
germa89, line 105: Interesting use of the deploy action.

.github/workflows/ci-build.yml
germa89: Suggested solution:

https://stackoverflow.com/a/71969517/6650211

germa89 (approved):

.github/workflows/ci-build.yml
MaxJPRey: Improved:
image

I will check if the backed up files are needed.

.github/workflows/ci-build.yml
MaxJPRey: For the first draft, it should be OK from the feedback I got.
But I completely agree that this could evolve in the next set of modifications where I will be handling examples for the "Full" documentation build for each release.

PR #1786, opened 2022-10-07
Added new methods to create ports on nets of the components and to create differential ports in Hfss 3d Layout

maxcapodi78:
…eate differential ports in Hfss 3d Layout


Samuelopez-ansys (approved):
lgtm

PR #1791, opened 2022-10-09
Add return code to Siwave.open_project

GarrStau:
Relative paths that pass `os.path.exists` can still fail to open.
Added a return value so calling code can throw an error if the project fails to open.


MaxJPRey (approved):
LGTM.

maxcapodi78 (approved):
LGTM

PR #1804, opened 2022-10-10
Wrong float conversion

Samuelopez-ansys:
This float conversion was affecting rotation in Circuit. For example, for a Port created if you tried to rotate it, it was failing because it expected an int and not a float


MaxJPRey (approved):
LGTM.

PR #1827, opened 2022-10-13
Fixed bug in create_rectangle_in_pad method

Alberto-DM:
Fixed bug in create_rectangle_in_pad. Routine was giving unexpected results in case the padstack poly definition was defined with points in clockwise direction.
Improved the unit tests.


maxcapodi78 (approved):
LGTM

PR #1831, opened 2022-10-14
Fixed extra parameter inside desktop.py causing AedtLogger.error to c…

GarrStau:
Attempts by Desktop() to log errors instead result in additional errors, masking the original cause.

Looks like there have been several refactors around AedtLogger, although I could not pin down exactly when this was introduced.
Best guess is this change was missed in this change: https://github.com/pyansys/pyaedt/commit/8b01103e1e7aa3536d0447d025b505bc931e6117#diff-084e551e13f981601616b1e08b400b8dede115a81e40332a72554cb03937a613


MaxJPRey:
@GarrStau yes, the `global` argument is not required anymore.
Thanks for the fixing it.

MaxJPRey (approved):
LGTM.
Thanks @GarrStau for your contribution.

PR #1832, opened 2022-10-14
Defaulting rlc value to zero when pinpairs are not present

svandenb-dev:
Note when importing zero values Rlc:
- Aedt will assign default values
- Siwave will keep zero values but deactivate the component.


MaxJPRey (approved):
LGTM.

PR #1851, opened 2022-10-19
fix export_results 

gmalinve:
- GetSweeps doesn't work with all design types (such as Maxwell 2D/3D)
- export_results clean


MaxJPRey (approved):
LGTM.

PR #1859, opened 2022-10-22
Q2D/Q3D export touchstone

gmalinve:
Q2D/Q3D export touchstone implementation in Analysis.py, cleaning + test


Samuelopez-ansys (approved):
LGTM

PR #1860, opened 2022-10-24
New class PinGroup and Terminal

maxcapodi78:
New property terminals part of edb
Refactoring of pingroups from list of Edb Object to dictionary of PinGroup Class

### DIFF
--- a/pyaedt/edb.py
+++ b/pyaedt/edb.py
@@ -263,7 +263,7 @@ class Edb(object):
         return names
 
     @property
-    def terminals(self):
+    def excitations(self):
         """Retrieve the Terminals of the active edb.
 
         Returns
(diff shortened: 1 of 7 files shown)


svandenb-dev (approved):

pyaedt/edb.py
```diff
@@ -260,6 +262,19 @@ def cell_names(self):
...
+    def excitations(self):
+        """Retrieve the Terminals of the active edb.
+
+        Returns
+        -------
+        dict
+        """
+        terminals = {}
+        for terminal in list(self.active_layout.Terminals):
+            terminals[terminal.GetName()] = Terminal(self, terminal)
```
ring630, line 275: A excitation consises of two terminals, so if we count the number of exicitation by couting terminals reference terminals should be excluded.

MaxJPRey (approved):

PR #1876, opened 2022-10-25
Refactored Hfss Sweep and Hfss3dLayout sweep to support Siwave and new version properties

maxcapodi78:
We used the same approach as in Setup with Sweep templates.


Samuelopez-ansys (approved):
LGTM

MaxJPRey (approved):
LGTM.

PR #1884, opened 2022-10-26
Add an example for the heat sink in Icepak.

MaxJPRey:
Add an example for the heat sink in Icepak.
Improve test for the heat sink in Icepak.

Use snake_case in icepak.py
Remove unnecessary 'pass' statement.
Remove new line in icepak tests.

### DIFF
--- a/pyaedt/icepak.py
+++ b/pyaedt/icepak.py
@@ -1073,14 +1073,14 @@ class Icepak(FieldAnalysis3D):
             Length of the heat sink. The default is ``40``.
         height : optional
             Height of the heat sink. The default is ``40``.
-        draftangle : optional
+        draftangle : int, float, optional
             Draft angle in degrees. The default is ``0``.
-        patternangle : optional
+        patternangle : int, float, optional
             Pattern angle in degrees. The default is ``10``.
         separation : optional
             The default is ``5``.
         symmetric : bool, optional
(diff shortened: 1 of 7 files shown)


pyaedt/icepak.py
```diff
@@ -1092,18 +1092,32 @@ def create_parametric_fin_heat_sink(
...
             Tolerance value. The default is ``0.001``.
 
         Returns
         -------
         bool
             ``True`` when successful, ``False`` when failed.
 
+        Examples
+        --------
+
```
PipKat, line 1111: I don't think a blank line is needed here. I'm not sure if it will cause a problem.

MaxJPRey, line 1111:
```suggestion
        --------
```

pyaedt/icepak.py
```diff
@@ -1092,18 +1092,32 @@ def create_parametric_fin_heat_sink(
...
+        Create a symmetric fin heat sink.
```
MaxJPRey, line 1112:
```suggestion
        Examples
        --------
        Create a symmetric fin heat sink.
```

pyaedt/icepak.py
maxcapodi78: No it's not. good catch.

maxcapodi78 (approved):

PR #1893, opened 2022-10-28
Extend ironpython test time

maxcapodi78:


MaxJPRey:
@maxcapodi78 I don't see the timeout extended for Linux, while it seems that you did commit it as the first commit.
Is it expected?

PR #1895, opened 2022-10-28
Improve Insert Matrix method in Q3D and Q2D 

Samuelopez-ansys:
Pass Net, source and Sink as an argument. This will give more control to the insert matrix method. In addition, JoinParallel operation now allows Net selection and not only Source selection


MaxJPRey (approved):
Great.

PR #1904, opened 2022-10-31
exclude reference terminal from port counting

ring630:


MaxJPRey:
Great catch@ring630!
Can you add a unit test? It seems we were not checking it. This could be the opportunity to avoid regressions in next modifications.

ring630:
> Great catch@ring630! Can you add a unit test? It seems we were not checking it. This could be the opportunity to avoid regressions in next modifications.

Done

MaxJPRey (approved):

PR #1920, opened 2022-11-02
port enhancement

ring630:


maxcapodi78 (approved):
maybe the only comment is related to the name "Ports". in pyaedt we used to call either sources or excitations. Not sure if would be better to use that name

pyaedt/edb_core/edb_data/sources.py
```diff
@@ -437,3 +439,51 @@ def rvalue(self, value):
     def source_type(self):
         """Source type."""
         return self._source_type
+
+
+class Port:
+    """Manages port properties"""
```
MaxJPRey, line 445: Documentation for the initializer is missing.
If possible, could you also add an example in the docstring.

pyaedt/edb_core/edb_data/sources.py
```diff
@@ -437,3 +439,51 @@ def rvalue(self, value):
...
+
+    @property
+    def vertical_extent_factor(self):
+        """Get vertical extent factor."""
+        txt = re.search(r"'Vertical Extent Factor'='.*?'", self._edb_properties).group()
+        return float(txt.split("=")[1].replace("'", ""))
+
+    @property
+    def radial_extent_factor(self):
+        """Get radial extent factor"""
```
MaxJPRey, line 481:
```suggestion
        """Get radial extent factor."""
```

pyaedt/edb_core/hfss.py
```diff
@@ -618,6 +631,16 @@ def create_edge_port_vertical(
...
+        hfss_type : str, optional
+            Type of the port. The default is ``Gap``. Options are ``"Gap"``, ``"Wave"``.
+        horizontal_extent_factor : int, float, optional
+            Horizontal extent factor.
+        vertical_extent_factor : int, float, optional
+            Vertical extent factor.
+        radial_extent_factor : int, float, optional
+            Radial exctent factor.
+        pec_launch_width : str, optional
+            Launch Width of PEC.
```
MaxJPRey, line 643:
```suggestion
        hfss_type : str, optional
            Type of the port. The default value is ``"Gap"``. Options are ``"Gap"``, ``"Wave"``.
        horizontal_extent_factor : int, float, optional
            Horizontal extent factor. The default value is ``5``.
        vertical_extent_factor : int, float, optional
            Vertical extent factor. The default value is ``3``.
        radial_extent_factor : int, float, optional
            Radial extent factor. The default value is ``0``.
        pec_launch_width : str, optional
            Launch Width of PEC. The default value is ``"0.01mm"``.
```

_unittest/test_00_EDB.py
```diff
@@ -2034,6 +2034,12 @@ def test_A118_edb_create_port(self):
                 prim_1_id, ["-60mm", "-4mm"], prim_2_id, ["-59mm", "-4mm"], "port_hori", 30, "Lower"
             )
             assert edb.core_hfss.get_ports_number() == 2
+            port_ver = edb.core_hfss.ports["port_ver"]
+            assert port_ver.hfss_type
+            assert port_ver.horizontal_extent_factor
+            assert port_ver.vertical_extent_factor
```
MaxJPRey, line 2040:
```suggestion
            assert isinstance(port_ver.horizontal_extent_factor, float)
            assert isinstance(port_ver.vertical_extent_factor, float)
```
@ring630 I am not sure but here we could even check the value of the float returned.

_unittest/test_00_EDB.py
```diff
@@ -2034,6 +2034,12 @@ def test_A118_edb_create_port(self):
                 prim_1_id, ["-60mm", "-4mm"], prim_2_id, ["-59mm", "-4mm"], "port_hori", 30, "Lower"
             )
             assert edb.core_hfss.get_ports_number() == 2
+            port_ver = edb.core_hfss.ports["port_ver"]
+            assert port_ver.hfss_type
```
MaxJPRey, line 2038: Could we check the value of the string returned?

ring630:
> maybe the only comment is related to the name "Ports". in pyaedt we used to call either sources or excitations. Not sure if would be better to use that name

Renamed to excitations.

pyaedt/edb_core/edb_data/sources.py
ring630: Can you help me with a template?

pyaedt/edb_core/edb_data/sources.py
```diff
@@ -437,3 +439,51 @@ def rvalue(self, value):
     def source_type(self):
         """Source type."""
         return self._source_type
+
+
+class Excitation:
+    """Manages excitation properties"""
```
MaxJPRey, line 445:
```suggestion
    """Manages excitation properties.

    Detailed description here.

    Parameters
    ----------
    pedb : str, optional
        Name of the project to select or the full path to the project
        or AEDTZ archive to open. The default is ``None``, in which
        case an attempt is made to get an active project. If no
        projects are present, an empty project is created.


    Examples
    --------
    Create an instance of HFSS and connect to an existing HFSS
    design or create a new HFSS design if one does not exist.

    >>> from pyaedt import Hfss
    >>> hfss = Hfss()
    pyaedt info: No project is defined...
    pyaedt info: Active design is set to...
```

MaxJPRey: @ring630 You might have to improve the example a little bit.

PR #1921, opened 2022-11-02
Fix/export siwave dc report

gmalinve:


Samuelopez-ansys (approved):
LGTM.  os.system was failing sometimes. subprocess should work for different OS

MaxJPRey (approved):
LGTM.

PR #1933, opened 2022-11-03
Enhancement delete nets

ring630:


MaxJPRey (approved):
LGTM.

pyaedt/edb_core/layout.py
```diff
@@ -639,6 +639,35 @@ def create_rectangle(
                 self._get_edb_value(rotation),
             )
 
+    @pyaedt_function_handler
+    def delete_primitives(self, net_names=None):
```
MaxJPRey, line 643: I think we can remove the default value here, because otherwise, we are going to create a list: list [None] line 661.
```suggestion
    def delete_primitives(self, net_names):
```

PR #1945, opened 2022-11-04
Specify the doc path for the html_context metadata.

MaxJPRey:


Samuelopez-ansys (approved):
LGTM

Samuelopez-ansys:
@MaxJPRey Why Codecov is checking everything? Could we change it and only check the modified code?

MaxJPRey:
> @MaxJPRey Why Codecov is checking everything? Could we change it and only check the modified code?

Yes @Samuelopez-ansys I will investigate that. So we have a better status for the PR.

PR #1965, opened 2022-11-09
fixed small bug in create_extent method

maxcapodi78:
there was a bug in extent computation that was preventing the usage of extent different from Conformal even if selected by the method arguments.


ring630 (approved):

MaxJPRey (approved):
LGTM.

PR #1971, opened 2022-11-10
create differential wave port

ring630:


pyaedt/edb_core/edb_data/sources.py
```diff
@@ -505,3 +505,20 @@ def pec_launch_width(self):
         """Get pec launch width."""
         txt = re.search(r"'PEC Launch Width'='.*?'", self._edb_properties).group()
         return txt.split("=")[1].replace("'", "")
+
+
+class ExcitationDifferential:
```
MaxJPRey, line 510: Just missing the doctring. Maybe it is because this class will only be used internally.

pyaedt/edb_core/edb_data/sources.py
```diff
@@ -505,3 +505,20 @@ def pec_launch_width(self):
...
+    def __init__(self, pedb, edb_boundle_terminal, name):
+        self._pedb = pedb
+        self._edb_boundle_terminal = edb_boundle_terminal
+        self._name = name
+
+    @property
+    def _edb(self):
```
MaxJPRey, line 517: @ring630 Why not naming the property `edb` instead?

MaxJPRey (approved):
This looks great. Thanks for adding the examples too.

pyaedt/edb_core/edb_data/sources.py
ring630: Done

pyaedt/edb_core/edb_data/sources.py
MaxJPRey: Perfect. Thanks

pyaedt/edb_core/edb_data/sources.py
```diff
@@ -505,3 +505,23 @@ def pec_launch_width(self):
         """Get pec launch width."""
         txt = re.search(r"'PEC Launch Width'='.*?'", self._edb_properties).group()
         return txt.split("=")[1].replace("'", "")
+
+
+class ExcitationDifferential:
+    """Manages differential excitation properties."""
+    def __init__(self, pedb, edb_boundle_terminal, name):
```
MaxJPRey, line 512:
```suggestion
    """Manages differential excitation properties."""

    def __init__(self, pedb, edb_boundle_terminal, name):
```

PR #1977, opened 2022-11-11
New feature to hide a list of variables

### LINKED ISSUE #1896: Hiding Variable is slow
I noticed that hiding variables in HFSS through pyaedt is very slow.  Tried 2 techniques

1:
hfss.hidden_variable(var_name, True)

2:
hfss.variable_manager.variables[var_name].hidden = True

Both techniques seem comparable.  It was taking on the order of 5 - 10 seconds to hide 80 variables.
In the HFSS GUI, I could hide all the variables, and it appeared to happen instantaneously.

Samuelopez-ansys:
Close #1896 

Allow a list as an input of the "hidden_variable method". Then the method applies all the modifications at once and it is faster than doing them one by one.


MaxJPRey (approved):
LGTM.

PR #1978, opened 2022-11-11
Add example to method "export_field_file_on_grid"

### LINKED ISSUE #1816: Improve example of the export_Field_file_on_grid

I am a new python user and this call threw me for a loop for a few days. Could you add a little more explanation to the example?

hfss.post.export_field_file_on_grid("E", solution=solnSetup, variation_dict=aedtapp.available_variations.nominal_w_values, filename=os.path.join(pathname, "GTEMFieldExport_export_Field_file_onGrid.fld"), gridtype='Cartesian', grid_center=[0, 0, 0], grid_start=[.7, .03, -.6], grid_stop=[.7, .07, .6], grid_step=[0.01, 0.01, 0.01], isvector=True, intrinsics="8GHz", phase=None)

Here are things I struggled with:
1) The solution setup is a string formatted as "Setup Name : LastAdaptive"
2) The variation dictionary is not actually a dictionary. It's not formatted as {Word : "Definition"}, but as ["Word=:Definition"]. It would be helpful to change the explanation for the variation_dict variable to match the format. If you use the aedtapp.available_variations.nominal_w_values, the frequency and phase definitions are not included with the dictionary--this should be noted.
3) The filename has to be formatted properly in order for this method to run. If it is not, you don't get an error message like "path not found" but an executive method fail that is really hard to debug. It would be nice if the error message was more specific when it fails at filename. It would be useful in the example to show the automatic formatting call "os.path.join"
4) gridtype -- if it set to Cartesian, the instructions say the grid center is ignored. However, in the export_field_file_on_grid method, that doesn't appear to be the case (again, I'm a new user and can be wrong about this)
5) intrinsics -- This is required if you use the variation call above, and intrinsics can only include "frequency". There is another example floating in the API that defines intrinsics differently, it would be helpful to include in the example the required format of "freq" + "GHz"

Feel free to use this snippet of code as the example!
#############################################################
spr = .01 #space between sampling points, in default units
GTEM1x, GTEM1y, GTEM1z = -0.3, 0, -0.13 #This coordinate system HAS to be smaller values than GTEM2
GTEM2x, GTEM2y, GTEM2z = -0.3, 0.1, 0.13 #The XYZ coordinates here must be larger than GTEM1

#Path name of this project and where the field exports will be saved, make a folder for these files
pathname = "C:\\Users\\examples\\"

proj = "GTEM_R2" #Project name
gtemDesign = "GTEMCell"
solutiontype="DrivenModal"
version = "2022.2"
setupname="Setup1" #Analysis Setup Name
freq = "8GHz" #Frequency of operation, solution solved at this frequency
solnSetup = setupname + " : LastAdaptive"
#Initialize design-object
hfssGTEM = Hfss(specified_version=version, projectname=proj, designname=gtemDesign, solution_type=solutiontype, setup_name=setupname, non_graphical=False, new_desktop_session=False, close_on_exit=False, student_version=False, machine='', port=0, aedt_process_id=None)
aedtapp = hfssGTEM
post = aedtapp.post #initialize post processing object

GTEMcoord1 = [GTEM1x, GTEM1y, GTEM1z]
GTEMcoord2 = [GTEM2x, GTEM2y, GTEM2z]
filenameGTEM = "GTEMFieldExport.fld"
filenameAir = "AirFieldExport.fld"

#Export Fields from GTEM Cell
hfssGTEM.post.export_field_file_on_grid("E", solution=solnSetup, variation_dict=hfssGTEM.available_variations.nominal_w_values, filename=os.path.join(pathname, filenameGTEM), gridtype='Cartesian', grid_center=[0, 0, 0], grid_start=GTEMcoord1, grid_stop=GTEMcoord2, grid_step=[spr, spr, spr], isvector=False, intrinsics=freq, phase=None)
hfssGTEM.release_desktop(close_projects=False, close_desktop=False)

_No response_

Samuelopez-ansys:
Close #1816

### DIFF
--- a/pyaedt/modeler/Modeler.py
+++ b/pyaedt/modeler/Modeler.py
@@ -2397,7 +2397,7 @@ class GeometryModeler(Modeler, object):
 
         Parameters
         ----------
-        bjid : str, int, or  Object3d
+        objid : str, int, or  Object3d
             Name or ID of the object.
         position : float
             List of the ``[x, y, z]`` coordinates or
--- a/pyaedt/modules/PostProcessor.py
+++ b/pyaedt/modules/PostProcessor.py
@@ -2288,6 +2288,15 @@ class PostProcessor(PostProcessorCommon, object):
         >>> oModule.EnterQty
         >>> oModule.EnterVol
         >>> oModule.ExportOnGrid
+
+        Examples
+        --------
+        >>> from pyaedt import Hfss
+        >>> hfss = Hfss()
+        >>> var = hfss.available_variations.nominal_w_values
+        >>> setup = "Setup1 : LastAdaptive"
+        >>> path = "Field.fld"
+        >>> hfss.post.export_field_file_on_grid("E", setup, var, path, 'Cartesian', [0, 0, 0],  intrinsics="8GHz")
         """
         self.logger.info("Exporting %s field. Be patient", quantity_name)
         if not solution:


MaxJPRey (approved):
LGTM?

PR #1992, opened 2022-11-15
feature/add mesh link

gmalinve:
Add mesh link for:
- HFSS
- Maxwell
- Mechanical


pyaedt/modules/SolveSetup.py
```diff
@@ -463,25 +463,42 @@ def disable(self, setup_name=None):
...
+        solution_name : str, optional
             Name of the solution in the format ``"setupname : solutionname"``.
-            Optionally use :attr:`appname.nominal_adaptive` to get the
-            nominal adaptive or :attr:`appname.nominal_sweep` to get the
-            nominal sweep.
-        parameters_dict : dict
-            Dictionary of the parameters. Optionally use
-            :attr:`appname.available_variations.nominal_w_values_dict`
-            to get the nominal values.
+            If None the default value is ``setupname : LastAdaptive``.
```
MaxJPRey, line 485:
```suggestion
            If ``None`` the default value is ``setupname : LastAdaptive``.
```

pyaedt/modules/SolveSetup.py
```diff
@@ -463,25 +463,42 @@ def disable(self, setup_name=None):
...
+        parameters_dict : dict, optional
+            Dictionary of the parameters.
+            If None the default value is `appname.available_variations.nominal_w_values_dict`.
```
MaxJPRey, line 488:
```suggestion
            If ``None`` the default value is `appname.available_variations.nominal_w_values_dict`.
```

pyaedt/modules/SolveSetup.py
```diff
@@ -493,24 +510,60 @@ def add_mesh_link(self, design_name, solution_name, parameters_dict, project_nam
...
+                ):
+                    meshlinks["Soln"] = "{} : LastAdaptive".format(solution_name.split()[0])
+                else:
+                    raise ValueError("Setup does not exist in current design.")
+
+            meshlinks["Params"] = OrderedDict({})
+            if parameters_dict is None:
+                parameters_dict = self.p_app.available_variations.nominal_w_values_dict
+                for el in parameters_dict:
+                    meshlinks["Params"][el] = el
```
MaxJPRey, line 552:
```suggestion
                    meshlinks["Params"][el] = parameters_dict[el]
```

gmalinve: @MaxJPRey  
actually I kept the original version: meshlinks["Params"][el] = el
because when you add mesh link if you don't map variables by name it gives back an error and the link is not created.
Look at this example:
image
I slightly modified the test though.

MaxJPRey: OK, makes sense. Thanks for the explanation @gmalinve .

Samuelopez-ansys (approved):

MaxJPRey (approved):
Looks great.

PR #1999, opened 2022-11-17
In a workflow, the same identifier cannot be used more than once within the same scope.

MaxJPRey:
That should fix the current issue with the `FullDocumentation` workflow when generating new releases.


Samuelopez-ansys (approved):
LGTM

PR #2009, opened 2022-11-19
fixed bug in X.analyze_setup(...)

nikolarobottesla:
Appears to be a copy/paste error.  Without the fix the .acf file will contain 'UseAutoSettings=<self.design_type>' e.g.  UseAutoSettings='HFSS'

With the fix the .acf file will contain UseAutoSettings=True or UseAutoSettings=False I tested import functionality through the HPC and Analysis options in the GUI. The capitalized True/False that are written to the .acf file import without error and demonstrate the expected behavior.


maxcapodi78 (approved):
LGTM. Thanks for your first contribution

MaxJPRey (approved):
LGTM. Thanks for your time and the improvement @nikolarobottesla .

PR #2013, opened 2022-11-21
Cutout extent defeaturing options

maxcapodi78:
now is possible to add defeaturing to the cutout extent creation in create_cutout_multithread method


MaxJPRey (approved):
LGTM.

PR #2020, opened 2022-11-22
Icepak method

ansys-satyajeet:
dded new method create_two_resistor_network_block.
Extended support for import_idf method to BDF and LDF


maxcapodi78 (approved):
LGTM

PR #2023, opened 2022-11-22
fix/emit_veritification_fail_changes_S722982: should pass with update…

cchan-ansys:
This is a fix for verification issues in a recent build of AEDT. This issue arises from the fact that my local build of delcross was behind the build version of delcross and there the wrapper on the PYAEDT side was having some issues.


Samuelopez-ansys (approved):
LGTM

Samuelopez-ansys (approved):
LGTM

PR #2029, opened 2022-11-23
Clr refactoring

maxcapodi78:
Move all clr loader into a new module file.
Added methods for lazy  modules load


MaxJPRey:
@maxcapodi78 It looks great. I will review a final part tomorrow first thing.

MaxJPRey (approved):

Issue #2030, opened 2022-11-23
Error when exporting an aedb to ipc2581


felipeescastro:
### Before submitting the issue

- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### Description of the bug

Hi, when trying to export an aedb to ipc2581, I get the following:

> In [1]: from pyaedt import Edb
> In [2]: inputfile = r"D:\aaSupport\test_six.aedb\edb.def"
> In [3]: edb = Edb(edbpath=inputfile,edbversion="2022.2")
> In [4]: edb.export_to_ipc2581(r"D:\aaSupport\test_six_ipc.xml","millimeter")
> pyaedt info: Export IPC 2581 is starting. This operation can take a while.
> pyaedt info: Error exporting IPC 2581.
> pyaedt info: Sequence contains no elements
>    at System.Linq.Enumerable.First[TSource](IEnumerable`1 source)
>    at EdbLib.IPC8521.IPCExporter.AddLayerSpecs(CadHeader CadHeader, Layout layout, String units)
>    at EdbLib.IPC8521.IPCExporter.BuildIPC2581(Layout l...
> Out[5]: False
> 2022-11-23 13:17:57,941 - INFO - Starting IPC2581 Export
> 2022-11-23 13:17:57,941 - INFO - Exporting Content Section
> 2022-11-23 13:17:57,941 - INFO - Exporting Layer Section

Test file can be shared internally to Ansys. Please reach out to get the file.
Kind regards,
Felipe

### Steps To Reproduce

Run command lines In [1] to In [4] in the bug description using database provided by me. 

### Which Operating System are you using?

Windows

### Which Python version are you using?

3.7

### Installed packages

aiohttp==3.8.3
aiosignal==1.2.0
alabaster==0.7.12
anyio==3.6.1
appdirs==1.4.4
argon2-cffi==21.3.0
argon2-cffi-bindings==21.2.0
arrow==1.2.3
astroid==2.12.11
async-timeout==4.0.2
asynctest==0.13.0
atomicwrites==1.4.1
attrs==22.1.0
autopep8==1.6.0
Babel==2.10.3
backcall==0.2.0
bcrypt==4.0.1
beautifulsoup4==4.11.1
binaryornot==0.4.4
black==22.10.0
bleach==5.0.1
certifi==2022.9.24
cffi==1.15.1
chardet==5.0.0
charset-normalizer==2.1.1
click==8.1.3
cloudpickle==2.2.0
[...]

PR #2038, opened 2022-11-25
support .aedb/edb.def as input

ring630:


maxcapodi78 (approved):

MaxJPRey (approved):
LGTM.
Linking to a former issue #2030 .

Issue #2040, opened 2022-11-25
Add method to get the list of object that belongs to net in Q3d


maxcapodi78:
### Description of the feature

Q3d Allows to group objects by nets. A method to retrieve objects by nets is needed.
In addition them method should filter the objects for specific materials.

PR #2041, opened 2022-11-25
Added new method objects_from_nets to Q3d

maxcapodi78:
to get the list of objects that belongs to a nets. Those can be also filtered by material.

#2040


felipeescastro (approved):
New function working as expected. Tested on multiple projects. Case sensitivity works fine.

PR #2058, opened 2022-11-29
Fix bug in general_methods.py when pyaedt is called as module

maxcapodi78:


cmesibov (approved):
Code changes look good and local testing passed.

PR #2064, opened 2022-11-30
Icepak method

ansys-satyajeet:
Enhanced ``create_two_resistor_network_block`` to automatically detect board side and case side. Enhanced to work with 3D component pcbs and primitive blocks in arbitrary orientation.

Added unit test for ``create_two_resistor_network_block``


maxcapodi78 (approved):
LGTM

PR #2076, opened 2022-12-02
Deprecate method

Samuelopez-ansys:
Copy_design method is not tested in the unittest and there is an equivalent method 'copy_design_from'. I think we should deprecate it.


MaxJPRey (approved):
LGTM.

PR #2084, opened 2022-12-04
Move all requirement files into a single folder.

### LINKED ISSUE #2074: Collect extra requirements inside a requirements/ folder
🐞 **Problem**
Before implementing #2071, a `requirements/` folder containing all requirements files is needed. Otherwise, since this project does not support `pyproject.toml`, those will not work for some reusable actions.

MaxJPRey:
A `requirements/` folder containing all requirements files is needed to use the pyansys [actions](https://github.com/pyansys/actions).
Fix #2074 .


jorgepiloto (approved):
Looks great! Make sure dependabot is also updated so it points to the new requirements folder instead to the base directory.

PR #2097, opened 2022-12-06
Bug via from to layer is inverted

ring630:


MaxJPRey (approved):
LGTM.

maxcapodi78 (commented):
please modify also Edb.py (line 1563 and on ) and example 04_edb_parametrized_design.py accordingly

PR #2123, opened 2022-12-09
Added import_nastran method to Model3D.py

maxcapodi78:


Samuelopez-ansys (approved):
This is great!

pyaedt/modeler/Model3D.py
```diff
@@ -556,9 +555,6 @@ def import_nastran(self, file_path, import_lines=True, lines_thickness=0):
         -------
         List of :class:`pyaedt.modeler.Object3d.Object3d`
         """
-        if is_ironpython:
-            self._app.logger.error("Method not supported in ironpython. Try import_3d_cad method.")
```
PipKat: The message string should be a complete sentence, such as:

"Method is not supported in IronPython. Try the import_3d_cad method."

maxcapodi78: I've removed and made the method compatible with ironpython.Thanks for suggestion

PR #2126, opened 2022-12-12
Fix bug in units computation on variables when a formula is entered.

maxcapodi78:


MaxJPRey (approved):
This will be very useful. LGTM.

gmalinve (approved):

Issue #2133, opened 2022-12-13
Add function that returns list of variables enabled for derivative analysis in solution setup in HFSS


p-krenz:
### Description of the feature

In the solution setup in HFSS, we can mark variables to be included in the derivative study (see Derivatives tab in solution setup).  During the solution process, HFSS will compute derivatives of output quantities with respect to those derivatives.
It will be helpful to have a function that can return a list of variables that are enabled for this derivative study.

### Steps for implementing the feature

In the IronPython scripting for HFSS, it is possible to query a solution setup and have the derivative variables returned:

oModule = oDesign.GetModule('AnalysisSetup')
oModule.GetDerivativeVariables('Setup1')

This command would return a list, for example ['Length1', 'Length2']

PR #2145, opened 2022-12-14
Added method to get_derivative_variables to Hfss Setup classes

maxcapodi78:
Refactored method add_derivatives and extended support to all HFSS setups.
#2133


Samuelopez-ansys (approved):
Great

PR #2148, opened 2022-12-14
Use the __version__ attribute to set the release and version value.

MaxJPRey:
The release and version value are not read from the `version.txt` file anymore.


maxcapodi78 (approved):
LGTM

PR #2155, opened 2022-12-16
FIX: turn doc-deploy-stable into a unique job

jorgepiloto:
As pointed out by @MaxJPRey, the latest stable release failed when pushing the stable documentation using the multi-version documentation. This is because deployment actions are expected to be executed under a Linux environment while PyAEDT is using a Windows self-hosted agent.

The following modifications guarantee that the deployment takes place.


MaxJPRey (approved):
LGTM. Thanks @jorgepiloto for the support.

PR #2204, opened 2022-12-26
Add plane to the modeler and primitives.

### LINKED ISSUE #1930: Add feature "draw plane"

In HFSS it is possible to draw planes via the "draw plane" function. It would be useful to have a similar function in PyAEDT.

_No response_

_No response_

MaxJPRey:
It is now possible to add a cut plane to the modeler.

image

A dictionary collecting all the the existing planes and the new ones is available. The planes' name are used as the key of this dictionary.
Planes properties such as name, color... are available .

Fix #1930 .


pyaedt/modeler/Object3d.py
```diff
@@ -4836,3 +4836,209 @@ def delete(self):
...
+        >>> oEditor.GetProperties
+        """
+        if not self._all_props:
+            self._all_props = _retry_ntimes(10, self.m_Editor.GetProperties, "Geometry3DPlaneTab", self._name)
+        return self._all_props
+
+    # Note: We currently cannot get the color property value because
+    # when we try to access it, we only get access to the 'edit' button.
+    # Following is the line that we would use but it currently returns 'edit'.
+    # color = _retry_ntimes(10, self.m_Editor.GetPropertyValue, "Geometry3DPlaneTab", self._name, "Color")
```
PipKat, line 4948:
```suggestion
    # Note: You currently cannot get the color property value because
    # when you try to access it, you only get access to the 'edit' button.
    # Following is the line that you would use, but it currently returns 'edit'.
    # color = _retry_ntimes(10, self.m_Editor.GetPropertyValue, "Geometry3DPlaneTab", self._name, "Color")
```

pyaedt/modeler/Primitives.py
```diff
@@ -1237,6 +1238,83 @@ def create_point(self, position, name=None, color="(143 175 143)"):
...
+        plane_base_z : str
+            Z coordinate of the plane base. The default value is ``"0mm"``.
+        plane_normal_x : str
+            X coordinate of the normal plane. The default value is ``"0mm"``.
+        plane_normal_y : str
+            Y coordinate of the normal plane. The default value is ``"0mm"``.
+        plane_normal_z : str
+            Z coordinate of the normal plane. The default value is ``"0mm"``.
+        color : str, optional
+            String exposing 3 int values such as "(value1 value2 value3)". Default value is ``"(143 175 143)"``.
```
PipKat, line 1273:
```suggestion
            String exposing the three integer values for the color of the plane. The
            default value is ``"(143 175 143)"``.
```
I don't think the "such as "(value1 value2 value3)" is needed as the default provides a clearer example on its own.

maxcapodi78 (approved):
LGTM

PR #2207, opened 2022-12-26
Remove any usage of unicode and basestring.

### LINKED ISSUE #2206: Reference to basestring can be removed
PyAEDT supports Python version >=3.7.
As a consequence, the comparison to `basestring` can be removed and only the comparison to `str` will prevail.

https://github.com/pyansys/pyaedt/blob/52e6938cf15ad773a148785d07b402ce3a02ab46/pyaedt/modeler/Modeler.py#L4531

MaxJPRey:
Since Python 3, all strings are unicode.
The `unicode()` function does not exist anymore.
`basestring` could be seen as a gateway to support both `unicode` and plain `str`.

Because PyAEDT is supporting Python>=3.7, there is no need to support anymore this structure for the `str`.

Fix #2206 .

### DIFF
--- a/pyaedt/application/AEDT_File_Management.py
+++ b/pyaedt/application/AEDT_File_Management.py
@@ -121,8 +121,8 @@ def change_objects_visibility(origfile, solid_list):
             )
             # Replacing string
             # fmt: off
-            ViewStr = u"Drawings[" + unicode(str(len(solid_list))) + u": " + unicode(str(solid_list).strip("["))
-            s = pattern.sub(r"\1" + ViewStr + r"\3", content)
+            view_str = u"Drawings[" + str(len(solid_list)) + u": " + str(solid_list).strip("[")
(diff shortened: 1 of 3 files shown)


pyaedt/application/AEDT_File_Management.py
```diff
@@ -121,8 +121,8 @@ def change_objects_visibility(origfile, solid_list):
             )
             # Replacing string
             # fmt: off
-            ViewStr = u"Drawings[" + unicode(str(len(solid_list))) + u": " + unicode(str(solid_list).strip("["))
-            s = pattern.sub(r"\1" + ViewStr + r"\3", content)
+            view_str = u"Drawings[" + str(len(solid_list)) + u": " + str(solid_list).strip("[")
```
maxcapodi78, line 124: is it also working with Ironpython?

maxcapodi78 (approved):

PR #2223, opened 2022-12-30
Changed way non-Graphical is evaluated when Desktop is initialized

maxcapodi78:
Now settings.non_graphical is evaluated directly from AEDT API


MaxJPRey (approved):
LGTM.

PR #2228, opened 2023-01-02
refacto / cleaning examples

gmalinve:


pyaedt/modules/PostProcessor.py
```diff
@@ -1132,26 +1133,26 @@ def create_rectangular_plot(
...
             ["X Component:=", primary_sweep_variable, "Y Component:=", expression],
         )
-        self.logger.info("Report %s correctly created.", plotname)
+        self.logger.info("Report %s correctly created.", plot_name)
         return True
 
     @pyaedt_function_handler()
     def get_solution_data_per_variation(
-        self, soltype="Far Fields", setup_sweep_name="", ctxt=None, sweeps=None, expression=""
+        self, solution_type="Far Fields", setup_sweep_name="", ctxt=None, sweeps=None, expression=""
```
MaxJPRey, line 1149: If we modify the name of the parameters to follow the snake_case convention, we might have to use a `deprecation warning` to allow the former scripts to keep working.

pyaedt/modules/PostProcessor.py
```diff
@@ -877,12 +877,12 @@ def copy_report_data(self, PlotName):
         return True
 
     @pyaedt_function_handler()
-    def delete_report(self, plot_name=None):
+    def delete_report(self, PlotName=None):
```
MaxJPRey, line 880: I would suggest to keep the parameter name to `plot_name` to follow snake_case naming convention.
```suggestion
    def delete_report(self, plot_name=None):
```

MaxJPRey (approved):

PR #2230, opened 2023-01-03
Add blacken-docs to pre-commit.

MaxJPRey:
In addition update hooks version.


RobPasMue (approved):
LGTM!

Would you consider using pre-commit.ci as well? We just implemented it in PyMAPDL. I can help you set it up, no problem regarding that.

This would allow to remove the code-style stage, automatically update the pre-commit hooks and a new fancy badge for the main README :)

PR #2233, opened 2023-01-04
Separate doc upload from generation into 2 stages

RobPasMue:
My feeling is that your workflow was failing because of this. It seems to me that the artifact is not accessible since you are still on the same stage/job. Separation of docs generation and upload into two jobs/stages will potentially solve the issue


MaxJPRey:
@RobPasMue This is excellent. Thanks for noticing it and fixing it.

MaxJPRey (approved):
LGTM.

PR #2238, opened 2023-01-05
pin group bug fix

ring630:
If pin group is created from SIwave GUI, it has not net object assigned to it. It leads to failure in port creation by pyaedt

### DIFF
--- a/pyaedt/edb_core/edb_data/sources.py
+++ b/pyaedt/edb_core/edb_data/sources.py
@@ -281,10 +281,13 @@ class PinGroup(object):
 
     def _create_terminal(self, is_reference=False):
         pg_term = self._edb_pin_group.GetPinGroupTerminal()
+        edb_net = self._edb_pin_group.GetNet()
+        if edb_net.IsNull():
+            edb_net = list(self._edb_pin_group.GetPins())[0].GetNet()
         if pg_term.IsNull():
             return self._pedb.edb.Cell.Terminal.PinGroupTerminal.Create(
                 self._active_layout,
-                self._edb_pin_group.GetNet(),
+                edb_net,
                 self.name,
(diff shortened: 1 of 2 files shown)


pyaedt/edb_core/edb_data/sources.py
```diff
@@ -281,10 +281,13 @@ def net_name(self):
 
     def _create_terminal(self, is_reference=False):
         pg_term = self._edb_pin_group.GetPinGroupTerminal()
+        edb_net = self._edb_pin_group.GetNet()
+        if edb_net.IsNull():  # pragma: no cover
+            edb_net = list(self._edb_pin_group.GetPins())[0].GetNet()
```
MaxJPRey, line 286: As we are in the `edb` module, it is not needed to prefix the name of the variable with `edb_`
So maybe it could be named:
```suggestion
        pin_group_net = self._edb_pin_group.GetNet()
        if pin_group_net.IsNull():  # pragma: no cover
            pin_group_net = list(self._edb_pin_group.GetPins())[0].GetNet()
```

pyaedt/edb_core/edb_data/sources.py
```diff
@@ -281,10 +281,13 @@ def net_name(self):
...
         if pg_term.IsNull():
             return self._pedb.edb.Cell.Terminal.PinGroupTerminal.Create(
                 self._active_layout,
-                self._edb_pin_group.GetNet(),
+                edb_net,
```
MaxJPRey, line 290:
```suggestion
                pin_group_net,
```

MaxJPRey (approved):
LGTM. Thanks for fixing the issue and the discussion @ring630 .

maxcapodi78 (approved):

PR #2258, opened 2023-01-10
Report layer copper area

ring630:


pyaedt/edb_core/edb_data/hfss_simulation_setup_data.py
```diff
@@ -1164,12 +1164,48 @@ def use_max_refinement(self, value):
...
+    def add_adaptive_frequency_data(self, frequency=0, max_num_passes=10, max_delta_s=0.02):
         """Add a setup for frequency data.
 
         Parameters
         ----------
-        frequency : str, float
+        low_frequency : str, float
+            Frequency with units or float frequency (in Hz).
+        high_frequency : str, float
+            Frequency with units or float frequency (in Hz).
```
MaxJPRey, line 1175:
```suggestion
        frequency : str, float
            Frequency with units or float frequency (in Hz).
```

pyaedt/edb_core/stackup.py
```diff
@@ -1210,6 +1210,36 @@ def place_a3dcomp_3d_placement(self, a3dcomp_path, angle=0.0, offset_x=0.0, offs
 
         return True
 
+    @pyaedt_function_handler
+    def report_copper_rate(self):
```
maxcapodi78, line 1214: Maybe a better and self explaining name would be better.
get_copper_rate_per_layer? or something similar

maxcapodi78 (commented):
Hi Hui, just a comment on the name of the method that is not clear to me

pyaedt/edb_core/stackup.py
ring630: Change to residual_copper_area_per_layer().

maxcapodi78 (approved):

PR #2279, opened 2023-01-12
Improved Documentation

maxcapodi78:
Various minor improvements on docs. 
Improved Desktop init in 23R1. Default wil be grpc but user will still be able to run AEDT using COM Engine (setting up settings.use_grpc_api = False)


Samuelopez-ansys (approved):
LGTM

MaxJPRey (approved):

PR #2309, opened 2023-01-17
Via platting fix

svandenb-dev:


pyaedt/edb_core/padstack.py
```diff
@@ -529,6 +529,32 @@ def set_all_antipad_value(self, value):
...
+    def check_and_fix_via_plating(self, minimum_value_to_replace=0.0, default_plating_ratio=0.2):
+        """Check and fix zero value plating ratio via definition with assigning default value.
+
+        Parameters
+        ----------
+        minimum_value_to_replace : float
+            Plating ratio below or equal to this value will be replaced by the default one.
+
+        default_plating_ratio : float
+            Default value used for plating ratio. Fefault value is 0.2.
```
MaxJPRey, line 542:
```suggestion
            Default value used for plating ratio. Default value is 0.2.
```

MaxJPRey (approved):
LGTM.

pyaedt/edb_core/padstack.py
```diff
@@ -529,6 +529,32 @@ def set_all_antipad_value(self, value):
                 padstack.edb_padstack.SetData(cloned_padstack_data)
             return all_succeed
 
+    @pyaedt_function_handler
+    def check_and_fix_via_plating(self, minimum_value_to_replace=0.0, default_plating_ratio=0.2):
+        """Check and fix zero value plating ratio via definition with assigning default value.
```
PipKat, line 534: I'm not sure I understand this description. Can it be shortened. Maybe something like this--and if it's wrong, it shows how I truly don't comprehend what it says:

'''Check and fix the zero value plating ratio.

This method allows you to assign a default value using a definition.

pyaedt/edb_core/padstack.py
```diff
@@ -529,6 +529,32 @@ def set_all_antipad_value(self, value):
...
+
+        Parameters
+        ----------
+        minimum_value_to_replace : float
+            Plating ratio below or equal to this value will be replaced by the default one.
```
PipKat, line 539:
```suggestion
            Plating ratio that is below or equal to this value is to be replaced
            with the value specified for the next parameter.
```

pyaedt/edb_core/padstack.py
```diff
@@ -529,6 +529,32 @@ def set_all_antipad_value(self, value):
...
+
+        default_plating_ratio : float
+            Default value used for plating ratio. Default value is ``0.2``.
```
PipKat, line 541:
```suggestion
            Default value to use for plating ratio. The default value is ``0.2``.
```

pyaedt/edb_core/padstack.py
```diff
@@ -529,6 +529,32 @@ def set_all_antipad_value(self, value):
...
+            if padstack_def.hole_plating_ratio <= minimum_value_to_replace:
+                padstack_def.hole_plating_ratio = default_plating_ratio
+                self._logger.info(
+                    "Padstack definition with zero plating ratio, defaulting to 20%".format(padstack_def.name)
+                )
+        return True
+
     @pyaedt_function_handler()
     def get_via_instance_from_net(self, net_list=None):
         """Get the list for Edb vias from net name list.
```
PipKat, line 558:
```suggestion
        """Get the list for EDB vias from a net name list.
```

pyaedt/edb_core/padstack.py
svandenb-dev: Yes your are right it's not clear. I will change it

PR #2312, opened 2023-01-17
Feature/s755061 move domain

apollo-crum:
Remove the domain wrapper and replace with direct use of emit python's domain class

### DIFF
--- a/_unittest/test_26_emit.py
+++ b/_unittest/test_26_emit.py
@@ -3,7 +3,7 @@ import os
 import sys
 
 # get aedt delcross python from custom path
-delcross_python_path = os.environ.get('ANSYS_DELCROSS_PYTHON_PATH')
+delcross_python_path = os.environ.get("ANSYS_DELCROSS_PYTHON_PATH")
 if delcross_python_path:
     sys.path.append(delcross_python_path)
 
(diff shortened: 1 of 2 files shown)


apollo-crum:
merge with (or after) https://tfs.ansys.com:8443/tfs/ANSYS_EBU_Collection/EBU/_git/delcross.git/pullrequest/98549

pyaedt/emit.py
```diff
@@ -487,7 +487,7 @@ def interaction_domain():
         Returns
         -------
         :class:`Emit.InteractionDomain`
-            Domain used to define a set of interacting interferers and receivers
+            Domain used to define a set of interacting interferers and receivers.
```
PipKat, line 490:
```suggestion
            Provides the domain used to define a set of interacting interferers and receivers.
```
FYI: The style is that class descriptions start with a verb ending in an "s".

pyaedt/emit.py
```diff
@@ -396,20 +206,20 @@ def run(self, domain):
...
+        interaction = eng.run(domain)
         return interaction
 
     @pyaedt_function_handler()
-    def get_max_simultaneous_frequencies(self):
+    def get_max_simultaneous_interferers(self):
 
         """
-        Get the number of maximum simultaneous frequencies
+        Get the number of maximum simultaneous interferers
```
PipKat, line 222:
```suggestion
        Get the number of maximum simultaneous interferers.
```

_unittest/test_26_emit.py
```diff
@@ -1,11 +1,16 @@
 # Import required modules
 import os
+import sys
+
+# get aedt delcross python from custom path
+delcross_python_path = os.environ.get("ANSYS_DELCROSS_PYTHON_PATH")
```
myoung301, line 6: @apollo-crum Are we adding this to the sys.path so that the `from pyaedt import Emit` line works? If so, can we move this into that module?

apollo-crum: Sounds good, I'll do that.  I also wasn't sure if there was a better way to do this in general.  I looked for something in common code that can redirect you to a local aedt install but couldn't find anything.

myoung301 (approved):

PR #2320, opened 2023-01-19
D758967: add domain validity and change a class name

apollo-crum:


pyaedt/emit.py
```diff
@@ -247,6 +247,20 @@ def set_max_simultaneous_interferers(self, val):
         eng = self.emit_obj._emit_api.get_engine()
         eng.max_simultaneous_interferers = val
 
+    @pyaedt_function_handler()
+    def is_domain_valid(self, ret_val, domain):
+
```
MaxJPRey, line 252:
```suggestion
```

pyaedt/emit.py
```diff
@@ -247,6 +247,20 @@ def set_max_simultaneous_interferers(self, val):
...
+        """
+        Return true if the given domain is valid for the current Revision
```
MaxJPRey, line 254:
```suggestion
        Return ``True`` if the given domain is valid for the current revision.
```

pyaedt/emit.py
```diff
@@ -247,6 +247,20 @@ def set_max_simultaneous_interferers(self, val):
...
+
+        Examples
+        ----------
+        >>> domain = aedtapp.interaction_domain()
+        >>> if aedtapp.results.is_domain_valid(domain):
```
MaxJPRey, line 259: This way we can also see the returned value.
```suggestion
        >>> aedtapp.results.is_domain_valid(domain)
        True
```

pyaedt/emit.py
```diff
@@ -247,6 +247,20 @@ def set_max_simultaneous_interferers(self, val):
...
+        """
+        eng = self.emit_obj._emit_api.get_engine()
+        return eng.is_domain_valid(domain)
```
MaxJPRey, line 262:
```suggestion
        return engine.is_domain_valid(domain)
```

myoung301 (approved):

PR #2324, opened 2023-01-19
create port per pin not working with simulationconfiguration

svandenb-dev:


pyaedt/edb_core/edb_data/simulation_configuration.py
```diff
@@ -174,6 +175,21 @@ def use_default_cutout(self):  # pragma: no cover
     def use_default_cutout(self, value):  # pragma: no cover
         self._use_default_cutout = value
 
+    @property
+    def do_pingroup(self):  # pragma: no cover
+        """Do pingroup on multi-pin component. ``True`` all pins from the same net are grouped, ``False`` one port
+        is created for each pin.
```
MaxJPRey, line 181:
```suggestion
        """Do pingroup on multi-pin component.
        If ``True`` all pins from the same net are grouped. Otherwise one port
        is created for each pin.
```

MaxJPRey (approved):
LGTM.

PR #2332, opened 2023-01-23
Added Desktop Back Compatibility to v221-

maxcapodi78:
Fixed minor issue that was blocking Desktop to start on 22.1 and below.


Samuelopez-ansys (approved):
LGTM

PR #2350, opened 2023-01-25
Fixes bug in 'count' entry of logscale in sweep

skandak-ansys:
Previously 'count' entry of log scale was always defaulting to 10 irrespective of user entry.


svandenb-dev (approved):

MaxJPRey (approved):
LGTM.

maxcapodi78 (approved):

PR #2359, opened 2023-01-25
Add emit support for units

jsalant22:
added support in emit for units, updated doc strings, added some test coverage

### DIFF
--- a/_unittest/conftest.py
+++ b/_unittest/conftest.py
@@ -52,7 +52,7 @@ sys.path.append(local_path)
 from _unittest.launch_desktop_tests import run_desktop_tests
 
 # Initialize default desktop configuration
-default_version = "2022.2"
+default_version = "2023.2"
 os.environ["ANSYSEM_FEATURE_SS544753_ICEPAK_VIRTUALMESHREGION_PARADIGM_ENABLE"] = "1"
 if inside_desktop and "oDesktop" in dir(sys.modules["__main__"]):
     default_version = sys.modules["__main__"].oDesktop.GetVersion()[0:6]
--- a/pyaedt/emit.py
+++ b/pyaedt/emit.py
@@ -7,6 +7,8 @@ from importlib import import_module
 from pyaedt import generate_unique_project_name
 from pyaedt.application.AnalysisEmit import FieldAnalysisEmit
 from pyaedt.generic.general_methods import pyaedt_function_handler
+from pyaedt.emit_core import EmitConstants
+import pyaedt.generic.constants as consts
 
 # global variable used to store module import
 mod = None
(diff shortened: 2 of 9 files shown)


_unittest/conftest.py
```diff
@@ -52,7 +52,7 @@
 from _unittest.launch_desktop_tests import run_desktop_tests
 
 # Initialize default desktop configuration
-default_version = "2022.2"
+default_version = "2023.2"
```
maxcapodi78, line 55: you cannot change Unit Test on Build machine since at the moment only 22R2 is installed. You can do it locally on your machine using local_config.json file containing modified default settings 
image

maxcapodi78 (dismissed):

_unittest/conftest.py
jsalant22: Sorry, didn't mean to commit that change.

apollo-crum (approved):

pyaedt/emit.py
```diff
@@ -545,3 +585,80 @@ def version(self, detailed=False):
...
+        Length : pm, nm, um, mm, cm, dm, meter, km, mil, in, ft, yd, mile
+        Time : ps, ns, us, ms, s
+        Voltage : mV, V
+        Data Rate : bps, kbps, Mbps, Gbps
+        Resistance : uOhm, mOhm, Ohm, kOhm, megOhm, GOhm
+
+        Returns
+        -------
+        Bool
+            ''True'' if the units were successfully changed and ''False''
```
MaxJPRey, line 611:
```suggestion
            ``True`` if the units were successfully changed and ``False``
```

pyaedt/emit.py
```diff
@@ -32,10 +34,22 @@ class Results:
 
     def __init__(self, emit_obj):
         self._result_loaded = False
+        """''True'' if the results are loaded and ''False'' if they are not."""
```
MaxJPRey, line 37:
```suggestion
        """``True`` if the results are loaded and ``False`` if they are not."""
```

pyaedt/emit.py
```diff
@@ -545,3 +585,80 @@ def version(self, detailed=False):
...
+        valid_system = EmitConstants.EMIT_UNIT_SYSTEM
+        valid_units = EmitConstants.EMIT_VALID_UNITS
+
+        if isinstance(unit_system, list):
+            for t, v in zip(unit_system, unit_value):
+                if t not in valid_system:
+                    print("[{}] units are not supported by EMIT. The options are: {}: ".format(t, valid_system))
+                    return False
+                if v not in valid_units[t]:
+                    print("[{}] are not supported by EMIT. The options are: {}: ".format(v, valid_units[t]))
```
MaxJPRey, line 623: Could we use the logger here instead of printing into the console?

pyaedt/emit.py
```diff
@@ -545,3 +585,80 @@ def version(self, detailed=False):
...
+            if unit_value not in valid_units[unit_system]:
+                print(
+                    "[{}] are not supported by EMIT. The options are: {}: ".format(unit_value, valid_units[unit_system])
+                )
+                return False
+            self.units[unit_system] = unit_value
+        return True
+
+    @pyaedt_function_handler()
+    def get_units(self, unit_system=""):
```
MaxJPRey, line 639:
```suggestion
    def get_units(self, unit_type=""):
```

PR #2407, opened 2023-02-01
refactor emit code

jsalant22:
move Results and Revision to separate modules

Haven't really deleted/moved any functions yet, but is this what we had in mind in terms of creating separate modules for Results/Revision?


pyaedt/emit_core/__init__.py
```diff
@@ -0,0 +1 @@
+from pyaedt.emit_core.results import Results
```
myoung301, line 1: I think it will be best not to bring these classes up into the package. For now at least, let's try keeping this __init__ file empty. I think that will be the least confusing. We might want to pull things into the pyaedt.emit package so the user can do pyaedt.emit.xyz, but I don't think I see a need to do it here in emit_core, or in results/__init__.py either. Does that make sense?

examples/07-EMIT/ComputeProtectionLevels.py
```diff
@@ -18,299 +18,299 @@
 import pyaedt
```
myoung301, line 18: In the future, can we put a version check at the top and exit if the version requirement isn't met, instead of commenting the whole thing? It will get hard to review and work with the diffs if we have to comment/uncomment. We could use the block comments (""") if needed, but maybe a version check would work?

pyaedt/emit.py
```diff
@@ -9,289 +9,14 @@
 from pyaedt import generate_unique_project_name
 from pyaedt.application.AnalysisEmit import FieldAnalysisEmit
 from pyaedt.emit_core import EmitConstants
+from pyaedt.emit_core.results.results import Results
+from pyaedt.emit_core.results.revision import Revision
```
myoung301, line 13: Is the Revision class needed here, or only from within Results?

examples/07-EMIT/ComputeProtectionLevels.py
jsalant22: Yea, that probably makes sense. I'm going to make that a separate DE, I just uncommented things here to do some basic testing and ensure things still work.

pyaedt/emit.py
jsalant22: As of now its needed by the main Emit class, but I haven't rearranged any of the class functions yet.

jsalant22:
@myoung301 @apollo-crum I don't have everything working yet, but here's my crack at refactoring. Thoughts? 

image

myoung301:
> @myoung301 @apollo-crum I don't have everything working yet, but here's my crack at refactoring. Thoughts?
> 
> image

- What is the color-code? (dark-blue vs light blue)
- How about current_design_revision() instead of current_design()?
- I think revisions is probably good (instead of revisions_list).
- If there are get_units and set_units methods in Emit class, what is the Emit.units for? (should it be _units if the get/set are the advertised way to interact?)
-

apollo-crum:
I think just "current_revision" instead of "current_design".  But we need to be careful with that, the latest result node in aedt might not have the latest changes to the design in it.  Maybe the "current design" should be the return value from "analyze" and that's it, any other ways to reference it are problematic.

We may want to move analyze, result_type, and interaction_domain from Emit to Results, since they really are results-only things.

revisions_list is going to turn into a get function that takes the revision name as an argument, but there's a separate ticket for that.

What is emit_obj?  If its just referring to the parent Emit object it should be user exposed, but maybe needs a different name.  Seems like things like project_path should be available from the Emit object?

I don't think emit_api needs to be in Revision as a user available thing, we should hide it (underscore or whatever).  This is what is currently the global called "mod" yes?  I can see the benefits of putting it Revision, but it still needs to be accessible at the higher levels (by us not users) for things like "interaction_domain()".  So I think it has to be an object in Results instead of Revision (even if the main thing it does is access the data for a single revision).

jsalant22:
@apollo-crum All the get_radio_names(), get_band_names(), etc need emit_api access so it seemed cleaner to keep a reference to it within Revision than to access it via the Results class. I'll make it private though and Results will need it to if I move interactionDomain and some other stuff there.

@myoung301 Colors were just from the Miro template I used. I think all the other suggestions make sense.

apollo-crum:
> @apollo-crum All the get_radio_names(), get_band_names(), etc need emit_api access so it seemed cleaner to keep a reference to it within Revision than to access it via the Results class. I'll make it private though and Results will need it to if I move interactionDomain and some other stuff there.

It will need to be referenced by both, but I see Results as the "owner" essentially

pyaedt/emit_core/results/results.py
```diff
@@ -0,0 +1,159 @@
...
+
+        Parameters
+        ----------
+        tx_rx : tx_rx_mode object
+            Used for determining whether to get ``rx`` or ``tx`` radio names.
+
+        Returns
+        -------
+        radios:class:`list of str`
+            list of tx or or rx radio names
```
MaxJPRey, line 83:
```suggestion
            List of tx or rx radio names.
```

pyaedt/emit_core/results/results.py
```diff
@@ -0,0 +1,159 @@
...
+
+        Returns
+        -------
+        """
+        print("This function is inaccessible when the Emit object has no revisions.")
+
+    @pyaedt_function_handler()
+    def get_radio_names(self, tx_rx):
+        """
+        Get a list of all ``tx'' or ``rx`` radios in the project.
```
MaxJPRey, line 73:
```suggestion
        Get a list of all ``tx`` or ``rx`` radios in the project.
```

pyaedt/emit_core/results/results.py
```diff
@@ -0,0 +1,159 @@
...
+        ----------
+        radio_name : str
+            Name of the radio.
+        tx_rx : tx_rx_mode object
+            Used for determining whether to get ``rx`` or ``tx`` radio names.
+
+        Returns
+        -------
+        bands:class:`list of str`
+            list of tx or or rx radio band names
```
MaxJPRey, line 111:
```suggestion
            List of ``tx`` or ``rx`` radio names.
```

pyaedt/emit_core/results/results.py
```diff
@@ -0,0 +1,159 @@
...
+           Name of the band.
+        tx_rx : tx_rx_mode object
+            Used for determining whether to get ``rx`` or ``tx`` radio names.
+        units : str
+            Units for the frequencies.
+
+        Returns
+        -------
+        freq:class:`list of float`
+            List of tx or or rx radio frequencies.
```
MaxJPRey, line 144:
```suggestion
            List of ``tx`` or ``rx`` radio names.
```

pyaedt/emit_core/results/revision.py
```diff
@@ -0,0 +1,125 @@
...
+    @pyaedt_function_handler()
+    def get_max_simultaneous_interferers(self):
+
+        """
+        Get the number of maximum simultaneous interferers.
+
+        Returns
+        -------
+        max_interferers : int
+            Maximum number of simultaneous interferers associated with engine
```
MaxJPRey, line 87:
```suggestion
            Maximum number of simultaneous interferers associated with engine.
```

pyaedt/emit_core/results/results.py
```diff
@@ -0,0 +1,125 @@
...
+    def analyze(self, revision_name=None):
+        """
+        Analyze the specified design.
+
+        Parameters
+        ----------
+        revision_name : str
+            Revision to analyze. If None, the most recent Revision
+            will be loaded if it matches the current design revision.
+
```
PipKat, line 88:
```suggestion
            Revision to analyze. The default is ``None``,in which case the most recent revision
            is loaded if it matches the current design revision.

```

pyaedt/emit_core/results/revision.py
PipKat: A period should appear at the end of the description for what is returned--just as Maxime indicated in his suggestion.

pyaedt/emit_core/results/revision.py
```diff
@@ -0,0 +1,271 @@
...
+        """
+        if not self.revision_loaded:
+            self.emit_project._emit_api.load_project(self.path)
+        engine = self.emit_project._emit_api.get_engine()
+        engine.max_simultaneous_interferers = val
+
+    @pyaedt_function_handler()
+    def is_domain_valid(self, domain):
+        """
+        Return ``True`` if the given domain is valid for the current Revision.
```
PipKat, line 165:
```suggestion
        Return ``True`` if the given domain is valid for the current revision.
```

pyaedt/emit_core/results/revision.py
```diff
@@ -0,0 +1,271 @@
...
+
+    @pyaedt_function_handler()
+    def get_radio_names(self, tx_rx):
+        """
+        Get a list of all ``tx'' or ``rx`` radios in the project.
+
+        Parameters
+        ----------
+        tx_rx : tx_rx_mode object
+            Used for determining whether to get ``rx`` or ``tx`` radio names.
```
PipKat, line 191:
```suggestion
            Specifies whether to get ``rx`` or ``tx`` radio names.
```

pyaedt/emit_core/results/revision.py
```diff
@@ -0,0 +1,271 @@
...
+
+        Returns
+        -------
+        radios:class:`list of str`
+            list of tx or or rx radio names
```
PipKat, line 196:
```suggestion
            List of ``tx'' or 'rtx'' radio names.
```

pyaedt/emit_core/results/revision.py
```diff
@@ -0,0 +1,271 @@
...
+        ----------
+        radio_name : str
+            Name of the radio.
+        tx_rx : tx_rx_mode object
+            Used for determining whether to get ``rx`` or ``tx`` radio names.
+
+        Returns
+        -------
+        bands:class:`list of str`
+            list of tx or or rx radio band names
```
PipKat, line 224:
```suggestion
            List of ``tx`` or ``rx`` radio band names.
```

pyaedt/emit_core/results/revision.py
```diff
@@ -0,0 +1,271 @@
...
+    def get_band_names(self, radio_name, tx_rx_mode):
+        """
+        Get a list of all ``tx`` or ``rx`` bands in a given radio.
+
+        Parameters
+        ----------
+        radio_name : str
+            Name of the radio.
+        tx_rx : tx_rx_mode object
+            Used for determining whether to get ``rx`` or ``tx`` radio names.
```
PipKat, line 219:
```suggestion
            Specifies whether to get ``tx`` or ``rx`` radio names.
```

pyaedt/emit_core/results/revision.py
```diff
@@ -0,0 +1,271 @@
...
+        Get a list of active frequencies for a ``tx`` or ``rx`` band in a radio.
+
+        Parameters
+        ----------
+        radio_name : str
+            Name of the radio.
+        band_name : str
+           Name of the band.
+        tx_rx : tx_rx_mode object
+            Used for determining whether to get ``rx`` or ``tx`` radio names.
```
PipKat, line 249:
```suggestion
            Specifies whether to get ``tx`` or ``rx`` radio names.
```

pyaedt/emit_core/results/revision.py
```diff
@@ -0,0 +1,271 @@
...
+        units : str
+            Units for the frequencies.
+
+        Returns
+        -------
+        freq:class:`list of float`
+            List of tx or or rx radio frequencies.
```
PipKat, line 256:
```suggestion
            List of ``tx`` or ``rx`` radio frequencies.
```

jsalant22:
@myoung301 @apollo-crum I'm going to keep this as a draft until I touch base with Jian, but I think it's at a point you (whoever draws the short straw) can start reviewing.

myoung301 (approved):

jsalant22:
@jianwei425 I added your code to this branch. Once you review it, you can close your branch (figured it was easier than trying to merge them).

_unittest/test_26_emit.py
```diff
@@ -559,17 +565,24 @@ def test_analyze_manually(self):
...
             assert engine is not None
             assert engine.is_domain_valid(domain)
+            assert rev.is_domain_valid(domain)
             interaction = engine.run(domain)
             assert interaction is not None
+            interaction2 = rev.run(domain)
+            assert interaction2 is not None
+            assert interaction2.is_valid()
+            self.aedtapp.results.delete_revision(rev.name)
+            assert not interaction.is_valid()
```
jianwei425, line 581: In this case, is it true that interaction and interaction2 both should not be valid? Should we add assertion for interaction2 ?

jianwei425 (approved):

PR #2434, opened 2023-02-07
Collapse navigation always

akaszynski:
It is possible to turn off the expandable navigation entirely by setting the collapse_navigation config option to `True`:

```py
html_theme_options = {
  "collapse_navigation": True
}
```

This tends to help improve documentation build times.

We previously did this in `sidebar-nav-bs.html`, but that option wasn't scalable as it didn't apply to all API pages. This PR removes that file and uses the cleaner option given by [pydata-sphinx-theme - Navigation](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/navigation.html).


MaxJPRey (approved):
LGTM. Thanks @akaszynski

PR #2453, opened 2023-02-13
2452 : Fix issue with plot of top/bottom components

JDufrenneAnsys:
- Fix retrieve of TOP/BOTTOM Layers
- Fix issue when no net is defined (case of IDF import)


svandenb-dev (approved):

MaxJPRey (approved):
LGTM.

PR #2455, opened 2023-02-13
2454 : Plot brd outline

JDufrenneAnsys:
- Allows to plot a layer not related to a net
- If layer name is "Outline" plot it as a contour


MaxJPRey (approved):
LGTM.

PR #2471, opened 2023-02-14
added source_group to report_templates.py

maxcapodi78:
now context supports Source Group for Hfss Report Templates


Samuelopez-ansys (approved):
LGTM

PR #2482, opened 2023-02-16
fix itervalues() not existing in python 3

rmawatson:
itervalues() has been removed in python 3.

### DIFF
--- a/pyaedt/modules/MaterialLib.py
+++ b/pyaedt/modules/MaterialLib.py
@@ -10,6 +10,7 @@ import json
 import math
 import os
 import re
+import sys
 
 from pyaedt import is_ironpython
 from pyaedt import settings
@@ -66,7 +67,7 @@ class Materials(object):
         return len(self.material_keys)
 
     def __iter__(self):
-        return self.material_keys.itervalues()
+        return iter(self.material_keys.values()) if sys.version_info.major > 2 else self.material_keys.itervalues()
 
     def __getitem__(self, item):
         matobj = self.checkifmaterialexists(item)


maxcapodi78 (approved):

MaxJPRey (approved):
Thanks for the fix @rmawatson

pyaedt/modules/MaterialLib.py
```diff
@@ -66,7 +67,7 @@ def __len__(self):
         return len(self.material_keys)
 
     def __iter__(self):
-        return self.material_keys.itervalues()
+        return iter(self.material_keys.values()) if sys.version_info.major > 2 else self.material_keys.itervalues()
```
MaxJPRey, line 70:
```suggestion
        return iter(self.material_keys.values())
```

pyaedt/modules/MaterialLib.py
```diff
@@ -10,6 +10,7 @@
 import math
 import os
 import re
+import sys
```
MaxJPRey, line 13:
```suggestion
```

pyaedt/modules/MaterialLib.py
maxcapodi78: yes but we are supporting ironpython

rmawatson: I noticed Iron Python is 2.7 is meant to be supported, and values() returns a copy of the list in Python 2. 

If that matters there are some options [here](https://python-future.org/compatible_idioms.html#dictionaries). Which is a bit cleaner.

It would add a dependency on `future` module though, and from a search for `itervalues()` and `iteritems()` in the code base, there appears to be only one other place this happens, which does it in a similar way to my original commit.

MaxJPRey: Very good point, thanks for thinking about IronPython.

PR #2483, opened 2023-02-16
allow adding both faces and objects to the same BoundaryObject

rmawatson:
Logic previously would only assign `Faces` array to `BoundaryObject`, even if `Objects` array contained items. It is possible to assign both in the UI. This fixes `update_assignment()` to allow both `Objects` and `Faces` to be assigned to the same `BoundaryObject`


maxcapodi78 (approved):
LGTM

PR #2503, opened 2023-02-17
get_dcir_solution_data

ring630:


pyaedt/hfss3dlayout.py
```diff
@@ -1998,3 +1999,56 @@ def edit_source_from_file(
                 return True
         self.logger.error("Port not found.")
         return False
+
+    def get_dcir_solution_data(self, setup_name, element_type, element_name):
+        """ReTrieve dcir solution data. Available element_names are dependent on element_type as below.
```
MaxJPRey, line 2004:
```suggestion
        """Retrieve dcir solution data. Available element_names are dependent on element_type as below.
```

pyaedt/hfss3dlayout.py
```diff
@@ -1998,3 +1999,56 @@ def edit_source_from_file(
...
+
+        Parameters
+        ----------
+        setup_name : str
+            Name of the setup.
+        element_type : str
+            Type of the element. Options are ``"Sources"`, ``"RL"`, ``"Vias"``, ``"Bondwires"``, and ``"Probes"``.
+        element_name : str
+            Name of the element. Options are ``"Voltage"`, ``"Current"`, ``"Power"``, ``"Loop Resistance"``,
+            ``"Path Resistance"``, ``"Resistance"``, ``"Inductance"``,v ``"X"``, ``"Y"``, ``"Limit"`` and ``"IR Drop"``.
```
MaxJPRey, line 2019: @ring630 @maxcapodi78 I think that instead of accepting `string`, we could have `enums`.
Indeed it would make is easier for the user to get a list of the accepted values through the auto-completion.
This is possible because it seems that the accepted values for those arguments is limited to a defined number.

It could be a good practice not only here but for PyAEDT post at large.

ring630: @MaxJPRey , replaced with enum now.

MaxJPRey: Great, many thanks @ring630 .

svandenb-dev (approved):

PR #2505, opened 2023-02-20
EDB get pingroup list

svandenb-dev:


pyaedt/edb_core/edb_data/sources.py
```diff
@@ -316,6 +316,16 @@ def create_port_terminal(self, impedance=50):
         terminal.SetIsCircuitPort(True)
         return terminal
 
+    def delete(self):
+        """Delete active pingroup.
```
MaxJPRey, line 320: Might be a wrong question, but would it delete the source terminal -that could be linked to the pin group- too, or just the pin group?

ring630: @MaxJPRey @svandenb-dev I have the same concern. All linked objects have to be deleted together. source terminal, pingroup ternminal, votalge/current source, etc.

maxcapodi78: good point. let's investigate it

svandenb-dev: @MaxJPRey @ring630 @maxcapodi78  That's correct. Normally terminal are defined as PingroupTerminal which belong to Pingroup object. I was assuming with deleting Pingroup it would delete also the terminal but I agree this has to be checked.

PipKat, line 320:
```suggestion
        """Delete active pin group.
```
Given the use of the underscore (pin_group), I assume it should be pin group (two words).

maxcapodi78: @svandenb-dev what we should do here. are you checking the terminals to delete?

svandenb-dev: @maxcapodi78  No I am not checking. I suggest ti postpone this PR the time I do it.

maxcapodi78 (approved):
Approved with addition of Terminal Delete

PR #2517, opened 2023-02-23
Rlc component handling

svandenb-dev:


pyaedt/edb_core/edb_data/components_data.py
```diff
@@ -463,6 +478,14 @@ def res_value(self):
                 return pair.R.ToString()
         return None
 
+    @res_value.setter
+    def res_value(self, value):
+        if value:
+            if self.rlc_values == [None, None, None]:
+                self.rlc_values = [value, 0, 0]
+            else:
+                self.rlc_values = [value, self.rlc_values[1], self.rlc_values[2]]
```
maxcapodi78, line 487: Hi @svandenb-dev here it says that the unit tests are not run for the new property setter rlc_values. probably the unit test is not working properly or the list you are trying to change values is emptyh

svandenb-dev: I think I must add it in unit test.

maxcapodi78 (approved):

PR #2521, opened 2023-02-23
Consistent input/output from unit_converter, some EMIT fixes

myoung301:
### units:
- The unit_converter function was returning a non-list when a list was passed with one entry. The change makes the output type match the input type (list in = list out, single element in = single element out).
- Improved the warning message to include the unrecognized units when unrecognized units are passed to unit_converter.

### EMIT:
- Emit inherited from FieldAnalysisEmit, but EMIT doesn't have any need for the FieldAnalysis functionality (no analysis setups in EMIT design). Emit will now inherit directly from Design. (S776155)
- Emit.modeler.components is now initialized automatically (refresh_all_ids is called)
- Units weren't initialized before the modeler so refresh_all_ids would fail during Emit initialization. Moved units initialization ahead of modeler initialization.


myoung301:
CI_Prerelease: https://github.com/pyansys/pyaedt/actions/runs/4256855860/jobs/7406255230

pyaedt/emit.py
```diff
@@ -154,6 +159,31 @@ def __init__(
 
             self.__emit_api_enabled = True
 
+        # Update modeler IDs now, after units are initialized.
+        # self.modeler.components.refresh_all_ids()
```
jsalant22, line 163:
```suggestion
```
Can we remove these?

jsalant22 (approved):

pyaedt/emit.py
myoung301: @jsalant22 - Yes! Good catch. Thank you.

myoung301:
There is one test failure:
image

This is due to an update to the Emit API that was merged 2/21 so the CI_Prerelease workflow hasn't caught up yet.

@apollo-crum, @jsalant22 FYI - in case you run that workflow.

PR #2558, opened 2023-03-01
Toolkit launchers: Use AEDT version-specific virtual environment; Update PyAnsys logo

kbhagat-ansys:
1. Use AEDT version-specific Virtual Environment for toolkits
    Currently, the toolkits have the AEDT version for the python virtual environments hardcoded. This change allows one to use separate virtual environments for the different AEDT versions.

2. Update PyAnsys logo

image


maxcapodi78 (approved):

pyaedt/misc/aedtlib_personalib_install.py
```diff
@@ -81,8 +81,11 @@ def install_toolkit(toolkit_dir, product):
     os.makedirs(lib_dir, exist_ok=True)
     os.makedirs(tool_dir, exist_ok=True)
     files_to_copy = ["Console", "Run_PyAEDT_Script", "Jupyter"]
-    jupyter_executable = sys.executable.replace("python.exe", "jupyter.exe")
-    ipython_executable = sys.executable.replace("python.exe", "ipython.exe")
+    # Remove hardcoded version number from python virtual environment path, so it runs version specific virtual
```
MaxJPRey, line 84:
```suggestion
    # Remove hard-coded version number from Python virtual environment path, so it runs version specific virtual
```

pyaedt/misc/Run_PyAEDT_Script.py_build
```diff
@@ -34,12 +34,13 @@ print("ret: " + repr(ret))
...
     if ret == ret.OK:
         # launch file
+        version = oDesktop.GetVersion()[2:6].replace(".", "")
         current_dir = os.path.dirname(os.path.abspath(os.path.realpath(__file__)))
         pyaedt_toolkit_dir = os.path.normpath(os.path.join(current_dir, "..", "..", r"##TOOLKIT_REL_LIB_DIR##"))
         config_file = open(
             os.path.join(pyaedt_toolkit_dir, "python_interpreter.bat"), "r"
         )  # will fail with a file not found error if it isn't here
-        python_exe = r"##PYTHON_EXE##"
+        python_exe = r"##PYTHON_EXE##" % version
```
MaxJPRey, line 43: I can't get this working locally @kbhagat-ansys . Is it working on your side?
Maybe you were expecting something like this:
```suggestion
        python_exe = r"##PYTHON_EXE##" + version
```

pyaedt/misc/aedtlib_personalib_install.py
PipKat: @MaxJPRey I find the ending of the comment in line 84 unclear. Can it be changed to something like this:
# Remove hard-coded version number from Python virtual environment path so it runs the version specified virtually

pyaedt/misc/Run_PyAEDT_Script.py_build
kbhagat-ansys: @MaxJPRey  that will not work. Refer to the Image shown in the PR description. We don't want to concatenate a string but replace a "%s" inside a string.

pyaedt/misc/aedtlib_personalib_install.py
kbhagat-ansys: @PipKat  replaced with 
```py
# Remove hard-coded version number from Python virtual environment path, and replace it with the corresponding AEDT
# version's Python virtual environment.
```

PR #2574, opened 2023-03-03
Improve dependencies permissiveness.

### LINKED ISSUE #2572: Fixing numpy version is causing incompatibilities in PyAnsys metapackage
Hi @maxcapodi78 @MaxJPRey

I understand that you may have your reasons for fixing the versions of the different packages you need in PyAEDT but this hard pinning is causing incompatibilities with the rest of the PyAnsys libraries. See https://github.com/pyansys/pyansys/pull/179

I recommend you "relax" your requirement conditions so that you provide support from a lower numpy version to the latest one. Something like ``numpy>=X.Y.Z``. Let me know if you have any doubts

MaxJPRey:
Fix #2572 .

The dependencies required by PyAEDT are too strict.


RobPasMue (approved):
LGTM but bear in mind that the minimum versions seem a bit too high. Aren't you compatible with older versions of those packages?

MaxJPRey:
> LGTM but bear in mind that the minimum versions seem a bit too high. Aren't you compatible with older versions of those packages?

Yes it is true especially for `matplotlib`. We needed some recent/new features for the post-processing.

PR #2577, opened 2023-03-04
Add multiplier to the log plot.

Samuelopez-ansys:
The user could choose between to which number multiple the field values (10, 20...)


maxcapodi78 (approved):
LGTM

PR #2594, opened 2023-03-07
Bug/export import existing hfss3dlayout setup

isaacansys:
Updates SetupTemplates.HFSS3DLayout and child templates to 2023 R1 and adds a test for export/import from an existing HFSS 3D Layout design and setup.  Reasons for draft status:

- The templates could/should be made version specific as it's not clear to me what happens if we try to update an older version setup with newer version properties
- An alternative "fix" could be to simply remove the check whether each key in the read JSON data is in the setup; this would allow any version to work at the expense of missing issues with hand-edited files


maxcapodi78:
Hi @isaacansys  I think the two approaches proposed by you are not excluding each other. I'm working to a potential fix for point 1 while we could make the import_from_json more tolerant during the import phase.
What do you think?

maxcapodi78:
@isaacansys please give a look to the modifications I've pushed and let me know if you agree on this approach

isaacansys (commented):
LGTM

PR #2595, opened 2023-03-08
edb design options

ring630:


pyaedt/edb_core/edb_data/design_options.py
```diff
@@ -0,0 +1,35 @@
+class EdbDesignOptions:
```
maxcapodi78, line 1: you need to add docstrings also in the class description

maxcapodi78 (approved):

PR #2604, opened 2023-03-09
2023 R1 COM Fallback

maxcapodi78:
Starting from 2023.1 Pyaedt uses GRPC as default connection method to AEDT. Anyway user can still use COM and now it's possible to fallback to active COM session in case no GRPC sessions are found and new_aedt_session is set to False.

### DIFF
--- a/pyaedt/generic/general_methods.py
+++ b/pyaedt/generic/general_methods.py
@@ -1017,6 +1017,67 @@ def grpc_active_sessions(version=None, student_version=False, non_graphical=Fals
     return sessions
 
 
+def active_sessions(version=None, student_version=False, non_graphical=False):
+    """Get information for the active COM AEDT sessions.
+
+    Parameters
+    ----------
+    version : str, optional
+        Version to check. The default is ``None``, in which case all versions are checked.
+        When specifying a version, you can use a three-digit format like ``"222"`` or a
+        five-digit format like ``"2022.2"``.
+    student_version : bool, optional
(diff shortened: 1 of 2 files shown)


Alberto-DM (approved):
LGTM

pyaedt/generic/general_methods.py
```diff
@@ -1017,6 +1017,67 @@ def grpc_active_sessions(version=None, student_version=False, non_graphical=Fals
...
+    """Get information for the active COM AEDT sessions.
+
+    Parameters
+    ----------
+    version : str, optional
+        Version to check. The default is ``None``, in which case all versions are checked.
+        When specifying a version, you can use a three-digit format like ``"222"`` or a
+        five-digit format like ``"2022.2"``.
+    student_version : bool, optional
+    non_graphical : bool, optional
```
PipKat, line 1030: We need descriptions here in the form "Whether to ...."

pyaedt/generic/general_methods.py
```diff
@@ -1017,6 +1017,67 @@ def grpc_active_sessions(version=None, student_version=False, non_graphical=Fals
...
+
+
+    Returns
+    -------
+    list
+        List of AEDT PIDs.
```
PipKat, line 1036: Am I the only one that doesn't know what a PID is?

MaxJPRey: @PipKat It stands for process ID.

PR #2616, opened 2023-03-10
support mult-terminal wave port (bundle port)

ring630:


pyaedt/edb_core/hfss.py
```diff
@@ -524,6 +525,65 @@ def create_differential_wave_port(
...
+        for p_id, loc in list(zip(primitives_id, points_on_edge)):
+            _, term = self.create_wave_port(
+                p_id,
+                loc,
+                port_name=_port_name,
+                horizontal_extent_factor=horizontal_extent_factor,
+                vertical_extent_factor=vertical_extent_factor,
+                pec_launch_width=pec_launch_width,
+            )
+            _port_name = None
```
MaxJPRey, line 580: Do we need to set `_port_name` back to `None` ?

ring630: yes, the terminals must not have the same name. I set it back to None to let create method to determine unique names

MaxJPRey: It seems in the code that _port_name is not used anymore after this line.

ring630: It is used in the for loop
_, term = self.create_wave_port(
                p_id,
                loc,
                port_name=**_port_name,**

MaxJPRey: I see it is used to reset the name of the port_name after the first wave port creation.
Make sense @ring630

maxcapodi78 (approved):
LGTM

PR #2619, opened 2023-03-12
Polyline refactoring

Alberto-DM:
Refactored the polylines code to be compatible with the Parasolid modeler.
With the parasolid modeler points are different from vertices if the polyline cross-section. Now points property always indicates the polyline definition points independently from the cross-section (i.e. even if the polyline is become a solid).
Points and segment_types properties are now dynamically updated.
Implemented the polyline object load when a project is loaded. 
All methods are now compatible with all segment types, before some of them were compatible only with polylines made of line segments.
Fixed many bugs. 
Fixed some bugs in the history. 
Fixed iterator in modeler.Position class.
Fixed examples.


maxcapodi78 (approved):
Great Refactoring

PR #2620, opened 2023-03-13
add expressions cache and save fields Maxwell

gmalinve:


pyaedt/modules/SolveSetup.py
```diff
@@ -2452,14 +2467,18 @@ def add_eddy_current_sweep(self, range_type="LinearStep", start=0.1, end=100, co
...
             Default value is ``False``.
+        save_all_fields : bool, optional
+            Save fields at all frequency points to save fields for the entire set of sweep ranges.
+            Default is ``True``.
+
 
         Returns
         -------
         bool
+            True if successful, False if it fails.
```
MaxJPRey, line 2481:
```suggestion
            ``True`` if successful, ``False`` if it fails.
```

Samuelopez-ansys (approved):

PR #2626, opened 2023-03-16
improved aedtlib_personalib_install.py to support run as standalone file or as a method of pyaedt

maxcapodi78:


pyaedt/misc/aedtlib_personalib_install.py
```diff
@@ -12,27 +12,15 @@
...
-sys_lib = True
-if len(sys.argv) == 3:
-    sys_lib = True if sys.argv[2] == "1" else False
+
 
 pid = 0
 
 
-def main():
+def add_pyaedt_to_aedt(aedt_version, is_student_version=True, use_sys_lib=False):
```
MaxJPRey, line 20: Maybe the default value for the `is_student_version` parameter should be `False` as most users are not using student version.
```suggestion
def add_pyaedt_to_aedt(aedt_version, is_student_version=False, use_sys_lib=False):
```

pyaedt/misc/console_setup
```diff
@@ -10,7 +10,15 @@ provided by the launcher
 """
 import sys
 import atexit
-from pyaedt import *
+try:
+   from pyaedt import *
+except ImportError:
+   import os
```
MaxJPRey, line 16: See my comment above.
```suggestion
```

Samuelopez-ansys (commented):
Why the file is called "aedtlib_....py"? Should it be pyaedt_personallib_install?

pyaedt/misc/console_setup
```diff
@@ -8,9 +8,17 @@ cpython_console.py should be run instead of this script.
...
 """
+import os
 import sys
 import atexit
-from pyaedt import *
+try:
+   from pyaedt import *
+except ImportError:
+   # debug only
+   console_path = os.path.dirname(__file__)
```
kbhagat-ansys, line 18: Suggestion to change variable name:
```suggestion
   console_setup_dir = os.path.dirname(__file__)
```
@maxcapodi78

pyaedt/misc/console_setup
```diff
@@ -8,9 +8,17 @@ cpython_console.py should be run instead of this script.
...
+   if "PersonalLib" in console_path:
+      sys.path.append(os.path.join(console_path, "..", "..", ".."))
```
kbhagat-ansys, line 20: @maxcapodi78  I think a comment here would be helpful on what we are trying to achieve - and what are we trying to apped to path if the file is in PersonalLib

kbhagat-ansys:
@maxcapodi78  Thoughts on renaming `console_setup` to `console_setup.py`  this way PyCharm and other IDEs can correct the format and do Pep8 parsing for us.

pyaedt/misc/console_setup
maxcapodi78: added

maxcapodi78:
> @maxcapodi78 Thoughts on renaming `console_setup` to `console_setup.py` this way PyCharm and other IDEs can correct the format and do Pep8 parsing for us.

I've done it. Thanks

kbhagat-ansys (approved):
LGTM.

kbhagat-ansys:
@maxcapodi78  I have a branch called `toolkit_copy` where I improve the paths of the Lib files required for the Toolkits and I correct the description of the `Jupter.py_build` file. 
I was planning on creating a new PR for it, or you can include it in your changes.

PR #2629, opened 2023-03-16
Fix variables_to_include argument in create_3dcomponent

### LINKED ISSUE #2628: Bug with variables_to_include argument in create_3dcomponent

### Description of the bug

If the argument is variables_to_include=[] all variables are exported. If the argument is None, the same happens. It is not coherent with the behavior of the other arguments.

### Steps To Reproduce

-

Windows

3.7

### Installed packages

-

lorenzovecchietti:
Fix `create_3dcomponent`.` variables_to_include` argument now works like the other arguments. `None `leads to all variables to be exported and `[]` leads to no variables being exported

Close #2628

### DIFF
--- a/pyaedt/modeler/modeler3d.py
+++ b/pyaedt/modeler/modeler3d.py
@@ -137,11 +137,11 @@ class Modeler3D(GeometryModeler, Primitives3D, object):
             is not set to ``False``.
         datasets : list, optional
             List of dataset names to export. The default is all datasets. This argument
-             is relevant only if ``auxiliary_dict_file`` is not set to ``False``.
+             is relevant only if ``auxiliary_dict_file`` is set to ``True``.
         native_components : list, optional
             List of native_components names to export. The default is all
             native_components. This argument is relevant only if ``auxiliary_dict_file``
-            is not set to ``False``.
+            is set to ``True``.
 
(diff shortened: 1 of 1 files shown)


pyaedt/modeler/modeler3d.py
```diff
@@ -127,18 +127,21 @@ def create_3dcomponent(
...
-            List of monitor objects' names to export. The default is the names of all monitor objects. This argument is
-            relevant only if ``auxiliary_dict_file`` is not set to ``False``.
+            List of monitor objects' names to export. The default is the names of all
+            monitor objects. This argument is relevant only if ``auxiliary_dict_file``
+            is not set to ``False``.
         datasets : list, optional
-            List of dataset names to export. The default is all datasets. This argument is relevant only if
-            ``auxiliary_dict_file`` is not set to ``False``.
+            List of dataset names to export. The default is all datasets. This argument
+             is relevant only if ``auxiliary_dict_file`` is not set to ``False``.
```
PipKat, line 140:
```suggestion
             is relevant only if ``auxiliary_dict_file`` is set to ``True``.
```
Wouldn't it be better to phrase this without a "not set to" construction? It would be easier to understand.

PR #2635, opened 2023-03-16
implement update trace name

gmalinve:


pyaedt/modules/report_templates.py
```diff
@@ -205,10 +205,44 @@ class Trace(object):
...
         self.LINESTYLE = LineStyle()
         self.TRACETYPE = TraceType()
         self.SYMBOLSTYLE = SymbolStyle()
 
+    @property
+    def trace_name_real(self):
+        report_name = self.trace_name.split(":")[0]
+        traces_in_report = self._oreport_setup.GetReportTraceNames(report_name)
+        for trace in traces_in_report:
+            if trace + ":" in self.trace_name:
```
MaxJPRey, line 218: Just a suggestion, but do we really need to check if `:` if part of the `trace_name`?
```suggestion
            if trace in self.trace_name:
```

_unittest/test_12_PostProcessing.py
```diff
@@ -498,7 +506,7 @@ def test_09f_update_traces_in_report(self):
     @pytest.mark.skipif(
         config["desktopVersion"] < "2022.2", reason="Not working in non-graphical mode in version earlier than 2022.2."
     )
-    def test_09d_create_monitor(self):  # pragma: no cover
+    def test_09h_create_monitor(self):  # pragma: no cover
```
MaxJPRey, line 509: I think we can remove `# pragma: no cover`. To be double checked.
```suggestion
    def test_09h_create_monitor(self):
```

MaxJPRey (approved):
LGTM.

pyaedt/modules/report_templates.py
gmalinve: yes because we can have trace names as dB(S(1:1,1:1)) and dB(S(1:1,1:1))_1 and trace_name is composed as:
plot_name:trace_name:variations_list

pyaedt/modules/report_templates.py
```diff
@@ -205,10 +205,44 @@ class Trace(object):
     def __init__(self, report_setup, trace_name):
         self._oreport_setup = report_setup
         self.trace_name = trace_name
+        self._trace_name_real = None
         self.LINESTYLE = LineStyle()
         self.TRACETYPE = TraceType()
         self.SYMBOLSTYLE = SymbolStyle()
 
+    @property
+    def trace_name_real(self):
```
maxcapodi78, line 214: please add docstring

pyaedt/modules/report_templates.py
maxcapodi78: I agree. Anyway I think we should choose a name that is more meaningful. It's not clear to me that real means...

maxcapodi78 (approved):
LGTM with a couple of comments

pyaedt/modules/report_templates.py
MaxJPRey: Thanks @gmalinve , it makes sense.

pyaedt/modules/report_templates.py
Samuelopez-ansys: @maxcapodi78 @gmalinve I think self.trace_name should be the name of the trace for instance "dB(S1:1)" and we should change the current self.trace_name, because it contains the following info: Report name + trace name + variation, for instance: "MyReport:dB(S1:1):$v=1,s=2", I think it is confusing. 

Then self.trace_name should be the name of the trace, and the other property could be called self.trace_aedt_name or something like this...

gmalinve: @maxcapodi78 @Samuelopez-ansys I'd change as:
- self.trace_name to be name of trace, i.e: "dB(S(1:1,1:1))".
- and something like self.trace_aedt_name what we have now for self.trace_name ("MyReport:dB(S1:1):$v=1,s=2")
Could it be a bit more clear?

PR #2677, opened 2023-03-22
Non stackup layers import export support

ring630:


pyaedt/edb_core/stackup.py
```diff
@@ -390,7 +390,7 @@ def _create_stackup_layer(self, layer_name, thickness, layer_type="signal"):
     def _create_nonstackup_layer(self, layer_name, layer_type):
         if layer_type == "conducting":  # pragma: no cover
```
ring630, line 391: This change is to match AEDT keywords.

PR #2696, opened 2023-03-23
Feat/issue 2695

dcrawforAtAnsys:
I updated syntax for Hfss3dLayout() so you can pass the aedb folder and don't always have to append "edb.def".  I couldn't find any unit tests and wasn't sure how to test this, so you may be able to point me to the unit tests.


pyaedt/edb.py
```diff
@@ -2437,6 +2437,10 @@ def hfss_setups(self):
         """
         return {i.name: i for i in self.setups if i.setup_type == "kHFSS"}
 
+        self.save_edb()
```
Samuelopez-ansys, line 2440: @dcrawforAtAnsys I think the following 3 lines should not be there. Please remove them.

maxcapodi78 (approved):

PR #2703, opened 2023-03-24
Graceful checks and Error dialogs if Toolkit files are missing.

kbhagat-ansys:
Graceful checks and Error dialogs if Toolkit files are missing.

Useful when a user Installs PyAEDT through one version of AEDT and then runs a different version of AEDT. In this case the virtual env. would not have been created for the second AEDT version.

The error message displayed will be `"<file_path>" does not exist. Please run the "Install PyAEDT" button in the Automation ribbon.`

PyAEDT Toolkit - Graceful checks

### DIFF
--- a/pyaedt/misc/Console.py_build
+++ b/pyaedt/misc/Console.py_build
@@ -11,7 +11,13 @@ interactive Python session.
 """
 import os
 import sys
+
+from System.Windows.Forms import MessageBox
+from System.Windows.Forms import MessageBoxButtons
+from System.Windows.Forms import MessageBoxIcon
+
 is_linux = os.name == "posix"
+
 if is_linux:
     import subprocessdotnet as subprocess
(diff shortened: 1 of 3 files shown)


pyaedt/misc/Console.py_build
```diff
@@ -88,5 +95,18 @@ def which(program):
     return None
 
 
+def check_file(file_path):
+    if not os.path.isfile(file_path):
+        show_error(
+            '"{}" does not exist. Please run the "Install PyAEDT" button in the Automation ribbon.'.format(file_path)
```
MaxJPRey, line 101:
```suggestion
            '"{}" does not exist. Please click on the "Install PyAEDT" button in the Automation ribbon.'.format(file_path)
```

maxcapodi78 (approved):

Issue #2667, opened 2023-03-21
Adding object priorities for Icepak components fails


StiftMC:
### Before submitting the issue

- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### Description of the bug
IcepakMesh.add_priority(entity_type, obj_list, comp_name=None, priority=3)

Function returns "Script macro error: Entity '2' is already in priority list '%2'"

![image](https://user-images.githubusercontent.com/100361281/226641058-51aebd56-3f84-45d1-8c9d-18bc2eb1d7d9.png)

If you have a PCB component in the design, there are already two entries in the priority list. 

![image](https://user-images.githubusercontent.com/100361281/226640053-263323f6-c2ba-4cc8-999c-c814235954ed.png)

### Steps To Reproduce

Import PCB with components. Call the function with list of components as obj_list and entity_type = 1 results in the mentioned error. Using the GUI with same inputs works. 

![image](https://user-images.githubusercontent.com/100361281/226640786-f187d1d6-c56c-4950-a214-03feb4223289.png)

### Which Operating System are you using?

Windows

### Which Python version are you using?

3.7

### Installed packages

AEDTLib==1.5.36
aiohttp==3.8.1
aiosignal==1.2.0
altgraph==0.17
anyio==3.0.1
appdirs==1.4.4
argon2-cffi==20.1.0
async-generator==1.10
async-timeout==4.0.2
asynctest==0.13.0
attrs==21.2.0
Babel==2.9.1
backcall==0.2.0
beautifulsoup4==4.10.0
bleach==3.3.0
cachetools==5.0.0
certifi==2020.12.5
cffi==1.15.1
chardet==4.0.0
charset-normalizer==2.1.1
click==8.1.3
clr-loader==0.2.4
colorama==0.4.4
cycler==0.11.0
dataframe-image==0.1.5
decorator==5.0.7
defusedxml==0.7.1
entrypoints==0.3
et-xmlfile==1.1.0
fonttools==4.28.2
frozenlist==1.3.1
future==0.18.2
fuzzywuzzy==0.18.0
idna==2.10
imageio==2.21.3
importlib-metadata==4.0.1
ipykernel==5.5.4
ipython==7.23.1
ipython-genutils==0.2.0
jedi==0.18.0
Jinja2==2.11.3
joblib==1.1.0
json5==0.9.5
jsonschema==3.2.0
jupyter-client==6.2.0
jupyter-core==4.7.1
[...]

PR #2731, opened 2023-03-29
Improve assign_priority method checking if passed objects are native

Samuelopez-ansys:
- Close issue #2667 
- Check if the object belongs to 3DComponent
- If object is not 3D, then it is not included


maxcapodi78 (approved):
LGTM

lorenzovecchietti (approved):

PR #2750, opened 2023-03-31
Speed up Icepak UT

Samuelopez-ansys:


MaxJPRey (approved):
Thanks for keeping the tests efficient. @Samuelopez-ansys .
This could also be helpful sometimes for other tests.

PR #2753, opened 2023-04-01
Update Edb.__setter__ to accept description.

dcrawforAtAnsys:
- Allow a 2-element list to be optionally passed as an argument to the ``Edb().__setter__``
- Updated unit test ``test_00_EDBtest_097b_variables``
- Added a method ``check_numeric_equivalence`` to check for almost identical numeric values considering relative tolerance.


dcrawforAtAnsys:
#2753

PR #2769, opened 2023-04-04
minor_fixes on cutout methods on point lists

maxcapodi78:


pyaedt/edb.py
```diff
@@ -1415,11 +1415,11 @@ def cutout(
             Full path and name for the new AEDB file. If None, then current aedb will be cutout.
         open_cutout_at_end : bool, optional
             Whether to open the cutout at the end. The default is ``True``.
-        use_legacy_cutout : bool, optional
+        use_pyaedt_cutout : bool, optional
             Whether to use new PyAEDT cutout method or EDB API method.
             New method is faster than native API method since it benefits of multithread.
         number_of_threads : int, optional
-            Number of thread to use. Default is 4. Valid only if `use_legacy_cutout` is set to `False`.
+            Number of thread to use. Default is 4. Valid only if `use_pyaedt_cutout` is set to `True`.
```
MaxJPRey, line 1422:
```suggestion
            Number of thread to use. Default is 4. Valid only if ``use_pyaedt_cutout`` is set to ``True``.
```

MaxJPRey (approved):

Issue #2770, opened 2023-04-04
Add Hfss3dLayout Center Line getters and setters for lines


maxcapodi78:
### Description of the feature

Add properties to modeler.lines in Hfss3dLayout to get and set center line.

### Steps for implementing the feature

```
from pyaedt import Hfss3dLayout

h3d = Hfss3dLayout(specified_version="2022.2")
h3d.modeler.lines["line_3"].center_line
h3d.modeler.lines["line_3"].center_line = {"Pt0":[1,0]}

```

PR #2771, opened 2023-04-04
Implement CenterLine into h3d modeler

maxcapodi78:
#2770

### DIFF
--- a/_unittest/test_41_3dlayout_modeler.py
+++ b/_unittest/test_41_3dlayout_modeler.py
@@ -287,6 +287,10 @@ class TestClass(BasisTest, object):
             "Bottom", [[0, 0], [10, 30], [20, 30]], lw=1, name="line1", net_name="VCC"
         )
         assert line.name == "line1"
+        assert isinstance(line.center_line, dict)
+        line.center_line = {"Pt0": [1, "0mm"]}
+        assert line.center_line["Pt0"] == ["1", "0"]
+        line.center_line = {"Pt0": ["0mm", "0mm"]}
 
     def test_13a_create_edge_port(self):
         port_wave = self.aedtapp.create_edge_port("line1", 3, False, True, 6, 4, "2mm")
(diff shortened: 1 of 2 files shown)


_unittest/test_41_3dlayout_modeler.py
```diff
@@ -287,6 +287,10 @@ def test_12_create_line(self):
             "Bottom", [[0, 0], [10, 30], [20, 30]], lw=1, name="line1", net_name="VCC"
         )
         assert line.name == "line1"
+        assert isinstance(line.center_line, dict)
+        line.center_line = {"Pt0": [1, "0mm"]}
```
MaxJPRey, line 291: @maxcapodi78 Shouldn't we provide list of `Points3dLayout` objects instead of a dictionary containing the name of the point as a key.

maxcapodi78: I made it as a dict since user may want to change just 1 point value and in this way he should not need to have the full list provided any time

MaxJPRey: Yes, that could avoid the users to have to do a `for loop` to get the point they are interested to

MaxJPRey (approved):
LGTM.

pyaedt/modeler/pcb/object3dlayout.py
```diff
@@ -1432,6 +1432,33 @@ def length(self):
...
+    @center_line.setter
+    def center_line(self, points):
+        u = self._primitives.model_units
+        for point_name, value in points.items():
+            vpoint = ["NAME:{}".format(point_name), "X:=", _dim_arg(value[0], u), "Y:=", _dim_arg(value[1], u)]
+            self.change_property(vpoint)
+
 
 class Points3dLayout(object):
     """Manages Hfss 3D Layout Points."""
```
PipKat, line 1464:
```suggestion
    """Manages HFSS 3D Layout points."""
```

Issue #2766, opened 2023-04-03
assign_rlc_model to allow only R or only L or only C and a combination of those


vishak-venkatraman:
### Description of the feature

Today assign_rlc_model() assumes that we need to provide all the values. However in the GUI we can give only one of those as well.

```
pyaedt ERROR: EDBComponent.assign_rlc_model() missing 2 required positional arguments: 'ind' and 'cap' - assign_rlc_model -  TYPE ERROR.
pyaedt ERROR: Arguments with values: 
pyaedt ERROR:     res = 0.05 
pyaedt ERROR:     ind = None 
pyaedt ERROR:     cap = None 
pyaedt ERROR:     is_parallel = False 
```

we would like to be able to do all of these as well

```
comp.assign_rlc_model(res, None, None, is_parallel=False)
comp.assign_rlc_model(res, ind, None, is_parallel=False)
comp.assign_rlc_model(None, ind, None, is_parallel=False)
comp.assign_rlc_model(None, ind, cap, is_parallel=False)
comp.assign_rlc_model(res, None, cap, is_parallel=False)
```
And other combinations

### Useful links and references

![image](https://user-images.githubusercontent.com/67716193/229616718-372dec29-d99d-44a7-9bb9-4a214f73a26e.png)

maxcapodi78:
Implemented in v0.6.71

PR #2772, opened 2023-04-04
Improved assign_rlc_model method

maxcapodi78:
#2766

### DIFF
--- a/pyaedt/edb_core/edb_data/components_data.py
+++ b/pyaedt/edb_core/edb_data/components_data.py
@@ -83,17 +83,17 @@ class EDBComponentDef(object):
         return {comp.refdes: comp for comp in comp_list}
 
     @pyaedt_function_handler
-    def assign_rlc_model(self, res, ind, cap, is_parallel=False):
+    def assign_rlc_model(self, res=None, ind=None, cap=None, is_parallel=False):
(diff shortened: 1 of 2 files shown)


pyaedt/edb_core/edb_data/components_data.py
```diff
@@ -895,27 +895,37 @@ def assign_s_param_model(self, file_path, name=None, reference_net=None):
         return self._set_model(model)
 
     @pyaedt_function_handler
-    def assign_rlc_model(self, res, ind, cap, is_parallel=False):
+    def assign_rlc_model(self, res=None, ind=None, cap=None, is_parallel=False):
```
MaxJPRey, line 898:
```suggestion
    def assign_rlc_model(self, res=0, ind=0, cap=0, is_parallel=False):
```

MaxJPRey (approved):

pyaedt/edb_core/edb_data/components_data.py
maxcapodi78: 0 can be a value. I need None to disable or enable

MaxJPRey: @maxcapodi78 OK, got it. It makes sense to allow disabling the creation or R, L or C.

PR #2782, opened 2023-04-05
Improve speed of history and box creation

maxcapodi78:


gmalinve (approved):

pyaedt/modeler/cad/object3d.py
```diff
@@ -1304,7 +1304,7 @@ def display_wireframe(self, fWireframe):
         self._change_property(vWireframe)
         self._wireframe = fWireframe
 
-    @property
+    @pyaedt_function_handler()
```
MaxJPRey, line 1307: Why is it not a property anymore?

MaxJPRey (approved):
LGTM.

pyaedt/modeler/cad/object3d.py
maxcapodi78: it's too slow due to recursiveness and makes both debugging and run very slow

Alberto-DM: Because it is a time and resource intensive operation. Since it gets called every time the object is inspected it was slow to debug the objects (pycharm sometimes goes in timeout).

PR #2784, opened 2023-04-05
Object 3d refactoring.

maxcapodi78:
made hidden some private methods (like m_Editor)
added property objects_by_names since most of users knows the object name and not the id

### DIFF
--- a/pyaedt/modeler/cad/Modeler.py
+++ b/pyaedt/modeler/cad/Modeler.py
@@ -1099,7 +1099,7 @@ class Lists(PropsManager, object):
                         object_list_new.append(int(element))
                     else:
                         if element in self._modeler.object_names:
-                            obj_id = self._modeler.object_id_dict[element]
+                            obj_id = self._modeler._object_id_to_names[element]
                             for sel in self._modeler.object_list:
                                 if sel.id == obj_id:
                                     for f in sel.faces:
@@ -1181,6 +1181,15 @@ class GeometryModeler(Modeler, object):
         self._user_lists = None
         self._planes = None
         self._is3d = is3d
+        self._solids = []
+        self._sheets = []
+        self._lines = []
+        self._points = []
+        self._unclassified = []
+        self._all_object_names = []
+        self.objects = {}
+        self.user_defined_components = {}
+        self._object_id_to_names = {}
 
     @property
     def coordinate_systems(self):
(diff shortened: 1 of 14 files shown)


pyaedt/modeler/cad/Primitives.py
```diff
@@ -1372,12 +1370,25 @@ def refresh(self):
...
+    @property
+    def objects_by_name(self):
+        """Object dictionary organized by name.
+
+        Returns
+        -------
+        dict
+        """
+        obj_dict = {}
+        for _, v in self.objects.items():
```
Alberto-DM, line 1388: why not like this?

```suggestion
        for v in self.objects.values():
```

pyaedt/modeler/cad/Modeler.py
```diff
@@ -2526,15 +2495,11 @@ def convert_to_selections(self, object_id, return_list=False):
...
-        elif not isinstance(object_id, list):
+        if not isinstance(object_id, list):
             object_id = [object_id]
         objnames = []
         for el in object_id:
-            if isinstance(el, int) and el in list(self.objects.keys()):
+            if isinstance(el, int) and el in self.objects:
                 objnames.append(self.objects[el].name)
             elif isinstance(el, int):
                 objnames.append(el)
```
Alberto-DM, line 2505: why there is this case? The returned list should contain the object names, here it is putting in an id.

pyaedt/modeler/cad/Primitives.py
```diff
@@ -1310,8 +1309,8 @@ def get_obj_id(self, objname):
             Object ID.
 
         """
-        if objname in self.object_id_dict:
-            return self.object_id_dict[objname]
+        if objname in self._object_id_to_names:
+            return self._object_id_to_names[objname]
```
Alberto-DM, line 1313: Here we pass the name as argument and we get back the id. So, should the name be 'object_name_to_id'?
I mean, the dictionary goes from name to id. Am I wrong?

Alberto-DM (approved):
Just a couple of comments. Otherwise LGTM.

pyaedt/modeler/cad/Primitives.py
```diff
@@ -592,11 +592,10 @@ def does_object_exists(self, object):
             else:
                 return False
```
MaxJPRey, line 593:
```suggestion
```

pyaedt/modeler/cad/Primitives.py
```diff
@@ -592,11 +592,10 @@ def does_object_exists(self, object):
...
         else:
-            for el in self.objects:
-                if self.objects[el].name == object:
-                    return True
-
-        return False
+            if object in self.objects_by_name:
+                return True
+            else:
+                return False
```
MaxJPRey, line 598:
```suggestion
        return False
```

pyaedt/modeler/cad/Primitives.py
```diff
@@ -592,11 +592,10 @@ def does_object_exists(self, object):
             else:
                 return False
         else:
```
MaxJPRey, line 594: Here the entire code:
```
        if isinstance(object, int):
            if object in self.objects:
                return True
            else:
                return False
        else:
            if object in self.objects_by_name:
                return True
            else:
                return False
```

Could become:
  ```
      if isinstance(object, int):
            if object in self.objects:
                return True
            elif object in self.objects_by_name:
                return True
            return False
```

maxcapodi78: thanks. Improved

pyaedt/modeler/cad/Primitives.py
maxcapodi78: because for Ironpython I should use list(self.objects.values()). I prefer this approach

PR #2785, opened 2023-04-05
add some functions to Revisions and fix some bugs

jsalant22:
Fixes B776164: revision changes not saved

Fixes S770448: revamp Revisions

Fixes B813304: issues saving to directory with multiple aedt projects

### DIFF
--- a/pyaedt/emit_core/results/results.py
+++ b/pyaedt/emit_core/results/results.py
@@ -140,16 +140,56 @@ class Results:
         """
         return [rev.name for rev in self.revisions]
 
-    @pyaedt_function_handler()
-    def analyze(self, revision_name=None):
+    @pyaedt_function_handler
+    def get_revision(self, revision_name=None):
         """
-        Analyze the specified design.
+        Load the specified revision.
 
         Parameters
         ----------
         revision_name : str
-            Revision to analyze. The default is ``None``,in which case the most recent revision
-            is loaded if it matches the current design revision.
+            Revision to load. If revision_name = None, the 
+            latest revision will be returned.
+
+        Returns
+        -------
+        rev:class:`pyaedt.modules.Revision`
+            Specified ``Revision`` object that was loaded.
+
+        Examples
+        --------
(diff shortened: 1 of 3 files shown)


pyaedt/emit_core/results/results.py
```diff
@@ -140,44 +140,37 @@ def revision_names(self):
...
-                    self.current_revision._load_revision()
-            else:
-                # there are changes since the current revision was analyzed, create
-                # a new revision
-                self.current_revision.revision_loaded = False
-                self.current_revision = self._add_revision()
+            # unload the current revision and load the latest
+            self.current_revision.revision_loaded = False
+            self.current_revision = self.revisions[-1]
+            self.current_revision._load_revision()
```
apollo-crum, line 173: @jsalant22 One thing I noticed is that _load_revision() prints a message if it is already loaded.  Doing that the way we are using _load_revision() here seems like a bad idea--depending on what the user is doing it could print that out a lot.  Do you think maybe we should get rid of that print statement?

pyaedt/emit_core/results/results.py
```diff
@@ -186,5 +179,44 @@ def analyze(self, revision_name=None):
...
+        ):
+            self.get_revision(self.revisions[-1].name)
+        else:
+            # there are changes since the current revision was analyzed, create
+            # a new revision
+            self.current_revision.revision_loaded = False
+            self.current_revision = self._add_revision()
+
+        # save the changes to the revision
+        self.emit_project._emit_api.save_project()
```
apollo-crum, line 221: @jsalant22 Is there a reason we are calling a save here?  The emit project isn't even loaded at this point is it?

apollo-crum (changes requested):

pyaedt/emit_core/results/results.py
apollo-crum: Might work to swap out the print for a logging.debug().  Seems like something we might want to start using.

apollo-crum (approved):

PR #2794, opened 2023-04-07
Improved plot Stackup

Alberto-DM:
Stackup plot is now supporting Overlapping and Laminate mode.
The stackup can be drawn with real layers thickness, or scaled thickness.


maxcapodi78 (approved):
Great

PR #2805, opened 2023-04-10
add icons

jsalant22:
add icons for the emit examples


jsalant22:
These are the icons that will be displayed on this page: https://aedt.docs.pyansys.com/release/0.6/examples/07-EMIT/index.html

myoung301:
@maxcapodi78 - I see the doc build was successful. Is there a way to view the build output so we can see how the example page will look before merging?

CC @jsalant22

maxcapodi78:
> @maxcapodi78 - I see the doc build was successful. Is there a way to view the build output so we can see how the example page will look before merging?
> 
> CC @jsalant22

The pyaedt documentation is published on https://aedt.docs.pyansys.com/version/stable/examples/index.html. Here you can find all examples including Emit published examples

The Dev documentaion doesn't build the examples  since it takes too much

myoung301:
Ok - I guess we need to push this in and then see what it looks like and adjust with another PR, if needed.

myoung301 (approved):

PR #2809, opened 2023-04-11
fixed existing points not enumerating

rmawatson:
It looks like app.modeler.planes will also not enumerate to a native object. It appears to return a dict of {name:comobject} and should really create a pyaedt Plane object also.


MaxJPRey:
> It looks like app.modeler.planes will also not enumerate to a native object. It appears to return a dict of {name:comobject} and should really create a pyaedt Plane object also.

Hi @rmawatson,
Do you intend to fix this issue regarding the plane in an other pull request? If so, could you create an issue for the time being?
Many thanks for working on the points.

PR #2821, opened 2023-04-13
assign_source function

lorenzovecchietti:
New function for assigning the source boundary condition. The old one (`create_source_power`) is deprecated. The new function has clearer arguments, and it is more general: it supports both temperature dependent assignment and transient ones.


maxcapodi78 (approved):
LGTM

PR #2828, opened 2023-04-14
fix extent creation when multiple polygons are generated from the computation.

maxcapodi78:
In case multiple polygons are generated the one with biggest area is taken as extent


amichel0205 (approved):
Tested with extent 10cm, 5cm and 8cm.
Working fine.

PR #2831, opened 2023-04-14
When Pyaedt is installed from Button in 23R2 there is no need to launch a new session of Desktop.

maxcapodi78:
…


kbhagat-ansys (approved):
LGTM

kbhagat-ansys:
@maxcapodi78 `pyaedt\pyaedt_with_IDE.bat` should be kept in sync with `pyaedt\doc\source\Resources\pyaedt_with_IDE.bat`

PR #2834, opened 2023-04-14
23R2 doesn't require anymore this property to be set as True.

maxcapodi78:


Samuelopez-ansys (approved):

MaxJPRey (approved):
LGTM.

pyaedt/edb.py
```diff
@@ -463,7 +463,8 @@ def open_edb(self, init_dlls=False):
             self._init_dlls()
         self.logger.info("EDB Path %s", self.edbpath)
         self.logger.info("EDB Version %s", self.edbversion)
```
MaxJPRey, line 465:
```suggestion
        self.logger.info("EDB path is: %s", self.edbpath)
        self.logger.info("EDB version is:  %s", self.edbversion)
```

maxcapodi78: Improved. Thaknks

PR #2846, opened 2023-04-17
Install doc dependencies.

MaxJPRey:


Revathyvenugopal162 (approved):
we can reduce line 27-33 in single actions. installing the additional dependencies will install the package is installed too:

PR #2847, opened 2023-04-17
Enable control program

gmalinve:


pyaedt/modules/SolveSetup.py
```diff
@@ -2585,6 +2587,76 @@ def add_eddy_current_sweep(
...
+        and allow you to control the source input, circuit elements, mechanical quantities, time step,
+        and stopping criteria, based on the updated solutions.
+
+        Parameters
+        ----------
+        control_program_path : str
+            File path of control program.
+        control_program_args : str, optional
+            Arguments to pass to control program.
+            Default value is ``" "``
```
MaxJPRey, line 2603:
```suggestion
            Default value is ``" "``.
```

pyaedt/modules/SolveSetup.py
```diff
@@ -2585,6 +2587,76 @@ def add_eddy_current_sweep(
...
+        call_after_last_step : bool, optional
+            if ``True`` the control program is called after the simulation is completed.
```
MaxJPRey, line 2605:
```suggestion
            If ``True`` the control program is called after the simulation is completed.
```

pyaedt/modules/SolveSetup.py
```diff
@@ -2585,6 +2587,76 @@ def add_eddy_current_sweep(
...
+        By default a control program script will be called by the pre-installed Python interpreter:
+        ``<install_path>\Win64\commonfiles\CPython\37\winx64\Release\python\python.exe``.
+        However, the user can specify a custom python interpreter to be used by setting following environment variable:
+        ``EM_CTRL_PROG_PYTHON_PATH=<path_to\python.exe>``
+
+        References
+        ----------
+        >>> oModule.EditSetup
+        """
+        if self.p_app.solution_type not in ["Transient", "TransientXY", "TransientZ"]:
```
MaxJPRey, line 2624: Aren't we missing a check for Maxwell 2D?

pyaedt/modules/SolveSetup.py
```diff
@@ -2585,6 +2587,76 @@ def add_eddy_current_sweep(
...
+
+        if not os.path.exists(control_program_path):
+            self._app.logger.error("Control Program file does not exist.")
+            return False
+
+        if not isinstance(control_program_args, str):
+            self._app.logger.error("Control Program arguments have to be a string.")
+            return False
+
+        self.props = copy.deepcopy(MaxwellTransient)
```
MaxJPRey, line 2636: Just for my knowledge, why do we need a deepcopy of MaxwellTransient

MaxJPRey (approved):

pyaedt/modules/SolveSetup.py
gmalinve: @MaxJPRey "TransientXY" and "TransientZ" are transient solution types for M2D :)

pyaedt/modules/SolveSetup.py
gmalinve: @MaxJPRey honestly I saw we are already using the deepcopy() to copy other setups so I used it here as well and base on the explanation it constructs a new compound object and then, recursively, inserts copies into it of the objects found in the original. Rather than a shallow copy which inserts references into it to the objects found in the original.

But if you or anyone else have a better solution please do tell me :)

pyaedt/modules/SolveSetup.py
MaxJPRey: Got it. Thanks @gmalinve!

pyaedt/modules/SolveSetup.py
MaxJPRey: This seems correct.

PR #2867, opened 2023-04-18
S778257: fix emit for multiple python versions

apollo-crum:


apollo-crum:
@myoung301 There are tests for class name that are failing because the full class name now depends on the python version (like "EmitApiPython310.InteractionDomain").  Are these tests still important?  If you think they are I can update them to just see if the full class name ends with the class name we expect.

apollo-crum:
@jsalant22 test_manual_revision_access_test_getters is failing because a frequency is off by 1000x, a units thing maybe?

pyaedt/emit_core/__init__.py
```diff
@@ -7,8 +7,19 @@
...
     sys.path.append(delcross_python_path)
-desktop_path = os.environ.get("ANSYSEM_ROOT232")
-if desktop_path and sys.version_info < (3, 8):
+desktop_path = os.environ.get("ANSYSEM_ROOT241")
+if not desktop_path:
+    desktop_path = os.environ.get("ANSYSEM_ROOT232")
+if desktop_path:
     path = os.path.join(desktop_path, "Delcross")
     sys.path.append(path)
-    EMIT_MODULE = import_module("EmitApiPython")
```
myoung301, line 14: @apollo-crum I don't think this change to the `import_module` call should be needed. We should still be able to `import_module("EmitApiPython")`. Instead of EmitApiPython referring to one EmitApiPython.pyd file, EmitApiPython is now a package (folder) in Delcross and the __init__.py in that folder takes care of importing the right version of pyd file (EmitApiPythonXY.pyd).

pyaedt/emit_core/EmitConstants.py
```diff
@@ -97,8 +97,10 @@ def unit_type():
 }
 """Valid units for each unit type."""
 
-desktop_path = os.environ.get("ANSYSEM_ROOT232")
-if desktop_path and sys.version_info < (3, 8):
+desktop_path = os.environ.get("ANSYSEM_ROOT241")
```
myoung301, line 100: @apollo-crum Can we use _aedt_version or something else within the Emit() class to retrieve this path so it keeps up with version changes automatically?

pyaedt/emit.py
```diff
@@ -151,7 +151,7 @@ def __init__(
         )
         self._modeler = ModelerEmit(self)
         self._couplings = CouplingsEmit(self)
-        if self._aedt_version >= "2023.1" and sys.version_info.major == 3 and sys.version_info.minor == 7:
+        if self._aedt_version >= "2023.1":
             self._emit_api = EMIT_MODULE.EmitApi()
             """Instance of the Emit api."""
```
PipKat, line 156:
```suggestion
            """Instance of the EMIT API."""
```

pyaedt/emit_core/EmitConstants.py
apollo-crum: I'm thinking I'll just remove it actually.  Unless someone knows why this check is needed?  @myoung301 @jsalant22

jsalant22: @apollo-crum I think it was needed to avoid calling unit_type() if only 23.1 or older is installed. So maybe change it to match the code used in emit.py:

if self._aedt_version >= "2023.1":

jsalant22:
@apollo-crum @myoung301 docs_build I think is failing due to this exception, but the only place I can find with **pyaedt.emit** is in design_types.py and I am at a loss for why the import would fail. Locally all the unit tests are passing for me on 24.1 (yesterday's build)

image

image

jsalant22 (approved):

PR #2883, opened 2023-04-20
FIX: Create versions.json

Revathyvenugopal162:
Add ``release/version.json`` file, which currently points to an older format, with the correct link only for the switcher. This should allow the current release to function properly. For future releases, we can ensure that the file points to the correct file ``versions.json``  from the start, and we can remove the old ``release/version.json`` file at that time.

### DIFF
--- a/release/versions.json
+++ b/release/versions.json
@@ -1,0 +1,12 @@
+[
+  {
+    "name": "dev",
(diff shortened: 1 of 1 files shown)


MaxJPRey (approved):
Thanks @Revathyvenugopal162

PR #2901, opened 2023-04-24
aedt edit terminal properties bug fix

svandenb-dev:


Samuelopez-ansys:
@svandenb-dev 

EditTerminal was not working?

PR #2923, opened 2023-04-27
Made simpler the _variable_dict method.

maxcapodi78:


Samuelopez-ansys (approved):
LGTM!

pyaedt/application/Variables.py
```diff
@@ -731,21 +731,42 @@ def _logger(self):
...
         self._dependent_project_variables = {}
 
+    @property
+    def _independent_variables(self):
+        all = {}
+        for k, v in self._independent_project_variables.items():
+            all[k] = v
+        for k, v in self._independent_design_variables.items():
+            all[k] = v
+        return all
```
MaxJPRey, line 746: @maxcapodi78 I think you could use the `update()` method here:
```suggestion
        all.update(self._independent_project_variables)
        all.update(self._independent_design_variables)

        return all
```

pyaedt/application/Variables.py
```diff
@@ -731,21 +731,42 @@ def _logger(self):
...
+
+    @property
+    def _dependent_variables(self):
+        all = {}
+        for k, v in self._dependent_project_variables.items():
+            all[k] = v
+        for k, v in self._dependent_design_variables.items():
+            all[k] = v
+        return all
```
MaxJPRey, line 755:
```suggestion
        all = {}
        all.update(self._dependent_project_variables)
        all.update(self._dependent_design_variables)
        return all
```

PR #2937, opened 2023-05-01
Update ComputeProtectionLevels.py.back

jsalant22:
modify example to ensure correct channel pair is used for computing powerAtRx

### DIFF
--- a/examples/07-EMIT/ComputeProtectionLevels.py.back
+++ b/examples/07-EMIT/ComputeProtectionLevels.py.back
@@ -138,8 +138,8 @@ for band in bands:
 rev = emitapp.results.analyze()
 modeRx = econsts.tx_rx_mode().rx
 modeTx = econsts.tx_rx_mode().tx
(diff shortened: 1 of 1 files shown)


myoung301 (approved):

examples/07-EMIT/ComputeProtectionLevels.py.back
```diff
@@ -138,7 +138,8 @@ for band in bands:
 rev = emitapp.results.analyze()
 modeRx = econsts.tx_rx_mode().rx
 modeTx = econsts.tx_rx_mode().tx
-resultMode = econsts.result_type().powerAtRx
+modePower = econsts.result_type().powerAtRx
+modeEmi = econsts.result_type().emi
```
MaxJPRey, line 142: I would just suggest to rename variables to follow Python snake_case convention.
```suggestion
mode_power = econsts.result_type().powerAtRx
mode_emi = econsts.result_type().emi
```

Issue #2916, opened 2023-04-25
Improve create_polygon_from_points() to allow ArcHeights


s-benter:
### Description of the current limitations

edb_core.modeler.get_polygon_points() returns a list of points.  Line endpoints are represented by two-value tuples for (x,y).  Arcs are represented by single-value tuples for ArcHeight.

edb.modeler.create_polygon_from_points() can only accept (x,y) pairs and throws an IndexError if ArcHeights are included included in the points list. 
 
for example:
```
 point_list = [[0 ,90], [0 ,10], [-2.9289321881345], [10 ,0], [90 ,0], [-2.9289321881345], [100 ,10], [100 ,90], [-2.9289321881345], [90 ,100], [10 ,100], [-2.9289321881345]]
 edb.modeler.create_polygon_from_points(point_list, "Layer") 
```

should return a polygon primitive of a 100mm square with rounded corners.

maxcapodi78:
@s-benter I've added the support for the height now. You have to pass a tuple of 3 values [xend, yend, height]. The Arc height will be created between previous point and [xend,yend] and height will be the Arc Height.

PR #2943, opened 2023-05-02
Added support of Arch Heights to create_polygon and create_polygon_from_points methods.

maxcapodi78:
#2916

### DIFF
--- a/pyaedt/edb_core/layout.py
+++ b/pyaedt/edb_core/layout.py
@@ -2,6 +2,7 @@
 This module contains these classes: `EdbLayout` and `Shape`.
 """
 import math
+import warnings
 
 from pyaedt.edb_core.edb_data.primitives_data import EDBPrimitives
 from pyaedt.edb_core.edb_data.utilities import EDBStatistics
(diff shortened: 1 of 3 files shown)


pyaedt/edb_core/layout.py
```diff
@@ -928,6 +938,20 @@ def _validatePoint(self, point, allowArcs=True):
...
+            if not allowArcs:
+                self._logger.error("Arc found but arcs are not allowed in _validatePoint.")
+                return False
+            if not isinstance(point[0], (int, float, str)):
+                self._logger.error("Point X value must be a number.")
+                return False
+            if not isinstance(point[1], (int, float, str)):
+                self._logger.error("Point Y value must be a number.")
+                return False
+            if not isinstance(point[1], (int, float)):
```
MaxJPRey, line 951: Why `z` cannot be a `str` like `x` and `y`?
```suggestion
            if not isinstance(point[1], (int, float,str)):
```

MaxJPRey (approved):

PR #2945, opened 2023-05-02
Bug fix create solder ball on polygon pad fixed

svandenb-dev:
Adding pads defined as polygon support. It happen for specific projects, when components pads are defined as polygon, the code was failing to evaluate the solder balls diameter. This is now supported with this pull request-


maxcapodi78 (commented):
Can you add a description on what was the problem and what has been fixed? Since this is not related to any issue is diffiult to understand what was the problem and what as been fixed simply from the title.

maxcapodi78 (approved):

PR #2948, opened 2023-05-02
Add a new method install_with_pip in generic.general_methods

maxcapodi78:
The methods is useful when pyaedt is used from AEDT console

### DIFF
--- a/pyaedt/generic/general_methods.py
+++ b/pyaedt/generic/general_methods.py
@@ -1583,7 +1583,7 @@ def install_with_pip(package_name, package_path=None, upgrade=False, uninstall=F
     package_name : str
         Name of the package to install.
     package_path : str, optional
-        Package path for GitHub download. Example ``git+https://......``.
+        Path for the GitHub package to download and install. For example, ``git+https://.....``.
     upgrade : bool, optional
         Whether to upgrade the package. The default is ``False``.
     uninstall : bool, optional


MaxJPRey (approved):
LGTM.

pyaedt/generic/general_methods.py
```diff
@@ -1574,6 +1574,48 @@ def _check_installed_version(install_path, long_version):
     return False
 
 
+def install_with_pip(package_name, package_path=None, upgrade=False, uninstall=False):  # pragma: no cover
+    """Install a new package using pip.
+    Useful to install a package from AEDT Console without launching the python environment.
```
PipKat, line 1579:
```suggestion
    This method can be used to install a package from the AEDT Console without launching the Python environment.
```

pyaedt/generic/general_methods.py
```diff
@@ -1574,6 +1574,48 @@ def _check_installed_version(install_path, long_version):
...
+
+    Parameters
+    ----------
+    package_name : str
+        Name of the package to install.
+    package_path : str, optional
+        Package path for GitHub download. Example ``git+https://......``.
```
PipKat, line 1586:
```suggestion
        Path for the GitHub package to download. For example, ``git+https://.....``.
```

pyaedt/generic/general_methods.py
```diff
@@ -1574,6 +1574,48 @@ def _check_installed_version(install_path, long_version):
...
+    upgrade : bool, optional
+        Whether to upgrade the package or not.
+    uninstall : bool, optional
+        Whether to install the package or uninstall the package.
```
PipKat, line 1590:
```suggestion
        Whether to install the package or uninstall the package.
```
This is confusing. The default is ``False``, but what does this do if you are choosing between installing or unintalling?

maxcapodi78: 2 options: install a package or uninstall a package. True, uninstall, False install

PR #2949, opened 2023-05-02
Use GetResultDirectory in Revision

myoung301:
Also:
- Use Revision constructor for _add_revision in Results
- Fix errors in ComputeChannelSeparation.py and ComputeProtectionLevels.py
- Add some logging output to EMIT API import

### DIFF
--- a/pyaedt/emit_core/results/results.py
+++ b/pyaedt/emit_core/results/results.py
@@ -48,7 +48,7 @@ class Results:
 
         Raises
         ------
-        RuntimeException if the name given is an ex
+        RuntimeError if the name given is not the name of an existing result set and a current result set already exists.
 
         Returns
         -------
--- a/pyaedt/emit_core/results/results.py
+++ b/pyaedt/emit_core/results/results.py
@@ -48,7 +48,8 @@ class Results:
 
         Raises
         ------
-        RuntimeError if the name given is not the name of an existing result set and a current result set already exists.
+        RuntimeError if the name given is not the name of an existing result set and a current result set already 
+        exists.
 
         Returns
         -------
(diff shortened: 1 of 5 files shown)


myoung301:
@maxcapodi78 @MaxJPRey I'm not sure who to ask about this, but the unit tests are timing out. It seems to be happening on non-EMIT tests, somewhere near the end, but that's all I can tell from the limited output shown. Is this a known, pre-existing issue with the unit tests?

jsalant22 (approved):

pyaedt/emit_core/results/results.py
```diff
@@ -38,27 +38,22 @@ def __init__(self, emit_obj):
...
         ----------
         name : str, optional
-            Name for the new revision. If None, it will
-            be named the current design revision.
+            Name for the new revision, if created. If name is ``None``, it will
+            be named after the current design revision.
+
+        Raises
+        ------
+        RuntimeException if the name given is an ex
```
PipKat, line 51: Incomplete sentence? This should be a complete sentence ending with a period (full stop).

pyaedt/emit_core/results/revision.py
```diff
@@ -19,6 +18,10 @@ class Revision:
         Name of the revision to create. The default is ``None``, in which case a
         default name is given.
```
PipKat, line 19: Should this have the same second sentence as in the previous file?

The default is ``None``, in which case the name of the current design revision is used.

myoung301:
@PipKat - thank you for the feedback. Do you want to review the updates and approve before this is merged?

MaxJPRey:
> @maxcapodi78 @MaxJPRey I'm not sure who to ask about this, but the unit tests are timing out. It seems to be happening on non-EMIT tests, somewhere near the end, but that's all I can tell from the limited output shown. Is this a known, pre-existing issue with the unit tests?
> 
> I had two timeout in a row and I've just started a 3rd run.

@myoung301 solved. It was due to an other modifications from an other PR.

PR #2950, opened 2023-05-02
Require N-1 to match domain size when interferer(s) are defined

myoung301:


jsalant22 (approved):

pyaedt/emit_core/results/revision.py
```diff
@@ -126,6 +126,8 @@ def run(self, domain):
         """
         self._load_revision()
         engine = self.emit_project._emit_api.get_engine()
+        if domain.interferer_names and engine.max_simultaneous_interferers != len(domain.interferer_names):
+            raise ValueError("The max_simultanous_interferers must equal the number of interferers in the domain.")
```
MaxJPRey, line 130:
```suggestion
            raise ValueError("The max_simultaneous_interferers must equal the number of interferers in the domain.")
```

PR #2951, opened 2023-05-03
added try except to shutil copy file for analyze method.

maxcapodi78:
In linux sometimes it gives SameFileError.


Samuelopez-ansys (approved):
LGTM

PR #2980, opened 2023-05-08
New eigenmode example

58Bahram:
this is the new eigenmode example
@Samuelopez-ansys


examples/02-HFSS/HFSS_eigenmode.py
```diff
@@ -0,0 +1,147 @@
...
+temp_folder = pyaedt.generate_unique_folder_name()
+project_path = pyaedt.downloads.download_file("eigenmode", "emi_PCB_house.aedt", temp_folder)
+
+
+###############################################################################
+# Launch AEDT
+# ~~~~~~~~~~~
+# Launch AEDT 2023 R1 in graphical mode.
+
+desktopVersion = "2023.1"
```
MaxJPRey, line 46: Just to follow Python snake_case convention.
```suggestion
desktop_version = "2023.1"
```

examples/02-HFSS/HFSS_eigenmode.py
```diff
@@ -0,0 +1,147 @@
...
+# You can set ``non_graphical`` either to ``True`` or ``False``.
+
+non_graphical = False
+
+###############################################################################
+# Launch AEDT
+# ~~~~~~~~~~~
+# Launch AEDT 2023 R1 in graphical mode.
+
+d = pyaedt.launch_desktop(desktopVersion, non_graphical=non_graphical, new_desktop_session=True)
```
MaxJPRey, line 62:
```suggestion
d = pyaedt.launch_desktop(desktop_version, non_graphical=non_graphical, new_desktop_session=True)
```

Samuelopez-ansys (approved):
LGTM

examples/02-HFSS/HFSS_eigenmode.py
```diff
@@ -0,0 +1,147 @@
+"""
+HFSS: Eigenmode Filter
```
PipKat, line 2:
```suggestion
HFSS: Eigenmode filter
```
Goggle developer style guide uses sentence case (not title case)

examples/02-HFSS/HFSS_eigenmode.py
```diff
@@ -0,0 +1,147 @@
...
+--------------------
+This example shows how you can use PyAEDT to automate eigenmode solver in HFSS.
+Eigenmode analysis can be applied to open, radiating structures
+using an absorbing boundary condition. This type of analysis can be very helpful
+to determine the resonant frequency of a geometry or an antenna and can be used to refine
+the mesh at the resonance, even when the resonant frequency of the antenna is not known.
```
PipKat, line 8:
```suggestion
This example shows how you can use PyAEDT to automate the eigenmode solver in HFSS.
Eigenmode analysis can be applied to open, radiating structures
using an absorbing boundary condition. This type of analysis is useful for
determining the resonant frequency of a geometry or an antenna and can be used to refine
the mesh at the resonance, even when the resonant frequency of the antenna is not known.
```

examples/02-HFSS/HFSS_eigenmode.py
```diff
@@ -0,0 +1,147 @@
...
+
+The challenge posed by this method is to identify and filter the non-physical modes
+resulting from reflection from boundaries of the main domain.
+Since the Eigenmode solver sorts by frequency and does not filter on the
+quality factor, these virtual modes will be present when the eigenmode approach is
+applied to nominally open structures.
+When looking for resonant modes over a wide frequency range for nominally
+enclosed structures, several iterations may be required as the minimum frequency
+is determined manually and simulations shall re-run until the complete frequency range is covered
+and all the important physical modes are calculated.
```
PipKat, line 18:
```suggestion
Because the Eigenmode solver sorts by frequency and does not filter on the
quality factor, these virtual modes are present when the eigenmode approach is
applied to nominally open structures.
When looking for resonant modes over a wide frequency range for nominally
enclosed structures, several iterations may be required because the minimum frequency
is determined manually and simulations re-run until the complete frequency range is covered
and all important physical modes are calculated.
```

examples/02-HFSS/HFSS_eigenmode.py
```diff
@@ -0,0 +1,147 @@
...
+
+The following script finds the physical modes of a model in a wide frequency range by automating the solution setup.
+During each simulation a user-defined number of modes are simulated and the modes with a Q higher than a user defined value are filtered.
+The next simulation will automatically continue to find modes having a frequency higher than the last mode of tge previous analysis.
+This will continue until the maximum frequency in the desired range is achieved.
```
PipKat, line 23:
```suggestion
During each simulation, a user-defined number of modes is simulated, and the modes with a Q higher than a user- defined value are filtered.
The next simulation automatically continues to find modes having a frequency higher than the last mode of the previous analysis.
This continues until the maximum frequency in the desired range is achieved.
```

examples/02-HFSS/HFSS_eigenmode.py
```diff
@@ -0,0 +1,147 @@
...
+###############################################################################
+# Perform required imports
+# ~~~~~~~~~~~~~~~~~~~~~~~~
+# Run through each cell. This cell imports the required packages.
+
+import sys
+import os
+import pyaedt
+
+# create temporary folder to download the example to this folder.
```
PipKat, line 35:
```suggestion
# Create a temporary folder to download the example to.
```

examples/02-HFSS/HFSS_eigenmode.py
```diff
@@ -0,0 +1,147 @@
...
+
+###############################################################################
+# Launch HFSS
+# ~~~~~~~~~~~
+# Launch HFSS 2023 R1 in graphical mode.
+
+hfss = pyaedt.Hfss(projectname=project_path, non_graphical=non_graphical)
+
+###############################################################################
+# Input Parameters for Eigenmode Solver
```
PipKat, line 72:
```suggestion
# Input parameters for eigenmode solver
```

examples/02-HFSS/HFSS_eigenmode.py
```diff
@@ -0,0 +1,147 @@
...
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+# The geometry and material should be already set. The anaylses will be generated by the code. Number of modes during each analysis, max allowed number is 20.
+# Entering a number higher than 10 might need long simulation time as the
+# eigenmode solver needs to converge on modes. fmin is the lowest frequency of interest, fmax highest frequesncy of interest.
+# parameter limit is the value, which based on it is decided which modes shall be ignored.
```
PipKat, line 77:
```suggestion
# The geometry and material should be already set. The analyses are generated by the code.
# Number of modes during each analysis, max allowed number is 20.
# Entering a number higher than 10 might need long simulation time as the
# eigenmode solver needs to converge on modes. ``fmin`` is the lowest frequency
# of interest. ``fmax`` is the highest frequency of interest.
# ``limit`` is the parameter limit that determines which modes are ignored.
```
The second line isn't a complete sentence. I'm not sure what is meant.

examples/02-HFSS/HFSS_eigenmode.py
```diff
@@ -0,0 +1,147 @@
...
+setup_nr = 1
+
+limit = 10
+resonance = {}
+
+###############################################################################
+# Finding the modes
+# ~~~~~~~~~~~~~~~~~
+# The following cell is a function and if called creates an eigenmode setup and solves it.
+# After solve each mode with its corresponding real frequency and quality factor will be saved for further processing.
```
PipKat, line 92:
```suggestion
# Find the modes
# ~~~~~~~~~~
# The following cell is a function.  If called, it creates an eigenmode setup and solves it.
# After the solve, each mode, along with its corresponding real frequency and quality factor,
# are saved for further processing.
```

examples/02-HFSS/HFSS_eigenmode.py
```diff
@@ -0,0 +1,147 @@
...
+        eigen_q_value = hfss.post.get_solution_data(expressions=eigen_q[cont], setup_sweep_name=setup_name+' : LastAdaptive', report_category = "Eigenmode")
+        eigen_mode_value = hfss.post.get_solution_data(expressions=eigen_mode[cont], setup_sweep_name=setup_name+' : LastAdaptive', report_category = "Eigenmode")
+        data[cont] = [eigen_q_value.data_real()[0], eigen_mode_value.data_real()[0]]
+        cont += 1
+
+    print(data)
+    return data
+
+###############################################################################
+# Automation
```
PipKat, line 122: This should be a verb to match the other section titles. Maybe something like this: Automate calls to cover frequency range

examples/02-HFSS/HFSS_eigenmode.py
```diff
@@ -0,0 +1,147 @@
...
+# ~~~~~~~~~~
+# Running the next cell will call the resonance function and save only those modes with a Q higher than the defined
+# limit. The find_resonance function will be called until the complete frequency range is covered.
+# At the end the physical modes in the whole frequency range will be reported.
```
PipKat, line 126:
```suggestion
# Running the next cell calls the resonance function and saves only those modes with a Q higher than the defined
# limit. The ``find_resonance`` function is called until the complete frequency range is covered.
# When the automation ends, the physical modes in the whole frequency range are reported.
```
We want to avoid future tense.

examples/02-HFSS/HFSS_eigenmode.py
58Bahram: changed the title

PR #2982, opened 2023-05-08
fix some emit bugs

jsalant22:
B829904: only split the first instance of '=' when getting component node props

B829824: make tx_rx_mode non-optional for get_active_frequencies and raise an error if 'both' is specified.

also added a function to get the band_node when given the radio_node and the band_name

### DIFF
--- a/_unittest/test_26_emit.py
+++ b/_unittest/test_26_emit.py
@@ -475,6 +475,8 @@ class TestClass(BasisTest, object):
         self.aedtapp = BasisTest.add_app(self, application=Emit)
         rad1, ant1 = self.aedtapp.modeler.components.create_radio_antenna("New Radio")
         rad2, ant2 = self.aedtapp.modeler.components.create_radio_antenna("Bluetooth Low Energy (LE)")
+        rad3, ant3 = self.aedtapp.modeler.components.create_radio_antenna("WiFi - 802.11-2012")
+        rad4, ant4 = self.aedtapp.modeler.components.create_radio_antenna("WiFi 6")
 
         # Check type
         rad_type = rad1.get_type()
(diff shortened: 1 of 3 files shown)


_unittest/test_26_emit.py
```diff
@@ -503,6 +513,46 @@ def test_radio_band_getters(self):
...
 
+        # Test error for trying to get BOTH tx and rx freqs
+        exception_raised = False
+        try:
+            freqs = rev.get_active_frequencies(radios[0], bands[0], econsts.tx_rx_mode().both, "MHz")
+        except:
+            exception_raised = True
+        assert exception_raised
+
+        # Get WiFi 2012 Tx Bands
```
myoung301, line 524:
```suggestion
        # Get WiFi 2012 Rx Bands
```

_unittest/test_26_emit.py
```diff
@@ -503,6 +513,46 @@ def test_radio_band_getters(self):
...
+
+        # Get WiFi 2012 All Bands (default args)
+        bands = rev.get_band_names(radios[2])
+        assert len(bands) == 32
+
+        # Get WiFi 6 All Bands (default args)
+        bands = rev.get_band_names(radios[3])
+        assert len(bands) == 192
+
+        # Get WiFi 6 Tx Bands
```
myoung301, line 544:
```suggestion
        # Get WiFi 6 Rx Bands
```

pyaedt/modeler/circuits/PrimitivesEmit.py
```diff
@@ -712,6 +712,22 @@ def bands(self):
...
+        band_name : name of the desired band node.
+
+        Returns
+        -------
+        band_node : Instance of the band node."""
+        band_nodes = self.bands()
+        for node in band_nodes:
+            if band_name == node.props["Name"]:
+                return node
+        return False
```
myoung301, line 729: Minor, but if the return type is an object and the expected object isn't found, `None` (no object) seems a little more fitting than False.

```suggestion
        return None
```

myoung301 (approved):

PR #3016, opened 2023-05-17
fix issue 3007 (get_messages failing with 3dcomp)

lorenzovecchietti:
Fix get_messages failing to get messages from projects stemming from the editing of 3d components. This happens because if a 3d component is open, oDesktop.GetMessages without the project name argument returns messages with "(3D Component)" appended to the project name.


lorenzovecchietti:
@maxcapodi78 codecov is not passing because the test is skipped in non-graphical. Works on my machine.

maxcapodi78 (approved):

PR #3024, opened 2023-05-19
add ctrl prg example

gmalinve:


maxcapodi78 (approved):
How long it takes to solve?

PR #3018, opened 2023-05-18
New block assignment function

lorenzovecchietti:
Close #2654

PR #3026, opened 2023-05-19
Improved docstring for block bc assignment functions

lorenzovecchietti:
@PipKat I applied your suggestions from https://github.com/ansys/pyaedt/pull/3018

The only one I did not applied was the one regarding the introduction of a link to the Icepak documentation. I'm currently thinking of introducing one or more function that may help an unfamiliar user to build those dictionaries. In this way we can also shorten those explanations which are repeated on several functions (and will be repeated on others) and can potentially become annoying to maintain.

### DIFF
--- a/pyaedt/icepak.py
+++ b/pyaedt/icepak.py
@@ -768,7 +768,7 @@ class Icepak(FieldAnalysis3D):
         >>> block.props["Nodes"]["Internal"][0]
         '2W'
         """
-        warnings.warn("This method is deprecated in 0.6.27. Use create_two_resistor_network_block", DeprecationWarning)
+        warnings.warn("This method is deprecated in 0.6.27. Use the create_two_resistor_network_block() method.", DeprecationWarning)
         if object_name in self.modeler.object_names:
             if gravity_dir > 2:
                 gravity_dir = gravity_dir - 3
--- a/pyaedt/icepak.py
+++ b/pyaedt/icepak.py
@@ -3848,7 +3848,7 @@ class Icepak(FieldAnalysis3D):
             )
             return None
         if ext_temperature != "AmbientTemp" and ext_temperature is not None and not htc:
-            self.logger.add_error_message("Set an argument for ``htc`` or remove ``ext_temperature`` argument.")
+            self.logger.add_error_message("Set an argument for ``htc`` or remove the ``ext_temperature`` argument.")
             return None
         if isinstance(ext_temperature, dict) and ext_temperature["Type"] == "Temp Dep":
             self.logger.add_error_message(
--- a/pyaedt/icepak.py
+++ b/pyaedt/icepak.py
@@ -684,7 +684,7 @@ class Icepak(FieldAnalysis3D):
         if input_power == 0:
             input_power = "0W"
         if not bool(input_power) ^ bool(thermal_dependent_dataset):
-            self.logger.error("Please assigned one input between ``thermal_dependent_dataset`` and  ``input_power``")
+            self.logger.error("Assign one input between ``thermal_dependent_dataset`` and  ``input_power``")
         if not source_name:
             source_name = generate_unique_name("Source")
         props = {}
@@ -768,9 +768,7 @@ class Icepak(FieldAnalysis3D):
         >>> block.props["Nodes"]["Internal"][0]
         '2W'
         """
-        warnings.warn(
-            "This method is deprecated in 0.6.27. Please use create_two_resistor_network_block", DeprecationWarning
-        )
+        warnings.warn("This method is deprecated in 0.6.27. Use create_two_resistor_network_block", DeprecationWarning)
         if object_name in self.modeler.object_names:
             if gravity_dir > 2:
                 gravity_dir = gravity_dir - 3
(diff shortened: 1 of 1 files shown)


PipKat:
> @PipKat I applied your suggestions from #3018
> 
> The only one I did not applied was the one regarding the introduction of a link to the Icepak documentation. I'm currently thinking of introducing one or more function that may help an unfamiliar user to build those dictionaries. In this way we can also shorten those explanations which are repeated on several functions (and will be repeated on others) and can potentially become annoying to maintain.

I'm all for avoiding maintenace. Let me know if I can help. We could consider putting the information in one place (in the user guide), and the linking to it where needed in the API doc. This way, the content would only need to be maintained in one place (but links would have to be placed in several places).

pyaedt/icepak.py
```diff
@@ -684,7 +684,7 @@ def create_source_power(
         if input_power == 0:
             input_power = "0W"
         if not bool(input_power) ^ bool(thermal_dependent_dataset):
-            self.logger.error("Please assigned one input between ``thermal_dependent_dataset`` and  ``input_power``")
+            self.logger.error("Assign one input between ``thermal_dependent_dataset`` and  ``input_power``")
```
PipKat, line 687: @lorenzovecchietti Thanks! I would like to see a period added before the closing double quote though!

pyaedt/icepak.py
```diff
@@ -768,9 +768,7 @@ def create_network_block(
         >>> block.props["Nodes"]["Internal"][0]
         '2W'
         """
-        warnings.warn(
-            "This method is deprecated in 0.6.27. Please use create_two_resistor_network_block", DeprecationWarning
-        )
+        warnings.warn("This method is deprecated in 0.6.27. Use create_two_resistor_network_block", DeprecationWarning)
```
PipKat, line 771: Can we use the suggested format in the second sentence of the warning: Use the ``create_two_resistor_network_block()`` method.

I ask because I'm not a programmer and do not know what the subsequent comma and DeprecationWarning does!

pyaedt/icepak.py
```diff
@@ -3789,28 +3787,40 @@ def assign_solid_block(
...
-            support the latter. ``"Values"`` contains a list of strings containing the parameters
-            required by the ``"Function"`` selection (e.g. ``"Linear"`` requires two parameters:
-            the value of the variable at t=0 and the slope of the line). The parameters required by
-            each ``Function`` option is in Icepak documentation. The parameters must contain the
-            units where needed.
+            String with the value and units of the power assignment or with
+            ``"Joule Heating"``. For a temperature-dependent or transient
+            assignment, a dictionary can be used. The dictionary should contain three keys:
+            ``"Type"``, ``"Function"``, and ``"Values"``.
+            - For the ``"Type"`` key, accepted values are ``"Temp Dep"`` and ``"Transient"``.
```
PipKat, line 3794: For a list to render properly, you must have a blank line before the first list element and after the last list element if there is more content.

pyaedt/icepak.py
```diff
@@ -3834,11 +3844,11 @@ def assign_solid_block(
         """
         if not self.modeler.get_object_from_name(object_name).solve_inside:
             self.logger.add_error_message(
-                "Please use ``assign_hollow_block`` function with this object as" "``solve_inside`` is ``False``."
+                "Use ``assign_hollow_block`` function with this object as" "``solve_inside`` is ``False``."
```
PipKat, line 3847:
```suggestion
                "Use the ``assign_hollow_block()`` method with this object as ``"solve_inside`` is ``False``."
```
I know with line limits, articles like "the" have often been dropped in the past. However, omission does reduce readability, so we are encouaging more readable messages but asking developers to add them.

pyaedt/icepak.py
```diff
@@ -3834,11 +3844,11 @@ def assign_solid_block(
...
             )
             return None
         if ext_temperature != "AmbientTemp" and ext_temperature is not None and not htc:
-            self.logger.add_error_message("Please set an argument for ``htc`` or remove ``ext_temperature`` argument.")
+            self.logger.add_error_message("Set an argument for ``htc`` or remove ``ext_temperature`` argument.")
```
PipKat, line 3851:
```suggestion
            self.logger.add_error_message("Set an argument for ``htc`` or remove the ``ext_temperature`` argument.")
```
Again--if you don't want to change the messages (and update existing tests) to add articles like "the", I do understand.

pyaedt/icepak.py
```diff
@@ -3902,27 +3912,37 @@ def assign_hollow_block(
...
-            required by the ``"Function"`` selection (e.g. ``"Linear"`` requires two parameters:
-            the value of the variable at t=0 and the slope of the line). The parameters required by
-            each ``Function`` option is in Icepak documentation. The parameters must contain the
-            units where needed.
+            String with value and units of the assignment. If ``"Total Power"`` is
+            assignment_type, ``"Joule Heating"`` can be used.
+            For a temperature-dependent or transient assignment, a dictionary can be used.
+            The dictionary should contain three keys:
+            ``"Type"``, ``"Function"``, and ``"Values"``.
+            - For the ``"Type"`` key, accepted values are ``"Temp Dep"`` and ``"Transient"``.
```
PipKat, line 3923: Again, for proper doc rending, we need to have a blank list before the first list item and after the last list item.

PipKat (approved):

pyaedt/icepak.py
MaxJPRey:
```suggestion
            self.logger.error("Assign one input between ``thermal_dependent_dataset`` and  ``input_power``.")
```

MaxJPRey (approved):

PR #3031, opened 2023-05-21
fix get_objects_by_material

gmalinve:


MaxJPRey (approved):
LGTM.

Samuelopez-ansys:
@gmalinve What was wrong with the previous implementation?

gmalinve:
@Samuelopez-ansys there was more than one error actually. 
First error was due to the fact that when parsing all objects, in some cases there were None objects.
Second error is that we are checking if obj.material_name it's equal to the material name given by the user but in some cases obj.material_name is a formula (see the parametric sweep case) so the object list is always empty.

maxcapodi78 (approved):

PR #3043, opened 2023-05-24
Fix create text method to support schematic_units

maxcapodi78:


amichel0205 (approved):
Branch tested successfully

PR #3071, opened 2023-05-31
New example post layout parameterization and new property edb.board_size

ring630:


ring630:
@Samuelopez-ansys @maxcapodi78 it looks like something's wrong in Icepak.

maxcapodi78 (approved):
Approved

PR #3076, opened 2023-05-31
Simulation configuration flexibility added

svandenb-dev:
adding more flexibility with SimulationConfiguration class
- add_frequency_sweep[bool] allowing custom frequency sweep assignement
- include_only_selected_nets only available when du_cutou_subdesign = False, to remove only nets which are not selected in signal and power nets 
- generate_excitations allowing custom port definition after build project

### DIFF
--- a/pyaedt/edb_core/edb_data/simulation_configuration.py
+++ b/pyaedt/edb_core/edb_data/simulation_configuration.py
@@ -606,7 +606,7 @@ class SimulationConfigurationBatch(object):
         Returns
         -------
         bool
-            ``True`` or ``False. Default value is ``False``.
+            ``True`` or ``False``. Default value is ``False``.
 
         """
         return self._include_only_selected_nets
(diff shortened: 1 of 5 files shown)


pyaedt/edb_core/edb_data/simulation_configuration.py
```diff
@@ -561,6 +564,58 @@ def signal_layers_properties(self, value):  # pragma: no cover
...
+        return self._add_frequency_sweep
+
+    @add_frequency_sweep.setter
+    def add_frequency_sweep(self, value):
+        if isinstance(value, bool):
+            self._add_frequency_sweep = value
+
+    @property
+    def include_only_selected_nets(self):
+        """Include only net selection in the project. Is only used when do_cutout is set to ``False``. Will also be
```
MaxJPRey, line 603:
```suggestion
        """Include only net selection in the project. It is only used when ``do_cutout`` is set to ``False``. Will also be
```

pyaedt/edb_core/edb_data/simulation_configuration.py
```diff
@@ -561,6 +564,58 @@ def signal_layers_properties(self, value):  # pragma: no cover
...
+        ignored if signal_nets and power_nets are ``None``, resulting project will have all nets included.
+
+        Returns
+        -------
+        bool
+            ``True`` or ``False. Default value is ``False``.
```
MaxJPRey, line 609:
```suggestion
            ``True`` or ``False``. Default value is ``False``.
```

maxcapodi78 (approved):

PR #3079, opened 2023-05-31
Edb refactor

maxcapodi78:


svandenb-dev (approved):

_unittest/test_26_emit.py
myoung301, line 1: Why is this file changing? It looks like the changes shown here are already on main, but if that's the case, I'm not sure why I'm seeing them in this PR.

pyaedt/emit_core/results/revision.py
myoung301, line 1: I'm seeing the same thing in this file - new sections are shown in the diff for this PR that are already on main. Am I looking at something wrong?

myoung301 (changes requested):
It looks like there are extra changes in the PR when I review the diffs. Is there a problem with the PR or am I looking at something wrong?

_unittest/test_26_emit.py
maxcapodi78: because I've merged from main

pyaedt/emit_core/results/revision.py
maxcapodi78: you see because I've merged from main

_unittest/test_26_emit.py
myoung301: Ok - I see it's a 'limitation' of GitHub. Looks like there are some ways to get a true diff though. 

https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch

This did the trick:

https://github.com/ansys/pyaedt/compare/main...edb_refactor

myoung301 (approved):

PR #3084, opened 2023-06-02
Remove Machine_learning_applied_to_Patch.py

dcrawforAtAnsys:
Will be refactored and republished after refactor. 
Currently this example has too many mistakes and should not be published.


MaxJPRey (approved):
LGTM.

PR #3086, opened 2023-06-05
Release COM object

Samuelopez-ansys:


pyaedt/desktop.py
```diff
@@ -1323,6 +1323,16 @@ def release_desktop(self, close_projects=True, close_on_exit=True):
...
+        props = [a for a in dir(self) if not a.startswith("__")]
+        for a in props:
+            self.__dict__.pop(a, None)
+        dicts = [self, sys.modules["__main__"]]
+        for dict_to_clean in dicts:
+            props = [a for a in dir(dict_to_clean) if "win32com" in str(type(dict_to_clean.__dict__.get(a, None)))]
+            for a in props:
+                dict_to_clean.__dict__[a] = None
+
+        gc.collect()
```
MaxJPRey, line 1335: Is this mandatory?

Samuelopez-ansys: @MaxJPRey After some tests, it is cleaner. COM object was not released in all cases, and we found this strategy the best one to reset all. With GRPC we will not have this problem, but for now PyAEDT must live with both.

MaxJPRey (approved):
LGTM.

PR #3122, opened 2023-06-14
Fix/emit test analyze manually

jsalant22:


apollo-crum (approved):

myoung301 (approved):

pyaedt/emit_core/results/revision.py
```diff
@@ -306,9 +306,8 @@ def get_active_frequencies(self, radio_name, band_name, tx_rx_mode, units=""):
             Name of the radio/emitter.
         band_name : str
            Name of the band.
-        tx_rx : :class:`emit_constants.TxRxMode`
-            Specifies whether to get ``tx`` or ``rx`` radio freqs. The default
-            is ``None``, in which case both ``tx`` and ``rx`` freqs are returned.
+        tx_rx_mode : :class:`emit_constants.TxRxMode`
+            Specifies whether to get ``tx`` or ``rx`` radio freqs.
```
PipKat, line 310:
```suggestion
            Specifies whether to get ``tx`` or ``rx`` radio frequencies.
```

pyaedt/emit_core/results/revision.py
```diff
@@ -306,9 +306,8 @@ def get_active_frequencies(self, radio_name, band_name, tx_rx_mode, units=""):
...
         units : str, optional
             Units for the frequencies. The default is ``None`` which uses the units
```
PipKat, line 312:
```suggestion
            Units for the frequencies. The default is ``None``, in which case the units
```

pyaedt/emit_core/results/revision.py
```diff
@@ -306,9 +306,8 @@ def get_active_frequencies(self, radio_name, band_name, tx_rx_mode, units=""):
...
             specified globally for the project.
```
PipKat, line 313:
```suggestion
            specified globally for the project are used.
```

myoung301: @PipKat I see that the merge already happened, but this would make it "uses ... are used" so I think that's a little awkward.

pyaedt/emit_core/results/revision.py
myoung301: @PipKat - We should be able to put this tweak in a future PR. (@jsalant22 @apollo-crum)

pyaedt/emit_core/results/revision.py
PipKat: @myoung301 What I was suggesting, and which can done sometime in the future for the sake of consistency was this description:

Units for the frequencies. The default is ``None``, in which case the units specified globally for the project are used."

No biggie! If I ever get time to skim PyAEDT doc for an overall review, I can make the change.

PR #3137, opened 2023-06-17
Siwave voltage probe

ring630:


pyaedt/edb_core/siwave.py
```diff
@@ -1295,6 +1295,39 @@ def create_voltage_source_on_pin_group(
...
+        Parameters
+        ----------
+        probe_name : str
+            Name of the probe.
+        pos_pin_group_name : str
+            Name of the positive pin group.
+        neg_pin_group_name : str
+            Name of the negative pin group.
+        impedance : int, float, optional
+            Phase of the source
```
MaxJPRey, line 1311:
```suggestion
            Phase of the source.
```

MaxJPRey: @ring630 Is it the phase of the impedance?

PR #3156, opened 2023-06-20
Optimized open_aedt_file

Alberto-DM:
- fixed bug and optimized subkey search
- simplified _decode_subkey method
- removed _decode_value_and_save
- optimized some methods using load_keyword_in_aedt_file instead of load_entire_aedt_file


maxcapodi78 (approved):
great catch

PR #3161, opened 2023-06-21
Improved desktop launcher

maxcapodi78:


pyaedt/desktop.py
```diff
@@ -99,18 +96,12 @@ def launch_desktop_on_port():
...
-        time.sleep(1)
-        active_after = grpc_active_sessions(
-            settings.aedt_version, non_graphical=settings.non_graphical, student_version=settings.is_student
-        )
-        if k > timeout:
-            return False, port
+    if port not in grpc_active_sessions(
+        settings.aedt_version, non_graphical=settings.non_graphical, student_version=settings.is_student
+    ):  # pragma: no cover
+        return False, port
```
Alberto-DM, line 104: if `port` is not in `grpc_active_sessions`, why do you return `port`? Maybe return `None` instead?

Alberto-DM (approved):
LGTM, good catch!

Samuelopez-ansys (approved):

pyaedt/desktop.py
maxcapodi78: to be coherent with expected format

PR #3166, opened 2023-06-22
Added UT for 3D Layout and Hfss Field Plots on Nets and Layers.

maxcapodi78:
Working from 23 R2


Samuelopez-ansys (approved):
LGTM

_unittest/test_41_3dlayout_modeler.py
```diff
@@ -729,8 +749,8 @@ def test_92_import_edb(self):
 
     @pytest.mark.skipif(config["desktopVersion"] < "2022.2", reason="Not Working on Version earlier than 2022R2.")
```
PipKat, line 750: Even in tests, messages should be short, complete sentences. Version for customer-facing doc should always be in this format: 2022R2. Consider rewriting the reason as this: This test does not work on versions earlier than 2022 R2.

maxcapodi78: I can change it but I didn't touch that UT and that comment in this PR

MaxJPRey (approved):
LGTM.

Issue #3165, opened 2023-06-22
In cutout method add an option to preserve padstack on 1pin component when option remove_single_pin_components=True,


amichel0205:
### Description of the feature

In cutout method when option remove_single_pin_components=True all component having only 1 pin are removed.
Both the component and the pin are removed.
I would like to have only the component removed but not the padstack of the pin.
The raison of this request is that in some design the padstack has two usages:

1. Connect the component
2. Use a stub for matching purpose

Removing the padstack will damage return loss in some case.

I think it is better to have this as an option to enable only if remove_single_pin_components=True

I will share test case directly with the assignee as I can share it on this site.

Issue #3164, opened 2023-06-22
Improve cutout method such all pins of component with more than two pins using spice model are preserved


amichel0205:
### Description of the feature

Cutout method is working with two list of nets: one list is used to define the cutout outline(signal_list) and the other list(reference_list) will be cutout by the defined outline.
Before to start cutout outline all nets that are not part of these two list are removed.
An issue occurs with SIwave when a component is associated to a Spice model and some of the pins of this component are connected to nets not part of one of the two list. I've excluded 2 pins components for this request as it should be handle differently and will be described in another improvement request.
I suggest the following steps:

1) Identify all components with:
- more than two pins and connected to a spice model and having at list on pin connected to net included in signal_list
2) From all identified components:
- Add nets that are connected to these components pins and not part of signal_list to reference_list
3) Remove all nets that are not part of signal_lis or reference_list.

I will provide directly to the assignee of this request a test case (can be shared here)

It may be needed to discuss about the process describe above to ensure this will not introduce some unexpected edge effect.

PR #3168, opened 2023-06-22
Preserve Spice Pins during cutout in EDB

maxcapodi78:
Added new argument to cutout method working only on pyaedt cutout which actually checks all components and find pins on components with associated models (spice, sparam, netlist) and preserve them to avoid errors in siwave simulation.

Option is set by default to False to preserve actual behaviour.
Added an argument to delete_single_pin_rlc to preserve the component and deactivate it
#3165 & #3164

### DIFF
--- a/pyaedt/edb.py
+++ b/pyaedt/edb.py
@@ -1501,7 +1501,7 @@ class Edb(Database):
             Default is `10`.
         preserve_components_with_model : bool, optional
             Whether to preserve all pins of components that have associated models (Spice or NPort).
-            It works only in pyaedt cutout (except point list).
+            This parameter is applicable only for a PyAEDT cutout (except point list).
 
 
         Returns
--- a/pyaedt/edb.py
+++ b/pyaedt/edb.py
@@ -1437,7 +1437,7 @@ class Edb(Database):
         preserve_components_with_model=False,
     ):
         """Create a cutout using an approach entirely based on PyAEDT.
-        This new method replaces all legacy cutout methods in pyaedt.
+        This method replaces all legacy cutout methods in PyAEDT.
         It does in sequence:
         - delete all nets not in list,
         - create a extent of the nets,
--- a/pyaedt/edb.py
+++ b/pyaedt/edb.py
@@ -1434,6 +1434,7 @@ class Edb(Database):
         include_pingroups=False,
         expansion_factor=0,
         maximum_iterations=10,
+        preserve_components_with_model=False,
     ):
         """Create a cutout using an approach entirely based on pyaedt.
         This new method replaces all legacy cutout methods in pyaedt.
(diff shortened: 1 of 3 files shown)


pyaedt/edb.py
```diff
@@ -1498,6 +1499,10 @@ def cutout(
         maximum_iterations : int, optional
             Maximum number of iterations before stopping in searching for a cutout with an error.
             Default is `10`.
+        preserve_components_with_model : bool, optional
+            Whether to preserve all pins of components that have associated models (Spice or NPort).
+            It works only in pyaedt cutout (except point list).
```
PipKat, line 1504:
```suggestion
            This parameter is applicable only for a PyAEDT cutout (except point list).
```

pyaedt/edb.py
```diff
@@ -1498,6 +1499,10 @@ def cutout(
         maximum_iterations : int, optional
             Maximum number of iterations before stopping in searching for a cutout with an error.
```
PipKat, line 1500:
```suggestion
            Maximum number of iterations before stopping a search for a cutout with an error.
```

pyaedt/edb.py
```diff
@@ -1434,6 +1434,7 @@ def cutout(
         include_pingroups=False,
         expansion_factor=0,
         maximum_iterations=10,
+        preserve_components_with_model=False,
     ):
         """Create a cutout using an approach entirely based on pyaedt.
```
PipKat, line 1439:
```suggestion
        """Create a cutout using an approach entirely based on PyAEDT.
```

pyaedt/edb.py
```diff
@@ -1822,6 +1829,7 @@ def _create_cutout_multithread(
         custom_extent_units="mm",
         check_terminals=False,
         include_pingroups=True,
+        preserve_components_with_model=False,
     ):
         if is_ironpython:  # pragma: no cover
             self.logger.error("Method working only in Cpython")
```
PipKat, line 1835: Error messages should be short but complete sentences with punctuation: This method works only in CPython.

pyaedt/edb_core/components.py
```diff
@@ -1509,9 +1509,15 @@ def create_pingroup_from_pins(self, pins, group_name=None):
...
     @pyaedt_function_handler()
-    def delete_single_pin_rlc(self):
+    def delete_single_pin_rlc(self, deactivate_only=False):
+        # type: (bool) -> list
         """Delete all RLC components with a single pin.
 
+        Parameters
+        ----------
+        deactivate_only : bool, optional
+            Whether to delete the components or keep them and deactivate only. Default is ``False`` to delete them.
```
PipKat, line 1519:
```suggestion
            Whether to only deactivate components rather than delete them.
            The default is ``False``, which deletes the components.
```

MaxJPRey (approved):
LGTM.

PR #3193, opened 2023-06-28
Fixes for assign block list input

lorenzovecchietti:


pyaedt/icepak.py
```diff
@@ -4118,7 +4114,12 @@ def assign_hollow_block(
 
         if not isinstance(object_name, list):
             object_name = [object_name]
-
+        for o_n in object_name:
+            if self.modeler.get_object_from_name(o_n).solve_inside:
+                self.logger.add_error_message(
+                    "Use ``assign_solid_block`` method with this object as ``solve_inside`` is ``True``."
```
PipKat, line 4120:
```suggestion
                    "Use the  ``assign_solid_block`` method with this object as ``solve_inside`` is ``True``."
```
The "the" appeared in the previous similar message for the ``False`` condition (line 3979).

maxcapodi78 (approved):

PR #3215, opened 2023-06-30
add docstrings and test coverage

jsalant22:
improve Coupling.py docstrings and add test coverage that was missing


pyaedt/emit_core/Couplings.py
```diff
@@ -9,12 +9,12 @@ class CouplingsEmit(object):
     """Provides for interaction with the EMIT ```coupling`` folder,
```
PipKat, line 9:
```suggestion
    """Provides for interaction with the EMIT ```coupling`` folder.
    
```

pyaedt/emit_core/Couplings.py
```diff
@@ -9,12 +9,12 @@ class CouplingsEmit(object):
...
     This class is accessible through the EMIT application results variable
     object (``emit.couplings``).
```
PipKat, line 11:
```suggestion
    This class is accessible through the results variable
    object (``emit.couplings``) for the EMIT app.
```
Here is a case where it was capitalized, but perhaps this should be rephrased ".... for Emit". We really should standardize on capitalization and whether we use "app" after the product/module name.

pyaedt/emit_core/Couplings.py
```diff
@@ -186,17 +186,17 @@ def cad_nodes(self):
...
+
         Parameters
         ----------
         None
-        
+
         Returns
         -------
         antenna_nodes : dict
             A dict containing all of the antenna nodes with their
```
PipKat, line 197:
```suggestion
            A dictionary containing all of the antenna nodes with their
```

pyaedt/emit_core/Couplings.py
jsalant22: When it's not referring to code, yes it's the product name and typically all caps.

_unittest/test_26_emit.py
```diff
@@ -380,18 +382,32 @@ def test_revision_generation(self):
...
         assert rev2.name == "Revision 13"
         assert notes == "Added Bluetooth and an antenna"
+
         # get the initial revision
         rev3 = self.aedtapp.results.get_revision("Revision 10")
         assert rev3.name == "Revision 10"
         assert rev3.revision_number == 10
         assert rev_timestamp == rev3.timestamp
+
+        # test result_mode_error(), try to access unloaded revision
```
MaxJPRey, line 392:
```suggestion
```

myoung301 (approved):
Looks good to me

Issue #3090, opened 2023-06-06
Airbox horizontal factor missing when assigning extents in 3d layout


skandak-ansys:
`Hfss3dLayout.edit_hfss_extents` command doesn't have a parameter for airbox horizontal factor.

maxcapodi78:
Added air_horizontal_padding to edit_hfss_extents

PR #3220, opened 2023-07-03
added argument air_horizontal_padding to edit_hfss_extent method in Hfss 3D Layout

maxcapodi78:
Implement #3090

### DIFF
--- a/pyaedt/hfss3dlayout.py
+++ b/pyaedt/hfss3dlayout.py
@@ -1775,6 +1775,7 @@ class Hfss3dLayout(FieldAnalysis3DLayout):
         air_vertical_positive_padding=None,
         air_vertical_negative_padding=None,
         airbox_values_as_dim=True,
+        air_horizontal_padding=None,
     ):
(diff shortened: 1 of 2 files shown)


pyaedt/hfss3dlayout.py
```diff
@@ -1820,6 +1824,9 @@ def edit_hfss_extents(
         if not air_truncate_model_at_ground_layer == "keep":
             arg.append("TruncAtGnd:=")
             arg.append(air_truncate_model_at_ground_layer)
+        if air_horizontal_padding:
+            arg.append("AirHorExt:=")
+            arg.append(["Ext:=", str(air_vertical_positive_padding), "Dim:=", airbox_values_as_dim])
```
MaxJPRey, line 1829: Where is `air_horizontal_padding` value used?

maxcapodi78: @MaxJPRey good catch. I fixed it

Samuelopez-ansys (approved):

PR #3230, opened 2023-07-05
added unit test for interference classification

madelinecoleman1:


jsalant22:
@madelinecoleman1 We should toggle some of the filtered states off and verify that the scripts still output the correct results. Does the example you set up have enough detail for that?

madelinecoleman1:
> @madelinecoleman1 We should toggle some of the filtered states off and verify that the scripts still output the correct results. Does the example you set up have enough detail for that?

@jsalant22 it tests each of the functions in interference_classification.py once with no filtering and once with filtering (1 categories deselected for each function). Is this good or should I test with each category deselected individually?

_unittest/test_26_emit.py
```diff
@@ -898,6 +898,143 @@ def test_availability_1_to_1(self):
...
+        # Reduce the bluetooth transmit power
+        bands = rad1.bands()
+        for band in bands:
+            band.set_band_power_level(-20)
+
+        # Disable L2 P(Y) GPS Band
+        bands = rad2.bands()
+        for band in bands:
+            for child in band.children:
+                if "L2 P(Y)" in band.node_name:
```
pumichat-kla, line 934: Do you want to disable if the band ="L2 P(Y)"?

_unittest/test_26_emit.py
```diff
@@ -898,6 +898,143 @@ def test_availability_1_to_1(self):
...
+                    band.enabled = True
```
pumichat-kla, line 935: Do you want to disable "L2 P(Y)"?

_unittest/test_26_emit.py
```diff
@@ -898,6 +898,143 @@ def test_availability_1_to_1(self):
...
+        expected_protection_power = [["N/A", "< -200", "< -200"], ["< -200", "< -200", "N/A"]]
+
+        interference_colors = []
+        interference_power_matrix = []
+        protection_colors = []
+        protection_power_matrix = []
+        interference_filter = [
+            "TxFundamental:Out-of-band",
+            ["TxHarmonic/Spurious:Out-of-band", "Intermod:Out-of-band", "Broadband:Out-of-band"],
+        ]
```
pumichat-kla, line 1020: Can you explain this? 
image

From this picture, you want to do only for TxFundamental:Out-of-band?

_unittest/test_26_emit.py
```diff
@@ -898,6 +898,143 @@ def test_availability_1_to_1(self):
...
+            band.set_band_power_level(-20)
+
+        # Disable L2 P(Y) GPS Band
+        bands = rad2.bands()
+        for band in bands:
+            for child in band.children:
+                if "L2 P(Y)" in band.node_name:
+                    band.enabled = True
+                else:
+                    band.enabled = False
```
pumichat-kla, line 937: Do you want to disable or enable "L2 P(Y)"?

madelinecoleman1: It's supposed to be enabled, I'll adjust the comments above it

_unittest/test_26_emit.py
madelinecoleman1: The gui/script classifies interference slightly different that By Problem Type. For the source side, it checks whether the signal is fundamental or not. On the victim side, it checks if the source signal falls within the band. For each case of interference, a string with the format of those in "interference_filter" is created and checked against the list of filter values to check if the result is within the scope of the filters.

jsalant22 (approved):

jsalant22:
@pumichat-kla You can merge if/when you approve.

PR #3241, opened 2023-07-10
Embed pydesktop plugin

maxcapodi78:


Alberto-DM (approved):
Very good improvement.

PR #3253, opened 2023-07-11
comment out emit examples which needs refactoring to run on CICD in full documentation

maxcapodi78:


jsalant22:
@maxcapodi78 How does the auto documentation work? Does it attempt to create a docs page for each *.py file within the 07-EMIT directory? 

We have a few examples that share some code and put that shared code into a separate file within the example directory. Should this common code be located elsewhere in the pyaedt structure? Can we create a subdirectory within the 07-EMIT directory that contains these files without them getting compiled into documentation? 

cc: @madelinecoleman1

jsalant22:
> it creates one html page for each python file.

@maxcapodi78 Does it recursively do this or only for the top level files? If I have the directory structure below, which pages will it create html pages (and jupyter notebooks) for? Directories are in **bold**

- **07 - EMIT**
  - Example 1.py
  - Example 2.py
  - **Support Files**
    - Support Function 1.py
    - Support Function 2.py
  - Example 3.py

MaxJPRey (approved):
LGTM.

PR #3254, opened 2023-07-11
Replace rlc by gap boundaries

svandenb-dev:
Added the command to convert RLC component in EDB to RLCBoundary type compatible with 3D modeler export (e.g HFSS).
Values from EDB are kept, only 2 pins RLC's are supported.


pyaedt/edb_core/components.py
```diff
@@ -996,6 +996,48 @@ def _get_closest_pin_from(self, pin, ref_pinlist):
                 closest_pin = ref_pin
         return closest_pin
 
+    @pyaedt_function_handler()
+    def replace_rlc_by_gap_boundaries(self, component=None):
+        """Replace RLC component by RLC gap boundaries. These boundary type are compatible with 3D modeler export.
```
MaxJPRey, line 1001:
```suggestion
        """Replace RLC component by RLC gap boundaries. These boundary types are compatible with 3D modeler export.
```

pyaedt/edb_core/components.py
```diff
@@ -1100,6 +1142,80 @@ def add_port_on_rlc_component(self, component=None):
             self._logger.info("Component {} has been replaced by port".format(component.refdes))
             return True
 
+    @pyaedt_function_handler()
+    def add_rlc_boundary(self, component=None, circuit_type=True):
```
MaxJPRey, line 1146: Could it help the user if we rename the method as:
```suggestion
    def add_rlc_gap_boundary(self, component=None, circuit_type=True):
```

svandenb-dev: actually this command is used either to define gap boundary or circuit one. It is called by the command reaplce_rlc_by_gap_boundaries

maxcapodi78 (approved):

PR #3267, opened 2023-07-13
Region properties functions

lorenzovecchietti:


Samuelopez-ansys:
@lorenzovecchietti Why is this method only available in Icepak.py? 

Region is common for all 3D modelers

lorenzovecchietti:
@Samuelopez-ansys You're right. Moved to 3d modeler. Is it the right place?

Samuelopez-ansys:
@lorenzovecchietti Yes it is the right place thanks.

You could also use this method to get the child object: 
get_oo_object

This is better to avoid using AEDT API directly.

One additional request, did you check if the property modification is possible with SetPropValue?

maxcapodi78 (approved):

lorenzovecchietti:
> One additional request, did you check if the property modification is possible with SetPropValue?

With `SetPropValue` I think we are setting one value one at a time. This can be a problem because order becomes important. For example, let us put ourselves in this situation:
immagine

Let's say I want to change `+X` to `'Percentage Offset'` and `-X` to `'Absolute Position'`. If I change `-X` first, I get an error (rightly so)
immagine
If I change `+X` first, everything works.

That's why I'm using `ChangeProperty`, I wanted to avoid tanking into account that complexity. Am I missing something here?

For sure that does not apply to` change_region_coordinate_system` and I'll update that.

maxcapodi78 (approved):

PR #3268, opened 2023-07-13
refactored downloads.py

maxcapodi78:


flamingto (approved):

Samuelopez-ansys (approved):

pyaedt/downloads.py
```diff
@@ -32,15 +31,15 @@ def _get_file_url(directory, filename=None):
         return EXAMPLE_REPO + "/".join([directory, filename])
 
 
-def _retrieve_file(url, filename, directory, destination=None):
+def _retrieve_file(url, filename, directory, destination=None, local_paths=[]):
```
MaxJPRey, line 34: I would suggest to not use list as default argument.
```suggestion
def _retrieve_file(url, filename, directory, destination=None, local_paths=None):
```

pyaedt/downloads.py
```diff
@@ -83,22 +82,26 @@ def _retrieve_file(url, filename, directory, destination=None):
             os.system(command)
     else:
         _, resp = urlretrieve(url, local_path)
-    return local_path
+    local_paths.append(local_path)
 
 
-def _retrieve_folder(url, directory, destination=None):
+def _retrieve_folder(url, directory, destination=None, local_paths=[]):
```
MaxJPRey, line 88:
```suggestion
def _retrieve_folder(url, directory, destination=None, local_paths=None):
```

PR #3274, opened 2023-07-14
show bitmap image option when creating Nport component

gmalinve:


Samuelopez-ansys (approved):
Could you add an example in the docstring?

PR #3278, opened 2023-07-14
Load stackup bug fix

ring630:


MaxJPRey (approved):
@ring630 I would just suggest to add a unit test but otherwise, LGTM.

ring630:
> @ring630 I would just suggest to add a unit test but otherwise, LGTM.

Will do

MaxJPRey (approved):
LGTM. Thanks for adding the test.

PR #3282, opened 2023-07-17
set temp dir

gmalinve:


Samuelopez-ansys (commented):
@gmalinve This method should not be at desktop.py level?

@maxcapodi78

gmalinve:
@Samuelopez-ansys I thought so but in Design.py we already have the property temporary_directory (to get the temp dir path) so I thought to put it here too :)

maxcapodi78 (approved):

PR #3287, opened 2023-07-18
improved grpc plugin call

maxcapodi78:


Alberto-DM (approved):
LGTM. This was a nice catch!

PR #3290, opened 2023-07-19
Fix get sparameter in circuit

Samuelopez-ansys:
app.excitations in circuit returns a dictionary. These methods and property were not working.


pyaedt/modeler/calculators.py
```diff
@@ -320,7 +320,7 @@ def find_waveguide(self, freq, units="GHz"):  # pragma: no cover
         freq : float
             Operational frequency.
         units : str
```
MaxJPRey, line 322:
```suggestion
        units : str, optional
```

MaxJPRey (approved):
LGTM.

PR #3293, opened 2023-07-19
expose props for face CS

gmalinve:


MaxJPRey (approved):
Maybe just add a small unit test to make sure that the CS is correct when accessing the props with the modeler.

maxcapodi78 (approved):

PR #3296, opened 2023-07-19
Sherlock example meshing settings, solve, post

flamingto:
I updated the original Sherlock example with global mesh settings, solve and post-processing step. The original one seems to have convergence issue.


examples/04-Icepak/Sherlock_Example.py
```diff
@@ -181,6 +194,30 @@
 
 ipk.assign_priority_on_intersections()
 
+###############################################################################
+# Compute power Budget
```
MaxJPRey, line 198:
```suggestion
# Compute Power Budget
```

MaxJPRey: or:
```suggestion
# Compute power budget
```

examples/04-Icepak/Sherlock_Example.py
```diff
@@ -181,6 +194,30 @@
...
+###############################################################################
+# Get Solution Data and Plot
+# ~~~~~~~~~~~~~~~~~~~~~~~
+
+face_u10 = [i.id for i in ipk.modeler["COMP_U10"].faces]
+plot1 = ipk.post.create_fieldplot_surface(face_u10, "SurfTemperature")
+plot1.export_image(os.path.join(ipk.working_directory, "temperature_u10.jpg"))
+face_u9 = [i.id for i in ipk.modeler["COMP_U9"].faces]
+plot2 = ipk.post.create_fieldplot_surface(face_u9, "SurfPressure")
+
```
MaxJPRey, line 219: Maybe we could export the image the same way we export the image for the temperature around `u10`.

examples/04-Icepak/Sherlock_Example.py
```diff
@@ -181,6 +194,30 @@
...
+###############################################################################
+# Compute power Budget
+# ~~~~~~~~~~~~~~~~~~~~~~~
+
+power_budget, total = ipk.post.power_budget("W" )
+print(total)
+
+###############################################################################
+# Compute power Budget
+# ~~~~~~~~~~~~~~~~~~~~~~~
```
maxcapodi78, line 206: all title should have a number of simbols under it equal to the size of the line
```suggestion
# Compute power Budget
# ~~~~~~~~~~~~~~~~~~~~
```

maxcapodi78 (changes requested):
Please apply to all titles my suggestion

examples/04-Icepak/Sherlock_Example.py
PipKat: Also, the first letter of the first word in a title is capitalized, but all other letters in the title that are not proper nouns are lower case. Here, "Budget" should be "budget".

PipKat (approved):
Looks great to me!

maxcapodi78 (approved):

PR #3304, opened 2023-07-21
splitted long Emit test in multiple smaller tests

maxcapodi78:


_unittest/test_26_emit.py
```diff
@@ -996,20 +995,83 @@ def test_interference_script(self):
...
+        for band in bands:
+            for child in band.children:
+                if "L2 P(Y)" in band.node_name:
+                    band.enabled = True
+                else:
+                    band.enabled = False
+
+        # Enable HR-DSSS Ch 1-13 Wifi band
+        bands = rad3.bands()
+        for band in bands:
```
jsalant22, line 1043: This code is now duplicated in 3 tests and is merely used to set up a simple project for the actual testing. We should just save an aedt file with this project configuration and then load it for each of these tests.

jsalant22: @maxcapodi78 If you want, myself or @madelinecoleman1 can create the project and make this change.

MaxJPRey (approved):
LGTM.

_unittest/test_26_emit.py
maxcapodi78: please do it. I'm doing this fix because is causing random crash on build machine. Tests too long can cause issue in build machine because AEDT is not multithread and could run in timeout

PR #3327, opened 2023-07-26
improve ut speed

maxcapodi78:


_unittest/test_22_Circuit_DynamicLink.py
```diff
@@ -210,31 +210,34 @@ def test_09_setup(self):
...
+        q3d = Q3d(self.q3d, specified_version=desktop_version)
 
         q3d_comp = self.aedtapp.modeler.schematic.add_subcircuit_dynamic_link(
             q3d, solution_name="Setup1 : LastAdaptive"
         )
         assert q3d_comp
         assert len(q3d_comp.pins) == 4
-        hfss = Hfss(proj_path, specified_version=desktop_version)
+
+    @pytest.mark.skipif(is_linux, reason="Skipped because Desktop is crashing")
```
MaxJPRey, line 233: @maxcapodi78 Is it crashing because of a bug in PyAEDT or because of a limitation in AEDT?

_unittest/test_41_3dlayout_modeler.py
```diff
@@ -509,7 +509,7 @@ def test_19A_analyze_setup(self):
     @pytest.mark.skipif(is_ironpython, reason="Not supported with IronPython")
     def test_19B_analyze_setup(self):
         assert self.solve.mesh.generate_mesh("Setup1")
-        assert self.solve.analyze_setup("Setup1")
+        assert self.solve.analyze_setup("Setup1", num_cores=1)
```
MaxJPRey, line 512: Could we raise the number of cores to 2 to increase the test speed even more?

MaxJPRey (approved):
LGTM.

pyaedt/edb_core/edb_data/simulation_configuration.py
```diff
@@ -261,7 +261,7 @@ def use_pyaedt_cutout(self, value):
 
     @property
     def use_default_cutout(self):  # pragma: no cover
-        """Whether the default EDB cutout or a new PyAEDT cutout is used.
+        """Whether the default EDB cutout or a new PyAEDT cutout is used. Default is ``False`` to use new pyaedt cutout.
```
PipKat, line 264:
```suggestion
        """Whether to use the default EDB cutout. The default is ``False``, in which case
        a new PyAEDT cutout is used.
```

_unittest/test_41_3dlayout_modeler.py
maxcapodi78: ideally yes. The problem is that build machines has only 8 cores and we use 8 concurrent pytest. I'm worried about some bottlenecks

PR #3338, opened 2023-07-28
Nitin refactoring scalar field value for objects

nnetake:
In this PR I have refactored the method "get_scalar_field_value" to expand its capability to use for specific objects and types of object like "surface", "volume" or "point". We can also query value on the adjacent side. 
I have refactored a way to read the final value from a calculator. Instead of writing a file and reading it, I am using the native command. Please review. Thanks!

I have added one solved test project and tests for the above scenarios.


lorenzovecchietti (approved):
Great enhancement

MaxJPRey (approved):
This is great @nnetake .

PR #3342, opened 2023-07-29
Desktop refactoring - unit test refactoring

Alberto-DM:
This is a draft pull request, to test the unit tests on server.
DO NOT REVIEW this PR.

Big refactoring of the `Desktop` class to achieve the following, among other things:
- having only one `Desktop` instance for each opened AEDT, this instance is referenced by all the applications.
- better management of the `Desktop` class when it is instantiated directly (i.e. `d=Desktop()`) or from an app (i.e. `hfss=Hfss()`).
- Desktop class has now the ground structure to manage multiple desktop sessions, so that we are ready when it will be possible to launch multiple AEDTs.
- Settings like `machine`, `port`, `non_graphical`, `aedt_process_id`, `is_student` are now `Desktop` attributes (since they are Desktop specific) and are removed from `settings`.

The  `Desktop` class refactoring required also a unit test refactoring:
- the unit tests are now using the `pytest` fixtures and do not use the setup and teardown method used for compatibility with Ironpython.
- the unit tests are consequently not compatible with IronPython. 
- IronPython unit tests are now few tests aiming to ensure that the several apps start correctly and all the modules are loaded. 
- Retry mechanism: reruns all the tests inside a class if any test inside the class fails (implemented using pytest-retry-class).
- Timeout mechanism: a unit test is aborted if it hangs too long (implemented using pytest-timeout)


maxcapodi78 (approved):
GREAT REFACTORING

maxcapodi78:
Coverage will be lower because we need to merge the 2 coverage of 2 Builds (Solvers and CI)

Samuelopez-ansys (approved):

MaxJPRey:
> Coverage will be lower because we need to merge the 2 coverage of 2 Builds (Solvers and CI) @MaxJPRey

Sure, I will work on the unification of those different coverage results.

PR #3343, opened 2023-07-31
id() is a fundamental build-in. Use PEP8 single trailing underscore convention instead.

MaxJPRey:
Some 'id' variables are still present in the code. It does not affect the behavior of the program. We will remove them iteratively.


Samuelopez-ansys (approved):
LGTM

PR #3347, opened 2023-07-31
Remove unused method _parse_signal_layer_properties().

MaxJPRey:
I don't think this method is used anymore.
The protected attribute `_parse_signal_layer_properties` seems to be used instead.


maxcapodi78 (approved):
LGTM

PR #3350, opened 2023-08-01
Add tangential

maxcapodi78:


pyaedt/maxwell.py
```diff
@@ -2743,6 +2743,132 @@ def assign_layout_force(
...
+        coordinate_system : str, optional
+            Coordinate system to use for UV vector.
+        origin : list, optional
+            Origin of the UV vector. Default is ``None`` to take the bottom left vertex.
+        u_pos : list, optional
+            Direction of the U vector. Default is ``None`` to take the top left vertex.
+        reverse : bool, optional
+            Whether if the vector is reversed or not. Default is ``False``.
+        bound_name : str, optional
+            Name of the end connection boundary. The default is ``None``, in which case the
```
PipKat, line 2783:
```suggestion
        x_component_real : float, str, optional
            X component value real part. The default is ``0``.
        x_component_imag : float, str, optional
            X component value imaginary part. The default is ``0``.
        y_component_real : float, str, optional
            Y component value real part. The default is ``0``.
        y_component_imag : float, str, optional
            Y component value imaginary part. The default is ``0``.
        coordinate_system : str, optional
            Coordinate system to use for the UV vector.
        origin : list, optional
            Origin of the UV vector. The default is ``None`, in which case the bottom
            left vertex is used.
        u_pos : list, optional
            Direction of the U vector. The default is ``None``, in which case the top
            left vertex is used.
        reverse : bool, optional
            Whether the vector is reversed. The default is ``False``.
        bound_name : str, optional
            Name of the end connection boundary. The default is ``None``, in which case the
```

MaxJPRey (approved):
LGTM.

IreneWoyna:
The method works fine, I wanted to add assign_zero_tangential_h_field support for Magnetostatic as well, but not sure if it worked.

PR #3365, opened 2023-08-02
Feat: add meilisearch index actions with version compactible.

Revathyvenugopal162:
- [x] update the action to scrap whenever the docs changes using pymeilisearch
- [x] update the index according to version


doc/source/conf.py
```diff
@@ -11,7 +11,7 @@
 import numpy as np
 import json
 from sphinx_gallery.sorting import FileNameSortKey
-from ansys_sphinx_theme import ansys_favicon, get_version_match, pyansys_logo_black
+from ansys_sphinx_theme import ansys_favicon, get_version_match, pyansys_logo_black, con
```
Revathyvenugopal162, line 14:
```suggestion
from ansys_sphinx_theme import ansys_favicon, get_version_match, pyansys_logo_black
```

.github/workflows/full_documentation.yml
```diff
@@ -89,3 +90,56 @@ jobs:
           cname: ${{ env.DOCUMENTATION_CNAME }}
           token: ${{ secrets.GITHUB_TOKEN }}
           python-version: ${{ matrix.python-version }}
+
+  
+  doc-index-stable:
```
jorgepiloto, line 95: Let us parametrize this job as two different API references are required. Also, make migrate this to use ansys/actions.

jorgepiloto (changes requested):

.github/workflows/full_documentation.yml
MaxJPRey: That would be great if we can reuse generic actions.

.github/workflows/full_documentation.yml
```diff
@@ -89,3 +90,54 @@ jobs:
...
+          echo "VERSION_MEILI=$VERSION_MEILI" >> $GITHUB_ENV
+
+      - name: "Deploy the stable documentation index for PyAEDT API"
+        uses: ansys/actions/doc-deploy-index@v4
+        with:
+          cname: ${{ env.DOCUMENTATION_CNAME }}/version/${{ env.VERSION }}
+          index-name: pyaedt-v${{ env.VERSION_MEILI }}
+          host-url: ${{ vars.MEILISEARCH_HOST_URL }}
+          api-key: ${{ env.MEILISEARCH_API_KEY }}
+          pymeilisearchopts: --stop_urls \"EDBAPI\"
```
MaxJPRey, line 133: Maybe add a comment here for later maintenance.
Some people might not be familiar with those pymeiliserach options and the consequences on EDB API.

MaxJPRey (approved):
LGTM.

Revathyvenugopal162:
Thank you for the review @MaxJPRey  and @jorgepiloto , merging!

PR #3385, opened 2023-08-07
Improve split options

gmalinve:


pyaedt/modeler/cad/Modeler.py
```diff
@@ -2773,21 +2773,32 @@ def convert_to_selections(self, object_id, return_list=False):
...
         ----------
         objects : str, int, or list
             One or more objects to split. A list can contain
             both strings (object names) and integers (object IDs).
-        plane : str
+        plane : str, optional
             Coordinate plane of the cut or the Application.PLANE object.
             Choices for the coordinate plane are ``"XY"``, ``"YZ"``, and ``"ZX"``.
+            If neither plane nor tool parameters are provided, the plane tool option is used.
+            The default value is ``"XY"``.
```
PipKat, line 2788:
```suggestion
            Coordinate plane of the cut or the ``Application.PLANE`` object.
            The default is ``None``, in which case the XY coordinate plane is
            used. Options for the coordinate plane are ``"XY"``, ``"YZ"``, and
            ``"ZX"``.  If plane or tool parameters are not provided, the plane
            tool option is used.
```
We try to list/explain the default first and then provide the options in alphabetical order.

pyaedt/modeler/cad/Modeler.py
```diff
@@ -2773,21 +2773,32 @@ def convert_to_selections(self, object_id, return_list=False):
...
         sides : str
             Which side to keep. Options are ``"Both"``, ``"PositiveOnly"``,
             and ``"NegativeOnly"``. The default is ``"Both"``, in which case
             all objects are kept after the split.
```
PipKat, line 2792:
```suggestion
            Which side to keep. The default is ``"Both"``, in which case
            all objects are kept after the split. Options are ``"Both"``,
            ``"NegativeOnly"``, and ``"PositiveOnly"``.
```

pyaedt/modeler/cad/Modeler.py
gmalinve: Hi @PipKat ,
The PR is still in draft so I've made some changes since you commented.
Please take a look now.
Thanks!

Samuelopez-ansys (approved):
LGTM

PR #3388, opened 2023-08-08
disabling examples post

maxcapodi78:
Those examples continue to fails on documentation build. Added environment variable to avoid disabling them


MaxJPRey (approved):

MaxJPRey:
LGTM.

jsalant22:
@maxcapodi78 Is there an error message of any kind from the failed doc builds? I know the UI aspects needed to be disabled, but it looks like maybe there's an issue with pyaedt finding our backend API too.

maxcapodi78:
@jsalant22 it simply hangs after the simulation has finished. it looks to me that the problem is in the methods defined into the example after the simulation is performed. The API commands looks good to me

jsalant22:
@maxcapodi78 Thanks. If there's a way for us to dig into these issues some more and try building the docs on a branch, I'd be happy to help.

maxcapodi78:
@jsalant22  I would suggest you try first running the examples in your machine in non-graphical (this is how Documentation builder works), then try to build documentation on your machine. this is how I typically debug examples on my laptop

PR #3396, opened 2023-08-09
Fix CreateAedtApplication argtypes for 24.1

myoung301:
A change was made to the arguments to CreateAedtApplication for 24.1 so that the port can be a port (int) or object.


pyaedt/generic/grpc_plugin_dll.py
```diff
@@ -59,7 +59,19 @@ def SetPyObjCalbacks(CreateAedtObj, CreateAedtBlockObj, GetAedtObjId):
     AedtAPI.SetPyObjCalbacks(callbackToCreateObj, callbackCreateBlock, callbackGetObjID)
 
 
-AedtAPI.CreateAedtApplication.argtypes = c_wchar_p, c_int, c_bool, c_bool
+# Find the version of AEDT from product info file
```
myoung301, line 62: @maxcapodi78 I'm not sure if there is a better way to get the version information here in the plugin initialization, but we need some way to handle the difference in argument types between 23R2 and 24R1.

myoung301: @Samuelopez-ansys - can you review and merge this PR if it looks good to you? It's holding up some related development.

myoung301: @jsalant22 Maybe you can take a look at this PR? Thanks!

jsalant22 (approved):

PR #3400, opened 2023-08-10
Core loss fix

IreneWoyna:
"Consider core loss effect on field" is now disabled by default, changed the documentation accordingly


pyaedt/maxwell.py
```diff
@@ -1729,6 +1729,308 @@ def export_element_based_harmonic_force(
         self.odesign.ExportElementBasedHarmonicForce(output_directory, setup_name, freq_option, f1, f2)
         return output_directory
 
+    @pyaedt_function_handler()
+    def heal_objects(
```
gmalinve, line 1733: @IreneWoyna please merge latest master into this branch.
heal_objects and simplify_objects have now been moved to modeler.py.
Thanks!

PR #3406, opened 2023-08-11
Fix/issue 3391

lorenzovecchietti:
I've added a few comments to explain the modifications I made.

The only needed modification is https://github.com/ansys/pyaedt/pull/3406/files#r1291539564.

I added the other two just for robustness. The fundamental problem is that accessing those function was not the intended usage if I'm understanding correctly the code.

The setter of thermal conductivity acts on the internal class (https://github.com/ansys/pyaedt/blob/ea3b6a16f20ba8c580bbf3c708341655dfac7935/pyaedt/modules/Material.py#L1422) AND on the props dictionary (https://github.com/ansys/pyaedt/blob/ea3b6a16f20ba8c580bbf3c708341655dfac7935/pyaedt/modules/Material.py#L1427).

But using the public thermal_conductivity property, the user can still directly modify the internal class (and without the two fixes this was leading to an incorrect assignment of the values). This is what led to this issue https://github.com/ansys/pyaedt/issues/3391.

The problem with accessing directly the thermal_condictivity class is that the props dictionary is not updated, so this should be not allowed or the logic to update the props dictionary should be implemented. @maxcapodi78 @Samuelopez-ansys  What do you suggest?


pyaedt/modules/Material.py
```diff
@@ -291,9 +293,15 @@ def type(self, type):
         if self._type == "simple":
             self._property_value = [self._property_value[0]]
         elif self._type == "anisotropic":
-            self._property_value = [self._property_value[0] for i in range(3)]
+            try:
+                self._property_value = [self._property_value[i] for i in range(3)]
+            except IndexError:
+                self._property_value = [copy.deepcopy(self._property_value[0]) for i in range(3)]
```
lorenzovecchietti, line 299: In the old version: if I modify the property_value, and then I modify the type, the property_value list was being overwritten and set to property_value[0].

Also: if I modify the type before modifying the property value, the same class was being referenced for three times (or 9 in the case of tensor type). So all the three (or nine) value were the same.

Samuelopez-ansys:
@maxcapodi78 @lorenzovecchietti 

I finished the issue when we updated the value property of the material, a collateral I had also to modified a bit the _update_properties method. Please test it, just in case I missed some combination. Thanks

Samuelopez-ansys (approved):
LGTM

pyaedt/modules/Material.py
```diff
@@ -196,10 +197,11 @@ class Dataset(object):
 class BasicValue(object):
     """Manages thermal and spatial modifier calculations."""
 
-    value = None
-    dataset = None
-    thermalmodifier = None
-    spatialmodifier = None
+    def __init__(self):
```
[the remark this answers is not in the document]
MaxJPRey: Agreed.

pyaedt/modules/Material.py
```diff
@@ -1417,15 +1409,16 @@ def thermal_conductivity(self):
...
-            self._update_props("thermal_conductivity", value)
+        self.physics_type = ["Electromagnetic", "Thermal", "Structural"]
+        self._thermal_conductivity.value = value
+        #
+        # if isinstance(value, list):
+        #     self._thermal_conductivity.type = "anisotropic"
+        #     self._update_props("thermal_conductivity", value)
+        # else:
+        #     self._thermal_conductivity.type = "simple"
+        #     self._update_props("thermal_conductivity", value)
```
MaxJPRey, line 1421:
```suggestion
```

maxcapodi78 (approved):
LGTM

PR #3445, opened 2023-08-21
add support to choose the notebook dir using env variable

kmahajan-cadfem:
- if user want to open the jupyter lab in any other directory using just button that are available, it is not possible
- added support for settings a environment variable with which user can set one directory for all notebooks and still use the buttons in AEDT Desktop
- `AEDT_NOTEBOOK_DIR` is the name of the variable


Samuelopez-ansys (approved):
Good improvement!

MaxJPRey:
> @MaxJPRey @jorgepiloto Why is Labeler failing? Thanks

@Samuelopez-ansys It is because, the PR is coming from a fork.

PR #3472, opened 2023-08-24
Two fixes for N-1 behavior in 2023 R2

myoung301:
- Throw error for N-1 runs with multiple interferers specified because this wasn't supported in 23R2
- Set max_simultaneous_interferers to 1 to properly disable N-1 when only 1 interferer is specified

### DIFF
--- a/pyaedt/emit_core/results/revision.py
+++ b/pyaedt/emit_core/results/revision.py
@@ -159,6 +159,11 @@ class Revision:
                     raise ValueError("The domain must not have channels specified.")
         self._load_revision()
         engine = self.emit_project._emit_api.get_engine()
+        if self.emit_project._aedt_version < "2024.1":
+            if len(domain.interferer_names) == 1:
+                engine.max_simultaneous_interferers = 1
+            if len(domain.interferer_names) > 1:
+                raise ValueError("Multiple interferers cannot be specified prior to AEDT version 2023 R2.")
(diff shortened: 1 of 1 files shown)


pyaedt/emit_core/results/revision.py
```diff
@@ -159,6 +159,11 @@ def run(self, domain):
                     raise ValueError("The domain must not have channels specified.")
         self._load_revision()
         engine = self.emit_project._emit_api.get_engine()
+        if self.emit_project._aedt_version < "2024.1":
+            if len(domain.interferer_names) == 1:
+                engine.max_simultaneous_interferers = 1
+            if len(domain.interferer_names) > 1:
+                raise ValueError("Multiple interferers cannot be specified prior to AEDT version 2023 R2.")
```
apollo-crum, line 166: @myoung301 "prior to 2032 R2"?  I'm guessing you meant 2024.1.  However, even that isn't accurate as we deferred the capability of specifying multiple specific tx's, so it is likely to be 2024.2 before we allow this.  But then again, none of that really matters since it isn't currently possible to hit that line (you can't specify multiple interferers in pyaedt and if you take a worst case domain that already has them specified it would run into the channels specified error on line 159).

All of which is to say, maybe you should change this, but only so as not to confuse the next person that works on this code.

apollo-crum (approved):

PR #3501, opened 2023-08-31
nitin feature get design settings

nnetake:
In this PR I have added a function to retrieve design settings based on prop names available in the object.
I have added UT for Icepak.


pyaedt/application/Design.py
```diff
@@ -3914,3 +3914,27 @@ def set_temporary_directory(self, temp_dir_path):
         """
         self.odesktop.SetTempDirectory(temp_dir_path)
         return True
+
+    @pyaedt_function_handler()
+    def get_design_settings(self):
+        """Get design settings for current aedt application
```
MaxJPRey, line 3920:
```suggestion
        """Get design settings for current AEDT application.
```

pyaedt/application/Design.py
```diff
@@ -3914,3 +3914,27 @@ def set_temporary_directory(self, temp_dir_path):
         """
         self.odesktop.SetTempDirectory(temp_dir_path)
         return True
+
+    @pyaedt_function_handler()
+    def get_design_settings(self):
```
MaxJPRey, line 3919: @nnetake Why not making it a property such as `design_settings`? Could it create conflicts with other attributes?

pyaedt/application/Design.py
nnetake: Corrected, thanks!

pyaedt/application/Design.py
```diff
@@ -3914,3 +3914,27 @@ def set_temporary_directory(self, temp_dir_path):
...
+        """Get design settings for current AEDT application.
+
+        Returns
+        -------
+        dict of valid design settings
```
PipKat, line 3924:
```suggestion
        dict
            Dictionary of valid design settings.
```

pyaedt/application/Design.py
```diff
@@ -3914,3 +3914,27 @@ def set_temporary_directory(self, temp_dir_path):
...
+        >>> oDesign.GetChildObject("Design Settings")
+        """
+        design_settings = self._odesign.GetChildObject("Design Settings")
+        prop_name_list = design_settings.GetPropNames()
+        design_settings_dict = {}
+        for prop in prop_name_list:
+            try:
+                design_settings_dict[prop] = design_settings.GetPropValue(prop)
+            except Exception as e:
+                design_settings_dict[prop] = "Could not retrieve the value"
```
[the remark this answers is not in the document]
nnetake: Done, thanks!

pyaedt/application/Design.py
maxcapodi78: @nnetake I think @MaxJPRey intended that the method name could be renamed design_settings from get_design_settings

nnetake: okay sure. I will modify the method name. Thanks!

MaxJPRey: > @nnetake I think @MaxJPRey intended that the method name could be renamed design_settings from get_design_settings. I think makes sense

Exactly. Thanks for adding some clarification @maxcapodi78 .

_unittest/test_01_Design.py
```diff
@@ -26,11 +27,20 @@ def aedtapp(add_app):
     return app
 
 
+@pytest.fixture(scope="class")
```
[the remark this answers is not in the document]
nnetake: corrected, thanks!

_unittest/test_01_Design.py
```diff
@@ -26,11 +27,20 @@ def aedtapp(add_app):
...
+    app = add_app(
+        project_name="icepak_proj", design_name="ipk_test_design", application=Icepak, subfolder=test_subfolder
+    )
+    return app
+
+
 class TestClass:
     @pytest.fixture(autouse=True)
-    def init(self, aedtapp, local_scratch):
+    def init(self, aedtapp, local_scratch, icepak_design):
```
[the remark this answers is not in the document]
nnetake: corrected, thanks!

_unittest/test_01_Design.py
```diff
@@ -392,3 +402,6 @@ def test_39_load_project(self, desktop):
         self.aedtapp.close_project(name="new")
         aedtapp = desktop.load_project(new_project)
         assert aedtapp
+
+    def test_40_get_design_settings(self):
```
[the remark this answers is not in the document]
nnetake: added few keys to be asserted.

pyaedt/application/Design.py
```diff
@@ -3916,3 +3916,28 @@ def set_temporary_directory(self, temp_dir_path):
...
+        -------
+        dict
+            Dictionary of valid design settings.
+
+        References
+        ----------
+
+        >>> oDesign.GetChildObject("Design Settings")
+        """
+        design_settings = self._odesign.GetChildObject("Design Settings")
```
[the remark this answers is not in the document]
nnetake: Makes sense. I have added try/except block.

Samuelopez-ansys:
@nnetake Please could you add the missing tests? You can find them in the Files Changed section. Basicaly, there are 2 missing tests in two Try / excepts. If you can not test it because you can not reproduce it, try to find a design to reproduce it, if not you should add the line # pragam no cover to exclude this part. You can find some PyAEDT lines with this comment.

_unittest/test_01_Design.py
```diff
@@ -392,3 +393,12 @@ def test_39_load_project(self, desktop):
         self.aedtapp.close_project(name="new")
         aedtapp = desktop.load_project(new_project)
         assert aedtapp
+
+    def test_40_get_design_settings(self, add_app):
+        ipk = add_app(application=Icepak)
+        design_settings_dict = ipk.design_settings()
+        assert design_settings_dict
```
gmalinve, line 400: @nnetake maybe better isinstance(design_settings_dict, dict).

pyaedt/application/Design.py
```diff
@@ -3916,3 +3916,32 @@ def set_temporary_directory(self, temp_dir_path):
...
+
+        References
+        ----------
+
+        >>> oDesign.GetChildObject("Design Settings")
+        """
+        try:
+            design_settings = self._odesign.GetChildObject("Design Settings")
+        except Exception:
+            raise Exception("Could not use GetChildObject('Design Settings')")
```
gmalinve, line 3937: @nnetake return False if the exception is raised. 
Try to cover this line as well with the code and fix the returns in the docstring so if it fails return False.

pyaedt/application/Design.py
```diff
@@ -3916,3 +3916,32 @@ def set_temporary_directory(self, temp_dir_path):
...
+
+        prop_name_list = design_settings.GetPropNames()
+        design_settings_dict = {}
+        for prop in prop_name_list:
+            try:
+                design_settings_dict[prop] = design_settings.GetPropValue(prop)
+            except Exception:
+                design_settings_dict[prop] = "Could not retrieve the value."
```
gmalinve, line 3945: @nnetake same as above, return False

_unittest/test_01_Design.py
nnetake: Yeah makes sense.

pyaedt/application/Design.py
nnetake: Done. thanks!

pyaedt/application/Design.py
nnetake: In this case, I don't want to return False because it may happen that out of 20 it is not able to retrieve a few properties. This is a partial success and I have added a warning message here.

pyaedt/application/Design.py
```diff
@@ -3921,3 +3921,34 @@ def set_temporary_directory(self, temp_dir_path):
...
+
+        References
+        ----------
+
+        >>> oDesign.GetChildObject("Design Settings")
+        """
+        try:
+            design_settings = self._odesign.GetChildObject("Design Settings")
+        except Exception:  # pragma: no cover
+            self.logger.error("Could not execute GetChildObject('Design Settings').")  # pragma: no cover
```
Samuelopez-ansys, line 3942:
```suggestion
            self.logger.error("Could not execute GetChildObject('Design Settings').") 
```

pyaedt/application/Design.py
```diff
@@ -3921,3 +3921,34 @@ def set_temporary_directory(self, temp_dir_path):
...
+            return False  # pragma: no cover
```
Samuelopez-ansys, line 3943:
```suggestion
            return False
```

pyaedt/application/Design.py
```diff
@@ -3921,3 +3921,34 @@ def set_temporary_directory(self, temp_dir_path):
...
+
+        prop_name_list = design_settings.GetPropNames()
+        design_settings_dict = {}
+        for prop in prop_name_list:
+            try:
+                design_settings_dict[prop] = design_settings.GetPropValue(prop)
+            except Exception:  # pragma: no cover
+                self.logger.warning("Could not retrieve property value in design settings")  # pragma: no cover
```
Samuelopez-ansys, line 3951:
```suggestion
                self.logger.warning("Could not retrieve property value in design settings") 
```

pyaedt/application/Design.py
```diff
@@ -3921,3 +3921,34 @@ def set_temporary_directory(self, temp_dir_path):
...
+                design_settings_dict[prop] = "Could not retrieve the value."  # pragma: no cover
```
Samuelopez-ansys, line 3952:
```suggestion
                design_settings_dict[prop] = "Could not retrieve the value."
```

Samuelopez-ansys (changes requested):

Samuelopez-ansys (approved):

pyaedt/application/Design.py
```diff
@@ -3921,3 +3921,34 @@ def set_temporary_directory(self, temp_dir_path):
...
+            return False
+
+        prop_name_list = design_settings.GetPropNames()
+        design_settings_dict = {}
+        for prop in prop_name_list:
+            try:
+                design_settings_dict[prop] = design_settings.GetPropValue(prop)
+            except Exception:  # pragma: no cover
+                self.logger.warning("Could not retrieve property value in design settings")
+                design_settings_dict[prop] = "Could not retrieve the value."
```
lorenzovecchietti, line 3952: `None` might be better than a string when the property is not retrieved. In this way one can directly evaluate the dictionary value with `if design_settings_dict[prop]` to check whether the setting was correctly loaded.

pyaedt/application/Design.py
```diff
@@ -3921,3 +3921,34 @@ def set_temporary_directory(self, temp_dir_path):
...
+
+        References
+        ----------
+
+        >>> oDesign.GetChildObject("Design Settings")
+        """
+        try:
+            design_settings = self._odesign.GetChildObject("Design Settings")
+        except Exception:  # pragma: no cover
+            self.logger.error("Could not execute GetChildObject('Design Settings').")
```
lorenzovecchietti, line 3942: I would use a more general message. Like: 
`"Failed to retrieve design settings."`

But that's up to you.

gmalinve: @lorenzovecchietti @nnetake agree with Loreno here.

pyaedt/application/Design.py
```diff
@@ -3921,3 +3921,34 @@ def set_temporary_directory(self, temp_dir_path):
...
+            self.logger.error("Failed to retrieve design settings.")
+            return False
+
+        prop_name_list = design_settings.GetPropNames()
+        design_settings_dict = {}
+        for prop in prop_name_list:
+            try:
+                design_settings_dict[prop] = design_settings.GetPropValue(prop)
+            except Exception:  # pragma: no cover
+                self.logger.warning("Could not retrieve property value in design settings")
```
lorenzovecchietti, line 3951: One last thing. Maybe specify the property not retrieved. Like:
`"Could not retrieve \"{}\" property value in design settings.".format(prop)`

nnetake: ok thanks for this comment, I used f string previously but it was failing in the test so I removed it.

PR #3502, opened 2023-08-31
Fixed Analysis.AvailableVariations.variations method

Alberto-DM:
Fixed Analysis.AvailableVariations.variations method. 
Added an option to output the variations families as dict. 
Added unit tests


MaxJPRey (approved):
LGTM.

PR #3506, opened 2023-09-01
Fix installation script

Samuelopez-ansys:
Script compatible from 2022R2
Only one AEDT needed for the installation


maxcapodi78 (approved):
LGTM

PR #3518, opened 2023-09-05
Added additional control to rpc connection to avoid errors in PyAEDT

maxcapodi78:


Samuelopez-ansys (approved):
LGTM

pyaedt/desktop.py
```diff
@@ -922,6 +929,18 @@ def _initialize(
 
     def _init_grpc(self, non_graphical, new_aedt_session, version, student_version, version_key):
         self.logger.info("Launching AEDT using the gRPC plugin.")
+        if settings.remote_rpc_session:  # pragma: no cover
+            settings.remote_api = True
+            if not self.machine:
+                try:
+                    self.machine = settings.remote_rpc_session.server_name
```
Alberto-DM, line 936: Since remote_rpc_session is not a setting but an object, we should consider to take it out from settings, put in a new class (together with remote_rpc_session_temp_folder and remote_rpc_service_manager_port) and then import in general_methods so that we can import directly from pyaedt.

Alberto-DM (approved):
LGTM. Consider the comment for future improvement if you agree.

pyaedt/desktop.py
maxcapodi78: I agree

PR #3520, opened 2023-09-06
EDB Bug fix preventing None object causing crash

svandenb-dev:
Small bug fix to prevent crash when primitives objects are None.


MaxJPRey (approved):
LGTM.

Alberto-DM (approved):
LGTM

PR #3522, opened 2023-09-06
Changed version comparison from "222" to "221" 

Samuelopez-ansys:
This update ensures correct version formatting when the version number is less than 221.


MaxJPRey (approved):
Thanks for the fix @Samuelopez-ansys

isaacansys:
Looks good here, thank you @Samuelopez-ansys

PR #3524, opened 2023-09-06
Edb cutout improvements

maxcapodi78:
New Edb cutout supports 2 options for padstack cutout:
- simple pad check based on padstack center (inside or outside extent)
- complete pad check (bounding box inside or outside extent or partially)
Complex pad check is much slower but provides a behaviour more similar to original edb api cutout method
Default is simple pad check to keep back compatibility.


svandenb-dev (approved):

ring630 (commented):
LGTM

PR #3526, opened 2023-09-07
Add tolerance to padding checking

jonas-norlin:
When we change the region padding, there is a validation check made to ensure the settings has been applied as expected. In some cases, the check fails due to floating point error. For example, if we give a padding of "10.0000000001mm", what we get back to check might be "10mm", and then using a simple "==" check fails. 

With this PR we extract the actual number from the string, and check with the *check_numeric_equivalence*  function, with a relative tolerance of 1e-9.  

If the applied setting does not contain a number, such as the setting "Absolute Position", we still default to the same == check. Since the checking becomes quite involved, I have refactored a little bit how the check is made. I have also added the validation failures to the log message in case of error. There are some unit tests added to ensure the functionality.


gmalinve:
Hi @jonas-norlin,

Thanks for submitting this PR! 
I have a couple of general questions for you that may be useful for me to understand your implementation :)
First, is there a reason why you don't use the pyaedt_function_handler decorator in your methods?
Second, regarding your method _check_value_from_string were you aware that we already have two methods called decompose_variable_value (Variables.py) and format_decimals() (in DataHandlers) that could be right for you?
if this is the case I think we need to improve the format_decimals() method.
If they are not right I'd move your method _check_value_from_string in the DataHandlers and maybe rename it.

gmalinve:
also @jonas-norlin please consider this method: check_numeric_equivalence it may be useful in your case.

jonas-norlin:
> Hi @jonas-norlin,
> 
> Thanks for submitting this PR! I have a couple of general questions for you that may be useful for me to understand your implementation :) First, is there a reason why you don't use the pyaedt_function_handler decorator in your methods? Second, regarding your method _check_value_from_string were you aware that we already have two methods called decompose_variable_value (Variables.py) and format_decimals() (in DataHandlers) that could be right for you? if this is the case I think we need to improve the format_decimals() method. If they are not right I'd move your method _check_value_from_string in the DataHandlers and maybe rename it.

Hi Giulia, thanks for pointing out the pyaedt_function_handler, I have added it. 

I have moved the new method to Variables.py. I have tried to apply decompose_variable_value in this case 

> Thanks for submitting this PR!
> I have a couple of general questions for you that may be useful for me to understand your implementation :)
> First, is there a reason why you don't use the pyaedt_function_handler decorator in your methods?
> Second, regarding your method _check_value_from_string were you aware that we already have two methods called decompose_variable_value (Variables.py) and format_decimals() (in DataHandlers) that could be right for you?
> if this is the case I think we need to improve the format_decimals() method.
> If they are not right I'd move your method _check_value_from_string in the DataHandlers and maybe rename it.

Hi Giulia, thanks for pointing me in the right direction. I have applied pyaedt_function_handler on the new function. And thanks for pointing out the decompose_variable_value method, I was able to apply it. I also moved all the *generate_validation_errors* new functionality into the Variables.py and now is just using it in the modeler. I don't know exactly what format_decimals is doing, or if we need it in this case, it seems to be currently unused?

jonas-norlin:
> also @jonas-norlin please consider this method: check_numeric_equivalence it may be useful in your case.

Is there a reason to prefer that over the  math.isclose function ?

Alberto-DM:
> Is there a reason to prefer that over the math.isclose function ?

math.isclose is not available in IronPython. You should use check_numeric_equivalence. Thanks

gmalinve:
@jonas-norlin correct me if I'm wrong but the purpose of this PR is to check if two values (as strings) are equal to each other. Is it not enough to reuse the decompose_variable_value to split number and unit in a string? and then use the method check_numeric_equivalence to set the tolerance?

jonas-norlin:
> > Is there a reason to prefer that over the math.isclose function ?
> 
> math.isclose is not available in IronPython. You should use check_numeric_equivalence. Thanks

Ah I see, have changed to check_numeric_equivalence

jonas-norlin:
> @jonas-norlin correct me if I'm wrong but the purpose of this PR is to check if two values (as strings) are equal to each other. Is it not enough to reuse the decompose_variable_value to split number and unit in a string? and then use the method check_numeric_equivalence to set the tolerance?

Well yes, it's only that we also need to check variables which are not numbers like "10mm", but also settings like "Absolute Offset". The check should work for both the case of number and string setting. And ideally you want something informative in the log for which validations failed.

pyaedt/application/Variables.py
```diff
@@ -312,10 +311,30 @@ def generate_property_validation_errors(property_name, expected, actual):
...
+        List of property names
+    expected_settings : List[str]
+        List of the expected settings.
+    actual_settings: List[str]
+        List of actual settings.
+
+    Returns
+    -------
+    List[str]
+        A list of validation errors for the given ettings
```
MaxJPRey, line 332:
```suggestion
        A list of validation errors for the given settings.
```

pyaedt/application/Variables.py
```diff
@@ -312,10 +311,30 @@ def generate_property_validation_errors(property_name, expected, actual):
...
+    """From the given property names, expected settings and actual settings, return a list of validation errors.
+    If no errors are found, an empty list is returned. The validation of values such as "10mm" 
+    ensures that they are close to within a relative tolerance. 
+    For example an expected setting of "10mm", and actual of "10.000000001mm" will not yield a validation error.
+    For values with no numerical value, an equivalence check is made.
+
+    Parameters
+    ----------
+    property_names : List[str]
+        List of property names
```
MaxJPRey, line 323:
```suggestion
        List of property names.
```

pyaedt/application/Variables.py
jonas-norlin: fixed

pyaedt/application/Variables.py
jonas-norlin: fixed

PR #3530, opened 2023-09-07
Added Dark mode to plot in pyvista

maxcapodi78:
fixed bug in create_3d_components when multiple coordinate where present.


MaxJPRey (approved):
Nice to see the dark mode added.

PR #3550, opened 2023-09-12
Update interference_gui.py

jsalant22:
Fixes DE892295, DE892289, DE893060, DE893040, DE893034


jsalant22:
@myoung301 Unfortunately this isn't self-contained since I also needed to tweak the ui file that it uses. That was apparently moved to the ansys/example-data repo which I'm trying to get write access too. Until then, you won't be able to test the warning label/message, but I think everything else can be verified. 

If you want to manually change your ui file, I can walk you through where it gets downloaded to and the changes that are needed. Basically just need to add this after Line 75:

`
    <item>
     <widget class="QLabel" name="warnings">
      <property name="text">
       <string>Warnings</string>
      </property>
      <property name="alignment">
       <set>AlignLeft</set>
      </property>
     </widget>
    </item>`

examples/07-EMIT/interference_gui.py
```diff
@@ -204,25 +242,50 @@ def open_file_dialog(self):
...
+            emit_designs = []
+            self.populating_dropdown = True
             self.design_name_dropdown.clear()
-            for d in emit_designs:
-                self.design_name_dropdown.addItem(d.GetName())
-                
-            self.design_name_dropdown.setEnabled(True)
+            self.populating_dropdown = False
+            for d in designs:
+                design_type = get_pyaedt_app(desktop_proj.project_name, d).design_type
```
myoung301, line 264: Is this pretty fast? I'm not sure if there is much overhead creating an App object, but there is some for EMIT. Can we avoid creating the app objects by using desktop.design_type()?

myoung301 (changes requested):

examples/07-EMIT/interference_gui.py
jsalant22: @myoung301 Unfortunately I don't think so. From what I saw, `desktop.design_type()` just returns the design type for the active design. I couldn't find a way to get the design type based on the design names. (doesn't mean there isn't something I missed)

jsalant22: Here's the other PR: https://github.com/ansys/example-data/pull/188

examples/07-EMIT/interference_gui.py
```diff
@@ -235,9 +298,27 @@ def open_file_dialog(self):
...
+        # check for at least 2 radios
+        radios = self.emitapp.modeler.components.get_radios()
+        self.warning_label.setHidden(True)
+        if len(radios) < 2:
+            self.warning_label.setText("Warning: The selected design must contain at least two radios.")
+            self.warning_label.setHidden(False)
+
     ###############################################################################
-    # Enable radio specific proteciton levels
+    # Enable radio specific protection levels
```
PipKat, line 321:
```suggestion
    # Enable radio-specific protection levels
```

examples/07-EMIT/interference_gui.py
```diff
@@ -235,9 +298,27 @@ def open_file_dialog(self):
...
     # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     # Activate radio selection dropdown and initialize dictionary to store protection levels
     # when the radio specific level dropdown is checked.
```
PipKat, line 324:
```suggestion
    # when the radio-specific level dropdown is checked.
```

examples/07-EMIT/interference_gui.py
myoung301: @jsalant22 - The function says it takes a project name and design name and seems to imply it can return the type for non-active projects and designs:
https://github.com/ansys/pyaedt/blob/7b96bc5df97390231df4ef118f0fd687948da83b/pyaedt/desktop.py#L1254-L1264

myoung301 (approved):

examples/07-EMIT/interference_gui.py
```diff
@@ -235,19 +298,37 @@ def open_file_dialog(self):
             self.radio_specific_levels.setEnabled(True)
             self.protection_results_btn.setEnabled(True)
             self.interference_results_btn.setEnabled(True)
-    
+
+    ###############################################################################
+    # Change design selection
+    # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+    # Refresh the warning messages when the selected design changes
```
PipKat, line 305:
```suggestion
    # ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    # Refresh the warning messages when the selected design changes.
```

PipKat, line 305: Just realized that a sentence in a new section of an example should have concluding punctuation. Would you please check these to add a period if needed? Thanks.

PR #3565, opened 2023-09-14
Get polygon points doc string

svandenb-dev:
doc string fix on get_polygon_points command


pyaedt/edb_core/layout.py
```diff
@@ -259,12 +259,16 @@ def get_polygon_points(self, polygon):
         Parameters
         ----------
         polygon :
-            Name of the polygon.
+            class: `pyaedt.edb_core.edb_data.primitives_data.EDBPrimitives`
```
PipKat, line 262: @svandenb-dev Shouldn't the class be used as the data type? I've not see anything formatted like this before.

pyaedt/edb_core/layout.py
```diff
@@ -259,12 +259,16 @@ def get_polygon_points(self, polygon):
...
 
         Returns
         -------
         list
-            List of doubles.
+            List of tuples. Each tuple provides x, y point coordinate. If the length of two consecutives tuples
+             from the list equals 2, a segment is defined. The first tuple defines the starting point while the second
+             tuple the ending one. If the length of one tuple equals one, that means a polyline is defined and the value
+             is giving the arc height. Therefore to polyline is defined as starting point for the tuple
+             before in the list, the current one the arc height and the tuple after the polyline ending point.
```
PipKat, line 271:
```suggestion
            List of tuples. Each tuple provides an x, y point coordinate. If the length of two consecutives tuples
             from the list equals two, a segment is defined. The first tuple defines the starting point, while the second
             tuple defines the ending point. If the length of one tuple equals one, that means a polyline is defined and
             that the value is giving the arc height. Therefore to polyline is defined as starting point for the tuple
             before in the list, the current one the arc height and the tuple after the polyline ending point.
```
The last sentence doesn't make sense--would you please revisit? Thanks!

PipKat (approved):

maxcapodi78 (approved):

PR #3593, opened 2023-09-19
fix import_dxf + tests

gmalinve:


Samuelopez-ansys (approved):

ngkazu:
Thanks @gmalinve @Samuelopez-ansys

PR #3595, opened 2023-09-19
better handling if users cancel file dlgs

jsalant22:
Don't error out if users cancel a file dlg, just do "nothing"


myoung301:
@jsalant22 What caused the change in the test result values?

jsalant22:
> @jsalant22 What caused the change in the test result values?

@myoung301 The interference tab was using PowerAtRx instead of normal EMI Margin.

myoung301 (approved):
Ok - I understand now. Thanks! Looks good.

PR #3600, opened 2023-09-20
fix split method

gmalinve:


Samuelopez-ansys (approved):
LGTM

pyaedt/modeler/cad/Modeler.py
```diff
@@ -3500,43 +3500,62 @@ def split(self, objects, plane=None, sides="Both", tool=None, split_crossing_obj
...
                 selections = ["NAME:Selections", "Selections:=", objects, "NewPartsModelFlag:=", "Model"]
             elif tool and not plane:
-                objs_selection = self.convert_to_selections(tool, False)
-                if isinstance(tool, str) or isinstance(tool, int):
-                    obj_name = objs_selection
-                    obj = [f for f in self.object_list if f.name == objs_selection][0]
+                if isinstance(tool, str):
+                    obj = [f for f in self.object_list if f.name == tool][0]
+                    obj_name = obj.name
+                    if isinstance(obj, Object3d) and obj.object_type != "Line":
```
MaxJPRey, line 3515: We might have to check if the object is a `Point` too. I don't remember if the `Point` can be an instance of `Object3d`.
Not mandatory though, as people will probably not try to split a `Point`.

MaxJPRey (approved):
LGTM.

pyaedt/modeler/cad/Modeler.py
gmalinve: @MaxJPRey points are not objects so this case would fall in another check but is it possible to split by point?

PR #3630, opened 2023-09-26
Fixed error in edb net plot.

Alberto-DM:
Fixed an error in edb.net plot function which was preventing the correct axis scale of the plot. Now the plot is automatically scaled based on the board dimensions.
Also the x and y axis of the plot were scaled differently and this resulted in plots that were out of proportion. This has been fixed too.


maxcapodi78 (approved):
LGTM

PR #3634, opened 2023-09-26
Remove Delcross from reference

### LINKED ISSUE #1704: Remove dependencies on libraries in $ANSYSEM_ROOT_PATH/Delcross or remove warning about this folder

If $ANSYSEM_ROOT_PATH/Delcross is not in $LD_LIBRARY_PATH on Linux, PyAEDT will issue a warning.  This folder contains, among other libraries, Qt 5 libraries that may prevent the use of later Qt 5 versions by downstream apps that depend on PyAEDT.  In testing with such an app I found that at least EDB-only usage of PyAEDT seems to work without the Delcross folder in $LD_LIBRARY_PATH.

_No response_

_No response_

maxcapodi78:
Fix #1704

### DIFF
--- a/pyaedt/generic/clr_module.py
+++ b/pyaedt/generic/clr_module.py
@@ -30,12 +30,12 @@ if is_linux and cpython:  # pragma: no cover
         json_file = os.path.abspath(os.path.join(pyaedt_path, "misc", "pyaedt.runtimeconfig.json"))
         load("coreclr", runtime_config=json_file, dotnet_root=os.environ["DOTNET_ROOT"])
         print("DotNet Core correctly loaded.")
-        if "Delcross" not in os.getenv("LD_LIBRARY_PATH", "") or "mono" not in os.getenv("LD_LIBRARY_PATH", ""):
-            warnings.warn("LD_LIBRARY_PATH needs to be setup to use pyaedt.")
-            warnings.warn("export ANSYSEM_ROOT222=/path/to/AnsysEM/v222/Linux64")
-            msg = "export LD_LIBRARY_PATH="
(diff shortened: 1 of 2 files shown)


pyaedt/generic/clr_module.py
```diff
@@ -30,11 +30,11 @@
         json_file = os.path.abspath(os.path.join(pyaedt_path, "misc", "pyaedt.runtimeconfig.json"))
         load("coreclr", runtime_config=json_file, dotnet_root=os.environ["DOTNET_ROOT"])
         print("DotNet Core correctly loaded.")
-        if "Delcross" not in os.getenv("LD_LIBRARY_PATH", "") or "mono" not in os.getenv("LD_LIBRARY_PATH", ""):
+        if "mono" not in os.getenv("LD_LIBRARY_PATH", ""):
             warnings.warn("LD_LIBRARY_PATH needs to be setup to use pyaedt.")
             warnings.warn("export ANSYSEM_ROOT222=/path/to/AnsysEM/v222/Linux64")
```
isaacansys, line 35: Should update this line for 232 as well, and maybe consider displaying the original message if on <232?

isaacansys (commented):
These changes look good but we should also update the Linux CI to not include this folder for 232 so that the tests run with what we are recommending for users.

pyaedt/generic/clr_module.py
maxcapodi78: at this point I don't know on which AEDT version the code is gonna run. so Either I put the message for everyone or I remove for everyone

maxcapodi78: @isaacansys  I've modified the message considering the option of 23.2. If you have a better idea let me know.

isaacansys (approved):
LGTM

MaxJPRey (approved):
LGTM.

PR #3688, opened 2023-10-02
test: pdf build time

Revathyvenugopal162:


Samuelopez-ansys:
@SMoraisAnsys I think we can close this PR, because you are already modifying the documentation build

SMoraisAnsys:
Closing this as #4284 handles the PDF creation.
@Revathyvenugopal162 Thanks for your help back then, it took us a while to handle everything but we converged with a solution working fine :)

PR #3718, opened 2023-10-06
Improved pyvista plot of fields for PCB

maxcapodi78:


pyaedt/modules/AdvancedPostProcessing.py
```diff
@@ -395,6 +397,8 @@ def plot_field_from_fieldplot(
             model.camera_position = view
         elif view != "isometric":
             self.logger.warning("Wrong view setup. It has to be one of xy, xz, yz, isometric.")
+        if is_pcb:
+            model.z_scale = 5
```
MaxJPRey, line 401: @maxcapodi78 Should it be a parameter or will it always work with this value for the `z_scale` property?

maxcapodi78: Ideally could be a parameter

PR #3725, opened 2023-10-09
method to extend trace 

ring630:
Method to extand trace path incrementally

### DIFF
--- a/pyaedt/edb_core/dotnet/database.py
+++ b/pyaedt/edb_core/dotnet/database.py
@@ -5,7 +5,6 @@ import sys
 
 from pyaedt import __version__
 from pyaedt.aedt_logger import pyaedt_logger
-from pyaedt.edb_core.edb_data.edbvalue import EdbValue
 from pyaedt.edb_core.general import convert_py_list_to_net_list
 from pyaedt.generic.general_methods import env_path
 from pyaedt.generic.general_methods import env_path_student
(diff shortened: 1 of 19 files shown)


pyaedt/edb_core/dotnet/database.py
```diff
@@ -80,6 +80,40 @@ def arcs(self):  # pragma: no cover
...
+        x: str, int, float
+            X coordinate.
+        y: str, in, float
+            Y coordinate.
+        incremental: bool
+            Add point incrementally. If True, coordinates of the added point is incremental to the last point.
+            The default value is ``False``.
+        Returns
+        -------
+        bool
```
PipKat, line 107:
```suggestion
        incremental: bool
            Whether to add the point incrementally. The default value is ``False``. When
            ``True``, the coordinates of the added point are incremental to the last point.
            

        Returns
        -------
        bool
            ``True`` when successful, ``False`` when failed.
```

pyaedt/edb_core/edb_data/primitives_data.py
```diff
@@ -715,6 +716,27 @@ def length(self):
                 length += GeometryOperators.points_distance(center_line[pt_ind], center_line[pt_ind + 1])
         return length
 
+    @pyaedt_function_handler
+    def add_point(self, x, y, incremental=False):
```
PipKat, line 720: This needs to have a description. It seems that this docstring should be the same as the one for ``def add_point`` in previous database.py file.

maxcapodi78 (approved):

Issue #3730, opened 2023-10-10
PyAEDT support wait for license flag with Desktop call


svandenb-dev:
### Description of the feature

Support the flag wait for license in case licenses are not available to prevent timeout error when Desktop is called from pyaedt.

PR #3732, opened 2023-10-10
added waitforlicense options to settings

maxcapodi78:
#3730 
Implemented waitforlicense flag into settings

```
from pyaedt import settings, Desktop
settings.wait_for_license=True
d = Desktop()
```


Samuelopez-ansys (approved):
Great

PR #3733, opened 2023-10-10
Implemented uniform points exports into methods export_report_to_csv and export_report_to_file

maxcapodi78:
Methods now support the export of uniform points data give a specified start, end and step values.

### DIFF
--- a/pyaedt/modules/PostProcessor.py
+++ b/pyaedt/modules/PostProcessor.py
@@ -1179,16 +1179,19 @@ class PostProcessorCommon(object):
         unique_file : bool
             If set to True, generates unique file in output_dit
         uniform : bool, optional
-            Whether if export uniform points to file or not.
-            Default is ``False``.
+            Whether the export uniform points to the file. The
+            default is ``False``.
         start : str, optional
-            If ``uniform`` is set then this is the start range with units for the sweep.
+            Start range with units for the sweep if the ``uniform`` parameter
+            is set to ``True``.
         end : str, optional
-            If ``uniform`` is set then this is the end range with units for the sweep.
+            End range with units for the sweep if the ``uniform`` parameter
+            is set to ``True``.
         step : str, optional
-            If ``uniform`` is set then this is the step range with units for the sweep.
+            Step range with units for the sweep if the ``uniform`` parameter is
+            set to ``True``.
         use_trace_number_format : bool, optional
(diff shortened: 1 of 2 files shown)


pyaedt/modules/PostProcessor.py
```diff
@@ -1167,6 +1178,17 @@ def export_report_to_file(self, output_dir, plot_name, extension, unique_file=Fa
...
+            Whether if export uniform points to file or not.
+            Default is ``False``.
+        start : str, optional
+            If ``uniform`` is set then this is the start range with units for the sweep.
+        end : str, optional
+            If ``uniform`` is set then this is the end range with units for the sweep.
+        step : str, optional
+            If ``uniform`` is set then this is the step range with units for the sweep.
+        use_trace_number_format : bool, optional
+            Whether to use trace number formats or not. Default is ``False``.
```
PipKat, line 1191:
```suggestion
        uniform : bool, optional
            Whether the export uniform points to the file. The
            default is ``False``.
        start : str, optional
            Start range with units for the sweep if the ``uniform`` parameter is set to ``True``..
        end : str, optional
            End range with units for the sweep if the ``uniform`` parameter is set to ``True``.
        step : str, optional
            Step range with units for the sweep if the ``uniform`` parameter is set to ``True``..
        use_trace_number_format : bool, optional
            Whether to use trace number formats. The default is ``False``.
```
Some of the already existing parameter descriptions should be cleaned up also to match our standards.

pyaedt/modules/PostProcessor.py
```diff
@@ -1222,6 +1250,17 @@ def export_report_to_csv(self, project_dir, plot_name):
...
```
PipKat, line 1263:
```suggestion
        uniform : bool, optional
            Whether the export uniform points to the file. The
            default is ``False``.
        start : str, optional
            Start range with units for the sweep if the ``uniform`` parameter
            is set to ``True``.
        end : str, optional
            End range with units for the sweep if the ``uniform`` parameter
            is set to ``True``.
        step : str, optional
            Step range with units for the sweep if the ``uniform`` parameter is
            set to ``True``.
        use_trace_number_format : bool, optional
            Whether to use trace number formats. The default is ``False``.
```

Samuelopez-ansys (approved):
LGTM

PR #3745, opened 2023-10-12
Validation rlc values

ring630:
RLC can have no value from database. This feature is to add default values if it is missing. 

In this PR, only inductors are check and fixed.


pyaedt/edb_core/edb_data/obj_base.py
```diff
@@ -0,0 +1,16 @@
...
+        self._edb_object = model
+
+    @property
+    def is_null(self):
+        """Flag indicating if this object is null."""
+        return self._edb_object.IsNull()
+
+    @property
+    def type(self):
+        """Get type."""
```
PipKat, line 15: Properties should be defined as nouns. For this ``type`` property, the description should be "Type of ...." (Type of the base object?) We don't need to call out GET or SET usage in the description. If needed, an "Examples" section in the docstring could provide usage examples.)

pyaedt/edb_core/layout_validation.py
```diff
@@ -230,3 +230,21 @@ def illegal_net_names(self, fix=False):
 
         self._pedb._logger.info("Found {} illegal net names.".format(len(renamed_nets)))
         return
+
+    def illegal_rlc_values(self, fix=False):
+        """Find and fix rlc illegal values."""
```
PipKat, line 235:
```suggestion
        """Find and fix RLC illegal values."""
```
Based on a search of the doc, it looks like we use RLC in all capital letters in descriptions.

svandenb-dev (approved):

PR #3754, opened 2023-10-13
remove installation from batch

gmalinve:


Samuelopez-ansys:
@gmalinve I think you should also remove the batch file from the repo?, it is under doc/source/Resources

Samuelopez-ansys (approved):

PR #3761, opened 2023-10-17
CI: Add edb label to labeler

SMoraisAnsys:
Update the Github workflow labeler to track changes related to `edb` (based on specific modified files).
This could me removed later on but would be of great help for the split between AEDT and EDB.


Samuelopez-ansys (approved):
LGTM

PR #3772, opened 2023-10-19
CI: Add edb label to labels.yml

SMoraisAnsys:
Previous PR (#3761) was missing an item in `.github/labels.yml` which resulted in:
- the deletion of the `edb` label on github (cf #3745) and the removal from each issue / PR
- no label added on PRs touching targeted files

This addition should patch it (tested on a fork cf https://github.com/SMoraisAnsys/pyaedt/pull/2)

### DIFF
--- a/.github/labels.yml
+++ b/.github/labels.yml
@@ -29,3 +29,7 @@
 - name: testing
   description: Anything related to testing
   color: 5802B8
+
+- name: edb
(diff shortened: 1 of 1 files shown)


Samuelopez-ansys (approved):

MaxJPRey:
Thanks @SMoraisAnsys for fixing that.

PR #3777, opened 2023-10-20
Disable PDF Documentation

Samuelopez-ansys:


maxcapodi78 (approved):
LGTM. @MaxJPRey This is needed until the pdf workflow is fixed

PR #3785, opened 2023-10-23
Add docstring for MeshRegion class.

### LINKED ISSUE #3721: Improve IcepakMesh class to allow setting 2D MLM plane

Add API access to set the 2D MLM plane when known, e.g. XY, XZ, YZ.   Or if there's already a method, add documentation on how to access it.

### Code sample expected

_No response_

_No response_

lorenzovecchietti:
Close #3721.

### DIFF
--- a/_unittest_solvers/test_00_analyze.py
+++ b/_unittest_solvers/test_00_analyze.py
@@ -1,3 +1,4 @@
+import csv
 import os
 import sys
 import time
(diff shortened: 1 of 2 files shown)


lorenzovecchietti:
As commented [here](https://github.com/ansys/pyaedt/issues/3721#issuecomment-1774961041):

> I added a docstring for the MeshRegion class.
> So when a user creates the mesh region, he can see the different attributes and their meaning with the help command. I also explicitly wrote down available options for particular arguments, like the one related to this issue.
> 
> The alternative was to create setter and getter, but this seems harder to maintain with little benefit.

Samuelopez-ansys (approved):

PR #3801, opened 2023-10-26
Now PyAEDT Cutout keeps the path as lines after the cutout

maxcapodi78:
Now PyAEDT Cutout keeps the path as lines after the cutout instead of converting them to polygon data.

### DIFF
--- a/pyaedt/edb.py
+++ b/pyaedt/edb.py
@@ -2026,6 +2026,7 @@ class Edb(Database):
                 i.net_object.Delete()
         reference_pinsts = []
         reference_prims = []
+        reference_paths = []
         for i in self.padstacks.instances.values():
             net_name = i.net_name
             id = i.id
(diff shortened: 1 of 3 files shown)


pyaedt/edb.py
```diff
@@ -1656,6 +1657,11 @@ def cutout(
             Whether to use the center of the pad to find the intersection with extent or use the bounding box.
             Second method is much slower and requires to disable multithread on padstack removal.
             Default is `True`.
+        keep_lines_as_path : bool, optional
+            Whether to keep the lines as Path after they are cutout or convert them to PolygonData.
+            This feature works only in Electronics Desktop (3D Layout).
+            If the flag is set to True it can cause issues in SiWave once the Edb is imported.
```
MaxJPRey, line 1663:
```suggestion
            If the flag is set to ``True`` it can cause issues in SiWave once the Edb is imported.
```

svandenb-dev (approved):

PR #3814, opened 2023-10-27
padstack name

ring630:
@maxcapodi78 @svandenb-dev there is a change in the padstack name. It may break customer code, but I think it is better to correct the way how the padstack name is returned in the lone run.


pyaedt/edb_core/edb_data/padstacks_data.py
```diff
@@ -1455,13 +1455,8 @@ def rotation(self):
...
     @property
     def name(self):
-        """Padstack Instance Name. If it is a pin, the syntax will be like in AEDT ComponentName-PinName."""
-        if self.is_pin:
-            comp_name = self._edb_padstackinstance.GetComponent().GetName()
-            pin_name = self._edb_padstackinstance.GetName()
-            return "-".join([comp_name, pin_name])
-        else:
-            return self._edb_padstackinstance.GetName()
+        """Padstack Instance Name."""
```
PipKat, line 1458:
```suggestion
        """Padstack instance name."""
```

pyaedt/edb_core/edb_data/padstacks_data.py
```diff
@@ -1504,6 +1499,7 @@ def metal_volume(self):
     @property
     def pin_number(self):
         """Get pin number."""
+        warnings.warn("`pin_number` is deprecated. Use `name` property instead.", DeprecationWarning)
```
svandenb-dev, line 1502: I guess you want to to say use `number`instead ?

svandenb-dev:
> @maxcapodi78 @svandenb-dev there is a change in the padstack name. It may break customer code, but I think it is better to correct the way how the padstack name is returned in the lone run.

I am not sure sure we can change the pin naming (e.g component.pin name). This will for sure break current flows. I remember we had to  sync EDB pin name with AEDT ones. @maxcapodi78 what do you think ?

pyaedt/edb_core/edb_data/padstacks_data.py
svandenb-dev: agreed there is no notion of pin number, the initial property pin_number is not correct. Let change the deprecation to `name` as suggested

PR #3819, opened 2023-10-30
Fix "split" method in Modeler.py

lorenzovecchietti:
The bug appears when `PLANE.YZ` is used as input for the `plane` argument.
Since `PLANE.YZ=0`, all the if statements like `if not plane` were evaluating as `True`. In the method implementation, it was supposed to happen only when `plane` argument was not provided, so when `plane=None`.


Samuelopez-ansys (approved):
Good catch!

PR #3827, opened 2023-10-31
parameterized Djordjevic Sarkar model in AEDT

ring630:
Now Djordjevic Sarkar model can mobe parameterized in AEDT


maxcapodi78 (approved):
LGTM

PR #3828, opened 2023-10-31
Add files related to contributions.

MaxJPRey:
Adding a list of the main contributors.
Obviously names can be added and/or removed based on contributors will.

**Thanks you all for your commitment and hard work.**

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,65 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+In the interest of fostering an open and welcoming environment, we as
+contributors and maintainers pledge to making participation in our
+project and our community a harassment-free experience for everyone,
+regardless of age, body size, disability, ethnicity, sex
+characteristics, gender identity and expression, level of experience,
+education, socio-economic status, nationality, personal appearance,
+race, religion, or sexual identity and orientation.
+
+## Our Standards
+
+Examples of behavior that contributes to creating a positive environment
+include:
+
+* Using welcoming and inclusive language
+* Being respectful of differing viewpoints and experiences
+* Gracefully accepting constructive criticism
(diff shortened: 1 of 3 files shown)


SMoraisAnsys:
@MaxJPRey Shouldn't we had every person present in here : https://github.com/ansys/pyaedt/graphs/contributors (excluding bots) ? The list could be split (if needed), e.g. in two parts : active collaborators and collaborators (not sure about the naming, something else might be more appropriate). Here is an example of what is performed in nodejs (https://github.com/nodejs/node/blob/main/README.md#current-project-team-members). 

Note that in nodejs there is a notion of technical steering committee which is directly linked to the project governance. Thus, while working on this PR, why not adding some project governance information ? (e.g. GOVERNANCE.md)

MaxJPRey:
> @MaxJPRey Shouldn't we had every person present in here : https://github.com/ansys/pyaedt/graphs/contributors (excluding bots) ? The list could be split (if needed), e.g. in two parts : active collaborators and collaborators (not sure about the naming, something else might be more appropriate). Here is an example of what is performed in nodejs (https://github.com/nodejs/node/blob/main/README.md#current-project-team-members).
> 
> Note that in nodejs there is a notion of technical steering committee which is directly linked to the project governance. Thus, while working on this PR, why not adding some project governance information ? (e.g. GOVERNANCE.md)

I would just add those limited for now as they are active contributors.
Of course the list will evolved but I would rather have people asking to be added in this file before adding them.

For the governance file. I would be interested to investigate that.
We would need some approval -e.g. legal- before to proceed.
Please bring the topic in the next team meeting.

Samuelopez-ansys (approved):

dcrawforAtAnsys (approved):
LGTM

gmalinve (approved):

lorenzovecchietti (approved):

PR #3830, opened 2023-10-31
Fix documentation style of hdm parser.

MaxJPRey:


pyaedt/sbrplus/hdm_parser.py
```diff
@@ -42,11 +42,11 @@ def __init__(self, filename):
         self.binarycontent = binarycontent
 
     def parse_message(self):
-        """Parses the binary content of the HDM file"""
+        """Parse the binary content of the HDM file."""
         return self._parse(self.message["type"])
 
     def _parse(self, type_name):
-        """Generic parser method, dispatches to specialized ones"""
+        """Generic parser method, dispatches to specialized ones."""
```
PipKat, line 49: Method description should start with a verb. Would we be able to use this instead: "Use generic parser method, which dispatches to specialized parsers."? What are we parsing?

pyaedt/sbrplus/hdm_parser.py
```diff
@@ -88,7 +88,7 @@ def _parse_list(self, type=None, base=None, size=1):
         """
         Parser for vector or list. 'vector's are to be interpreted in the linear algebra sense,
         and converted to numpy.array. 'list's are Python lists. Only simple base types can be
```
PipKat, line 90:
```suggestion
        Parser for vector or list. A vector is interpreted in the linear algebra sense
        and converted to a NumPy array. A list is converted to a Python list. Only simple base types can be
```

pyaedt/sbrplus/hdm_parser.py
```diff
@@ -88,7 +88,7 @@ def _parse_list(self, type=None, base=None, size=1):
...
-        interpreted as a numpy array
+        interpreted as a numpy array.
```
PipKat, line 91:
```suggestion
        interpreted as a NumPy array.
```

pyaedt/sbrplus/hdm_parser.py
MaxJPRey, line 49:
```suggestion
        """Use a generic parser method, which dispatches to appropriate and specialized parsers."""
```

MaxJPRey:
Thanks @PipKat for the review.

Samuelopez-ansys:
@zebanaqviWAT @jsalant22 Could you commit the comments and merge this PR? Thanks

Samuelopez-ansys (approved):

PR #3842, opened 2023-11-03
added layout_component_names property

maxcapodi78:


Samuelopez-ansys (approved):
Why it was working before?

maxcapodi78:
> Why it was working before?

@Samuelopez-ansys it's randomly failing. I tried to isolate them and make more robust and independent from other tests

PR #3845, opened 2023-11-03
Improved the way the information are plotted from the __str__ methods

maxcapodi78:


pyaedt/application/Design.py
```diff
@@ -133,6 +154,16 @@ def __setitem__(self, variable_name, variable_value):
         self.variable_manager[variable_name] = variable_value
         return True
 
+    @property
+    def info(self):
+        """Return the PyAEDT session info.
```
PipKat, line 159:
```suggestion
        """Dictionary of the PyAEDT session information.
```

pyaedt/application/Design.py
```diff
@@ -103,15 +103,36 @@ class Design(AedtObjects):
...
-        pyaedt_details = "      pyaedt API\n"
-        pyaedt_details += "pyaedt running AEDT Version {} \n".format(settings.aedt_version)
-        pyaedt_details += "Running {} tool in AEDT\n".format(self.design_type)
-        pyaedt_details += "Solution Type: {} \n".format(self.solution_type)
-        pyaedt_details += "Project Name: {} Design Name {} \n".format(self.project_name, self.design_name)
+    @property
+    def _pyaedt_details(self):
+        import platform
+
+        from pyaedt import __version__ as release
```
MaxJPRey, line 110: Just to avoid confusion with the AEDT release, I would suggest:

```suggestion
        from pyaedt import __version__ as pyaedt_version
```

If you apply this change , don't forget to do the same few lines below.

Samuelopez-ansys (approved):

PR #3854, opened 2023-11-07
Add is used flag to material

### LINKED ISSUE #3815: add is_used for materials used in the design

Samuelopez-ansys:
Close #3815


Samuelopez-ansys:
@gmalinve  This new property detects if the material is used in the project or not, it is not specific for the design, but in a new PR, it could be implemented. But it does not exist in the AEDT API the method to find if the material is used in a specific design, with a couple of loops we could implement it in the future.

gmalinve (approved):

PR #3856, opened 2023-11-08
Emit license session

bryankaylor:
Exposes the EMIT API's new license_session object through pyaedt starting with 24R2.

Without it, when doing many runs, the license has to be checked in and out every time. 
With it, the user controls when checkout and checkin happen and can do any number of runs between. 

Also adds a new test verifying the license session is faster than individual runs for a test project.


pyaedt/emit_core/results/revision.py
```diff
@@ -704,3 +704,22 @@ def set_emi_category_filter_enabled(self, category: EmiCategoryFilter, enabled:
             raise RuntimeError("This function is only supported in AEDT version 2024 R1 and later.")
         engine = self.emit_project._emit_api.get_engine()
         engine.set_emi_category_filter_enabled(category, enabled)
+
+    def license_session(self):
```
MaxJPRey, line 708: I would suggest to either rename it `get_license_session` or use a property instead of a method.

bryankaylor: @MaxJPRey - Thanks for the idea! Changed it to a property and updated the test/docstring example accordingly.

_unittest_solvers/test_26_emit.py
```diff
@@ -1131,3 +1132,92 @@ def test_result_categories(self, add_app):
         with pytest.raises(RuntimeError) as e:
             instance.get_largest_emi_problem_type()
             assert "An EMI value is not available so the largest EMI problem type is undefined." in str(e)
+
+    @pytest.mark.skipif(
+        config["desktopVersion"] < "2024.2", reason="Skipped on versions earlier than 2024.2"
```
PipKat, line 1137:
```suggestion
        config["desktopVersion"] < "2024.2", reason="Skipped on versions earlier than 2024 R2."
```
Versions in public-facing documentation should be in the format shown in the edits. Messages should have concluding punctuation.

_unittest_solvers/test_26_emit.py
```diff
@@ -1131,3 +1132,92 @@ def test_result_categories(self, add_app):
...
+                                       ": " + str(tx_freq) + " and " + receiver + ": " + 
+                                       rx_band + ": " + str(rx_freq))
+                            
+                            combinations_run += 1
+
+                            if combinations_run >= 20:
+                                return [best_band, best_freq]
+            return [best_band, best_freq]
+        
+        # Warmup incase something isn't solved
```
PipKat, line 1201:
```suggestion
        # Warmup in case something isn't solved
```

pyaedt/emit_core/results/revision.py
```diff
@@ -732,3 +732,23 @@ def set_emi_category_filter_enabled(self, category: EmiCategoryFilter, enabled:
             raise RuntimeError("This function is only supported in AEDT version 2024 R1 and later.")
         engine = self.emit_project._emit_api.get_engine()
         engine.set_emi_category_filter_enabled(category, enabled)
+
+    @property
+    def license_session(self):
+        """Get a license_session object.
+
+        A license session can be started with checkout(), and ended with checkin().
+        The `with` keyword can also be used, where checkout() is called on enter, and checkin() is called on exit.
```
PipKat, line 741:
```suggestion
        You can start a license session with the ``checkout()`` method and end it with the ``checkin()`` method.
        You can also use the ``with`` keyword, where the ``checkout()`` method is called on enter and the ``checkin()``
        method is called on exit.
```

pyaedt/emit_core/results/revision.py
```diff
@@ -732,3 +732,23 @@ def set_emi_category_filter_enabled(self, category: EmiCategoryFilter, enabled:
...
+
+        Avoids having to wait for license checkin and checkout when doing many runs.
```
PipKat, line 743:
```suggestion
        Using a license session avoids having to wait for license checkin and checkout when doing many runs.
```

pyaedt/emit_core/results/revision.py
MaxJPRey: Great, thanks @bryankaylor .

bryankaylor:
Thanks for the review @PipKat - I've applied your edits.

PipKat:
[like]  Kathy Pippert reacted to your message:
________________________________

pyaedt/emit_core/results/revision.py
myoung301: This is a pretty good explanation of properties vs. methods: https://stackoverflow.com/a/20506966

The checkout() called by this "property" definitely has side effects so it might be best to make license_session a method (`get_license_session()` as proposed). @bryankaylor - what do you think?

pyaedt/emit_core/results/revision.py
```diff
@@ -732,3 +732,25 @@ def set_emi_category_filter_enabled(self, category: EmiCategoryFilter, enabled:
             raise RuntimeError("This function is only supported in AEDT version 2024 R1 and later.")
         engine = self.emit_project._emit_api.get_engine()
         engine.set_emi_category_filter_enabled(category, enabled)
+
+    @property
+    def license_session(self):
+        ""License session object.
```
Alberto-DM, line 738:
```suggestion
        """License session object.
```
One quote is missing.

bryankaylor: Thanks Alberto- fixed.

pyaedt/emit_core/results/revision.py
bryankaylor: Thanks @myoung301 , changed from a property to a get_license_session() function.

pyaedt/emit_core/results/revision.py
```diff
@@ -732,3 +732,22 @@ def set_emi_category_filter_enabled(self, category: EmiCategoryFilter, enabled:
             raise RuntimeError("This function is only supported in AEDT version 2024 R1 and later.")
         engine = self.emit_project._emit_api.get_engine()
         engine.set_emi_category_filter_enabled(category, enabled)
+
+    def get_license_session(self):
+        """Get a license_session object.
```
PipKat, line 737:
```suggestion
        """Get a license_session object.
```Try to avoid including a code entity in the brief description: Get a license_session.
If you feel that the code entity must be used, place it in double backticks: Get a ``license_session`` object.

File: pyaedt/emit_core/results/revision.py (hunk 4 above)
Comment (PipKat) [reply pyaedt/emit_core/results/revision.py:740]: @bryankaylor The previous converstion that you marked as resolved has suggestions to bring the content inline with our style guide. Please incorporate the suggested changes.

File: pyaedt/emit_core/results/revision.py (hunk 7 above)
Comment (bryankaylor) [reply pyaedt/emit_core/results/revision.py]: Thanks @PipKat - fixed in the most recent commit.

File: _unittest_solvers/test_26_emit.py [hunk 8]
```diff
@@ -1170,3 +1171,92 @@ def test_result_categories(self, add_app):
...
+        # Get the time using the license session
+        start = time.perf_counter()
+        best_case_rx_ch = {}
+        with revision.get_license_session():
+            for rx in [receivers[0]]:
+                best_case_rx_ch[rx] = get_best_rx_channel(results, rx)
+        end = time.perf_counter()
+        licenseSessionTime = end - start
+
+        assert licenseSessionTime < noLicenseSessionTime
```
myoung301, line 1262: @bryankaylor Should we make sure it's at least 2x faster? A simple `<` could pass due to the times being just slightly different (noise essentially). 

Also, if the results are cached by the first loop, we might not be solving in the second loop which could also cause it to be faster without really verifying the license session is helping. I'm not sure this could happen right now, but a change to caching could mean it starts happening. Ideally we'd ensure nothing is cached/solved the second time through by purging results. I don't think there is a way to do that right now, but would reversing the order help? If we run the license session first which should be faster, things would fail if the second uses a cache and is faster because it doesn't actually have to run and thus doesn't do license checkout. Does that make sense?

bryankaylor: @myoung301 - There is an untimed warmup loop on line 1240, before the two timed loops. I put it there in case there is or will be some caching behavior that affects the times. With it, either the results will be cached for both timed loops or for neither. Does that sound right?

I do agree we should change it to licenseSessionTime*2 < noLicenseSessionTime, so if this caching starts happening and the times are nearly equal (or if the times are nearly equal for any other reason), it will fail consistently instead of being almost random.

Samuelopez-ansys (approved):

Samuelopez-ansys:
@jsalant22 This branch was automatically merged after I approved it. Please check if everything is correct. Thanks

PR #3863, opened 2023-11-10
improve the speed of IL computation in Touchstone parser

maxcapodi78:
on big touchstone files it was taking too much time to compute insertion losses indexes.


Samuelopez-ansys (approved):
LGTM

PR #3886, opened 2023-11-15
get materials in project

gmalinve:


Samuelopez-ansys (commented):
@gmalinve this returns the project materials and not the design materials, right? 

What’s the difference with the is_used method?

MaxJPRey (approved):
LGTM.

gmalinve:
@Samuelopez-ansys I should have used the method GetInUseProjectMaterialNames() instead but for me this method is also useful to get the used project materials as a list whereas is_used is a property of the material.
I could retrieve all materials in a project and then parse them in a for loop to check if they are used with the is_used prop but this method automatically retrieve the list, which in same cases can be useful.

pyaedt/modules/MaterialLib.py
```diff
@@ -870,3 +870,19 @@ def import_materials_from_excel(self, material_file):
             materials_added.append(new_material)
 
         return materials_added
+
+    @pyaedt_function_handler
+    def get_used_project_materials(self):
+        """Get list of material names in current design.
```
Alberto-DM, line 876: Few comments:
- does it returns the materials used in the design or in the project?
- this function should return a list of material `Material` objects.
- if you want you can also create a class which returns the list of material names (`get_used_project_material_names`)

Alberto-DM (changes requested):

pyaedt/modules/MaterialLib.py
gmalinve: @Alberto-DM @Samuelopez-ansys 
- This method return a list or material names used in project. I could rename it as `get_design_materials` and change the logic. In this case, can we keep this method in `MaterialLib`?
- why material objects rather than material names?

gmalinve: @Alberto-DM @Samuelopez-ansys for me it's still not clear how to retrieve only materials used in a specific design. The `is_used` prop doesn't specify if the material is used in my design (in case of multiple designs in a project).

Samuelopez-ansys, line 876:
```suggestion
        """Get list of material names in current project.
```

Samuelopez-ansys (approved):

pyaedt/modules/MaterialLib.py
Samuelopez-ansys: @Alberto-DM @gmalinve What this method is doing is getting the material names used in the project, so @Alberto-DM suggestion is correct, and change the name of the method to "get_used_project_material_names" because it is consistent to other pyaedt methods.

I think we can merge this pull request because it is a useful method, but this will follow in the next implementation, to get the materials used in a design, because this is more useful than getting the project materials.

_unittest/test_03_Materials.py
```diff
@@ -245,3 +245,9 @@ def test_12_material_model(self):
         self.aedtapp["$dk"] = 3
         self.aedtapp["$df"] = 0.01
         assert mat.set_djordjevic_sarkar_model(dk="$dk", df="$df")
+
+    def test_13_get_materials_in_design(self):
```
Samuelopez-ansys, line 249:
```suggestion
    def test_13_get_materials_in_project(self):
```

_unittest/test_03_Materials.py
```diff
@@ -245,3 +245,9 @@ def test_12_material_model(self):
...
+        used_materials = self.aedtapp.materials.get_used_project_materials()
```
Samuelopez-ansys, line 250:
```suggestion
        used_materials = self.aedtapp.materials.get_used_project_material_names()
```

pyaedt/modules/MaterialLib.py
```diff
@@ -870,3 +870,19 @@ def import_materials_from_excel(self, material_file):
             materials_added.append(new_material)
 
         return materials_added
+
+    @pyaedt_function_handler
+    def get_used_project_materials(self):
```
Samuelopez-ansys, line 875:
```suggestion
    def get_used_project_material_names(self):
```

Samuelopez-ansys (commented):
@gmalinve

pyaedt/modules/MaterialLib.py
```diff
@@ -852,3 +852,19 @@ def import_materials_from_excel(self, material_file):
...
         return materials_added
+
+    @pyaedt_function_handler
+    def get_used_project_material_names(self):
+        """Get list of material names in current project.
+
+        Returns
+        -------
+        List of str
+            List of material names in current design.
```
Alberto-DM, line 863:
```suggestion
            List of material names used in the current project.
```

Alberto-DM (approved):

PR #3898, opened 2023-11-21
Improve method get_solution_data to avoid intrinsics as arguments and…

maxcapodi78:
… to add the method to the setups objects


Samuelopez-ansys (approved):
This is great! It will simplify a lot our lives.

PR #3899, opened 2023-11-22
Add jupytext header

dcrawforAtAnsys:
Added the jupytext header for sphinx to the example 02-HFSS/Array.py.  This makes it possble to directly view and edit the example with the jupyter notebook editor while developing examples if the jupytext plugin is installed. Markdown cells use markdown instead of restructured text.

### DIFF
--- a/examples/02-HFSS/Array.py
+++ b/examples/02-HFSS/Array.py
@@ -1,3 +1,17 @@
+# ---
+# jupyter:
+#   jupytext:
+#     text_representation:
+#       extension: .py
+#       format_name: sphinx
+#       format_version: '1.1'
+#       jupytext_version: 1.14.5
+#   kernelspec:
+#     display_name: Python 3 (ipykernel)
+#     language: python
+#     name: python3
+# ---
+
 """
 HFSS: component antenna array
 -----------------------------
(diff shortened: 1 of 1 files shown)


Samuelopez-ansys:
@MaxJPRey @SMoraisAnsys @jorgepiloto 

@dcrawforAtAnsys has implemented this PoC to use jupytext for the examples, I think it is really great because content and example creation will be much easier.

After installing jupytext, this is how it looks:

image

I am wondering if this will affect the CI/CD, and if we could test it locally before pushing it. It passed all checks, but I think we could have problems when publishing the documentation, maybe at least we could add jupytext in the dependencies?

Thanks!

MaxJPRey (commented):
Hi @dcrawforAtAnsys . This is an excellent idea. I started to look at it on my side too.

SMoraisAnsys:
Great idea !

@Samuelopez-ansys I don't know if that would affect the CI/CD, I can try to test it locally to see if everything works fine. If this changes are to be added, we might want to add some tests or change the one associated to the examples.

SMoraisAnsys:
@dcrawforAtAnsys I've added the upload, see the artifact down https://github.com/ansys/pyaedt/actions/runs/7058239367

PR #3905, opened 2023-11-23
Solder balls height control with simconfig

svandenb-dev:
Adding custom solder balls height support with SImulatiomCnnfiguration class.


ring630 (commented):
LGTM

SMoraisAnsys (approved):
Codecov seems to indicate that the modified method `build_simulation_project` is not tested for `dict` components. When I read the code, I feel like it should be done with your new test and json file. Can you double check that your test works fine ?

PR #3910, opened 2023-11-24
added "DecadeCount" "OctaveCount" "ExponentialCount" in parametric sw…

IreneWoyna:
…eep, removed unit from "step"


IreneWoyna:
It seems that "LogScale" is actually not used at all and could be removed

        elif variation_type == "LogScale":
            sweep_range = "DEC {} {} {}".format(start_point, end_point, step)

MaxJPRey (approved):
LGTM.

Samuelopez-ansys (changes requested):
@IreneWoyna Great implementation, but you need to add some unit test to pass the CodeCov: 
image

It is not your fault, it was not covered before, but now you touched this part of the code and the bot detected it :) You just need to add these different sweeps in the unittest.

gmalinve:
@IreneWoyna I'd remove the LogScale and add unit test to cover all new cases.

Samuelopez-ansys (approved):
LGTM

PR #3915, opened 2023-11-27
Assign FEBI method

Samuelopez-ansys:
Object-oriented not implemented in AEDT API when Hybrid region

### DIFF
--- a/pyaedt/hfss.py
+++ b/pyaedt/hfss.py
@@ -4589,7 +4589,7 @@ class Hfss(FieldAnalysis3D, object):
         Examples
         --------
 
-        Create a box and assign a FEBI boundary to it.
+        Create a box and assign an FEBI boundary to it.
 
         >>> box = hfss.modeler.create_box([0, -200, -200], [200, 200, 200],
         ...                                         name="Radiation_box")
--- a/pyaedt/hfss.py
+++ b/pyaedt/hfss.py
@@ -4589,7 +4589,7 @@ class Hfss(FieldAnalysis3D, object):
         Examples
         --------
 
-        Create a box and assign an FEBI boundary to it.
+        Create a box and assign an FE-BI boundary to it.
 
         >>> box = hfss.modeler.create_box([0, -200, -200], [200, 200, 200],
         ...                                         name="Radiation_box")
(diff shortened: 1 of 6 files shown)


Samuelopez-ansys:
Added Transient composite solution type.

It is missed the default setup. I realized it would fail because the transient setup is different than the frequency one. It will be refactored in a different PR.

pyaedt/hfss.py
```diff
@@ -4561,6 +4565,49 @@ def assign_hybrid_region(self, obj_names, boundary_name="", hybrid_region="SBR+"
...
+
+        References
+        ----------
+
+        >>> oModule.AssignFEBI
+
+        Examples
+        --------
+
+        Create a box and assign a FEBI boundary to it.
```
PipKat, line 4592:
```suggestion
        Create a box and assign an FEBI boundary to it.
```

pyaedt/hfss.py
```diff
@@ -4561,6 +4565,49 @@ def assign_hybrid_region(self, obj_names, boundary_name="", hybrid_region="SBR+"
...
+    @pyaedt_function_handler()
+    def assign_febi(self, obj_names, boundary_name=""):
+        """Assign an FE-BI region to one or more objects.
+
+        Parameters
+        ----------
+        obj_names : str or list or int or :class:`pyaedt.modeler.cad.object3d.Object3d`
+            One or more object names or IDs.
+        boundary_name : str, optional
+            Name of the boundary. The default is ``""``.
```
PipKat, line 4577: Perhaps we should include the bevavior that occurs when ``""`` is used, something like this:
The default is ``""``, in which case a name is automatically assigned.

pyaedt/hfss.py
```diff
@@ -4561,6 +4565,49 @@ def assign_hybrid_region(self, obj_names, boundary_name="", hybrid_region="SBR+"
...
+
+        References
+        ----------
+
+        >>> oModule.AssignFEBI
+
+        Examples
+        --------
+
+        Create a box and assign an FEBI boundary to it.
```
PipKat, line 4592:
```suggestion
        Create a box and assign an FE-BI boundary to it.
```
Check for consistency. It looks like FE-BI is preferred.

pyaedt/hfss.py
Samuelopez-ansys, line 4577:
```suggestion
            Name of the boundary. The default is ``""``, in which case a name is automatically assigned.
```

MaxJPRey (approved):
LGTM.

PR #3924, opened 2023-11-29
Improve variations in create_report

gmalinve:
At the moment if variations is not provided as input to the method and there are no available variations in the design, report is empty.

### DIFF
--- a/pyaedt/modules/PostProcessor.py
+++ b/pyaedt/modules/PostProcessor.py
@@ -1615,7 +1615,7 @@ class PostProcessorCommon(object):
             ]
         report.expressions = expressions
         report.domain = domain
-        if not variations and domain == "Sweep" :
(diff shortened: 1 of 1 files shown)


Samuelopez-ansys (approved):
LGTM. 

We will need to cover more cases in another pull request (like domain = "Time")

amichel0205:
Hello All,

I just wanted to point out that the primary sweep is not a variation. Variation for me is when you add any variable and define a sweep for this variable in optimetrics. A variation will be specified with a corresponding primary sweep. 
So I can't tell the consequence of considering the primary sweep as a variation but I just wanted to share my thought.

Samuelopez-ansys:
@amichel0205 Talk with AEDT API development, this is the behavior since AEDT API exists :)

amichel0205 (approved):

PR #3636, opened 2023-09-26
improved method create_frequency_sweep to support all optional arguments.

maxcapodi78:
Fix #2381

PR #3929, opened 2023-11-30
Fix parameter check in `create_frequency_sweep` & add related test

atlanswer:
The implementation introduced in #3636 for filling `nums_of_freq_points` is incorrect.
https://github.com/ansys/pyaedt/blob/62b6750a6b6e803cefa15c87e80329b1cb929f36/pyaedt/modules/SolveSetup.py#L2248-L2255
The exception will occur if `num_of_freq_points` isn't `None`, no matter what `sweep_type` is. The exception type of `ValueError` is more appropriate since it indicates correct types but wrong values.

A related test is included.


Samuelopez-ansys (approved):
Good catch! Thank you

Samuelopez-ansys (approved):

PR #3939, opened 2023-12-04
Bug fix refresh padsatck def after layer name changed

svandenb-dev:
Update layer for padstack definition when layer.name setter is used.


pyaedt/edb_core/edb_data/layer_data.py
```diff
@@ -46,17 +46,6 @@ def is_via_layer(self):
...
-        Returns
-        -------
-        Database padstacks definitions:class:`pyaedt.edb_core.edb_data.padstacks.EdbPadstacks`
-
-        """
-        return self._pclass._pedb.padstacks.definitions
-
     @property
     def color(self):
         """Retrieve color of the layer.
```
PipKat, line 51:
```suggestion
        """Color of the layer.
```

ring630 (approved):
LGTM

PR #3950, opened 2023-12-06
Bug fix merge nets polygon

svandenb-dev:
Bug fix see related issue linked.


pyaedt/edb_core/layout.py
```diff
@@ -1103,7 +1103,7 @@ def parametrize_trace_width(
         return True
 
     @pyaedt_function_handler()
-    def unite_polygons_on_layer(self, layer_name=None, delete_padstack_gemometries=False, net_list=None):
+    def unite_polygons_on_layer(self, layer_name=None, delete_padstack_gemometries=False, net_list=[]):
         """Try to unite all Polygons on specified layer.
```
PipKat, line 1107:
```suggestion
        """Try to unite all polygons on the specified layer.
```

pyaedt/edb_core/layout.py
```diff
@@ -1112,6 +1112,8 @@ def unite_polygons_on_layer(self, layer_name=None, delete_padstack_gemometries=F
             Layer Name on which unite objects. If ``None``, all layers will be taken.
         delete_padstack_gemometries : bool, optional
             ``True`` to delete all padstack geometry.
+        net_list : list[str] : optional
+            Net list filter, If ``None``, all nets will be taken.
```
PipKat, line 1116:
```suggestion
            Net list filter. The default is ``[]``, in which case all nets are taken.
```
Is this a list of filters or a list of filters to exclude from X? I find the current description unclear.

pyaedt/edb_core/nets.py
```diff
@@ -1185,42 +1184,11 @@ def merge_nets_polygons(self, net_list):
             net name of list of net name.
```
PipKat, line 1184:
```suggestion
            Net name of list of net name.
```
I have no idea what this means. Should it be changed to something like "Name of net created from a list of nets"? instead.

Issue #2688, opened 2023-03-22
Replace all os.name checks with is_windows, is_linux booleans


kbhagat-ansys:
### Description of the current limitations

This [PR](https://github.com/pyansys/pyaedt/pull/2670/files) #2670 introduced `is_windows` and `is_linux` booleans. 
We can replace all `os.name == "posix"`, `os.name != "posix"`, `os.name == "nt"`, `os.name != "nt"` with the booleans

PR #3956, opened 2023-12-07
Replace all os.name with boolean checks is_windows, is_linux

kbhagat-ansys:
Fixes Issue #2688


Samuelopez-ansys:
@kbhagat-ansys From which AEDT version is available libPyDesktopPlugin.so?

@maxcapodi78 Any comments?

Samuelopez-ansys (approved):

PR #3977, opened 2023-12-13
Adding padstackinstance terminal position getter

svandenb-dev:
Padstack instance terminal position getter added


pyaedt/edb_core/edb_data/terminals.py
```diff
@@ -476,6 +476,13 @@ class PadstackInstanceTerminal(Terminal):
     def __init__(self, pedb, edb_object):
         super().__init__(pedb, edb_object)
 
+    @property
+    def position(self):
```
MaxJPRey, line 480: Just missing the docstring. Otherwise LGTM.

MaxJPRey (approved):

pyaedt/edb_core/edb_data/terminals.py
```diff
@@ -476,6 +476,13 @@ class PadstackInstanceTerminal(Terminal):
...
+        edb_padstack_instance = self._edb_object.GetParameters()
+        if edb_padstack_instance[0]:
+            return self._pedb.padstacks.instances[edb_padstack_instance[1].GetId()].position
```
ring630, line 483: I suggest to create a new object 

`return EdbPadstack(edb_padstack_instance[1]).position`

instead of querying the how instances.

pyaedt/edb_core/edb_data/terminals.py
svandenb-dev: oh yes thanks

pyaedt/edb_core/edb_data/terminals.py
svandenb-dev: Yes I agree thanks

PR #3995, opened 2023-12-18
Make license session test more consistent and meaningful

bryankaylor:
Before, the license session test compared times in different scenarios. This relies on the license checkout/checkin times being consistent, which caused some intermittent failures depending on machine and license setup. The new test counts the number of checkout/checkins instead.

### DIFF
--- a/_unittest_solvers/test_26_emit.py
+++ b/_unittest_solvers/test_26_emit.py
@@ -1203,6 +1203,8 @@ class TestClass:
         with os.scandir(dot_ansys_directory) as dir:
             for file in dir:
                 filename_pieces = file.name.split('.')
+                # Since machine names can contain periods, there may be over five splits here
+                # We only care about the first split and last three splits
                 if len(filename_pieces) >= 5:
                     if (filename_pieces[0] == 'ansyscl' and 
                         filename_pieces[-3] == str(pid) and
--- a/_unittest_solvers/test_26_emit.py
+++ b/_unittest_solvers/test_26_emit.py
@@ -1182,81 +1182,34 @@ class TestClass:
         results = self.aedtapp.results
         revision = self.aedtapp.results.analyze()
 
-        receivers = revision.get_receiver_names()
-        modeRx = TxRxMode.RX
-        modeTx = TxRxMode.TX
-        modeEmi = ResultType.EMI
         self.aedtapp.set_units("Frequency", "GHz")
 
-        def get_best_rx_channel(results, receiver):
+        def do_run():
             domain = results.interaction_domain()
             rev = results.current_revision
-            # get interferers
-            interferers = rev.get_interferer_names()
-            best_emi = 300.0
-            best_band = None
-            best_freq = None
-
-            combinations_run = 0
-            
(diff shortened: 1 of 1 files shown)


_unittest_solvers/test_26_emit.py
```diff
@@ -1182,81 +1181,62 @@ def test_license_session(self, add_app):
...
+            for i in range(number_of_runs):
+                do_run()
+        
+        end_checkouts, end_checkins = count_license_actions(license_file_path)
+
+        checkouts = end_checkouts - start_checkouts
+        checkins = end_checkins - start_checkins
 
-        assert (licenseSessionTime*2) < noLicenseSessionTime
+        assert (checkouts == number_of_runs+1) and checkouts == checkins
```
myoung301, line 1242: @bryankaylor I see the do_run method is called `number_of_runs` times twice (two separate loops). The second loop is inside a license_session so it should only have one license checkout for that loop. However, there is also a do_run on line 1194. So wouldn't there be `number_of_runs + 2` license checkouts?

_unittest_solvers/test_26_emit.py
```diff
@@ -1182,81 +1181,62 @@ def test_license_session(self, add_app):
...
-                            if combinations_run >= 20:
-                                return [best_band, best_freq]
-            return [best_band, best_freq]
+        number_of_runs = 5
+
+        # Do a run to ensure the license log exists
+        do_run()
+
+        # Find the license log for this process
+        appdata_local_path = os.getenv('LOCALAPPDATA')
```
myoung301, line 1197: @bryankaylor - does this resolve to the same as your approach: `tempfile.gettempdir()`

If so, using tempfile might be better. It is more compact and has cross-platform support. We don't support Linux with EMIT yet, but  it's best to write things in a way that will minimize the rework when we do.

bryankaylor: @myoung301 - Thanks for the suggestion! I didn't know about that import/function, I've updated the PR using it. It leads to `C:\Users\<username>\AppData\Local\Temp`.

_unittest_solvers/test_26_emit.py
bryankaylor: @myoung301 - The do_run on line 1194 is there because, if you run this test in isolation, the license file for the process doesn't exist yet. So, to ensure the test works whether it's run in isolation or with all the other tests, I added this first do_run that just makes the license file exist if it doesn't already.

After that do_run on line 1194, we parse the file and record the initial checkouts/checkins. Then, we do `number_of_runs` runs with and without a license session, re-parse the file, and record the final chckouts/checkins. So, since that first do_run is before we record the initial counts, the difference in checkouts should still be number_of_runs+1.

bryankaylor:
@myoung301 - This is ready for another look whenever you get a chance, let me know if you have any questions!

The log filename search should be a bit more resilient now, and the new checkout/checkin counting has been tested with various EMIT HPC options (Pool + 4 threads, Pool + 16 threads, and Pack).

myoung301 (approved):
I see the tests all passed now. Code looks good.

PR #3997, opened 2023-12-20
core loss example

gmalinve:


examples/03-Maxwell/Maxwell_Transformer_Coreloss.py
```diff
@@ -0,0 +1,136 @@
...
+# Perform required imports.
+
+from pyaedt import downloads
+from pyaedt import generate_unique_folder_name
+from pyaedt import Maxwell3d
+from pyaedt.generic.constants import unit_converter
+from pyaedt.generic.general_methods import read_csv
+
+
+def _convert_tuple_to_float(curves_csv):
```
gmalinve, line 19: @MaxJPRey @Samuelopez-ansys @Alberto-DM  any advice here for this provate method needed to remove the header of the csv file and convert all string numbers to float? I'm not sure this is the correct place to put this but it's not even general enough to put it in general_methods.py.

gmalinve: sorted!

_unittest/test_03_Materials.py
```diff
@@ -363,6 +363,13 @@ def test_14_set_core_loss(self):
                 150: [[0, 0], [1, 10], [2, 19]],
             }
         )
+        try:
+            self.aedtapp.materials["mat_test"].set_coreloss_at_frequency(
+                points_list_at_freq={80: [[0, 0], [1, 3.5], [2, 7.4]]}, core_loss_model_type="Power Ferrite"
+            )
+            assert False
+        except ValueError:
+            assert True
```
SMoraisAnsys, line 372:
```suggestion
        with pytest.raises(ValueError):
            self.aedtapp.materials["mat_test"].set_coreloss_at_frequency(
                points_list_at_freq={80: [[0, 0], [1, 3.5], [2, 7.4]]}, core_loss_model_type="Power Ferrite"
            )
```

SMoraisAnsys: I would recommend using pytest `raises` to ensure that an exception of a specific type is raised.
Btw, I can see a few lines before the following code :
```python
            assert False
        except TypeError:
            assert True
```
I can't understand why such lines are here. If you agree with the proposed modification, it would be great to also use `raises` in that case and to remove the assert `False` and `True`.

pyaedt/modules/Material.py
```diff
@@ -2057,6 +2057,8 @@ def get_core_loss_coefficients(
             value, unit = decompose_variable_value(thickness)
             if not is_number(value) and not unit:
                 raise TypeError("Thickness must be provided as a string with value and unit.")
+        if len(points_list_at_freq) == 1 and core_loss_model_type == "Power Ferrite":
```
SMoraisAnsys, line 2060:
```suggestion
        if len(points_list_at_freq) <= 1 and core_loss_model_type == "Power Ferrite":
```

SMoraisAnsys: This would also cover the empty dict case.

pyaedt/modules/Material.py
```diff
@@ -2185,6 +2193,8 @@ def set_coreloss_at_frequency(
         """
         if not isinstance(points_list_at_freq, dict):
             raise TypeError("Points list at frequency must be provided as a dictionary.")
+        if len(points_list_at_freq) == 1 and core_loss_model_type == "Power Ferrite":
```
SMoraisAnsys, line 2196:
```suggestion
        if len(points_list_at_freq) <= 1 and core_loss_model_type == "Power Ferrite":
```

SMoraisAnsys: Same as previous comment

pyaedt/modules/Material.py
```diff
@@ -2194,9 +2204,14 @@ def set_coreloss_at_frequency(
...
-                {"property_type": "ChoiceProperty", "Choice": "Electrical Steel"}
-            )
+            if core_loss_model_type == "Electrical Steel":
+                self._props["core_loss_type"] = OrderedDict(
+                    {"property_type": "ChoiceProperty", "Choice": "Electrical Steel"}
+                )
+            else:
+                self._props["core_loss_type"] = OrderedDict(
+                    {"property_type": "ChoiceProperty", "Choice": "Power Ferrite"}
+                )
```
SMoraisAnsys, line 2214:
```suggestion
            choice = "Electrical Steel" if core_loss_model_type == "Electrical Steel" else ""Power Ferrite"
            self._props["core_loss_type"] = OrderedDict(
                {"property_type": "ChoiceProperty", "Choice": choice}
            )
```

SMoraisAnsys (changes requested):
Here are some changes I would suggest.

_unittest/test_08_Primitives3D.py
```diff
@@ -701,12 +701,11 @@ def test_44_create_polyline_basic_segments(self):
...
         )
-        try:
+        with pytest.raises(ValueError) as execinfo:
             prim3D.create_polyline(position_list=test_points[0:3], segment_type="Spline", name="PL03_spline_str_3pt")
-        except ValueError as e:
-            assert str(e) == "The position_list argument must contain at least 4 points for segment of type Spline."
-        else:
-            assert False
+            assert (
+                str(execinfo) == "The position_list argument must contain at least 4 points for segment of type Spline."
```
PipKat, line 707:
```suggestion
                str(execinfo) == "The `position_list` argument must contain at least four points for segment of type `Spline`."
```
Usually single quotes surround code entities in message strings.

_unittest/test_08_Primitives3D.py
```diff
@@ -1023,12 +1022,9 @@ def test_54a_create_spiral_and_add_segments(self):
...
         p11 = polyline_points[11]
         position_lst = [[-142, 130, 0], [-126, 63, 0], p11]
-        try:
+        with pytest.raises(ValueError) as execinfo:
             ind.insert_segment(position_lst, "Arc")
-        except ValueError as e:
-            assert str(e) == "Adding the segment result in an unclassified object. Undoing operation."
-        else:
-            assert False
+            assert str(execinfo) == "Adding the segment result in an unclassified object. Undoing operation."
```
PipKat, line 1027: Should the error be "Cannot add the segment result ...."?

_unittest/test_20_HFSS.py
```diff
@@ -337,12 +335,10 @@ def test_06c_create_linear_step_sweep(self):
                 step_size=step_size,
                 sweep_type="Incorrect",
             )
-        except AttributeError as e:
-            exception_raised = True
             assert (
-                e.args[0] == "Invalid value for `sweep_type`. The value must be 'Discrete', 'Interpolating', or 'Fast'."
+                execinfo.args[0]
+                == "Invalid value for `sweep_type`. The value must be 'Discrete', 'Interpolating', or 'Fast'."
```
PipKat, line 340:
```suggestion
                == "Invalid value for 'sweep_type'. The value must be 'Discrete', 'Interpolating', or 'Fast'."
```
Here the first was in double backticks so changed to single quote for consistency.

_unittest/test_20_HFSS.py
```diff
@@ -1208,12 +1195,9 @@ def test_49_port_creation_exception(self):
...
         self.aedtapp.solution_type = "Modal"
         # Spiral lumped port can only be created in a 'Terminal' solution.
-        try:
+        with pytest.raises(Exception) as execinfo:
             self.aedtapp.create_spiral_lumped_port(box1, box2)
-        except Exception as e:
-            exception_raised = True
-            assert e.args[0] == "This method can be used only in Terminal solutions."
-        assert exception_raised
+            assert execinfo.args[0] == "This method can be used only in Terminal solutions."
```
PipKat, line 1200:
```suggestion
            assert execinfo.args[0] == "This method can be used only in 'Terminal' solutions."
```

_unittest/test_41_3dlayout_modeler.py
```diff
@@ -448,12 +447,10 @@ def test_18b_create_linear_step_sweep(self):
                 sweep_type="Incorrect",
                 save_fields=True,
             )
-        except AttributeError as e:
-            exception_raised = True
             assert (
-                e.args[0] == "Invalid value for `sweep_type`. The value must be 'Discrete', 'Interpolating', or 'Fast'."
+                execinfo.args[0] == "Invalid value for `sweep_type`. The value must be 'Discrete', "
```
PipKat, line 451:
```suggestion
                execinfo.args[0] == "Invalid value for 'sweep_type'. The value must be 'Discrete', "
```

SMoraisAnsys (approved):

Samuelopez-ansys (approved):
LGTM

PR #4019, opened 2023-12-29
Feat/4010 pass primitive features

dcrawforAtAnsys:


dcrawforAtAnsys:
@PipKat : I've added ``**kwargs`` as an argument when primitives are created thereby making it possible to set a property when the primitive is instantiated.  This will simplify syntax and improve consistency as some methods use explicit keyword arguments that have inconsistent naming. For example ``non_model=True`` which is used to set the property ``model`` to ``False``. There is, however, a conflict between PEP and IronPython syntax. 
```python
   **kwargs,
   )
```
is allowed for PEP but causes the IronPython unit test to fail. If I fix this so IronPython passes, then we fail the style check.

pyaedt/modeler/cad/Primitives2D.py
```diff
@@ -182,6 +196,7 @@ def create_rectangle(self, position, dimension_list, is_covered=True, name=None,
         ...                                          name="MyCircle", matname="Copper")
 
         """
+        # TODO: Primitives in Maxwell 2D must have Z=0, otherwise the transparency cannot be changed. (issue 4071)
```
gmalinve, line 199: @dcrawforAtAnsys regarding the transparency issue I'd put a check in _default_object_attributes because we set the transparency either 0.2 or 0.8 depending the material name, so here I'd add an additional check for the coordinates.

maxcapodi78 (approved):
Nice improvement

dcrawforAtAnsys:
Allow primitive properties to be set on instantiation by passing named arguments.

PR #4031, opened 2024-01-05
pyvista parallel projection

ring630:
enable_parallel_projection() gives a traditional view of the object.

Traditional view
image

compare to 
image


MaxJPRey (approved):
LGTM.

PR #4035, opened 2024-01-08
Path length correction

svandenb-dev:
Adjusting path length computation with taking in account arcs and cap end style. Values has been compared with the one from aedt and both are matching.


SMoraisAnsys (approved):
LGTM

ring630 (approved):
LGTM

PR #4043, opened 2024-01-09
Ipc exporter false error issue

svandenb-dev:
Fixing false error messages when padstack definitions are not defined


ring630 (approved):
LGTM

PR #4052, opened 2024-01-09
Component is top

svandenb-dev:
Component.is_top property added.


pyaedt/edb_core/edb_data/components_data.py
```diff
@@ -796,6 +796,20 @@ def placement_layer(self):
...
+
+        Returns
+        -------
+        bool
+            ``True`` component is placed at top ``False`` at bottom.
+
+        """
+        if self.placement_layer == list(self._pedb.stackup.signal_layers.values())[0].name:
+            return True
+        return False
```
ring630, line 811: It returns False if the component is place on layer 2. It conflicts with the docstring.

ring630 (approved):

pyaedt/edb_core/edb_data/components_data.py
svandenb-dev: right let me fix this

pyaedt/edb_core/edb_data/components_data.py
```diff
@@ -796,6 +796,21 @@ def placement_layer(self):
         """
         return self.edbcomponent.GetPlacementLayer().Clone().GetName()
 
+    @property
+    def is_top(self):
+        """Check if a component is placed at top layer.
```
MaxJPRey, line 801:
```suggestion
        """Check if a component is placed on the top layer.
```

MaxJPRey (approved):

pyaedt/edb_core/edb_data/components_data.py
svandenb-dev: thanks I changed property name and description to avoid confusion

PR #4057, opened 2024-01-11
group_name property gets updates from the UI

Alberto-DM:
Refactored object3d group_name property to be updated in real time if changes are done in the UI.


Samuelopez-ansys (approved):
LGTM

PR #4059, opened 2024-01-11
set up Edb from json configure file

ring630:
New class to support JSON as a configure file to set up Edb. The schema of JSON is larged borrowed from 3DLayout XML schemas.


pyaedt/edb_core/configuration.py
```diff
@@ -0,0 +1,201 @@
...
+            Config dictionary.
+
+        Examples
+        --------
+        >>> from pyaedt import Edb
+        >>> edbapp = Edb("ansys_pcb.aedb")
+        >>> edbapp.configuration.load("configure.json")
+        """
+
+        if isinstance(config_file, SimulationConfiguration):
```
svandenb-dev, line 34: argument for load method is called config_file but can also accept SimulationConfiguration object. I would rename the argument. Can we rather embed the new json file into SimulationConfig class (inherrited) so that we handle only one json for all ?

ring630: @svandenb-dev as we discussed, I will remove SimulationConfig support from this method.

svandenb-dev (approved):

PR #4061, opened 2024-01-11
Create port on pin name selection adjustment

svandenb-dev:
Fixing pin selection in component.create_port_on_pin() for instance pin A1 and A15 can be wrongly both selected.


SMoraisAnsys (approved):
LGTM

MaxJPRey (approved):
LGTM.

PR #4069, opened 2024-01-12
Add q3d dynamic link component

gmalinve:


pyaedt/twinbuilder.py
```diff
@@ -309,5 +312,238 @@ def set_sim_setup_parameter(self, var_str, expression, analysis_name="TR"):
...
+            The default value is ``1meter``
+        maximum_order : float, optional
+            The Maximum Order value specifies the highest order state space
+            system that you can choose while fitting to represent the system behavior.
+            A lower order may lead to less accuracy but faster simulation.
+            The default value is ``10000``.
+        error_tolerance : float, optional
+            Error Tolerance sets the error tolerance for S-Matrix fitting.
+            The default value is ``0.005``.
+        z_ref : float,optional
```
Samuelopez-ansys, line 357:
```suggestion
        z_ref : float, optional
```

Samuelopez-ansys (approved):

pyaedt/q3d.py
```diff
@@ -1423,7 +1423,12 @@ def net_sinks(self, net_name):
                 net_id = i.props.get("ID", None)  # pragma: no cover
                 break  # pragma: no cover
         for i in self.boundaries:
-            if i.type == "Sink" and i.props.get("Net", None) == net_name or i.props.get("Net", None) == net_id:
+            if (
+                i.type == "Sink"
+                and i.props.get("Net", None) == net_name
+                or i.type == "Sink"
+                and i.props.get("Net", None) == net_id
+            ):
```
SMoraisAnsys, line 1431:
```suggestion
            if (
                i.type == "Sink"
                and any(map(lambda val: i.props.get("Net", None) == val, [net_name, net_id]))
            ):
```

pyaedt/twinbuilder.py
```diff
@@ -309,5 +312,258 @@ def set_sim_setup_parameter(self, var_str, expression, analysis_name="TR"):
...
+        ...                             "sweep1",
+        ...                             "Original",
+        ...                             "100mm")
+        >>> tb.release_desktop()
+        """
+        dkp = self.desktop_class
+        is_loaded = False
+        app = None
+        if os.path.isfile(source_project):
+            if os.path.exists(source_project):
```
SMoraisAnsys, line 404: If you valid os.path.isfile i assume that the second if statement isn't needed ?

pyaedt/twinbuilder.py
```diff
@@ -309,5 +312,258 @@ def set_sim_setup_parameter(self, var_str, expression, analysis_name="TR"):
...
+                project_name = os.path.splitext(os.path.basename(source_project))[0]
+                if project_name in dkp.project_list():
+                    app = dkp[[project_name, source_design_name]]
+                else:
+                    app = dkp.load_project(project_path, source_design_name)
+                    is_loaded = True
+        elif source_project in self.project_list:
+            project_name = source_project
+            project_path = os.path.join(self.project_path, project_name + ".aedt")
+            project_path = project_path.replace("/", "\\")
```
SMoraisAnsys, line 415: Doesn't that break with linux ?

pyaedt/twinbuilder.py
```diff
@@ -309,5 +312,258 @@ def set_sim_setup_parameter(self, var_str, expression, analysis_name="TR"):
...
+            app = dkp[[source_project, source_design_name]]
+        else:
+            raise ValueError("Invalid project name or path provided.")
+        if app is None:
+            raise ValueError("Invalid project or design name.")
```
SMoraisAnsys, line 420: This statement should be removed since it's not reachable with previous if, elif, else statements

gmalinve: @Alberto-DM correct me if I'm wrong but this has to be kept not just because project or design name is invalid.

pyaedt/twinbuilder.py
gmalinve: I don't know honestly

Alberto-DM: I see that it is different from what we saw together. 
@SMoraisAnsys the recording shows that in windows the backslash "\" are escaped in the string itself, so they should be "\\\\" four of them in python. 
Maybe everything works also with a single slash "/", in this case no escape is necessary and we don't have to worry for Linux. 
@gmalinve can you try if it works with a single forward slash "/" in the path? Thanks

pyaedt/twinbuilder.py
Alberto-DM: @SMoraisAnsys Desktop class `__getitem__` can return `None` if the project is not open or the specified design is not in the project. So we have to check for that.

pyaedt/twinbuilder.py
```diff
@@ -309,5 +312,257 @@ def set_sim_setup_parameter(self, var_str, expression, analysis_name="TR"):
...
+                app = dkp.load_project(project_path, source_design_name)
+                is_loaded = True
+        elif source_project in self.project_list:
+            project_name = source_project
+            project_path = os.path.join(self.project_path, project_name + ".aedt")
+            project_path = project_path.replace("/", "\\")
+            app = dkp[[source_project, source_design_name]]
+        else:
+            raise ValueError("Invalid project name or path provided.")
+        if app is None:
```
Alberto-DM, line 418:
```suggestion
        if app is None:  # pragma: no cover
```
Codecov will not complain. 
Or, better, test this branch asking for a design that is not included in your example project.

pyaedt/twinbuilder.py
```diff
@@ -309,5 +312,257 @@ def set_sim_setup_parameter(self, var_str, expression, analysis_name="TR"):
...
+            raise ValueError("Invalid project or design name.")
+        setup = [s for s in app.setups if s.name == setup_name]
+        if not setup:
+            raise ValueError("Invalid setup in selected design.")
+        else:
+            if [sweep for sweep in setup[0].sweeps if sweep.name == sweep_name]:
+                coupling_solution_name = "{} : {}".format(setup_name, sweep_name)
+            else:
+                raise ValueError("Invalid sweep name.")
+        if not [m for m in app.matrices if m.name == coupling_matrix_name]:
```
Alberto-DM, line 428: Consider testing this if clause. If it is not possible, then add `# pragma: no cover`

pyaedt/twinbuilder.py
```diff
@@ -309,5 +312,257 @@ def set_sim_setup_parameter(self, var_str, expression, analysis_name="TR"):
...
+            q3d_model_type = 1
+            ref_pin_style = 3
+            enforce_passivity = True
+        elif state_space_dynamic_link_type == "EQ":
+            state_space_dynamic_link_type = "{}SmlLink".format(design_type)
+            q3d_model_type = 0
+            ref_pin_style = 3
+            enforce_passivity = True
+        else:
+            raise TypeError("Link type is not valid.")
```
Alberto-DM, line 492: Consider testing this part also.

Alberto-DM (changes requested):
LGTM

pyaedt/twinbuilder.py
SMoraisAnsys: @Alberto-DM @gmalinve There might be a misunderstanding here, I just wanted to pinpoint the replace call. The call `os.path.join` is compatible with both windows and linux so I though that using replace would break the linux compatibility. However, if there is something specific to `self.project_path` that might need such a modification, it might be interesting to only perform the replace on `self.project_path`. Forgive my ignorance if my remark is unfounded.

```python
# In windows
path = 'C:\\Users\\ansys'
res = os.path.join(path, 'some_dir') # res would be equal to 'C:\\Users\\ansys\\some_dir'
# In linux
path = '/home/ansys'
res = os.path.join(path, 'some_dir') # res would be equal to '/home/ansys/some_dir'
```

pyaedt/twinbuilder.py
SMoraisAnsys: > @SMoraisAnsys Desktop class `__getitem__` can return `None` if the project is not open or the specified design is not in the project. So we have to check for that.

@Alberto-DM Thanks, I didn't took into account this fact.

pyaedt/twinbuilder.py
gmalinve: @SMoraisAnsys @Alberto-DM tests pass anyway so I'll remove the replace method and leave it as is. Thanks both for feedback! :)

Alberto-DM (approved):
LGTM

SMoraisAnsys (approved):
LGTM

PR #4079, opened 2024-01-15
Edb config json enhancement

ring630:
1. setup refactor
2. support siwave syz and dc
3. fix bug in stackup.load


svandenb-dev (approved):
LGTM

PR #4083, opened 2024-01-16
component.create() metod rlc values and location bug fix

svandenb-dev:
component.create had some issue:
- component wrong value
- wrong location
This PR is resolving these issues


Samuelopez-ansys (approved):
LGTM

PR #4088, opened 2024-01-16
Fix ac example

Samuelopez-ansys:


MaxJPRey (approved):
LGTM.

examples/00-EDB/15_ac_analysis.py
```diff
@@ -111,20 +133,34 @@
 # ~~~~~~~~~~~~~~~~~~~
 # To do SYZ analysis, we need to launch HFSS 3D Layout and import EDB into it.
```
PipKat, line 134:
```suggestion
# To do SYZ analysis, you must launch HFSS 3D Layout and import EDB into it.
```
You want to use "you" and not "we".

PR #4091, opened 2024-01-17
Update full_documentation.yml

Samuelopez-ansys:


SMoraisAnsys (approved):
Don't know how i missed that in my previous PR. Good catch.
This should solve our documentation issue !

PR #4100, opened 2024-01-18
Icepak csv example improvement

lorenzovecchietti:
I improve the example with these modifications:

- remove 2R network assignment function from the example to use the one already present in pyAEDT
- remove the board side column from the csv file and remove the line for the board creation (https://github.com/ansys/example-data/pull/220)
- remove and substitution of deprecated methods
- coding styles improvements


Samuelopez-ansys (approved):
LGTM

PR #4103, opened 2024-01-19
issue-4102-reverting-implementation-in-get_scalar_field_value

nnetake:
The simplified implementation GetTopEntryValue works fine in small models but it is failing in the big model (mesh ~27M). Therefore I am reverting this implementation to the previous one with temp file exporting and reading it to extract value which we verified that works fine with bigger models. GetTopEntryValue was working fine with versions <=22R2 but something happened in later versions on the native scripting side.


Samuelopez-ansys (approved):
LGTM

lorenzovecchietti (approved):

PR #4114, opened 2024-01-22
Remove duplicate setup keys for IcepakSteadyState.

MaxJPRey:


SMoraisAnsys (approved):
LGTM

Samuelopez-ansys:
@MaxJPRey @SMoraisAnsys 

I think we need to keep them.
They are there because Icepak has 3 different setups fro Transient (36,37,38)

MaxJPRey:
@Samuelopez-ansys @lorenzovecchietti OK, got it.

PR #4116, opened 2024-01-22
TEST: add test on create polygon

svandenb-dev:
Bug fix on create_polygon method


SMoraisAnsys:
@svandenb-dev Just to be sure, the fix you propose is only related to a test right ? It isn't related to the implementation of method `create_polygon` ? I'm asking this because the PR title made me expect a modification in the method, not in the test.

If that's what you wanted to do, the changes look good to me. Can you wait for #4118 to be merged ? This PR should fix the CI issue that we are currently having.

SMoraisAnsys (approved):
LGTM

MaxJPRey (approved):
LGTM.

PR #4118, opened 2024-01-23
MAINT: fix CI by fixing scipy and scikit-rf versions

SMoraisAnsys:
There is a side effect on scikit-rf 0.30.0 as it uses scipy without bounding the version and that method linspace, used by scikit-rf is no longuer available through scipy import in scipy 1.12.0. (see https://github.com/scikit-rf/scikit-rf/issues/1005)

**Note**: this should be 'reverted' later on to use scikit-rf new version as it should fix the current issue.

**Extra note**: a commit revert is also added to this PR as it was added on main by mistake and is failing (see https://github.com/ansys/pyaedt/commit/37218a5c6d1da2b8149a40abf8e4655c0011d2a3)


MaxJPRey (approved):
LGTM.

Issue #4001, opened 2023-12-21
Bug located in ...get_object_material_properties


tungnguyen1003:
### Before submitting the issue

- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### Description of the bug

Test command (did not work)
hfss.get_object_material_properties()

Found issue (worked)
hfss.materials
hfss.get_object_material_properties()

### Steps To Reproduce

As in description

### Which Operating System are you using?

Windows

### Which Python version are you using?

3.10

### Installed packages

pyaedt latest version

SMoraisAnsys:
Hi @tungnguyen1003, thanks for the feedback. To make it easier for us to identify and resolve the bug, please answer the following question.

What do you mean by:

Test command (did not work)
hfss.get_object_material_properties()

Found issue (worked)
hfss.materials
hfss.get_object_material_properties()

Does the second part is a workaround to make `get_object_material_properties` work ?

Could you provide us with a simple reproducer to show the bug ?

SMoraisAnsys:
@tungnguyen1003 Hi and happy new year, could you please answer the previous comment ? This would help us to provide you with the right answer.

tungnguyen1003:
Hi

Sorry for my late reply.
Also, my sincere apologies for the lack of explanation on the issue.

We thought that the issue was not with the command itself, but with the explanation on how to make it work, described in the below URL.
https://aedt.docs.pyansys.com/version/0.7/API/_autosummary/pyaedt.hfss.Hfss.get_object_material_properties.html

As you can see in the screenshot below, at first, we opened a simple test design and tested the command, but errors popped up. Then we had to look deeply in the code for this command and got it up and running by adding “app.materials” before calling this command.

We assumed that the project materials had to be loaded before this generic object-materials command can be used. It took me more than 30 minutes to figure out this simple process, so I thought a little more explanation would be very helpful. That s all.

***@***.***

Thank you for reaching out.

Happy New Year
**------------------------------------------------------------------------**
Tung Nguyen (トゥン　グェン), Ph.D. / Lead Tech. Support Engineer - RF Modeling, Electronics BU, ACE Japan  / Technical Division
***@***.***<https://www.ansys.com/>
アンシス・ジャパン株式会社 / 〒160-0023 東京都新宿区西新宿6-10-1 / 日土地西新宿ビル18F
tel: 03-5324-7357 / fax : 03-5324-7308
[...]

PR #4147, opened 2024-01-29
MISC: add logger error when object not found

SMoraisAnsys:
Following user feedback (#4001) this PR adds an extra error log to clarify the fact that some calls might not work as expected since some objects don't exist (or at least are not found)

### DIFF
--- a/pyaedt/modeler/cad/Primitives.py
+++ b/pyaedt/modeler/cad/Primitives.py
@@ -92,6 +92,7 @@ class GeometryModeler(Modeler):
             return self.user_defined_components[partId]
         elif isinstance(partId, Object3d) or isinstance(partId, UserDefinedComponent):
             return partId
+        self.logger.error("Object '{}' not found.".format(partId))
         return None
 
     def __init__(self, app, is3d=True):


Samuelopez-ansys (approved):
LGTM

Samuelopez-ansys:
@SMoraisAnsys Maybe we should add a test

SMoraisAnsys:
> @SMoraisAnsys Maybe we should add a test

I added a test over the log message content. Is that what you wanted or had you something else in mind ?

_unittest/test_14_AedtLogger.py
```diff
@@ -219,6 +219,21 @@ def test_05_disable_stdout(self):
         assert stream_content[2] == "PyAEDT INFO: Info after re-enabling the stdout handler.\n"
 
 
+class TestLogMessages:
+    """Class used to test log messages."""
+
+    def test_log_when_accessing_non_existing_object(self, caplog):
+        """Check that accessing non existing object"""
```
MaxJPRey, line 226:
```suggestion
        """Check that accessing non existing object log an error message."""
```

MaxJPRey (approved):
LGTM.

Samuelopez-ansys (approved):
LGTM

PR #4188, opened 2024-02-05
improve m2d DC Conduction example - pdf report

gmalinve:


examples/03-Maxwell/Maxwell2D_DCConduction.py
```diff
@@ -178,6 +195,43 @@
...
+# Export model picture
+# ~~~~~~~~~~~~~~~~~~~~
+# Export model picture.
+
+model_picture = m2d.post.export_model_picture()
+
+################################################################################
+# Generate pdf report
+# ~~~~~~~~~~~~~~~~~~~
+# Generate a pdf report with output of simulation.
```
PipKat, line 208:
```suggestion
# Generate PDF report
# ~~~~~~~~~~~~~~~~~~~
# Generate a PDF report with output of simulation.
```
Rule is to place file extension letters in all capital letters (and also not to use something like *.pdf)

Samuelopez-ansys (approved):
LGTM

PR #4226, opened 2024-02-13
Fix & Improvement to parametric heatsink function

lorenzovecchietti:
New function allows multiple parametric heatsink with any position and orientation, automatically placing it on top of a face. It is also possible to express dimensions relative to the base dimensions.

The new function also fixes a few bugs from the old one.


maxcapodi78:
> New function allows multiple parametric heatsink with any position and orientation, automatically placing it on top of a face. It is also possible to express dimensions relative to the base dimensions.
> 
> The new function also fixes a few bugs from the old one.

Hi @lorenzovecchietti is it possible to derive deprecated method from the new one in order to reduce amount of code to tests and mantain?

pyaedt/icepak.py
```diff
@@ -1450,6 +1450,9 @@ def create_parametric_heatsink_on_face(
         -------
         :class:`pyaedt.modeler.cad.object3d.Object3d`
             Heatsink created or ``False`` when failed.
+        dict
+            Variable mapping. Keys are the different parameters names and values
```
PipKat, line 1350:
```suggestion
            Variable mapping. Keys are the different parameters names, and values
```

pyaedt/icepak.py
```diff
@@ -1256,130 +1256,359 @@ def create_parametric_fin_heat_sink(
...
+        """Create a parametric heat sink.
+
+        Parameters
+        ----------
+        top_face : modeler.cad.elements3d.FacePrimitive
+            Face on which to build the heatsink.
+        relative : bool, optional
+            Whether the dimensions used as arguments of the function are
+            absolute or relative to the width and the height of the
+            top_face.
```
PipKat, line 1311:
```suggestion
            top face.
```

pyaedt/icepak.py
```diff
@@ -1256,130 +1256,359 @@ def create_parametric_fin_heat_sink(
...
+        hs_basethick : float, optional
+            Thickness of the heat sink base. If ``relative==True``, it is the
```
PipKat, line 1313: why are the double equal signs required here in these statements: ``relative==True``?

lorenzovecchietti: We can alternatively write
`If ``relative`` is True`

== is the comparison operator, just one = is the assignment operator

PipKat: @lorenzovecchietti I never knew this! If its technically correct, then leave it! (I hope I've never incorrectly changed it to one equal sign over the past two or three years and not have someone catch it!)

Samuelopez-ansys (approved):

PR #4231, opened 2024-02-14
set solve_inside to False for pec, added a unit test

IreneWoyna:


_unittest/test_08_Primitives3D.py
```diff
@@ -1308,6 +1308,9 @@ def test_67_assign_material(self):
         assert self.aedtapp.modeler[cyl1].material_name == "aluminum"
         assert self.aedtapp.modeler[cyl2].material_name == "aluminum"
 
+        assert self.aedtapp.modeler[box1].material_name == "pec"
```
gmalinve, line 1311: @IreneWoyna remove these assert statements from here. They will fail in Primitives3D

_unittest/test_28_Maxwell3D.py
```diff
@@ -75,6 +75,9 @@ def test_01_create_primitive(self):
         assert plate.solve_inside
         assert plate.material_name == "aluminum"
 
+        assert self.aedtapp.assign_material(plate, "pec")
```
gmalinve, line 78: @IreneWoyna other than asserting if the method is successful, it's better to assert if the property solve inside is now set to the expected valuen

IreneWoyna: sure, makes sense, I changed it:

self.aedtapp.assign_material(plate, "pec")
assert not plate.solve_inside

gmalinve (approved):

PR #4232, opened 2024-02-14
Update ci 2024 r1

### LINKED ISSUE #4205: Update CI unittest to 2024R1

Samuelopez-ansys:
Pull request to pass CI to 2024R1

Close #4205


SMoraisAnsys (changes requested):
Could you please move the skip method to the conftest.py file ?

Could we also define a constant AEDT_VERSION for the examples ? This would avoid us the pain of having to do the version changes manually every new AEDT release.

PR #4233, opened 2024-02-14
New feature from 24.1 - via_mesh_plating

ring630:


Samuelopez-ansys (changes requested):
@ring630 THe unittest can not be changed to 2024.1 for now, I am working in the transition, so please roll back the modification in conftest.py

If the property you created only works for 2024.1, you should add in the property an "if" to check the version, and if the version is lower than 2024.1, return a warning. For the unit test the same, check the version because I guess in 2023.2, the test will fail

_unittest/test_00_EDB.py
```diff
@@ -2146,12 +2146,14 @@ def test_129_hfss_simulation_setup(self):
...
         via_settings = setup1.via_settings
         via_settings.via_density = 1
+        via_settings.via_mesh_plating = True
         via_settings.via_material = "pec"
         via_settings.via_num_sides = 8
         via_settings.via_style = "kNum25DViaStyle"
 
         via_settings = edbapp.setups["setup1"].via_settings
         assert via_settings.via_density == 1
+        assert not via_settings.via_mesh_plating
```
Samuelopez-ansys, line 2156: here I would add an if 2024.1 then assert is True right? then we do not need to modify the unitest once we migrate to the new version. The version is already there, because it is imported: from _unittest.conftest import desktop_version

Samuelopez-ansys (changes requested):

MaxJPRey (approved):
LGTM.

Samuelopez-ansys (approved):

PR #4234, opened 2024-02-14
added erl parameter to VirtualCompliance

maxcapodi78:
Add config support to erl calculation


pyaedt/circuit.py
```diff
@@ -1692,32 +1692,36 @@ def import_edb_in_circuit(self, edb_path):
...
+        modulation_type=None,
     ):
         """Compute effective return loss (erl) using Ansys SPISIM from a S-parameter file.
 
         Parameters
         ----------
         touchstone_file : str
             Path to the touchstone file.
+        config_file : str, optional
+            Config file to be used as reference. This is optional and it can be skipped.
```
PipKat, line 1718:
```suggestion
            Configuration file to use as a reference. The default is ``None``, in
            which case this parameter is ignored.
```

pyaedt/circuit.py
```diff
@@ -1854,6 +1885,8 @@ def compute_erl(
             Name of the setup to use as the nominal. The default is
             ``None``, in which case the active setup is used or
             nothing is used.
+        config_file : str, optional
+            Config file to be used as reference. This is optional and it can be skipped.
```
PipKat, line 1889:
```suggestion
            Configuration file to use as reference. The default is ``None``, in which
            case this parameter is ignored.
```

pyaedt/generic/compliance.py
```diff
@@ -109,7 +110,7 @@ def add_aedt_report(self, name, report_type, config_file, design_name, traces, s
         name : str
             Name of the report.
         report_type : str
-            Report type. It can be "frequency", "eye diagram", "statistical eye", "contour eye diagram", "transient".
+            Report type. It can be "frequency", "eye diagram", "statistical eye", "contour eye diagram", "time".
```
PipKat, line 113:
```suggestion
            Report type. Options are "contour eye diagram", "eye diagram", "frequency", "statistical eye", and "time".
```
Should be an alphabetized list.

Samuelopez-ansys (dismissed):
Great

pyaedt/generic/spisim.py
```diff
@@ -0,0 +1,208 @@
...
+        Parameters
+        ----------
+        config_file : str, optional
+            Configuration file to use as a reference. The default is ``None``, in
+            which case this parameter is ignored.
+        port_order : str, optional
+            Whether to use "``EvenOdd``" or "``Incremental``" numbering for ``s4p`` files.
+            Ignored if the ports are greater than 4.
+        specify_through_ports : list, optional
+            Input and Output ports on which compute the erl. Those are ordered like ``[inp, inneg, outp, outneg]``.
```
PipKat, line 111:
```suggestion
            Input and output ports to compute the erl on. Those are ordered like ``[inp, inneg, outp, outneg]``.
```
What is "erl"? should it be ERL or spelled out.

pyaedt/generic/spisim.py
```diff
@@ -0,0 +1,208 @@
...
+            Ignored if number of ports are 4.
```
PipKat, line 112:
```suggestion
           The default is ``None``. This parameter is ignored if there are more than four ports.
```

pyaedt/generic/spisim.py
```diff
@@ -0,0 +1,208 @@
...
+        bandwidth : float, str, optional
+            Application bandwidth: inverse of one UI (unit interval). Can be a float or str with unit ("m", "g").
+            Default is ``30e9``.
```
PipKat, line 115:
```suggestion
            Application bandwidth, which is the inverse of one UI (unit interval). The value
            can be a float or a string with the unit ("m", "g"). The default is ``30e9``.
```
It would be helpful if the string examples included numbers to show formatting, something like this: ``"1m"`` or ``50g``. However, I don't know what realistic examples would be.

pyaedt/generic/spisim.py
```diff
@@ -0,0 +1,208 @@
...
+        tdr_duration : float, optional
+            TDR duration (in second): How long the TDR tailed data should be applied. Default is ``5``.
```
PipKat, line 117:
```suggestion
            TDR duration in seconds, meaning how long the TDR tailed data should be applied.
            The default is ``5``.
```
TDR should be spelled out before using this abbreviation, to ensure that the user knows what TDR is. (I know that I don't.)

pyaedt/generic/spisim.py
```diff
@@ -0,0 +1,208 @@
...
+        z_terminations : float, optional
+            Z-Terminations: termination (Z11 and Z22) when TDR is calculated. Default is ``50``.
```
PipKat, line 119:
```suggestion
            Z-terminations (Z11 and Z22) when TDR is calculated. The default is ``50``.
```
We don't want to use the colon within the description as is often done in this section.  I'm not going to comment on the following--but please revise them all as per my previous instructions. Again, the biggest issues are that the defaults    in the descriptions don't' match those shown in the method signature. Don't use a colon in a description. Don't use abbreviations without first defining them. Always use complete sentences after the first brief description.

pyaedt/generic/spisim.py
```diff
@@ -0,0 +1,208 @@
...
+        fixture_delay : float, optional
+            Fixture delay: delay when input starts transition from 0 to Vcc. Default is ``500e-12``.
+        input_amplitude : float, optional
+            Input amplitude: Vcc volt of step input. Default is ``1.0``.
+        ber : float, optional
+            Specified BER: At what threshold ERL is calculated. Default is ``1e-4``.
+        pdf_bin_size : float, optional
+            PDF bin size: how to quantize the superimposed value. Default is ``1e-5``.
+        signal_loss_factor : float, optional
+            Signal loss factor (Beta). See SPISIM Help for info. Default is ``1.7e9``.
```
PipKat, line 131:
```suggestion
            Signal loss factor (Beta). For more information, see the SPISIM Help. The default is ``1.7e9``.
```
Here, should it be SPISIM documentation? Can we add a link? You always want to put the reason to click a link before the link.

pyaedt/generic/spisim.py
```diff
@@ -0,0 +1,208 @@
...
+        permitted_reflection : float, optional
+            Permitted reflection (Rho). See SPISIM Help for info. Default is ``0.18``.
+        reflections_lenght : float, optional
```
PipKat, line 134: Length is misspelled in the parameter name? The description needs to be something like: Length of the reflections, which are used to calculate ERL. The default is ...." Here, ERL is capitalized. We need to define on the first reference and use the abbreviation or acronym consistently thereafter.

pyaedt/generic/spisim.py
Samuelopez-ansys: It is defined in the definition of the method _Compute effective return loss (erl)_

pyaedt/generic/spisim.py
Samuelopez-ansys: The default values are loaded from the cfg_dict, if the user does not define any of these parameters, it will use the default ones defined on this dictionary

pyaedt/generic/spisim.py
Samuelopez-ansys: it is

Samuelopez-ansys (approved):

Samuelopez-ansys (approved):
Looks really great

Samuelopez-ansys:
@ring630 @maxcapodi78 Is it ready to merge?

ring630:
> @ring630 @maxcapodi78 Is it ready to merge?

@Samuelopez-ansys  just wait a bit more. @maxcapodi78 has somethong more to add.

PR #4239, opened 2024-02-15
fix: declare regex as raw string to avoid escape sequence deprecation

JonathanRayner:
Declare regex as raw string to avoid escape sequence deprecation.


JonathanRayner:
@ansys-reviewer-bot review

Samuelopez-ansys (approved):
LGTM

PR #4285, opened 2024-02-27
CI: Update workflows runs-on

SMoraisAnsys:
Notes: to refact the CI/CD we need to avoid using pyaedt self-hosted runners when not needed. However, the current workflow files leverage "windows-latest" and "pyaedt" which leads to using pyaedt self-hosted runners instead of github runners. Here we remove the use of "windows-latest" when the CI/CD should run on self-hosted runners.

On top of these changes, the self-hosted runners tag are going to be update to remove "windows-latest" from their labels. It does not make sense to keep it and might be confusing with Github runners.

### DIFF
--- a/.github/workflows/full_documentation.yml
+++ b/.github/workflows/full_documentation.yml
@@ -33,7 +33,7 @@ jobs:
   full_documentation:
     # The type of runner that the job will run on
     name: full_documentation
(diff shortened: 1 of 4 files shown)


Samuelopez-ansys (approved):
LGTM

PR #4293, opened 2024-02-28
fix view and plot label in plot_field_from_fieldplot

gmalinve:


Samuelopez-ansys (commented):
Why did you change the name label to scalar_name? Is not possible to keep label?

Samuelopez-ansys (dismissed):

_unittest/test_12_1_PostProcessing.py
```diff
@@ -526,6 +527,19 @@ def test_14_Field_Ploton_cutplanedesignname(self):
...
+            imageformat="jpg",
+            view="xy",
+            show=False,
+            plot_label=plot1.name + " label",
+            file_format="aedtplt",
+        )
+        assert os.path.exists(plot_obj.image_file)
+        plot_obj.plot(plot_obj.image_file)
+
     @pytest.mark.skipif(is_linux or sys.version_info < (3, 8), reason="Not running in ironpython")
```
PipKat, line 543:
```suggestion
    @pytest.mark.skipif(is_linux or sys.version_info < (3, 8), reason="Not running in IronPython.")
```

pyaedt/modules/AdvancedPostProcessing.py
```diff
@@ -377,6 +378,12 @@ def plot_field_from_fieldplot(
             Whether to display the legend or not. The default is ``True``.
         plot_as_separate_objects : bool, optional
             Plot each object separately. It may require more time to export from AEDT.
+        file_format : str, optional
+            File format of the exported plot.
+            Available options are: ``"case"`` or ``"aedtplt"``.
```
PipKat, line 383:
```suggestion
            Whether to display the legend. The default is ``True``.
        plot_as_separate_objects : bool, optional
            Whether to plot each object separately, which can require
            more time to export from AEDT. The default is ``True``.
        file_format : str, optional
            File format to export the plot to. The default is ``"case".
            Options are ``"aedtplt"`` and ``"case"``.
```

Samuelopez-ansys (approved):

PR #4298, opened 2024-02-28
Add lazy load flag

### LINKED ISSUE #4268: Lazy load setting in PyAEDT

Add a setting in PyAEDT to disable lazy load. Now PyAEDT does not load post and modeler objects to avoid long load, but users could like to load this data since the pyaedt object creation. 

_No response_

_No response_

Samuelopez-ansys:
Close #4268

### DIFF
--- a/pyaedt/generic/settings.py
+++ b/pyaedt/generic/settings.py
@@ -67,7 +67,7 @@ class Settings(object):
         self._number_of_grpc_api_retries = 6
         self._retry_n_times_time_interval = 0.1
         self._wait_for_license = False
-        self._lazy_load = True
+        self.__lazy_load = True
 
     @property
     def lazy_load(self):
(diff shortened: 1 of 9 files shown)


maxcapodi78 (dismissed):

pyaedt/generic/settings.py
```diff
@@ -67,6 +67,22 @@ def __init__(self):
...
+
+        Returns
+        -------
+        bool
+        """
+        return self._lazy_load
+
+    @lazy_load.setter
+    def lazy_load(self, value):
+        self._lazy_load = value
```
SMoraisAnsys, line 85:
```suggestion
        self.__lazy_load = True

    @property
    def lazy_load(self):
        """Flag for enabling and disabling the lazy load.
        The default is ``True``.

        Returns
        -------
        bool
        """
        return self.__lazy_load

    @lazy_load.setter
    def lazy_load(self, value):
        self.__lazy_load = value
```
If you don't mind, we could use double underscore `__` to hint that this attribute is private. Don't take it into account if you want to be compliant with the way things are already implemented. However, we should think about refactoring the code to make it more pythonic and this changes would be done during this hypothetical refactoring :)

Samuelopez-ansys: I agree :) So you are volunteering right? ;)

maxcapodi78 (approved):

PR #4309, opened 2024-03-01
improve speedup of Modeler

maxcapodi78:


Samuelopez-ansys (approved):
Good improvement for big projects

Samuelopez-ansys (approved):

PR #4316, opened 2024-03-04
Disable EMIT tests 2024R1

Samuelopez-ansys:


MaxJPRey (dismissed):
LGTM.

SMoraisAnsys (approved):
Do we know what extended this workflow so much ? It ran in self than 30 minutes in the last working run (https://github.com/ansys/pyaedt/actions/runs/8138121240/job/22238148636)

Samuelopez-ansys (commented):
Block PR

pyaedt/desktop.py
```diff
@@ -175,13 +175,22 @@ def _check_grpc_port(port, machine_name=""):
...
+        rand_port = random.randint(50051, 51000)
+        try:
+            sock.bind(("", rand_port))
+            sock.close()
+            return rand_port
+        except OSError:
+            tries += 1
+            if tries == 1000:
+                break
+    raise IOError("No free ports")
```
PipKat, line 193:
```suggestion
    raise IOError("No ports are free.")
```
String should be a complete sentence, including punctuation.

_unittest_solvers/test_26_emit.py
```diff
@@ -18,6 +18,7 @@
 from pyaedt.modeler.circuits.PrimitivesEmit import EmitComponents
 
 test_subfolder = "T26"
+test_review_flag = True
```
MaxJPRey, line 21: I would suggest to rename the constants using upper case naming convention.
```suggestion
TEST_SUBFOLDER = "T26"
TEST_REVIEW_FLAG = True
```

maxcapodi78 (approved):

PR #4317, opened 2024-03-04
Circuit vpwl

### LINKED ISSUE #4315: Add piecewise linear voltage source in Circuit

Create a piecewise linear voltage source in Circuit. 

1. create an V_PWL componnet
2. assign time and voltage series 

_No response_

boyang2022:
close #4315


pyaedt/modeler/circuits/PrimitivesNexxim.py
```diff
@@ -865,6 +865,55 @@ def create_voltage_pulse(self, compname=None, value_lists=[], location=[], angle
...
+        Parameters
+        ----------
+        compname : str, optional
+            Name of the voltage pulse. The default is ``None``.
+        time_list : list, optional
+            List of time points for the pwl voltage source. The default is ``[0]``.
+        voltage_list : list, optional
+            List of voltages for the pwl voltage source. The default is ``[0]``.
+        location : list of float, optional
+            Position on the X axis and Y axis.
```
PipKat, line 883: While I prefer capitalizing X and Y as you have done, in this library (and virtually all libraries), a lower-case x and y are almost always used.

pyaedt/modeler/circuits/PrimitivesNexxim.py
```diff
@@ -865,6 +865,55 @@ def create_voltage_pulse(self, compname=None, value_lists=[], location=[], angle
...
+            compname,
+            component_library="Independent Sources",
+            component_name="V_PWL",
+            location=location,
+            angle=angle,
+            use_instance_id_netlist=use_instance_id_netlist,
+        )
+
+        if len(time_list) != len(voltage_list):
+            raise ValueError("length of time not equal to length of voltage")
```
PipKat, line 909:
```suggestion
            raise ValueError("Length of time is not equal to length of voltage.")
```
Messages should be complete sentences, with verbs and concluding punctuation.

boyang2022:
@Samuelopez-ansys @svandenb-dev 
do I need to fix this problem?
image

pyaedt/modeler/circuits/PrimitivesNexxim.py
boyang2022: thanks!

pyaedt/modeler/circuits/PrimitivesNexxim.py
```diff
@@ -865,6 +865,55 @@ def create_voltage_pulse(self, compname=None, value_lists=[], location=[], angle
...
+            compname,
+            component_library="Independent Sources",
+            component_name="V_PWL",
+            location=location,
+            angle=angle,
+            use_instance_id_netlist=use_instance_id_netlist,
+        )
+
+        if len(time_list) != len(voltage_list):
+            raise ValueError("Length of time is not equal to length of voltage")
```
SMoraisAnsys, line 909: Minor comment: to my knowledge, the current way of doing things in pyaedt is to return `False` or `None` when something goes wrong. @Samuelopez-ansys this line should return either oh them right ?
Note that I think we should raise Exception instead of returning None / False but I think that this should be done in some refactoring of the whole code base. IMO it would be better to be consistent here.

Samuelopez-ansys: Right, @boyang2022  please return False in this case, and also add a logger.error with the message.

Samuelopez-ansys:
@boyang2022 Regarding the Codacy issue, yes you need to fix it. The issue is that you can not have a empty list a default parameter. Put None, and in the first line of the method you can add:

```
if location is None:
    location = []
```

boyang2022:
@Samuelopez-ansys unit test timed out, Any idea?

Error: Timeout of 2400000ms hit

SMoraisAnsys:
@boyang2022 It's not coming from your side. We are aware of the problem and working on making things back on track.

pyaedt/modeler/circuits/PrimitivesNexxim.py
```diff
@@ -865,6 +865,61 @@ def create_voltage_pulse(self, compname=None, value_lists=[], location=[], angle
...
+            component_name="V_PWL",
+            location=location,
+            angle=angle,
+            use_instance_id_netlist=use_instance_id_netlist,
+        )
+
+        if (time_list is not None) and (voltage_list is not None):
+
+            if len(time_list) != len(voltage_list):
+                self.logger.error("Length of time is not equal to length of voltage.")
```
MaxJPRey, line 914:
```suggestion
                self.logger.error("The number of time points is different than the number of voltages.")
```

Samuelopez-ansys (approved):
LGTM

PR #4319, opened 2024-03-05
passivity causality check

svandenb-dev:
Passivity and causality check added using genequiv run.


pyaedt/generic/touchstone_parser.py
```diff
@@ -489,7 +489,7 @@ def check_touchstone_files(folder="", passivity=True, causality=True):
         folder path.
     passivity : bool, optional
         ``True`` passivity check enabled, ``False`` disabled. Default value is ``True``
-    causality
+    causality : bool, optional
         ``True`` causality check enabled, ``False`` disabled. Default value is ``True``
```
PipKat, line 493:
```suggestion
        Whether the causality check is enabled. The default is ``True``.
```

pyaedt/generic/touchstone_parser.py
```diff
@@ -489,7 +489,7 @@ def check_touchstone_files(folder="", passivity=True, causality=True):
         folder path.
     passivity : bool, optional
         ``True`` passivity check enabled, ``False`` disabled. Default value is ``True``
```
PipKat, line 491:
```suggestion
        Whether the passivity check is enabled, The default is ``True``.
```

felipeescastro (approved):

pyaedt/generic/touchstone_parser.py
```diff
@@ -475,3 +477,74 @@ def read_touchstone(file_path):
...
+
+    Parameters
+    ----------
+    folder : str
+        folder path.
+    passivity : bool, optional
+        Whether the passivity check is enabled, The default is ``True``.
+    causality : bool, optional
+        Whether the causality check is enabled. The default is ``True``.
+    Returns : dict
```
Samuelopez-ansys, line 494: Please follow the docstring style

Samuelopez-ansys:
There are also Codacy issues

pyaedt/generic/touchstone_parser.py
```diff
@@ -475,3 +477,74 @@ def read_touchstone(file_path):
     """
     data = TouchstoneData(touchstone_file=file_path)
     return data
+
+
+@pyaedt_function_handler()
+def check_touchstone_files(folder="", passivity=True, causality=True):
+    """Check passivity and causality for all touchstone files included in folder.
```
PipKat, line 484:
```suggestion
    """Check passivity and causality for all Touchstone files included in the folder.
```
I did a Google search and see that Touchstone is a proper noun and so should be capitalized in descriptive text.

pyaedt/generic/touchstone_parser.py
```diff
@@ -475,3 +477,74 @@ def read_touchstone(file_path):
...
+
+    Parameters
+    ----------
+    folder : str
+        folder path.
```
PipKat, line 489:
```suggestion
    folder : str
        Folder path. The default is ``""``.
```
We should indicate what the behavior is when the default is used: "The default is ``""``, in which case the active folder is used." I'm just guessing here at what the default behavior might be, so please revise.

pyaedt/generic/touchstone_parser.py
```diff
@@ -475,3 +477,74 @@ def read_touchstone(file_path):
...
+    passivity : bool, optional
+        Whether the passivity check is enabled, The default is ``True``.
+    causality : bool, optional
+        Whether the causality check is enabled. The default is ``True``.
+    Returns : dict
+        Dictionary with snp file name as key, list for passivity and causality is enabled. First element from list
+        is str with ``"passivity"`` or ``"causality"`` as value. Second element bool with ``True`` when criteria passed
+        ``False`` otherwise. Last element str with the log information.
```
PipKat, line 497:
```suggestion
        Dictionary with the SNP file name as the key and a list if the passivity and/or causality checks are enabled.
        The first element in the list is a str with ``"passivity"`` or ``"causality"`` as a value. The second element
        is a Boolean that is set to ``True`` when the criteria passed or ``False`` otherwise. The last element
        is a string with the log information.
```
It seems that both the passivity and causality checks can be enabled as there is no indication in the parameter descriptions that it can only be one or the other. Consequently, I used "and/or" in the first line of the Returns description. However, the description for the first line seems to imply it is one or another. Please clarify.

pyaedt/generic/touchstone_parser.py
svandenb-dev: What is not correct ?

Samuelopez-ansys: The format, it should be something like:

Returns
 -------
        dict
            ....

felipeescastro: I just changed that on my last commit.

Samuelopez-ansys (approved):
Good job @felipeescastro

PR #4340, opened 2024-03-08
Fix angle in degrees example

Samuelopez-ansys:
It was correct in the documentation ("degrees") but it was not correct until now.


SMoraisAnsys (approved):
LGTM, did you try to run the example ?

PR #4349, opened 2024-03-11
delete unclassified

gmalinve:


gmalinve:
@Samuelopez-ansys @maxcapodi78 
Why do we undo the design if the result of the intersection is an unclassified object?
It's in the delete() method in Primitives.py: 
`unclassified1 = list(self.oeditor.GetObjectsInGroup("Unclassified"))
if unclassified != unclassified1:
    self._odesign.Undo()
    self.logger.error("Error in intersection. Reverting Operation")
    return`
it doesn't necessarily means that the intersect method failed.
In fact in test _test_26_unclassified_object_ we use the native API to be able to get the unclassified objects.

_unittest/test_07_Object3D.py
```diff
@@ -537,6 +537,11 @@ def test_26_unclassified_object(self):
         self.aedtapp.modeler.oeditor.Intersect(vArg1, vArg2)
         assert box1 in self.aedtapp.modeler.unclassified_objects
 
+    def test_26a_delete_unclassified_object(self):
+        unclassified = self.aedtapp.modeler.unclassified_objects
+        assert self.aedtapp.modeler.delete(unclassified)
+        assert len(self.aedtapp.modeler.unclassified_objects) != unclassified
```
SMoraisAnsys, line 543: Shouldn't you test that the value is equal to 0 since the got deleted ?

gmalinve: @SMoraisAnsys yes I could but it's the same at the end. I can add a new line.

Samuelopez-ansys:
@gmalinve I think if the number of unclassified objects is different before the operation and after, it means the operation failed, right?

Samuelopez-ansys (approved):

gmalinve:
> @gmalinve I think if the number of unclassified objects is different before the operation and after, it means the operation failed, right?

@Samuelopez-ansys does the intersection fail if you have an unclassified object as a result?

Samuelopez-ansys:
@gmalinve Unclassified object is not desired. It means the boolean operation failed and it was not correct. This extra step, is just avoiding the user to Undo later, but the method is returning False in this case, so for me it is expected behavior.

PR #4350, opened 2024-03-11
MAINT: Update pyedb deps

SMoraisAnsys:
In order to have both pyaedt and pyedb in pyansys metapackage, we need more flexibility. The proprosed changes consist in:
- freezing pyedb version for python 3.7;
- freezing pyedb version to the latest compatible when testing;
- allowing flexibility when installing pyaedt to retrieve any patch version in between 0.5 (included) and 0.6 (excluded)

Note: This should be working well with dependabot and we might need to update pyedb to remove python version 3.7 compatibility to avoid unwanted version upgrades.


pyproject.toml
```diff
@@ -51,6 +52,8 @@ tests = [
     "pandas==1.3.5; python_version == '3.7'",
     "pandas==2.0.3; python_version == '3.8'",
     "pandas==2.2.1; python_version > '3.9'",
+    "pyedb==0.5.2; python_version > '3.7'",
+    "pyedb==0.4.2; python_version == '3.7'",
```
MaxJPRey, line 56: For the optional dependencies should we allow patch releases too to have flexibility?

Samuelopez-ansys (approved):
LGTM. Good idea @SMoraisAnsys

PR #4356, opened 2024-03-12
MAINT: Fix wrong version of pyedb

SMoraisAnsys:
For some reasons, pyedb version 0.4.2 was not published in pypi. Therefore, when one wants to work with Python 3.7, we must specify version 0.4.0 instead.

NB: Caught this error while refactoring the CI/CD. Once used, we'll avoid such error to happen !


Samuelopez-ansys (approved):

MaxJPRey:
Nice catch @SMoraisAnsys !

PR #4366, opened 2024-03-15
Lazy load of parametrics, setups and optimizations

Samuelopez-ansys:
I had to add these 3 properties to the lazy load because, for Big project, it takes time to load the objects. If they are needed later, it will take the same previous time to load the objects.

### DIFF
--- a/_unittest/test_11_Setup.py
+++ b/_unittest/test_11_Setup.py
@@ -68,6 +68,7 @@ class TestClass:
         self.aedtapp.duplicate_design("auto_open")
         for setup in self.aedtapp.get_setups():
             self.aedtapp.delete_setup(setup)
+        assert not self.aedtapp.setups
         self.aedtapp.set_auto_open()
         setup1 = self.aedtapp.get_setup("Auto1")
         setup1.enable_adaptive_setup_multifrequency([1.9, 2.4], 0.02)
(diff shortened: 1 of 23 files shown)


pyaedt/application/Analysis.py
```diff
@@ -192,6 +192,49 @@ def materials(self):
...
+                self._setups = [self.get_setup(setup_name) for setup_name in self.setup_names]
+        return self._setups
+
+    @property
+    def parametrics(self):
+        """Setups in the project.
+
+        Returns
+        -------
+        :class:`pyaedt.modules.DesignXPloration.ParametricSetups`
```
PipKat, line 216: Alignment under "Returns" heading?

SMoraisAnsys (approved):
I like the idea of not doing things eagerly but lazily. I would even enforce doing it by default and not using a flag to activate it.

Just a quick question, should one of the attributes `_setups`, `_parametrics` or `_optimizations` change after it has been set once ? If no, e.g. the Analysis is always done as post processing and nothing should change, ignore my comment. Else, the implementation should be updated as it does not allow for value modification.

_unittest/test_11_Setup.py
```diff
@@ -67,6 +68,7 @@ def test_01c_create_hfss_setup_auto_open(self):
         self.aedtapp.duplicate_design("auto_open")
         for setup in self.aedtapp.get_setups():
             self.aedtapp.delete_setup(setup)
```
SMoraisAnsys, line 70:
```suggestion
        for setup in self.aedtapp.get_setups():
            self.aedtapp.delete_setup(setup)
            assert setup not in self.aedtapp.setups
```

SMoraisAnsys (dismissed):
Good job with this refactoring !

SMoraisAnsys (approved):

PR #4367, opened 2024-03-15
Add network export to jep30

lorenzovecchietti:


lorenzovecchietti:
> @lorenzovecchietti Do we need this new library ? Is it not possible to use pandas? What do you think @SMoraisAnsys ?

I'm converting to draft, so I can check if pandas is suitable for this. Honestly I didn't even think of pandas.

PR #4369, opened 2024-03-18
MAINT: Update dependencies with ranges

SMoraisAnsys:
The current way we handle dependencies is too "tight" and can easily lead to conflicts when installing pyaedt with other projects. This is even more problematic as it happens with other ansys projects, e.g. pyedb, pyaedt toolkits, ...
This PR proposes to use ranges to specify a wider compatibility of dependencies.

This changes have been tested in https://github.com/ansys/pyaedt/actions/runs/8345060900 where used python 3.7, 3.8, 3.9, 3.10 and 3.11, installed pyaedt with test dependencies and run the tests. Except for EMIT that is not available in Python3.7 (explaining the `skipif` decorator) every tests were passing correctly for every versions !


pyproject.toml
```diff
@@ -26,112 +26,112 @@ classifiers = [
 ]
 
 dependencies = [
-    "rpyc==6.0.0",
-    "psutil",
     "fpdf2",
     "jsonschema",
-    "pytomlpp; python_version < '3.12'",
-    "pyedb==0.4.0; python_version == '3.7'",
+    "psutil",
```
MaxJPRey, line 31: Thanks for sorting package by alphabetical order.

MaxJPRey (approved):
LGTM.

Samuelopez-ansys (approved):

PR #4379, opened 2024-03-19
CI: Fix nightly documentation build

SMoraisAnsys:
Note: There was an incompatibility between the actions to download and upload artifacts. The versions being different, the artifact generated was not compatible.


MaxJPRey (approved):
Thanks for catching this @SMoraisAnsys

Samuelopez-ansys (approved):

Samuelopez-ansys:
Can we merge?

PR #4384, opened 2024-03-19
Lazy load on objects

maxcapodi78:


Samuelopez-ansys (dismissed):
LGTM

Samuelopez-ansys (approved):

pyaedt/application/Design.py
```diff
@@ -262,8 +263,11 @@ def load_aedt_thread(path):
         self._logger.odesign = self.odesign
         AedtObjects.__init__(self, is_inherithed=True)
         self.logger.info("Aedt Objects correctly read")
-        if t:
-            t.join()
+        # if t:
+        #     t.join()
```
Alberto-DM, line 267:
```suggestion

```

pyaedt/modeler/cad/Primitives.py
```diff
@@ -6596,9 +6784,9 @@ def get_object_from_name(self, objname):
             3D object returned.
 
         """
-        if objname in self._object_names_to_ids:
-            object_id = self.get_obj_id(objname)
-            return self.objects[object_id]
+        if objname in self.object_names:
+            # object_id = self.get_obj_id(objname)
```
Alberto-DM, line 6788:
```suggestion

```

_unittest/conftest.py
```diff
@@ -94,7 +94,7 @@
 desktop_version = config["desktopVersion"]
 new_thread = config["NewThread"]
 settings.use_grpc_api = config["use_grpc"]
-
+settings.objects_lazy_load = False
```
Alberto-DM, line 97: Is it not tested? I think we should run at least Primitives test with and without lazy load

Alberto-DM (commented):
LGTM, just a couple of comments

PR #4390, opened 2024-03-21
Add export option to export_file and export_on_grid

### LINKED ISSUE #4383: Add export_options to export_field_file()

`export_field_file()` in pyaedt/modules/PostProcessor.py asks for an optional argument `export_with_sample_points`, set to `True` by default. Under the hood, it calls `ofieldsreporter.ExportToFile()`, and passes` export_with_sample_points` as the fifth argument.

Recent versions of AEDT have additional options while exporting field, e.g. selecting which coordinate system to use. According to HFSS scripting guide, this can be set by passing a list to the fifth argument, instead of the legacy Boolean option. Example from the HFSS scripting guide (section 20-28, under **ExportToFile [Fields Calculator]**:
```python
oModule.ExportToFile("C:\\offset_file_model_unit_ref.fld", "C:\\offset_SI.pts",
"4500MHz : LastAdaptive",
  [
    "Freq:=" , "4.5GHz",
    "Phase:=" , "0deg"
  ],
  [
    "NAME:ExportOption",
    "IncludePtInOutput:=" , True,
    "RefCSName:=" , "offset",
    "PtInSI:=" , False,
    "FieldInRefCS:=" , True
  ])
```
The Scripting Guide also has the following note, which explains why the current pyaedt implementation works:

> Regarding the legacy ExportToFile script which only has “IncludePtInOutput” argument (the last Boolean one), AEDT can
> still read it and assign other new arguments as default values. Those default values are RefCSName = “global”, PtInSI =
> “True”, FieldInRefCS = “False

**Current workaround:**
Since `export_field_file()` directly passes `export_with_sample_points` to `ofieldsreporter.ExportToFile()` without any modification, user should be able to pass a list instead of a Boolean, and get desired results. eg:
```python
app.post.export_field_file("ComplexMag_E", filename="E_Field.fld", 
                           sample_points_file="data_points.pts", intrinsics="2.45GHz",
                          export_with_sample_points= ['IncludePtInOutput:=', False, 'RefCSName:=', 'CustomCS'])
```

**Suggested enhancement**
Going forward, it'll be nice to have additional arguments available to `export_field_file()` for each export option. According to HFSS scripting guide, the available options are: 

> Whether to include points in output (Boolean), RefCSName (String), Pts in SI units (Boolean), Field in RefCS (Boolean).

_No response_

_No response_

Samuelopez-ansys:
Close #4383

### DIFF
--- a/_unittest_solvers/test_00_analyze.py
+++ b/_unittest_solvers/test_00_analyze.py
@@ -331,6 +331,20 @@ class TestClass:
             sample_points_lists=[[0, 0, 0], [3, 6, 8], [4, 7, 9]],
         )
         assert os.path.exists(fld_file_2)
+        cs = self.icepak_app.modeler.create_coordinate_system()
+        fld_file_3 = os.path.join(self.local_scratch.path, "test_fld_3.fld")
+        self.icepak_app.post.export_field_file(
+            quantity_name="Temp",
+            solution=self.icepak_app.nominal_sweep,
+            variation_dict=self.icepak_app.available_variations.nominal_w_values_dict,
+            filename=fld_file_3,
+            obj_list="box",
+            sample_points_lists=[[0, 0, 0], [3, 6, 8], [4, 7, 9]],
+            reference_coordinate_system=cs.name,
+            export_in_si_system=False,
+            export_field_in_reference=False,
+        )
+        assert os.path.exists(fld_file_3)
 
     def test_04a_3dl_generate_mesh(self):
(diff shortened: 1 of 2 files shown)


gmalinve:
@Samuelopez-ansys you could add UTs for both.

Samuelopez-ansys:
@gmalinve I have pushed a new test, but it is not needed to increase the coverage, because if it fails, it will be an exception in the AEDP API, not in PyAEDT. But for safety, I have added one test to cover the non-default values.

gmalinve (approved):

PR #4425, opened 2024-03-28
Various maintainability fixes.

MaxJPRey:


pyaedt/modeler/cad/Primitives.py
```diff
@@ -4651,16 +4651,16 @@ def import_spaceclaim_document(self, SCFile):
...
-            if "AWP_ROOT" in l:
-                if l > latestversion:
-                    latestversion = l
-        if not latestversion:
+        env_var = os.environ
+        latest_version = ""
+        for variable in env_var:
+            if "AWP_ROOT" in variable:
+                if variable > latest_version:
+                    latest_version = variable
```
SMoraisAnsys, line 4659:
```suggestion
                    latest_version = variable
                    break
```
It should stop once the first variable is found right ?

MaxJPRey: Yes, I think so.

PR #4436, opened 2024-03-31
Fix/issue 3913

dcrawforAtAnsys:
Setting ``design_name`` to a valid design name in the project now sets that one to the active design.


Samuelopez-ansys (approved):
LGTM. @maxcapodi78  Do we need the timeout? It is hardcoded, I understand the reason, but not sure if this is safe.

PR #4457, opened 2024-04-03
fix for mesh regions

### LINKED ISSUE #4456: Cannot set manual settings in mesh region

### Description of the bug

-

### Steps To Reproduce

-

Windows

3.7

### Installed packages

-

lorenzovecchietti:
Close #4456


Samuelopez-ansys:
@Alberto-DM Any reason this variable was set to "not" ?

lorenzovecchietti:
> @Alberto-DM Any reason this variable was set to "not" ?

Is the question actually to @Alberto-DM or to me? I set it to "not" just because at the beginning the property was "automatic_settings" and not "manual_settings". Then I changed my mind during the refactoring and this is just a leftover.

Samuelopez-ansys (approved):
LGTM

PR #4463, opened 2024-04-05
Refactor function handler

maxcapodi78:


.github/workflows/unit_tests.yml
```diff
@@ -132,13 +132,13 @@ jobs:
...
         with:
-          max_attempts: 3
+          max_attempts: 2
           retry_on: error
           timeout_minutes: 50
           command: |
             testenv\Scripts\Activate.ps1
             Set-Item -Path env:PYTHONMALLOC -Value "malloc"
-            pytest -n 6 --dist loadfile --durations=50 -v --cov=pyaedt --cov-report=xml --cov-report=html --junitxml=junit/test-results.xml _unittest
+            pytest -n 4 --dist loadfile --durations=50 -v --cov=pyaedt --cov-report=xml --cov-report=html --junitxml=junit/test-results.xml _unittest
```
SMoraisAnsys, line 141:
```suggestion
            pytest -n auto --dist loadfile --durations=50 -v --cov=pyaedt --cov-report=xml --cov-report=html --junitxml=junit/test-results.xml _unittest
```

SMoraisAnsys (changes requested):
Since the line of codes
```
            if not settings.enable_error_handler:
                if settings.release_on_exception:
                    from pyaedt.generic.desktop_sessions import _desktop_sessions

                    for v in list(_desktop_sessions.values())[:]:
                        v.release_desktop(v.launched_by_pyaedt, v.launched_by_pyaedt)
                raise e
            else:
                return False
```
are used in every exception handling, could we have an inner function in `wrapper` to refactor that and call it every time it is needed ?

.github/workflows/unit_tests.yml
maxcapodi78: here the problem is that is not related to the number of cores available but on the issues caused by running parallel AEDT sessions. I'm trying to limit to make the test more stable and reduce randomic failures

maxcapodi78:
> Since the line of codes
> 
> ```
>             if not settings.enable_error_handler:
>                 if settings.release_on_exception:
>                     from pyaedt.generic.desktop_sessions import _desktop_sessions
> 
>                     for v in list(_desktop_sessions.values())[:]:
>                         v.release_desktop(v.launched_by_pyaedt, v.launched_by_pyaedt)
>                 raise e
>             else:
>                 return False
> ```
> 
> are used in every exception handling, could we have an inner function in `wrapper` to refactor that and call it every time it is needed ?

Implemented

SMoraisAnsys (approved):
LGTM

Samuelopez-ansys (approved):

SMoraisAnsys (approved):

PR #4464, opened 2024-04-05
Refactoring scattering

maxcapodi78:
Created a new class to share Scattering methods across Circuit, Hfss and Hfss3dLayout classes


Samuelopez-ansys (dismissed):
Great!

Samuelopez-ansys (dismissed):

Samuelopez-ansys (approved):

PR #4284, opened 2024-02-27
CI: Refact and extend CI/CD workflows to leverage ansys actions

SMoraisAnsys:
> [!IMPORTANT]  
> One of the new steps requires that a new `ansys/actions` is released. This PR shouldn't be merged as long as `5.1.9` is the latest version released.

List of the worfklows to validate in the new one (aggregation into one workflow):
- [x] Doc Style
- [x] Doc Build (see #4296 ~if we keep print_errors.py or leverage option `-W` of `sphinx-build`, or something else~)
  - [x] Add install option [doc-noexamples] to remove a lot of unneeded dependencies
- [x] Test
- [x] Test solver
- [x] Iron Python
- [x] Wheelhouse and smoke tests
- [x] Full documentation (moved to CICD)
- [x] Doc build upload and update dev index (moved to CICD)
- [x] Cpython linux (test and test solver)
- [x] PR title 

Extra workflows:
- ~[Nightly] Cpython linux (test and test solver)~
- ~[Nightly] Doc build with upload and update dev index~
- [Nightly] Test pre release

PR #4468, opened 2024-04-05
Fix pdf 

Samuelopez-ansys:


SMoraisAnsys (approved):
LGTM

Closely related to #4284 which (hopefully) will be accepted soon

PR #4451, opened 2024-04-03
Improve h3d layout methods

maxcapodi78:
add few methods to hfss3dlayout.py:
- dissolve_component
- create_pec_on_component_by_nets
- modeler.create_components_on_pins
Added new property to set_solderball method to add an offset for the port in component solderball.
added method create_components_on_pins

PR #4469, opened 2024-04-08
FIX: Only one component per pins is created.

MaxJPRey:
Just renaming the method from `create_components_on_pins()` to `create_component_on_pins()` because based on the code, only one component can be created.


MaxJPRey:
@maxcapodi78 I don't think that those changes made in #4451 were part of the release 0.8.7. So no need to handle deprecation.

SMoraisAnsys (approved):
Nice catch

Samuelopez-ansys (approved):
Should we create a deprecation warning?

MaxJPRey:
> Should we create a deprecation warning?

@Samuelopez-ansys I don't think so as the method was created 5 days ago and it is not part of the last release.
so no one should have ever used it.

PR #4471, opened 2024-04-08
Refactored Mesh class arguments

maxcapodi78:


pyaedt/modules/Mesh.py
```diff
@@ -325,7 +325,7 @@ class Mesh(object):
     >>> from pyaedt import Hfss
     >>> aedtapp = Hfss()
```
gmalinve, line 326: @maxcapodi78 that's small but I started to be consistent in methods docstrings. For example, rather then calling an instance of a class "app" I use the name of the class:
hfss = Hfss()
m2d = Maxwell2d()
etc.

gmalinve, line 326: @maxcapodi78 that's small but I started to be consistent in methods docstrings. For example, rather then calling an instance of a class "app" I use the name of the class:
hfss = Hfss()
m2d = Maxwell2d()
etc.
```suggestion
    >>> hfss= Hfss()
```

and for all the docstrings below..

pyaedt/modules/Mesh.py
```diff
@@ -566,50 +566,50 @@ def assign_surface_mesh(self, names, level, meshop_name=None):
...
-        mop = MeshOperation(self, meshop_name, props, "SurfApproxBased")
+        mop = MeshOperation(self, name, props, "SurfApproxBased")
         mop.create()
         self.meshoperations.append(mop)
         return mop
 
-    @pyaedt_function_handler()
-    def assign_surface_mesh_manual(self, names, surf_dev=None, normal_dev=None, aspect_ratio=None, meshop_name=None):
+    @pyaedt_function_handler(names="objects", meshop_name="name")
+    def assign_surface_mesh_manual(self, objects, surf_dev=None, normal_dev=None, aspect_ratio=None, name=None):
```
gmalinve, line 599: @maxcapodi78 I'd prefer to use full names: surface_deviation instead for example.
```suggestion
    def assign_surface_mesh_manual(self, objects, surface_deviation=None, normal_dev=None, aspect_ratio=None, name=None):
```

pyaedt/modules/Mesh.py
```diff
@@ -595,15 +595,17 @@ def assign_surface_mesh(self, objects, level, name=None):
...
+    ):
         """Assign a surface mesh to a list of faces.
 
         Parameters
         ----------
         objects : list or str or :class:`pyaedt.modeler.elements3d.FacePrimitive`
             List of faces to apply the surface mesh to.
-        surf_dev : float or str, optional
+        surface_deviation : float or str, optional
             Surface deviation. The default is ``None``. Allowed values are float, number with units or `"inf"`.
```
PipKat, line 609:
```suggestion
            Surface deviation. The default is ``None``. Allowed values are float, number with units, or `"inf"`.
```

gmalinve (approved):

PR #4482, opened 2024-04-10
DOC: Fix pdf download

SMoraisAnsys:
Merged PR #4468 was not entirely working, this should add the extra step required to make it work.

Note: Extra changes include the removal of not used assets. They were used some months ago to build the documentation but we moved to use font awesome icons.


MaxJPRey (approved):
LGTM.

PR #4484, opened 2024-04-10
replaced default HFSS Solution type to Terminal for version >= 2024.1

maxcapodi78:


MaxJPRey:
@maxcapodi78 In the title of the PR, you mention that the default Solution type should be `Terminal`.
But in the code it seems it is `Modal`.

It is just to be sure that it is expected.

Samuelopez-ansys:
@MaxJPRey You can see in desing_types that @maxcapodi78 changed the default HFSS design, the main reason is because from AEDT 2023R2, the default is Terminal.

Samuelopez-ansys (approved):
LGTM

MaxJPRey:
> @MaxJPRey You can see in desing_types that @maxcapodi78 changed the default HFSS design, the main reason is because from AEDT 2023R2, the default is Terminal.

Got it. Thanks @Samuelopez-ansys !

PR #4488, opened 2024-04-11
Use isinstance() rather than type() for a typecheck.

MaxJPRey:


Samuelopez-ansys (approved):
LGTM

SMoraisAnsys (approved):
LGTM, could you update the code to leverage the fact that `isinstance` can work with a tuple of classes for the second ?  It'll return true if the first argument is an instance of any of the types in that sequence.

MaxJPRey:
> LGTM, could you update the code to leverage the fact that `isinstance` can work with a tuple of classes for the second ? It'll return true if the first argument is an instance of any of the types in that sequence.

Excellent idea @SMoraisAnsys . I sort of automated the process.

pyaedt/modeler/cad/Primitives.py
```diff
@@ -2451,7 +2451,7 @@ def get_group_bounding_box(self, group):
         >>> oEditor.GetObjectsInGroup
         >>> oEditor.GetModelBoundingBox
         """
-        if type(group) is not str:
+        if isinstance(group, str):
```
MaxJPRey, line 2454:
```suggestion
        if not isinstance(group, str):
```

pyaedt/modules/Mesh.py
```diff
@@ -244,10 +244,11 @@ def update_assignment(self):
         if "Faces" in self.props:
             faces = self.props["Faces"]
             faces_out = []
-            if type(faces) is not list:
+            if not isinstance(faces, list):
                 faces = [faces]
             for f in faces:
-                if type(f) is EdgePrimitive or type(f) is FacePrimitive or type(f) is VertexPrimitive:
+
```
MaxJPRey, line 250:
```suggestion
```

SMoraisAnsys (dismissed):

myoung301 (dismissed):
The EMIT-related portions look good to me.

Samuelopez-ansys (approved):
LGTM

PR #4490, opened 2024-04-11
MAINT: Fix various code issues like the direct iteration over dictionaries.

MaxJPRey:
Add `assert` statement in tests.
Remove unused variables.
Do not use mutable variables in method's signatures.


Samuelopez-ansys (approved):
LGTM

PR #4495, opened 2024-04-15
fix point_in_polygon method when the asum was a multiple of 2*phi

maxcapodi78:


svandenb-dev (dismissed):
LGTM

Alberto-DM (changes requested):
Do you have an example of when this is happening? 
It should never happen, geometrically.

Alberto-DM (approved):
LGTM

PR #4498, opened 2024-04-15
Enhancement to parametric heatsink function

lorenzovecchietti:
Even if the number of columns is 1, we can leave this as a parameter so that it can be changed down the line


SMoraisAnsys (approved):
LGTM

Samuelopez-ansys (approved):

PR #4500, opened 2024-04-15
DOC: Use GIF in documentation

SMoraisAnsys:
When working on creating PDF files through CICD (#4468), we removed GIF files as they were not correctly handled.
This changes revert back the use of GIF and should allow our PDF file to work with GIF files.

### DIFF
--- a/doc/source/User_guide/modeler.rst
+++ b/doc/source/User_guide/modeler.rst
@@ -57,3 +57,8 @@ All objects support executing any modeler operation, such as union or subtractio
      cyl = hfss.modeler["mycyl"]
      box.unite(cyl)
      box.subract(cyl)
+
+
+.. image:: ../Resources/objects_operations.gif
+  :width: 800
+  :alt: Object Modeler Operations
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -29,7 +29,7 @@ import shutil
 # Additionally, when documenting images in formats other than the supported ones, 
 # make sure to specify their types.
 from sphinx.builders.latex import LaTeXBuilder
-LaTeXBuilder.supported_image_types = ["image/png", "image/pdf", "image/svg+xml", "image/webp" ]
+LaTeXBuilder.supported_image_types = ["image/png", "image/pdf", "image/gif", "image/svg+xml", "image/webp" ]
 
 from sphinx.writers.latex import CR
 from sphinx.writers.latex import LaTeXTranslator


Samuelopez-ansys (approved):

SMoraisAnsys:
@Samuelopez-ansys I think the best approach might be to just get rid of the GIF files. We only had ony before #4468 and it seems to be a pain to add a GIF into the PDF file built through `sphinx-build latex`.

A work around could be to use `pillow` to split the GIF into multiples PNG files and then to use `animate` to create an animation afterward. However, this is time consuming. For example, it took multiple minutes to split `object_operations.gif` and the result was 374 PNG files. Moreover, if new GIF files are added, it might start to be too much time consuming.

I would advise to keep the GIF for html documentation and drop the GIF when creating the PDF.

PipKat:
> @Samuelopez-ansys I think the best approach might be to just get rid of the GIF files in the PDF documentation. We only had ony before #4468 and it seems to be a pain to add a GIF into the PDF file built through `sphinx-build latex`.
> 
> A work around could be to use `pillow` to split the GIF into multiples PNG files and then to use `animate` to create an animation afterward. However, this is time consuming. For example, it took multiple minutes to split `object_operations.gif` and the result was 374 PNG files. Moreover, if new GIF files are added, it might start to be too much time consuming.
> 
> I would advise to keep the GIF for html documentation and drop the GIF when creating the PDF.

The product documentation also has this issue. You might consider the approach that the writers for the Mechanical application use. They include this phrase above the GIF: "The following demo is presented as an animated GIF. View online if you are reading the PDF version of the help."  We could tweak it a bit to something like this:  "The following demo is presented as an animated GIF. View online if you are reading the PDF version of this documentation."

Samuelopez-ansys (approved):

doc/source/User_guide/modeler.rst
```diff
@@ -57,3 +57,15 @@ All objects support executing any modeler operation, such as union or subtractio
...
+.. only:: latex
+
+    The following demo is presented as an animated GIF.
+    `View online <https://aedt.docs.pyansys.com/version/stable/_images/objects_operations.gif>`_
+    if you are reading the PDF version of this documentation.
+
+.. only:: html
+  .. image:: ../Resources/objects_operations.gif
+    :width: 800
+    :alt: Object Modeler Operations
```
SMoraisAnsys, line 71:
```suggestion
    :alt: Object Modeler Operations

```

Samuelopez-ansys (approved):

Samuelopez-ansys (approved):

PR #4517, opened 2024-04-17
DOC: Remove download button from release

SMoraisAnsys:
Currently the PDF button is present in dev mode (as expected) but also when a new release is created. However, we currently don't  create the PDF for the full documentation as it takes too long to generate.
This should restrict PDF download to dev mode documentation.


Samuelopez-ansys (approved):
This will change when we move the examples to the new repository, am I right?

SMoraisAnsys:
> This will change when we move the examples to the new repository, am I right?

Yes, at that point we will redirect to the associated examples.

MaxJPRey (approved):
Excellent @SMoraisAnsys

PR #4582, opened 2024-04-23
FIX: Export from 3D layout to 3D modeler with keeping net names bug fix

svandenb-dev:
Initial setting used were causing failure on some design. du to precision difference between 3D modeler and EDB.
Taking coarser value resolve the issue


SMoraisAnsys (approved):
LGTM

PR #4586, opened 2024-04-24
TESTS: Apdate solvers Linux testing

SMoraisAnsys:
Applying changes of @Samuelopez-ansys from #4284

Changes consist in commenting/disabling some tests that are currently not working in Linux when used in non graphical mod.


Samuelopez-ansys:
Closing this PR, it will fixed with this #4589

PR #4587, opened 2024-04-24
Fix active design selection behavior for Emit designs

myoung301:
If the project and design names are not specified, the active project and active design are supposed to be used. However, a new project is always being created when the Emit() object is created.

### DIFF
--- a/_unittest_solvers/test_26_emit.py
+++ b/_unittest_solvers/test_26_emit.py
@@ -7,6 +7,7 @@ from _unittest_solvers.conftest import config
 import pytest
 
 from pyaedt import Emit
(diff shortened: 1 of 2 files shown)


jsalant22 (approved):

SMoraisAnsys:
@myoung301 Thanks for the contribution ! Could you also update the tests such that your modification passes our CI ?
Seems like modifying the behavior of automatically creating project was leveraged in some tests.

MaxJPRey (approved):
LGTM.

SMoraisAnsys (approved):
LGTM, thanks for updating the tests

PR #4595, opened 2024-04-26
FEAT: Convert relevant columns of fields summary pandas data from str to floats

lorenzovecchietti:


lorenzovecchietti:
Those columns will always contain numbers (unit is specified by the entity column)

Samuelopez-ansys (approved):

PR #4611, opened 2024-04-30
DOC: Fix blank area

SMoraisAnsys:
On top of fixing the blank area, this PR also drops the dependency to sphinx-jinja which is no longer required / leveraged.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -93,7 +93,6 @@ doc = [
     #"sphinx-autodoc-typehints",
     "sphinx-copybutton>=0.5.0,<0.6",
     "sphinx-gallery>=0.14.0,<0.17",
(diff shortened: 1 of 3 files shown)


Samuelopez-ansys (approved):

SMoraisAnsys:
Here is the resulting rendering

image

pyproject.toml
```diff
@@ -93,7 +93,6 @@ doc = [
     #"sphinx-autodoc-typehints",
     "sphinx-copybutton>=0.5.0,<0.6",
     "sphinx-gallery>=0.14.0,<0.17",
-    "sphinx-jinja>=2.0,<2.1",
```
MaxJPRey, line 96: One less dependency 👍

PR #4615, opened 2024-04-30
Fix export to q3d

maxcapodi78:


examples/05-Q3D/Q3D_DC_IR.py
```diff
@@ -11,7 +11,7 @@
 
 import os
 import pyaedt
-
+from pyedb import Edb
```
PipKat, line 14:
```suggestion
From PyEDB import EDB
```

maxcapodi78: @PipKat sorry but this code will not wor. Python requires "from" and the exact name of the package

svandenb-dev (dismissed):
LGTM

pyaedt/modules/SolveSetup.py
```diff
@@ -1973,21 +1974,61 @@ def _get_primitives_points_per_net(self):
...
+                    primitive_dict[net].append(pt)
+                    next_p = int(len(points[0]) / 2)
+                    pt = [points[0][next_p], points[1][next_p]]
+                    pt.append(z)
+                    primitive_dict[net].append(pt)
+
+                else:
+                    n = 0
+                    while n < 1000:
+                        n += 10
```
SMoraisAnsys, line 2025: A week ago we had to switch back to `n += 20`. Do your changes handle the reason for that switch back ?

Samuelopez-ansys: @SMoraisAnsys This is an algorithm issue. @maxcapodi78 figured out how to make it more robust

Samuelopez-ansys (approved):

PR #4616, opened 2024-05-01
Revert "Fix active design selection behavior for Emit designs"

myoung301:
Reverts ansys/pyaedt#4587


SMoraisAnsys:
@myoung301 Is there a reason to revert those changes ?

myoung301:
> @myoung301 Is there a reason to revert those changes ?

We noticed a test failure after the original PR. I created a couple draft PRs to evaluate different approaches. The other approach was successful and allowed the fix to remain so I will abandon this PR and publish the other.

myoung301:
Found another approach that worked and left the fix intact.

PR #4617, opened 2024-05-01
Make sure a new design is created for test_23_result_categories in test_26_emit.py

myoung301:


SMoraisAnsys (approved):
LGTM, thanks for catching this one. It would have been a headache to understand the problem in a few weeks when we move to 2024.2

PR #4589, opened 2024-04-25
REFACT: Fix Circuit in linux 2024 R1

Samuelopez-ansys:
Add a common method for active project and active design

PR #4618, opened 2024-05-02
TESTS: Temporary disabling SPISIM test with linux

SMoraisAnsys:
Seems like #4589 did not handle this tests and that it might be worth it to reopen #4590.

The skipped tests are currently blocking #4284 from passing CICD as tests would be performed on Linux.


MaxJPRey (approved):
LGTM.
Feel free to merge this PR if it can help to merge #4284 .

PR #4631, opened 2024-05-06
CI: Fix nightly-docs needs

SMoraisAnsys:
As title says. An incorrect job need was used in the YAML file.

### DIFF
--- a/.github/workflows/nightly-docs.yml
+++ b/.github/workflows/nightly-docs.yml
@@ -20,7 +20,6 @@ jobs:
   doc-build:
     name: Documentation build without examples
     runs-on: ubuntu-latest
-    needs: [doc-style]
     steps:
       - name: Install Git and checkout project
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


Samuelopez-ansys (approved):

.github/workflows/nightly-docs.yml
```diff
@@ -20,7 +20,6 @@ jobs:
   doc-build:
     name: Documentation build without examples
     runs-on: ubuntu-latest
-    needs: [doc-style]
```
MaxJPRey, line 23: Why not checking the doc style before to build it?

Is it because we are assuming that the style is always correct because it is checked in every PR merged in main?

SMoraisAnsys: This can also be performed but, as you mentioned, we shouldn't have a PR merged without passing the documentation style. Moreover, I feel like adding doc style could lead us to cases where we wouldn't have a dev documentation up to date because the CI failed due to some changing in the `vale` action.

PR #4636, opened 2024-05-06
REFACTOR: Improve try/except statements in `_get_design_mesh_regions` and `_get_design_mesh_operations`

### LINKED ISSUE #4632: Error is logged by _get_design_mesh_regions and _get_design_mesh_operations

### Description of the bug

`PyAEDT ERROR: 'NoneType' object is not subscriptable`  is logged by `_get_design_mesh_regions` and `_get_design_mesh_operations`

### Steps To Reproduce

-

Windows

3.7

### Installed packages

-

lorenzovecchietti:
Close #4632.
Current implementation of `_get_design_mesh_regions` and `_get_design_mesh_operations` lead to errors being logged, I removed this improving try/except statements


Samuelopez-ansys (approved):

SMoraisAnsys (commented):
Are you able to know why the TypeError is raised ? If not I would rather use `self._app.logger.debug("Failed to get design mesh operations")` for every exception.
For the Keyerror handling, I would also use the previous log as it is considered bad practice to use a try: ... except: pass pattern. Tools that we'll use for vulnerability checking would report such things as a vulnerability.

lorenzovecchietti:
> Are you able to know why the TypeError is raised ?

`TypeError` gets raised because, when the file is not saved, self._app.design_properties is None, so it cannot be accessed like a dictionary. (Mesh regions/operations are unfortunately one of the few things we still need to parse from the file)

> For the Keyerror handling, I would also use the previous log as it is considered bad practice to use a try: ... except: pass pattern. 
I can think of something else to log? Because when KeyError is raised, we know we have a design dictionary parsed, but it simply doesn't have the keys needed to define a mesh region/mesh operations.

In both cases, the user was previously seeing an error being logged when the code was actually working as intended (empty mesh region/empty mesh operation list) all the time he was accessing the mesh class with a new proj/a project without any region.

I can try to rewrite everything from scratch, but I wouldn't leave an error being raised and logged when the code is working properly (and in addition given that the error comes from an internal function) as this could throw off a lot of beginner/intermediate users.

SMoraisAnsys:
> > Are you able to know why the TypeError is raised ?
> 
> `TypeError` gets raised because, when the file is not saved, self._app.design_properties is None, so it cannot be accessed like a dictionary. (Mesh regions/operations are unfortunately one of the few things we still need to parse from the file)
> 
> > For the Keyerror handling, I would also use the previous log as it is considered bad practice to use a try: ... except: pass pattern.
> > I can think of something else to log? Because when KeyError is raised, we know we have a design dictionary parsed, but it simply doesn't have the keys needed to define a mesh region/mesh operations.
> 
> In both cases, the user was previously seeing an error being logged when the code was actually working as intended (empty mesh region/empty mesh operation list) all the time he was accessing the mesh class with a new proj/a project without any region.
> 
> I can try to rewrite everything from scratch, but I wouldn't leave an error being raised and logged when the code is working properly (and in addition given that the error comes from an internal function) as this could throw off a lot of beginner/intermediate users.

When I was speaking about the previous log I wanted to refer to the one I did type. Sorry for not being clear enough. What you did is great and the right thing to do. We should just avoid using the try: except: pass pattern.
Logging something in debug wouldn't be caught be users AFAIK and it could be helpful for debugging

Samuelopez-ansys (approved):
LGTM

PR #4647, opened 2024-05-07
DOCS: Add thumbnails manually

SMoraisAnsys:
@lorenzovecchietti This is a WIP to add thumbnails manually:
- pros: only requires one comment in the example and it does not appear when the documentation is built
- cons: we require this comment in the example (might not be something that we want to keep with respect to users)

Another approach could be to add an "expected" result at the beginning of the example. That way, since the expected result is plotted at first, it will be used as thumbnail.


SMoraisAnsys:
Closing this as changes are going to be performed in pyaedt-examples with another approach

PR #4651, opened 2024-05-08
CHORE: allow latest PyEDB version

RobPasMue:
As of now, you cannot use together the main branches of PyAEDT and PyEDB. This will allow you to do so

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,6 @@ tests = [
     "pytest>=7.4.0,<8.3",
     "pytest-cov>=4.0.0,<5.1",
     "pytest-xdist>=3.5.0,<3.7",
-    "pyedb>=0.4.0,<0.11",
     "pyvista>=0.38.0,<0.44",
     "scikit-learn>=1.0.0,<1.5",
     "scikit-rf>=0.30.0,<1.1",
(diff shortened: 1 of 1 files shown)


Samuelopez-ansys (commented):
If I am not wrong we have to remove it from the tests dependencies

SMoraisAnsys (dismissed):
Thanks for fixing that

pyproject.toml
```diff
@@ -29,7 +29,7 @@ dependencies = [
     "fpdf2",
     "jsonschema",
     "psutil",
-    "pyedb>=0.4.0,<0.11",
+    "pyedb>=0.4.0,<0.12",
```
SMoraisDev, line 32:
```suggestion
    "pyedb>=0.4.0",
```
In fact, we might just removed the upper bound since we won't be able to test the new release of pyedb through dependabot. Indeed, we would already be allowing this version.
Moreover, `pyedb` release process tests that the current pyedb dev version does not fail pyaedt units tests.

SMoraisDev (changes requested):

RobPasMue:
Thank you both! I just removed the upper bound following @SMoraisAnsys suggestion. Feel free to approve and merge whenever you prefer

SMoraisDev (approved):

Samuelopez-ansys (approved):

PR #4655, opened 2024-05-10
DOCS: Reduce documentation size

### LINKED ISSUE #2291: Image optimization opportunities

### Description of the bug

Looking over all images contained in the web documentation of PyAEDT, I found a bunch of issues and opportunities to improve. I used the [JSON export ](https://github.com/pyansys/devportal-doc-archive/tree/drupal)that you provide for import into the developer portal.

Optimizing the entire _images folder with open source tools set to 80% quality for PNG and 70% for GIF saved 21.6 MB out of 34.9 MB. 61.2% per file on average (up to 75.4%).

In particular, `sphx_glr_SBR_Time_Plot_001.png` in the Examples section is in reality an animated GIF, just mislabeled. It will display correctly in most browsers regardless of the mislabeling, they are quite forgiving. However, this image is half of the filesize of the total JSON export. I ran this GIF through Gifsicle at 70% quality setting, which is generally a good optimization compromise for display on the web, and brought it down to 8MB from 22.5MB.

`sphx_glr_Hfss_Icepak_Coupling_001.png` is also an animated GIF. However, the 18 frames display no change as far as I can tell, so I exported the first frame only as a PNG, which cut its size in half. 

I then transformed all images except for objects_operations.gif to webp, saving yet another 60%. The remaining gif image is smaller than any webp I could produce, so it remains as is. The webp versions, although lossy, look fine to my eye when used on the web.

Results:
```
myprompt % du -ks _images/ images-opt/ webp/
34364	_images/
13104	images-opt/
5164	webp/
```

Attached is a compressed tar archive with two folders containing the resulting images if you want to make use of it. 
[images.tgz](https://github.com/pyansys/pyaedt/files/10414029/images.tgz)

### Steps To Reproduce

n/a

MacOS

3.7

### Installed packages

n/a

Samuelopez-ansys:
Close #2291


Samuelopez-ansys:
image

Samuelopez-ansys:
@maxcapodi78 @SMoraisAnsys What do you think? I have applied the image modification requested in #2291 , but it is just 8 MB of improvement. I saw that HTML API directory is the bigger one, And we can not reduce it (262 MB).

SMoraisAnsys (approved):
Good job @Samuelopez-ansys IMO it is worth the shot even if the size reduction is limited from the documentation point of view. The size reduction should also apply to the repo itself (which can take long if you don't have a good connection)
From what I could see it should be ok to move from PNG/JPEG to WEBP as long as our users do not suffer from it (which, I assume, won't be the case)
Can you check the documentation good rendering in HTML and PDF format ?
Btw, you removed a lot of other PNG files, are you sure they are not used in the full documentation, i.e. through examples ?

Samuelopez-ansys:
I have tested it locally, and it works (HTML and PDF) with examples. The removed images were not used in the full repo, they were probably used in previous documentation versions.

PR #4662, opened 2024-05-11
FEAT: Add enforce touchstone model passive function.

HaiwenZhang:


SMoraisAnsys (commented):
Hi @HaiwenZhang and thanks for this contribution!
Seems to be legit to me but I'll let others handle the reviewing.
@Samuelopez-ansys Seems like something has changed as we are now able to perform test triggering from CICD

Samuelopez-ansys:
@HaiwenZhang To run unit test locally, you first need to install tests dependencies.

pip install .[tests]

Then you can run the unit test. 

Regarding the code, you can use ModelParameters which has already the EditWithComps call, then you can do something like:

t2 = self.aedtapp.modeler.schematic.create_touchstone_component(touch)

passive = OrderedDict(
            [
                ("DCOption", -1),
                ("InterpOption", 1),
                ("ExtrapOption", 3),
                ("Convolution", 0),
                ("Passivity", 6),
                ("Reciprocal", False),
                ("ModelOption", ""),
                ("DataType", 2),
            ]
        )

props["NexximCustomization"] = passive
props["ModTime"] = int(time.time())
t2.model_data.props = props
t2.model_data.update()

You can change your code with a similar approach to avoid duplications. 

And please push the unit tests that you mentioned you created.

pyaedt/modeler/circuits/object3dcircuit.py
```diff
@@ -871,6 +872,44 @@ def change_property(self, property_name, names=None):
...
+        props["NexximCustomization"] = passive
+        props["ModTime"] = int(time.time())
+        try:
+            arg = []
+            _dict2arg(props, arg)
+            arg.insert(0, "Name:{}".format(self.model_name))
+            self._circuit_components.o_model_manager.EditWithComps(self.model_name, arg, [])
+            return True
+        except Exception:
+            self._circuit_components.logger.warning("Failed to enforce % touchstone model passive ", self.model_name)
```
MaxJPRey, line 910:
```suggestion
            self._circuit_components.logger.warning("Failed to enforce % touchstone model passive.", self.model_name)
```

HaiwenZhang:
Hi @Samuelopez-ansys @MaxJPRey, thank you for your review. I changed the code and pushed the unit test code.

pyaedt/modeler/circuits/object3dcircuit.py
```diff
@@ -871,6 +872,37 @@ def change_property(self, property_name, names=None):
             return self._oeditor.ChangeProperty(vOut)
         return False
 
+    @pyaedt_function_handler()
+    def enforce_touchstone_model_passive(self):
+        """Enforce touchstone model passive.
+
+        Returns
+        -------
+        bool
```
Samuelopez-ansys, line 881: You should add something like (check other methods):
 bool

pyaedt/modeler/circuits/object3dcircuit.py
```diff
@@ -871,6 +872,37 @@ def change_property(self, property_name, names=None):
...
+                ("Reciprocal", False),
+                ("ModelOption", ""),
+                ("DataType", 2),
+            ]
+        )
+        props["NexximCustomization"] = passive
+        props["ModTime"] = int(time.time())
+        self.model_data.props = props
+        self.model_data.update()
+
```
Samuelopez-ansys, line 905: it should return True if the update() is correct,so do :

return self.model_data.update()

HaiwenZhang:
Hi @Samuelopez-ansys, I pushed new code. Thanks!!!

pyaedt/modeler/circuits/object3dcircuit.py
Samuelopez-ansys: @HaiwenZhang Please in the docstring add this:

        Returns
        -------
        bool

Samuelopez-ansys (dismissed):
Thank you for your contribution!

_unittest/test_21_Circuit.py
```diff
@@ -898,3 +898,18 @@ def test_49_automatic_ami(self):
...
+        self.aedtapp.modeler.schematic_units = "mil"
+        s_parameter_component = self.aedtapp.modeler.schematic.create_touchstone_component(self.touchstone_file)
+        assert s_parameter_component.enforce_touchstone_model_passive()
+        nexxim_customization = s_parameter_component.model_data.props["NexximCustomization"]
+        assert -1 == nexxim_customization["DCOption"]
+        assert 1 == nexxim_customization["InterpOption"]
+        assert 3 == nexxim_customization["ExtrapOption"]
+        assert 0 == nexxim_customization["Convolution"]
+        assert 6 == nexxim_customization["Passivity"]
+        assert False == nexxim_customization["Reciprocal"]
```
SMoraisAnsys, line 913:
```suggestion
        assert not nexxim_customization["Reciprocal"]
```

SMoraisAnsys (approved):
Thanks for your contribution !

HaiwenZhang:
@Samuelopez-ansys @SMoraisAnsys Thanks for your review. BTW, I have a problem, when I import the S parameter in Ansys Designer, it has a default pin location, I want to change the pin location, do you have any good suggestions?

PR #4673, opened 2024-05-14
FIX: design_type in desktop.py

myoung301:
For the case when the `design_name` argument is provided to `design_type()`, the wrong arguments were being provided to `self.active_design()`. It expects a project object as the first object and design_name string as the second argument.


SMoraisAnsys (approved):
Thanks for the fix !

PR #4672, opened 2024-05-14
REFACTOR: kwargs

gmalinve:
refactoring kwargs maxwellcircuit.py, CableModeling.py, report_templates.py, general_methods.py

PR #4626, opened 2024-05-03
FEAT: Improve antenna array processing and plotting

Devin-Crawford:
Minor improvements to plotting and functionality to support array phase excitations for beam-steering in HFSS.

PR #4702, opened 2024-05-21
CHORE: Use matplotlib Agg backend when testing

SMoraisAnsys:
Multiple recent PRs struggled because figures get shown in the CICD and one would have to close them manually (which does not work through CICD). This leads to tests failing due to time out as no one closes the figures manually.

Here we use `Agg` backend which is a non-GUI backend and will avoid future PR to struggle with this issue.

For example, this happened in https://github.com/ansys/pyaedt/pull/4626
Also related to https://github.com/ansys/pyaedt/pull/4672


SMoraisAnsys:
Since this changes might also impact the documentation, we might want to rename the environment variable a bit and see it's impact with the documentation.

Samuelopez-ansys (approved):

doc/source/conf.py
```diff
@@ -277,12 +277,17 @@ def setup(app):
     # Must be less than or equal to the XVFB window size
     pyvista.global_theme["window_size"] = np.array([1024, 768])
 
-    # suppress annoying matplotlib bug
+    # Suppress annoying matplotlib bug
     warnings.filterwarnings(
         "ignore",
         category=UserWarning,
         message="Matplotlib is currently using agg, which is a non-GUI backend, so it cannot show the figure.",
```
PipKat, line 284:
```suggestion
        message="Because Matplotlib is using `agg`, which is a non-GUI backend, it cannot show the figure.",
```
The message is a run-on sentence as initially written.

doc/source/conf.py
```diff
@@ -277,12 +277,17 @@ def setup(app):
...
     )
+    warnings.filterwarnings(
+        "ignore",
+        category=UserWarning,
+        message="FigureCanvasAgg is non-interactive, and thus cannot be shown",
```
PipKat, line 289:
```suggestion
        message="Because 'FigureCanvasAgg' is non-interactive, it cannot be shown.",
```

pyaedt/generic/plot.py
```diff
@@ -36,6 +36,12 @@
     try:
         from matplotlib.patches import PathPatch
         from matplotlib.path import Path
+
+        # Use matplotlib Agg backend (non-interactive) when the CI is running.
```
PipKat, line 40:
```suggestion
        # Use matplotlib ``Agg`` backend (non-interactive) when the CI is running.
```

gmalinve (approved):

doc/source/conf.py
SMoraisAnsys: This is a warning message from Matplotlib, not from our side so we have to keep the same content to filter it.
Thanks for the rewriting though :)

doc/source/conf.py
SMoraisAnsys: This is a warning message from Matplotlib, not from our side so we have to keep the same content to filter it.
Thanks for the rewriting though :)

SMoraisAnsys:
Changes: I removed the use of the environment variable `'ON_CI'` and added a fixture instead. This will ensure that the documentation is not impacted.

@Samuelopez-ansys If you want, we can update the test to remove the `show=False` argument. Now, even if `show=True`, the tests should be passing correctly. Let me know if you want to extend this PR with those changes.

Samuelopez-ansys (approved):

MaxJPRey (dismissed):
LGTM.

SMoraisAnsys:
Removing show=False proved to have impact on the tests and led me to notice "misses" in our recent refactoring.
This PR should stay as a draft as long as https://github.com/ansys/pyaedt/pull/4723 is not merged.
Then tests should be updated accordingly.

SMoraisAnsys:
I'll close this PR as adding those changes increases our tests by a huge overhead.
However, it was not entirely a waste of time as it allowed to catch https://github.com/ansys/pyaedt/issues/4730 !

PR #4705, opened 2024-05-21
FEAT: Add discovery link function

IreneWoyna:


pyaedt/modeler/cad/Primitives.py
```diff
@@ -4924,6 +4924,70 @@ def import_spaceclaim_document(self, input_file):
...
+
+        >>> oEditor.CreateUserDefinedModel
+        """
+        version = settings.aedt_version[-4:].replace(".", "")
+
+        aedt_install_dir = os.environ.get("ANSYSEM_ROOT{}".format(version), "")
+        ansys_install_dir = os.environ.get("AWP_ROOT{}".format(version), "")
+
+        if ansys_install_dir == "":
+            self.logger.error("Same version of AEDT and Discovery is needed.")
```
MaxJPRey, line 4951: Here it seems that you are checking if ansys unified is installed.
I think we could modify the message to be:

```suggestion
        if ansys_install_dir == "":
            self.logger.error("Ansys is not installed. Consequently, Discovery installation is missing too.")
```

IreneWoyna: Thanks @MaxJPRey for your comment! AEDT can be installed independently from Discovery, so my intention is to check if the same version of AEDT and Disco are installed. However it's possible that I'm not doing it correctly.

MaxJPRey: You are doing it properly @IreneWoyna . I am just suggesting to modify the error message.
I think we should just mention in the log file which version of Ansys is missing, such as:
`self.logger.error("Ansys {} is not installed. Consequently, Discovery installation is missing too.".format(version))`

pyaedt/modeler/cad/Primitives.py
```diff
@@ -4924,6 +4924,70 @@ def import_spaceclaim_document(self, input_file):
...
+            ``True`` when successful, ``False`` when failed.
+
+        References
+        ----------
+
+        >>> oEditor.CreateUserDefinedModel
+        """
+        version = settings.aedt_version[-4:].replace(".", "")
+
+        aedt_install_dir = os.environ.get("ANSYSEM_ROOT{}".format(version), "")
```
gmalinve, line 4947: @IreneWoyna. Why do you need to check the aedt install dir and the ansys_install_dir?
the aedt_install_dir is not even used.
Bear in mind that you can get it directly from settings as:
`settings.aedt_install_dir`
without even getting the version as you are doing.

pyaedt/modeler/cad/Primitives.py
```diff
@@ -4924,6 +4924,70 @@ def import_spaceclaim_document(self, input_file):
         self.refresh_all_ids()
         return True
 
+    def discovery_import(self, input_file):
```
gmalinve, line 4927: @IreneWoyna I'd keep the method name as in AEDT: add_discovery_link

gmalinve:
@IreneWoyna I posted a couple of general questions about the implementation but I was wondering.. why did you put it in Primitives?

pyaedt/modeler/cad/Primitives.py
IreneWoyna: @gmalinve thanks for your comments! AEDT and Discovery versions need to be the same, that's why I'm checking it. But surely there's a smarter way to do it.

pyaedt/modeler/cad/Primitives.py
IreneWoyna: Ok, that's fine. I changed the name because I didn't implement the actual "dynamic link", only static import (because it works better)

IreneWoyna:
> @IreneWoyna I posted a couple of general questions about the implementation but I was wondering.. why did you put it in Primitives?

Because that's where the spaceclaim link is, there's no further intelligence behind that decision :D

pyaedt/modeler/cad/Primitives.py
Alberto-DM: In misc/misc.py there is a function to get all AEDT installed versions on the system, it retrurns a dictioary containing their installation paths (it is used in PyAEDT initialization). It is called `installed_version`. 
I suggest to create a similar function to get all Discovery installations and then just check if `settings.aedt_version` in the list of Disco installations.
Keep in mind that checking if Discovery is installed is a bit tricky because a user can install Ansys structures without Discovery, so it would have the environment variable, but Disco won't be installed.

pyaedt/modeler/cad/Primitives.py
Alberto-DM: I agree with @MaxJPRey, but the situation is a bit trickier. A user can have the correct Ansys version installed, but without having Discovery installed on the system (Discovery can be deselected in the installation). So my suggestion is to check if Disco is installed (rather than Ansys unified) and log the message accordingly, just about the Discovery version missing.

pyaedt/modeler/cad/Primitives.py
```diff
@@ -4924,6 +4924,70 @@ def import_spaceclaim_document(self, input_file):
...
+                "SACADIntegUDM",
+                "Library:=",
+                "installLib",
+                "Version:=",
+                "1.0",
+                "ConnectionID:=",
+                "",
+            ]
+        )
+        self.refresh_all_ids()
```
Alberto-DM, line 4988: Since the method implement only the static import, I would break the link immediately after the import. and before refreshing all the IDs.

Alberto-DM (dismissed):

pyaedt/modeler/cad/Primitives.py
MaxJPRey: You are correct @Alberto-DM.

gmalinve:
@SMoraisAnsys @Samuelopez-ansys to make pass the new test we need Workbench to be installed in the VM. Would this be possible? 

@IreneWoyna

pyaedt/modeler/cad/Primitives.py
```diff
@@ -5063,6 +5063,75 @@ def import_spaceclaim_document(self, input_file):
...
+        ----------
+
+        >>> oEditor.CreateUserDefinedModel
+        """
+        version = self._app.aedt_version_id[-3:]
+
+        ansys_install_dir = os.environ.get("AWP_ROOT{}".format(version), "")
+
+        if ansys_install_dir:
+            if "Discovery" not in os.listdir(ansys_install_dir):
```
Samuelopez-ansys, line 5089:
```suggestion
            if "Discovery" not in os.listdir(ansys_install_dir):  # pragma: no cover
```

pyaedt/modeler/cad/Primitives.py
```diff
@@ -5063,6 +5063,75 @@ def import_spaceclaim_document(self, input_file):
...
+                self.logger.error("Discovery installation not found.")
+        else:
```
Samuelopez-ansys, line 5091:
```suggestion
        else:  # pragma: no cover
```

Samuelopez-ansys (dismissed):

Samuelopez-ansys (dismissed):

SMoraisAnsys (approved):

Samuelopez-ansys (approved):

PR #4706, opened 2024-05-21
FIX: Function handler to return False

### LINKED ISSUE #4707: Bug located in pyaedt_function_handler

### Description of the bug

Previous behavior of `pyaedt_function_handler` allowed to return `False` when an exception was raised and the `settings` were correctly set up (cf `enable_error_handler`). However, this is not the case.

### Steps To Reproduce

Try any decorated function raising an exception. It would return None instead of False (which explain why our testing was not too impacted)

Windows

3.10

### Installed packages

Not of interest

SMoraisAnsys:
Following https://github.com/ansys/pyaedt/pull/4463, it seems that the `pyaedt_function_handler` function was not working correctly. Indeed, if an exception was raised and the settings were such that one would want the function handler to return False, no value was returned by the function (leading to a None return).

Some tests have been added to ensure that the behavior of the function does not changes without notice. They cover the following behaviors of the function handler:
- enable error handler;
- release on exception;
- argument deprecation mechanism.

Closes #4707

### DIFF
--- a/_unittest/test_utils.py
+++ b/_unittest/test_utils.py
@@ -16,7 +16,7 @@ SETTINGS_ENABLE_ERROR_HANDLER = settings.enable_error_handler
 
 @pyaedt_function_handler(deprecated_arg="trigger_exception")
 def foo(trigger_exception=True):
-    """Some dummy function used for testing"""
+    """Some dummy function used for testing."""
     if trigger_exception:
         raise Exception("Dummy message.")
 
--- a/_unittest/test_utils.py
+++ b/_unittest/test_utils.py
@@ -23,7 +23,7 @@ def foo(trigger_exception=True):
 
 @patch("pyaedt.generic.desktop_sessions._desktop_sessions")
 def test_handler_release_on_exception(mock_sessions):
-    """Test handler while activating/deactivating error handler."""
+    """Test handler while activating or deactivating error handler."""
     mock_session = MagicMock()
     mock_sessions.values.return_value = [mock_session]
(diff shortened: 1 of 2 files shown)


_unittest/test_utils.py
```diff
@@ -0,0 +1,70 @@
...
+
+SETTINGS_RELEASE_ON_EXCEPTION = settings.release_on_exception
+SETTINGS_ENABLE_ERROR_HANDLER = settings.enable_error_handler
+
+
+@pyaedt_function_handler(deprecated_arg="trigger_exception")
+def foo(trigger_exception=True):
+    """Some dummy function used for testing."""
+    if trigger_exception:
+        raise Exception("Dummy message.")
```
MaxJPRey, line 21: Do we test the content of this message in our tests?
Just to be sure it is still working.

_unittest/test_utils.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+@patch("pyaedt.generic.desktop_sessions._desktop_sessions")
+def test_handler_release_on_exception(mock_sessions):
+    """Test handler while activating/deactivating error handler."""
```
PipKat, line 26:
```suggestion
    """Test handler while activating or deactivating error handler."""
```
Avoid using a forward slash as it isn't always clear whether it represents an "or" or an "and".

maxcapodi78 (approved):

_unittest/test_utils.py
SMoraisAnsys: Done

PR #4708, opened 2024-05-21
CHORE: expose show for plotting

dcrawforAtAnsys:
Expose the keyword argument "show" for all methods that create matplotlib figures. Default behavior is unchanged. Plots will be rendered when the method is called.

### DIFF
--- a/pyaedt/generic/plot.py
+++ b/pyaedt/generic/plot.py
@@ -435,7 +435,7 @@ def plot_3d_chart(plot_data, size=(2000, 1000), xlabel="", ylabel="", title="", 
     fig.set_size_inches(size[0] / dpi, size[1] / dpi)
     if snapshot_path:
         fig.savefig(snapshot_path)
-    if show and not is_notebook():
+    if show:
         fig.show()
     return fig
 
--- a/pyaedt/generic/plot.py
+++ b/pyaedt/generic/plot.py
@@ -378,14 +378,14 @@ def plot_polar_chart(
     fig.set_size_inches(size[0] / dpi, size[1] / dpi)
     if snapshot_path:
         fig.savefig(snapshot_path)
-    if show:
+    if show and not is_notebook():
         fig.show()
     return fig
 
 
 @pyaedt_function_handler()
 @update_plot_settings
-def plot_3d_chart(plot_data, size=(2000, 1000), xlabel="", ylabel="", title="", snapshot_path=None):
+def plot_3d_chart(plot_data, size=(2000, 1000), xlabel="", ylabel="", title="", snapshot_path=None, show=True):
     """Create a Matplotlib 3D plot based on a list of data.
 
     Parameters
@@ -403,6 +403,9 @@ def plot_3d_chart(plot_data, size=(2000, 1000), xlabel="", ylabel="", title="", 
         Plot Title label.
     snapshot_path : str
         Full path to image file if a snapshot is needed.
+    show : bool, optional
+        Whether to render the figure. The default is ``True``. If ``False``, the
+        figure is not drawn.
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


pyaedt/generic/plot.py
```diff
@@ -542,7 +550,8 @@ def plot_matplotlib(
         List of annotations to add to the plot. The format is [x, y, string, dictionary of font options].
         Default is `None`.
```
PipKat, line 551:
```suggestion
        The default is ``None``.
```

pyaedt/generic/plot.py
```diff
@@ -638,8 +647,8 @@ def plot_contour(
     snapshot_path : str, optional
         Full path to save the image save. The default is ``None``.
```
PipKat, line 648:
```suggestion
        Full path to save the image to. The default is ``None``.
```

pyaedt/modules/solutions.py
```diff
@@ -778,6 +778,7 @@ def plot(
         title="",
         snapshot_path=None,
         is_polar=False,
+        show=True,
     ):
         """Create a matplotlib plot based on a list of data.
```
PipKat, line 783:
```suggestion
        """Create a Matplotlib plot based on a list of data.
```

pyaedt/modules/solutions.py
```diff
@@ -808,6 +809,9 @@ def plot(
             Full path to image file if a snapshot is needed.
```
PipKat, line 809:
```suggestion
            Full path to the image file if a snapshot is needed.
```

pyaedt/modules/solutions.py
```diff
@@ -808,6 +809,9 @@ def plot(
...
         is_polar : bool, optional
             Set to `True` if this is a polar plot.
```
PipKat, line 811:
```suggestion
            Whether this is a polar plot. The default is ``False``.
```

pyaedt/modules/solutions.py
```diff
@@ -868,6 +872,7 @@ def plot_3d(
         formula=None,
         size=(2000, 1000),
         snapshot_path=None,
+        show=True,
     ):
         """Create a matplotlib 3d plot based on a list of data.
```
PipKat, line 877:
```suggestion
        """Create a Matplotlib 3D plot based on a list of data.
```

dcrawforAtAnsys: Added all suggestions from @PipKat .

pyaedt/generic/plot.py
```diff
@@ -378,14 +378,14 @@ def plot_polar_chart(
     fig.set_size_inches(size[0] / dpi, size[1] / dpi)
     if snapshot_path:
         fig.savefig(snapshot_path)
-    if show:
+    if show and not is_notebook():
```
SMoraisAnsys, line 381:
```suggestion
    if show:
```

dcrawforAtAnsys: @SMoraisAnsys Thanks. I missed that one.

SMoraisAnsys (commented):
To extend your changes and be consistent, should we also update `plot` from `pyaedt.generic.plot.ModelPlotter` ? The current code is

```python
        if export_image_path:
            self.pv.show(screenshot=export_image_path, full_screen=True)
        elif show and self.is_notebook:  # pragma: no cover
            self.pv.show()  # pragma: no cover
        elif show:
            self.pv.show(full_screen=True)  # pragma: no cover
```

Since we are starting to follow the approach of allowing both writting a figure in a file AND plotting it, I would update the previous snipper into

```python
        if export_image_path:
            self.pv.show(screenshot=export_image_path, full_screen=True)
        if show:  # pragma: no cover
            self.pv.show(full_screen=not self.is_notebook)
```

What do you think ?

SMoraisAnsys (approved):
LGTM, I just left a minor comment

maxcapodi78 (approved):

PR #4712, opened 2024-05-22
FIX: fixed _get_primitives_points_per_net since it cannot handle bondwires.

maxcapodi78:
@svandenb-dev  I've disable at the moment the bondwire check. But we need to figure out how to handle it

### DIFF
--- a/pyaedt/modules/SolveSetup.py
+++ b/pyaedt/modules/SolveSetup.py
@@ -1981,8 +1981,7 @@ class Setup3DLayout(CommonSetup):
             primitive_dict[net] = []
             self.p_app.logger.info("Processing net {}...".format(net))
             for prim in primitives:
-
-                if prim.layer_name not in layers_elevation:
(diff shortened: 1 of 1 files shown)


svandenb-dev (approved):
LGTM agreed we will have to check how to handle it.

SMoraisAnsys (approved):
@maxcapodi78 Could you provide a bit of information to justifying these changes ? I'm sure you have great reasons but that would help reviewing future changes if we know why the current one are being performed !

maxcapodi78:
> @maxcapodi78 Could you provide a bit of information to justifying these changes ? I'm sure you have great reasons but that would help reviewing future changes if we know why the current one are being performed !

@SMoraisAnsys Actually the method is computing the bounding box of primitives in EDB. This method doesn't apply to bondwires which are not planar surface. We can have a method in the future to estimate the bounding box but will require different computation and doesn't have native equivalent method.

PR #4713, opened 2024-05-22
CHORE: Remove dependency to scikit-learn

SMoraisAnsys:
While looking at #4710 changes in our code base, I noticed that we no longer use `sktlearn` in `pyaedt`. If I'm wrong, feel free to remove this PR and accept #4710, otherwise accept this one and close #4710.


SMoraisAnsys:
Seems like the package is used indirectly through unpickling file which required it, see bellow.
image

I'll close this PR and update the other one to have more information and also add the package to target `all`

PR #4714, opened 2024-05-22
REFACTOR: update mirror argument from "position" to "origin"

marcobiasizzo:
In the mirror function of components_3d and object3d one of the argument is called position. To be coherent with other functions (also with mirror in primitives), the variable name is updated to origin


SMoraisAnsys (changes requested):
Thanks @marcobiasizzo for these changes !

We are smoothly transitioning from deprecated arguments into new ones and your proposition seems to align with that effort of standardization !
Do you mind reworking your changes to use our function handler ? This will notify users of the method signature changes and warn them that it will be discarded in the coming weeks/months.

For example, with the `mirror` method, you would have
```python
    @pyaedt_function_handler(position="origin")
    def mirror(self, position, vector):
```
You can have a look at `duplicate_and_mirror` in the same file.

marcobiasizzo:
Sure, thanks! I updated the function handler.
Is it ok now?

marcobiasizzo:
Oh, sorry maybe you meant to cancel my changes and use only the function handler. Let me know!

maxcapodi78:
> Oh, sorry maybe you meant to cancel my changes and use only the function handler. Let me know!

@marcobiasizzo that's sound good now. Let's see if all UT passes

SMoraisAnsys:
> Oh, sorry maybe you meant to cancel my changes and use only the function handler. Let me know!

No, you did exactly what I had in mind, thanks !
Just so you know, our CICD requires access to github secrets (to run the tests on self hosted runners) and, since you are not part of ansys, you can't access the secrets and the tests can't run. I'll Create a new PR with your changes (cherry-picked) that way you'll stay the author of the changes and test will be available to run.

SMoraisAnsys:
Closing this PR for its duplicate #4715 where tests should be running.

PR #4715, opened 2024-05-23
REFACTOR: Update mirror argument from "position" to "origin"

SMoraisAnsys:
Mirroring #4714 from https://github.com/marcobiasizzo (who cannot access secrets)

"""
In the mirror function of components_3d and object3d one of the argument is called position. To be coherent with other functions (also with mirror in primitives), the variable name is updated to origin.
"""

@marcobiasizzo Thanks again for your contribution !


MaxJPRey (approved):
LGTM.

PR #4721, opened 2024-05-24
CHORE: Add security policy.

MaxJPRey:
Adding a file to specify the policy and the process to the  PYAEDT users and contributors when they discover a vulnerability.

@SMoraisAnsys I will add the versions table later in a new PR.


SMoraisAnsys (approved):
LGTM, thanks Maxime !

MaxJPRey:
> @SMoraisAnsys Why the check title is failing?

@Samuelopez-ansys  Because I used 'MAINT' instead of 'CHORE'. 
Thanks @SMoraisAnsys. I just saw you already fixed it.

PR #4722, opened 2024-05-24
FIX: submit_job improvement to allow the possibility to pass a server path

maxcapodi78:
#4662

### DIFF
--- a/pyaedt/desktop.py
+++ b/pyaedt/desktop.py
@@ -1769,26 +1769,29 @@ class Desktop(object):
         Parameters
         ----------
         project_file : str
-            Full path to the project. The path should be visible from the server running the simulation.
-            If the path is a client path, then make sure that the setting_file contains the
-            mapping between client and server path.
+            Full path to the project. The path should be visible from the server where the
+            simulation will run.
+            If the client path is used then the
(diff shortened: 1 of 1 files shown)


pyaedt/desktop.py
```diff
@@ -1816,43 +1825,47 @@ def submit_job(
                 return False
         else:
             if not os.path.exists(aedt_full_exe_path):
-                self.logger.error("AEDT shared path does not exist. Provide a full path.")
-                return False
+                self.logger.warning("The AEDT executable path not visible from the client.")
```
dcrawforAtAnsys, line 1828: Is this client or server? Shouldn't the exe be visible on the server?

dcrawforAtAnsys (approved):
One minor point regarding the message to be resolved. Otherwise all good with some minor linguistic modifications.

pyaedt/desktop.py
maxcapodi78: yes that is the server, so the client not necessarly is able to see it. That's why we cannot force the return False if we are not able to see it from the client

dcrawforAtAnsys (commented):
LGTM.

PR #4723, opened 2024-05-24
REFACTOR: Extend consistency on plot methods

SMoraisAnsys:
Following https://github.com/ansys/pyaedt/pull/4626 and https://github.com/ansys/pyaedt/pull/4708, theses changes aim at making our code base consistent.

The changes consist in returning `matplotlib` figure objects on methods that were returning a list of value.
In the case of `polar_plot_3d_pyvista`, a `pyvista` plotter object is returned when, before, a boolean was returned if one wanted to show the figure or save it in a file.

This changes should be performed before merging #4702


pyaedt/modules/solutions.py
```diff
@@ -938,8 +938,8 @@ def plot_3d(
 
         Returns
         -------
-        :class:`matplotlib.plt`
-            Matplotlib fig object.
+        :class:`matplotlib.figure.Figure`
```
dcrawforAtAnsys, line 941:
```suggestion
        :class:`matplotlib.pyplot.Figure`
```

pyaedt/modules/solutions.py
```diff
@@ -1849,30 +1847,29 @@ def plot_2d_cut(
...
-            else:
-                return plot_2d_chart(
-                    curves,
-                    xlabel=x_key,
-                    ylabel=quantity,
-                    title=title,
-                    snapshot_path=image_path,
-                    show_legend=show_legend,
-                )
+        show_legend = show
```
dcrawforAtAnsys, line 1850: show_legend is different from show so I don't think show_legend should be set to show. They are two different things.

pyaedt/modules/solutions.py
```diff
@@ -1913,14 +1910,12 @@ def polar_plot_3d(
             Full path for the image file. The default is ``None``, in which case a file is not exported.
         show : bool, optional
             Whether to show the plot. The default is ``True``.
-            If ``False``, the Matplotlib instance of the plot is shown.
+            If ``False``, the Matplotlib instance of the plot is not shown.
```
dcrawforAtAnsys, line 1913:
```suggestion
            If ``False``, the ``plt.Figure`` instance of the plot is not shown.
```
Just semantics, but it is the figure rendering that in changed by `show`

dcrawforAtAnsys (approved):
See minor suggestions, otherwise looks good.

pyaedt/modules/solutions.py
SMoraisAnsys: I do agree and, as a matter of fact, I just wanted to follow previous logic (if show then show_legend = True).
I'll update it such that they are different things

pyaedt/modules/solutions.py
SMoraisAnsys: If you look at the source code of matplotlib, matplotlib.pyplot imports Figure from matplotlib.figure.
Do you still want to keep this change though ?

pyaedt/modules/solutions.py
SMoraisAnsys: Thanks for the remark ! I'll handle this comment depending on your answer to my previous question.

pyaedt/modules/solutions.py
dcrawforAtAnsys: @SMoraisAnsys : matplotlib.figure.Figure is correct. You can ignore my comments.

pyaedt/modules/solutions.py
dcrawforAtAnsys: @SMoraisAnsys : It wasn't clear to me what the ``pyplot`` interface was doing. Now it's clear. It is the Figure instance that is rendered, and ``pyplot`` is just along for the ride to do the bookkeeping.

dcrawforAtAnsys:
All good.

dcrawforAtAnsys (commented):
LGTM.

dcrawforAtAnsys:
Inadvertently closed. LGTM.

Samuelopez-ansys (approved):
LGTM

PR #4726, opened 2024-05-27
FIX: changed default behavior of import_nastran method

maxcapodi78:
Changed method arguement enable_planar_merge to "True" in order to keep default behaviour as in AEDT


Samuelopez-ansys (approved):
LGTM

PR #4727, opened 2024-05-27
FEAT: PCB component

lorenzovecchietti:


pyaedt/modules/Boundary.py
```diff
@@ -330,6 +331,19 @@ def delete(self):
...
 
+def disable_auto_update(func):
+    def wrapper(self, *args, **kwargs):
+        auto_update = self.auto_update
+        self.auto_update = False
+        out = func(self, *args, **kwargs)
+        self.update()
+        self.auto_update = auto_update
+        if out is not None:
+            return out
```
SMoraisAnsys, line 342: Is there a reason for this if statement ? Why wouldn't you return `None` if that's what the function returns ? Btw, that's the basic return value of any function in python.

pyaedt/modules/Boundary.py
```diff
@@ -354,6 +368,334 @@ class NativeComponentPCB(NativeComponentObject, object):
...
+    @disable_auto_update
+    def footprint_filter(self, minimum_footprint):
+        """Set minimum component footprint for filtering.
+
+        Parameters
+        ----------
+        minimum_footprint : str
+            Value with unit of the minimum component footprint for filtering.
+        """
+        self.auto_update = False
```
SMoraisAnsys, line 392:
```suggestion
```
Is there a reason to keep this `auto_update = False` since you use the decorator to temporarily disable auto update ? If you want this method to ensure that the auto_update is set to False afterward then fine, but I don't think that's the case as it's the only method that gets this call.

pyaedt/modules/Boundary.py
```diff
@@ -354,6 +368,334 @@ class NativeComponentPCB(NativeComponentObject, object):
...
+        if minimum_footprint is not None:
+            new_filters.append("FootPrint")
+            self.props["NativeComponentDefinitionProvider"]["FootPrint"] = minimum_footprint
+        self.props["NativeComponentDefinitionProvider"]["Filters"] = new_filters
+
+    @property
+    @pyaedt_function_handler()
+    def power_filter(self):
+        """Minimum component power for filtering."""
+        return self.filters.get("Power", {}).get("Value", None)
```
SMoraisAnsys, line 407:
```suggestion
        return self.filters.get("Power", {}).get("Value")
```
dict `get` returns `None` by default if the key is not found

pyaedt/modules/Boundary.py
```diff
@@ -354,6 +368,334 @@ class NativeComponentPCB(NativeComponentObject, object):
...
+        if minimum_power is not None:
+            new_filters.append("Power")
+            self.props["NativeComponentDefinitionProvider"]["PowerVal"] = minimum_power
+        self.props["NativeComponentDefinitionProvider"]["Filters"] = new_filters
+
+    @property
+    @pyaedt_function_handler()
+    def type_filters(self):
+        """Types of component that are filtered."""
+        return self.filters.get("Types", None)
```
SMoraisAnsys, line 432:
```suggestion
        return self.filters.get("Types")
```

SMoraisAnsys (approved):
LGTM, I left some minor changes / comments
Great job !

pyaedt/modules/Boundary.py
lorenzovecchietti: You are absolutely right

pyaedt/modules/Boundary.py
lorenzovecchietti: Yes, it is a residual from before I implemented the wrapper :)

pyaedt/modules/Boundary.py
lorenzovecchietti, line 342:
```suggestion
        return out
```

lorenzovecchietti:
> LGTM, thanks for taking into account my comments :)

Thanks to you for the review!

PR #4734, opened 2024-05-28
CHORE: Update codespell version

Samuelopez-ansys:
Fix codespell issues


SMoraisAnsys (dismissed):

SMoraisAnsys:
@Samuelopez-ansys From the docstring, I though that `checkin` and `checkout` are methods of EMIT engine ? If that's the case you should keep the same name right ?

MaxJPRey (approved):
LGTM.

PR #4736, opened 2024-05-28
TEST: Improve workflow coverage

Samuelopez-ansys:


maxcapodi78 (dismissed):
Looks great

myoung301 (commented):
Looks good from an EMIT perspective. A couple methods in emit_core/revision.py don't have coverage right now, but we can address that later. CC @jsalant22

maxcapodi78 (approved):

Samuelopez-ansys:
@myoung301 @jsalant22 What do you think regarding the extensions/workflows?

I think EMIT could take advantage of this. Using Tkinter we can create automated workflows inside PyAEDT.  

If the extension requires complex stuff then I would use the pyaedt-common-toolkit approach.

SMoraisAnsys (approved):
LGTM, great job !
I'd recommend avoiding using environment variables as it is done right now and only use `argparse` (which is something that is already in the workflows/misc.py file.
Using `argparse` to handle CLI arguments is common and good practice IMO

SMoraisAnsys (dismissed):

maxcapodi78 (approved):

maxcapodi78 (approved):

PR #4742, opened 2024-05-30
CHORE: Extend wheelhouse with installer target

SMoraisAnsys:
Currently if one wants to use the python script we provide to install pyaedt (see pyaedt_installer_from_aedt.py) without accessing internet and only using our wheelhouse, there is an issue.

This PR propose changes to add the missing packages into a new install target named `installer`.

We could also add them to target `all` or rework the pyaedt_installer script but this might be for another discussion ?


Samuelopez-ansys (dismissed):

doc/source/Resources/pyaedt_installer_from_aedt.py
```diff
@@ -237,7 +234,7 @@ def install_pyaedt():
 
             run_command('"{}" install --no-cache-dir --no-index --find-links={} pyaedt'.format(pip_exe, unzipped_path))
```
Samuelopez-ansys, line 235: It should be here: 

run_command('"{}" install --no-cache-dir --no-index --find-links={} pyaedt[installer]'.format(pip_exe, unzipped_path))

doc/source/Resources/pyaedt_installer_from_aedt.py
```diff
@@ -237,7 +234,7 @@ def install_pyaedt():
...
         else:
-            run_command('"{}" --default-timeout=1000 install pyaedt[all]'.format(pip_exe))
```
Samuelopez-ansys, line 240: I would keep it, because it is when the zip path does not exist, then it tries to perform an online installation. 

I think this one should be: 
run_command('"{}" --default-timeout=1000 install pyaedt[installer]'.format(pip_exe))

Samuelopez-ansys (dismissed):

Samuelopez-ansys (dismissed):

Samuelopez-ansys (approved):

pyproject.toml
```diff
@@ -131,6 +131,25 @@ all = [
...
 ]
+installer = [
+    "imageio>=2.30.0,<2.35",
+    "matplotlib>=3.5.0,<3.9",
+    "numpy>=1.20.0,<2",
+    "openpyxl>=3.1.0,<3.3",
+    "osmnx>=1.1.0,<1.10",
+    "pandas>=1.1.0,<2.3",
+    "pyvista>=0.38.0,<0.44",
+    "fast-simplification>=0.1.7",
```
Samuelopez-ansys, line 142:
```suggestion

```

Samuelopez-ansys (approved):

PR #4751, opened 2024-05-31
FEAT: Create rings on cone shape

Samuelopez-ansys:
This "shape" could be useful for some antenna application:
image


maxcapodi78 (approved):

SMoraisAnsys (approved):
LGTM

PR #4760, opened 2024-06-03
FEAT: Increment retries of ERL in unittests

Samuelopez-ansys:


maxcapodi78 (dismissed):

codecov.yml
```diff
@@ -23,3 +23,4 @@ coverage:
     - "pyaedt/workflows/installer"  # ignore folders and all its contents
     - "pyaedt/workflows/templates"  # ignore folders and all its contents
     - "pyaedt/common_rpc.py"  # ignore folders and all its contents
+    - "pyaedt/EDB.py"  # ignore folders and all its contents
```
SMoraisAnsys, line 26:
```suggestion
    - "pyaedt/edb.py"  # ignore folders and all its contents
```

pyaedt/generic/spisim.py
```diff
@@ -252,6 +252,8 @@ def compute_erl(
                 fp.write("# {}: {}\n".format(k, k))
                 fp.write("{} = {}\n".format(k, v))
         retries = 3
+        if "PYTEST_CURRENT_TEST" in os.environ:
+            retries = 10
```
SMoraisAnsys, line 256: Why aren't you just increasing `retries` without looking at the environment variables ? It might also be helpful for users no ?

SMoraisAnsys (commented):
Btw, I don't like using environment variables when not needed but it is a nice idea to use `"PYTEST_CURRENT_TEST"` to see if this is run in a test or not !

pyaedt/generic/spisim.py
Samuelopez-ansys: @SMoraisAnsys There is a random issue in the unit tests, that we can not reproduce locally. We think it is related to the number of retries. But if for some reason it really fails, we do not want users to wait 10 times until they see the error. Hopefully, this issue only impacts with pytest, the VMs and SPISIM.

SMoraisAnsys (approved):

Issue #4730, opened 2024-05-27
Bug located in plot methods related to pyvista


SMoraisAnsys:
### Before submitting the issue

- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### Description of the bug

It seems that recent changes allowing to both plot something and save something are not as straightforward as expected when dealing with `pyvista`. Indeed, when one creates an instance of `Plotter` embedded in any of our public methods related to plotting, there is an error that gets triggered if one wants to save and image in a file and render it.

A work around to this would be to use `auto_close=False` when calling `show()`. Here is an exemple illustrating my point:

```python
from pyvista import Plotter

p = Plotter()
p.show("something")
# The following line triggers a runtime error with message "This plotter has been closed and cannot be shown."
p.show()

p = Plotter()
p.show("something", auto_close=False)
# The following line does not trigger any error
p.show()
# But this one will
p.show()
```

Unless there is a need for our users to work with the resulting Plotter and try to call `show` later on by themselves, could we update our code base related to pyvista as follow:

```python
auto_close_plotter = not show
if img_path:
    p.show(screen_shoot=img_path, auto_close=auto_close_plotter)
if show:
    p.show()
```

?

### Steps To Reproduce

Run test test_70_far_field_data while adding `show=True` to line 615

```python
ffdata.polar_plot_3d_pyvista(
            quantity="RealizedGain",
            image_path=img4,
            background=[255, 0, 0],
            show_geometry=False,
            convert_to_db=True,
            show=True,
        )
```

### Which Operating System are you using?

Windows

### Which Python version are you using?

3.10

### Installed packages

Not relevant

PR #4762, opened 2024-06-04
CHORE: Added support to Python 3.12

maxcapodi78:
Note: some tests which uses VTK (through Pyvista) fails. They have been disabled. This is probably due to vtk 9.3.0 which is the only compatible with Python 3.12 but has some issues


SMoraisAnsys:
@maxcapodi78 Do you mind putting this PR as a draft to let me some time to investigate the issue with `vtk` and `pyvista` ?
I could also tackle #4730 while doing so !

maxcapodi78:
> @maxcapodi78 Do you mind putting this PR as a draft to let me some time to investigate the issue with `vtk` and `pyvista` ? I could also tackle #4730 while doing so !

@SMoraisAnsys Done. Consider that all test actually passes (apart the one I've commented)

SMoraisAnsys:
@maxcapodi78 I investigated the problem and I fear that we won't be able to test with `pyvista>=0.43` (which are the only versions compatible with python 3.12). I'll investigate a bit more and open an issue in `pyvista` repo if required.

Samuelopez-ansys (dismissed):

SMoraisAnsys (approved):

SMoraisAnsys (approved):

Samuelopez-ansys (approved):
LGTM

SMoraisAnsys (approved):

PR #4771, opened 2024-06-05
REFACTOR: Encryption password handling

SMoraisAnsys:
Hardcoding a default password is considered bad practice because it degrades security (easy to target for attackers and reverse engineering), flexibility (harder to rotate password), ....

As an alternative approach, we could use config files secrets or environment variables. This PR follows the environment variables that was used in #4672 to handle codacy warning.


Samuelopez-ansys (approved):

MaxJPRey (approved):
Good practice @SMoraisAnsys 👍

PR #4773, opened 2024-06-05
REFACTOR: improve readability of error messages

maxcapodi78:


SMoraisAnsys (approved):
LGTM if it doesn't remove too much information.
Also, I think that removing the message redirecting to the online documentation makes sense as it happened to me that the final query was not really relevant.

Samuelopez-ansys (approved):

PR #4794, opened 2024-06-11
FEAT: Check License feature

### LINKED ISSUE #4753: Provide meaningful error message when licence is not available

When licence is invalid error comes back that "designname" is empty even though it is assigned in the code. This is because of licence error in AEDT.

### Code sample expected

_No response_

_No response_

Samuelopez-ansys:
Close #4753 

@MelanieMichon We have implemented this method to check if the license server is down or if the license increment is available. The user should first check it, because for now is dangerous to add it inside the Desktop initialization.

### DIFF
--- a/pyaedt/generic/general_methods.py
+++ b/pyaedt/generic/general_methods.py
@@ -963,7 +963,6 @@ def is_license_feature_available(feature="electronics_desktop", count=1):  # pra
          ``None`` when licenser server is down.
     """
     import subprocess  # nosec B404
-    import tempfile
 
     aedt_install_folder = list(installed_versions().values())[0]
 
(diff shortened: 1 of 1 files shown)


maxcapodi78 (dismissed):

pyaedt/generic/general_methods.py
```diff
@@ -944,6 +945,74 @@ def is_project_locked(project_path):
...
+    is_server_down = False
+    with open_file(tempfile_status, "r") as f:
+        for line in f:
+            if line == "ansysli_server process could not be found.\n":
+                is_server_down = True
+                break
+
+    if is_server_down:
+        pyaedt_logger.info("License server process could not be found.")
+        return None
```
SMoraisAnsys, line 996:
```suggestion
        pyaedt_logger.info("License server process could not be found.")
        return None
```
Just wondering, would it make sense to log a warning here and return `False` instead of the current code ?
I kinda feel like this function will be used in if statement like  `if is_license_feature_available(): ...` thus I'm not sure if anyone will differentiate between False and None.

SMoraisAnsys (approved):
LGTM, I left a minor question (really no pressure in following what I mention)

pyaedt/generic/general_methods.py
Samuelopez-ansys: In fact I did it in purpose just in case the user needs to check if the server is alive or not. But I agree. I will modify it

SMoraisAnsys (approved):

PR #4799, opened 2024-06-12
FEAT: Get profile data

Samuelopez-ansys:


maxcapodi78 (dismissed):
LGTM

pyaedt/modules/SolveSetup.py
```diff
@@ -221,6 +221,20 @@ def name(self, name):
...
         self.props["Name"] = name
 
+    @pyaedt_function_handler()
+    def get_profile(self):
+        """Solution profile.
+
+        Returns
+        -------
+        dict of :class:`pyaedt.modeler.cad.elements3d.BinaryTree` when solved setups,
+        ``None`` when not solved setup or not compatible application.
```
PipKat, line 231:
```suggestion
        ``None`` when not solved setups or not compatible application.
```
I assume that the second reference to not solved setups should be plural like the reference to "solved setups"?. Can this perhaps be rewritten: dict of class:`pyaedt.modeler.cad.elements3d.BinaryTree` when solved setups exist,
        ``None`` when no solved setups or no compatible application exists.

maxcapodi78 (approved):

PR #4803, opened 2024-06-13
FEAT: Sort flag in dataset

### LINKED ISSUE #4801: Make sorting optional for `DataSet.update()`

The `update()` method in `application.Variables.DataSet` class sorts the dataset by it's X coordinate values - both for 1D and 3D datasets. While this makes sense for 1D, could be problematic for 3D datasets. I was only updateing the `v` column of a 3D dataset, and seeing weird results because of this :)

- At the very least, [improving documentation ](https://aedt.docs.pyansys.com/version/0.9/API/_autosummary/pyaedt.application.Variables.DataSet.update.html) to reflect this sorting behavior would be helpful.
- Make the sorting optional, if possible through a passed argument.

_No response_

Samuelopez-ansys:
Close #4801 

By default is sorting with X values, but in 3D Dataset could be dangerous, so we have added a disable of the sorting.

### DIFF
--- a/_unittest/test_01_Design.py
+++ b/_unittest/test_01_Design.py
@@ -221,6 +221,8 @@ class TestClass:
         assert self.aedtapp.dataset_exists("Test_DataSet", is_project_dataset=True)
         assert ds2.delete()
         assert not self.aedtapp.dataset_exists("Test_DataSet", is_project_dataset=True)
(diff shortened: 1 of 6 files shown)


maxcapodi78 (approved):

SMoraisAnsys (approved):
LGTM. Would it make sense to remove the `key` argument in calls to `sorted` ? That should sort following X, then Y, then Z, ... If that's something that you would like, it might be worth the shot.

PR #4805, opened 2024-06-14
CHORE: mutable args

gmalinve:


pyaedt/modules/solutions.py
```diff
@@ -3911,13 +3911,13 @@ def __init__(
...
         bounces=5,
-        intrinsics={},
+        intrinsics=None,
     ):
         self.is_creeping_wave = is_creeping_wave
         self._postprocessor = postprocessor
         self._ofield = postprocessor.ofieldsreporter
         self.quantity = quantity
-        self.intrinsics = intrinsics
+        self.intrinsics = [] if intrinsics is None else intrinsics
```
MaxJPRey, line 3920:
```suggestion
        self.intrinsics = {} if intrinsics is None else intrinsics
```

MaxJPRey (commented):
Thanks @gmalinve for taking care of that.

SMoraisAnsys (dismissed):
LGTM

pyaedt/modules/solutions.py
gmalinve: @SMoraisAnsys I do agree, based on that if check the only possible option is for self.intrinsics to be a dictionary.
Can we remove the check:
`if isinstance(self.intrinsics, list):` ?
also we have to fix the return type of intrinsicsVar that always returns a string and not a list or dict as stated.
Any comments? @maxcapodi78 @Samuelopez-ansys

SMoraisAnsys: Yeah, I assume that this should be how to handle it.

Samuelopez-ansys: I agree with all, that intrinsics is difficult because the AEDT API returns different types depending on the case, but I agree, we should standardize.

maxcapodi78 (approved):
LGTM

SMoraisAnsys (approved):
LGTM, this will help with future maintenance !

Samuelopez-ansys (approved):

PR #4784, opened 2024-06-10
FIX: metadata in pyproject.toml

jorgepiloto:
This pull-request fixes the metadata included in the `pyproject.toml` file. In addition, new classifiers have been added. Related with #4775.

PR #4812, opened 2024-06-15
FIX: Project classifier to be PyPI compatible

SMoraisAnsys:
Seems like changes in #4784 introduced a typo in the package classifier which lead to [release failure](https://github.com/ansys/pyaedt/actions/runs/9518607641/job/26260047110) when uploading to PyPI.

Here is a fix following [PyPI's classifiers](https://pypi.org/classifiers/).


Samuelopez-ansys (approved):
Thank you!

MaxJPRey (approved):
LGTM. We could show it to the rest of the team in order to increase proper visibility of the libraries.

PR #4813, opened 2024-06-15
FEAT: e field tangential m2d

gmalinve:


Samuelopez-ansys (approved):

_unittest_solvers/test_45_workflows.py
```diff
@@ -228,6 +229,19 @@ def test_09_advanced_fields_calculator_general(self, add_app):
 
         aedtapp.close_project(aedtapp.project_name)
 
+        aedtapp = add_app(application=pyaedt.Maxwell2d,
```
SMoraisAnsys, line 232: Unless I'm wrong, this could be in a different test instead of extending the existing one right ?
In that case, you might want to move `from pyaedt.workflows.project.advanced_fields_calculator import main` to the top of the file.

SMoraisAnsys (commented):
LGTM, I left a minor comment

_unittest_solvers/test_45_workflows.py
Samuelopez-ansys: For now, is Ok extending the UT. But if this extension increases, we will need to split the unit test.

Samuelopez-ansys (approved):

Issue #4815, opened 2024-06-17
Problem to use pyvista >= 0.43.0


SMoraisAnsys:
While making `pyaedt` available in CPython3.12 (see #4759 and #4769) we faced a problem related to `pyvista`.

Indeed, this package has been officially available for CPython3.12 starting with release `0.43.0`. However, starting from this release, problem started to arise (on our side) due to some inner problems related to the way `pyvista` and `vtk`. This problem also arise when using `vtk-osmesa` (in CICD) whatever the version we tried.

Here is a link to the issue opened in `pyvista` repo https://github.com/pyvista/pyvista/issues/6234.

We should leave this issue open to track the state of the `pyvista` and the changes we have to perform (and discard later) to make the repo working with CPython3.12

Tests impacted:
- _unittest/test_04_SBR.py : test_16_read_hdm
- _unittest/test_12_PostProcessing.py : test_70_far_field_data
- _unittest/test_98_Icepak.py : test_02b_PCB_filters

PR #4770, opened 2024-06-05
WIP: Python 3.12 testing

SMoraisAnsys:

PR #4816, opened 2024-06-17
FIX: imp module deprecated

Samuelopez-ansys:
(cherry picked from commit cd8c3ad2a2b62ad71f4eb2cc187ce6c59a348ee9)

@atlanswer I have cherry picked your PR to trigger the CICD. Thank you for your contribution

**Problem**: The `imp` module has been deprecated since version 3.4 and removed in 3.12[^1][^2].

It's used in #2949:
https://github.com/ansys/pyaedt/blob/da0f3f4070070d6da5b480369160f9608803c7df/pyaedt/emit_core/__init__.py#L25

**Changes**: Replace the usage of `imp.find_module()` according to the official guide[^2].

[^1]: https://docs.python.org/3.10/library/imp.html
[^2]: https://docs.python.org/3.12/whatsnew/3.12.html#imp

### DIFF
--- a/pyaedt/emit_core/__init__.py
+++ b/pyaedt/emit_core/__init__.py
@@ -22,8 +22,8 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-import imp
 from importlib import import_module
+from importlib.util import find_spec
 import os
 import sys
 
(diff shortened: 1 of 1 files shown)


SMoraisAnsys (changes requested):
@atlanswer Thanks for the changes proposition ! This is indeed something that must be done to make `pyaedt` compatible with CPython3.12

@Samuelopez-ansys Since we also require to be IronPython3.7 compatible for a few weeks, we can't use the changes proposed as they are as `find_spec` does not exist. This is part of the changes I was working on in #4770.
I added the changes to @maxcapodi78 PR's (#4762) as it seems that we won't be able to have something working with `pyvista` in the coming days. See  #4815 for more details.
I'd recommend to close this PR and merge #4762 after it gets reviewed by someone else than me :)

@atlanswer Thanks again for you changes, this would have been added right away if we didn't already have other jobs lined up.

atlanswer:
Oh, I missed #4762. Nice to see that 3.12 support has been worked on.

PR #4829, opened 2024-06-18
CI: Fix nightly build missing env var

### LINKED ISSUE #4731: CI: Nightly update of the documentation is not working as expected
There seem to be a problem with our nightly documentation build, see [nightly-docs](https://github.com/ansys/pyaedt/actions/workflows/nightly-docs.yml)

Investigation must be performed to understand where this problem is coming from. What appears in the CICD is a time out - current upload of the dev documentation takes ~4 more hours than the upload of the last stable version ...

SMoraisAnsys:
As title says.

Seems like the issue was that we were missing an environment variable.

Closes #4731


Samuelopez-ansys (approved):
Wow! Good catch!

PR #4834, opened 2024-06-19
CHORE: Fix TOML Topic

Samuelopez-ansys:
TOML topic failed again with @SMoraisAnsys modifications. I have triggered again the release, I think it should work with these modifications.

https://github.com/ansys/pyaedt/actions/runs/9587855486

Once the release is created, we can merger this PR.


Samuelopez-ansys:
Release created, and it worked. What do you think @SMoraisAnsys ?

MaxJPRey (approved):

SMoraisAnsys (approved):
Indeed #4812 was not enough as there was still typos... Nice catch an thanks for the fix !

PR #4835, opened 2024-06-20
CI: Temporary remove assets

SMoraisAnsys:
This is a temporary change that should let us continue releasing and which should be reverted once we reduce the size of the assets. See https://github.com/ansys/pyaedt/actions/runs/9587855486/job/26451274049 for effective failure.

Screenshot from raw content
image

Options for such goal are to:
- extract the examples in another repo;
- use autoapi extension to limit the number of files (and thus the size).


Samuelopez-ansys (dismissed):

.github/workflows/ci_cd.yml
```diff
@@ -270,7 +270,7 @@ jobs:
           .venv\Scripts\Activate.ps1
           pytest --durations=50 -v --cov=pyaedt --cov-report=xml --cov-report=html --junitxml=junit/test-results.xml _unittest_solvers
 
-      - uses: codecov/codecov-action@v4
+      - uses: codecov/c aodecov-action@v4
```
SMoraisAnsys, line 273:
```suggestion
      - uses: codecov/codecov-action@v4
```

PR #4849, opened 2024-06-25
REFACTOR: Architecture to match pyansys guide lines

SMoraisAnsys:
This PR is a huge refactoring of the code base. It includes the following changes:
1. move the code from `pyaedt` to `src/ansys/aedt/core`;
2. duplicate the files of `src/ansys/aedt/core` in `src/pyaedt` to allow users to keep being able to do `import pyaedt`;
3. rename files to follow PEP8 convention on module names and harmonize with other python files;
4. change build system from `flint` to `setuptools` to allow building while having 2. working correctly (otherwise, only one target can be used)

Note: a warning is now triggered when one does `import pyaedt` so that every user is aware of this changes. On top of that, the renaming of the files shouldn't be impacting users as we are doing 2 and 4 at the same time. This way, old filenmaes will temporarily remain available as aliases pointing to the newly renamed files, ensuring temporary backward compatibility.


src/pyaedt/__init__.py
```diff
@@ -0,0 +1,31 @@
+from ansys.aedt.core import __version__
+from ansys.aedt.core import *
+
+__version__ = __version__
+
+import warnings
+
+WARNING_MESSAGE = """Module 'pyedt' has become an alias to the new package structure.
```
Samuelopez-ansys, line 8:
```suggestion
WARNING_MESSAGE = """Module 'pyaedt' has become an alias to the new package structure.
```

src/pyaedt/hfss.py
```diff
@@ -0,0 +1 @@
+from ansys.aedt.core...hfss import *
```
Samuelopez-ansys, line 1:
```suggestion
from ansys.aedt.core.hfss import *
```

Samuelopez-ansys (approved):

.gitignore
```diff
@@ -395,6 +395,7 @@ model.index\+
 
 # local environment settings used by e.g. Visual Studio Code
 /.env
+/.venv*
```
myoung301, line 398: The added line looks to be redundant with line 400.

```suggestion
```

Samuelopez-ansys (approved):

PR #4857, opened 2024-06-28
FIX: Design's save_project fails to save project when file_name's parent dir does not exist

alexciurana:
Fixed a bug in `Design.save_project()` where the parent directory of `file_name` was created but the project was not saved. Added a test to verify the fix.

Additionally, modified the method to account for scenarios where `remote_rpc_session` is enabled, utilizing its file manager for path operations.


Samuelopez-ansys (approved):

SMoraisAnsys (commented):
Not sure if we should add the `create_parent_dir` argument.
Saying that, adding a check to create the directory should be the default approach in my opinion (like you just did)
What about restricting the changes to modifying `elif file_name` into `if file_name` like you did ?

pyaedt/application/Design.py
```diff
@@ -3739,6 +3739,8 @@ def save_project(self, file_name=None, overwrite=True, refresh_ids=False):
         refresh_ids : bool, optional
             Whether to refresh object IDs after saving the project.
             The default is ``False``.
+        create_parent_dir : bool, optional
+            Whether to create the parent directory of file_name if it does not exist.
```
PipKat, line 3743:
```suggestion
            Whether to create the parent directory for the project if it does not exist.
```
Whether than use the code entity, consider rewriting in plain English: Whether to create the parent directory of the ``file_name`` parameter
            if it does not exist.

alexciurana:
> Of course, if there is a use case that I'm not thinking of where one could want to use `create_parent_dir=False`, then you can ignore this comment.

Hello, thanks for the quick review. The use case is for client-server scenarios where PyAEDT and AEDT may have access to different file systems, or if they share a file volume, it might be accessible through different paths.
 
For example:
 
- AEDT runs on Windows.
- PyAEDT client runs on WSL.
 
Both environments share a file system volume, but the path to access it is different.
 
Assume our project file is at `C:\my_projects\my_project.aedt`. This path should be passed to `save_project` since it's the correct path for AEDT. However, from WSL, this isn't a valid path, and we shouldn't try to operate on it, such as creating directories. That can be handled externally by the application. To access the file from WSL, we should use `/mnt/c/my_projects/my_project.aedt`. This scenario can be extended to other situations, such as running PyAEDT client in a Docker container on WSL, AEDT running on other servers, etc.

SMoraisAnsys (dismissed):
Awesome, the changes follows valid requirements for remotes rpc sessions !

SMoraisAnsys:
Did you run the tests locally ? I'll rerun the tests but there seem to be something failing.

alexciurana:
> Did you run the tests locally ? I'll rerun the tests but there seem to be something failing.

Yes. For example, for the file that I modified:
```
❯ pytest -n 1 .\_unittest\test_01_Design.py
59 passed, 1 warning in 51.05s 
```

SMoraisAnsys (approved):
LGTM
@alexciurana Thanks for updating to main branch, I was about to merge the doc fix.

SMoraisAnsys:
Closing this in favor of #4868 to avoid secret issues.

PR #4863, opened 2024-07-01
DOCS: Fix title to follow google headings

SMoraisAnsys:
Previous PR on extension seem to have added a warning that does not pass the CICD anymore.
This PR should fix the typo problem


SMoraisAnsys:
> @SMoraisAnsys The proper name for the app is HFSS 3D Layout. Perhaps try adding all three words as one line entry to Vale's accept list? The heading would then be HFSS 3D Layout extensions.

Yeah I though that all of them where added but we were missing 3D... I also added 2D to avoid future issues and removed full expressions.

MaxJPRey (approved):
LGTM.

PR #4881, opened 2024-07-04
FIX: Vale usage and its integration in CICD

### LINKED ISSUE #4880: CI: Vale warning can get merged but prevent later merges
Some recent PR added changes that trigger `vale` warning(s) with our configuration of vale.
While it is not a problem for the PR introducing this changes, it becomes a problem for following PR(s) as, afterward, the warning make the CICD fail. See the two behaviors bellow:
- the first one is the PR adding the change (no issue)
- the second one is a PR created from main after the warning got merged

image

image

SMoraisAnsys:
Changes:
- upgrade of the CICD and changes in `vale` related folder to use version `>=3` of `vale`;
- fix headings;
- remove google rules (duplicated from online);
- **modify the severity of google rules using severity level of `warning` into `error`.**

The reason for the last change is related to #4880.
The problem this PR tackles is the fact that a PR from a fork, dependabot, ... ends up failing the `doc-style` job of our CICD.

From my understanding of the problem, it is related to the permissions associated to GitHub token in the event [pull_request_target](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target). Since this permissions can differ from one context to another, some PR might add changes passing `doc-style` but doing so while adding `vale` warnings. Those are pending bombs as they will impact future PR(s) where the permissions are not enough (for example, cases where GitHub token doesn't have write permission of Review API). Indeed, in such cases, the report results are performed via logging command and creating annotations similar to `github-pr-check` reporter as a fallback. However, the behavior is different as in this case warnings cause reviewdog to exit with status code 1 and make the `documentation-style` fail.

Proof that author access has no impact: PR #4882 comes from my fork of pyaedt and has the same permission issue.

This seems to be related to https://github.com/reviewdog/reviewdog/issues/1408 and the workaround we are applying here is to update every `warning` severity to `error`.

Closes #4880


MaxJPRey (approved):
Nice work @SMoraisAnsys . We can now properly use the standard and up-to-date doc-style action.

SMoraisAnsys:
Temporary passing to draft due to CI typo (deleted breaking lines)

Samuelopez-ansys (approved):

PR #4884, opened 2024-07-04
DOCS: Added visualization to documentation

maxcapodi78:
Class TouchstoneData was missed in documentation
All visualization classes are not in a separate section


Samuelopez-ansys (approved):

pyaedt/generic/plot.py
```diff
@@ -1733,8 +1733,13 @@ def s_callback():  # pragma: no cover
             self.pv.screenshot(exp, return_img=False)
 
         self.pv.add_key_event("s", s_callback)
-        if export_image_path:
-            self.pv.show(screenshot=export_image_path, full_screen=True)
+        if export_image_path:  # pragma: no cover
+            supported_export = [".svg", ".pdf", ".eps", ".ps", ".tex"]
+            extension = os.path.splitext(export_image_path)[1]
+            if extension in supported_export:
+                self.pv.save_graphic(export_image_path, raster=raster, painter=painter)
```
SMoraisAnsys, line 1740: @maxcapodi78 I think there is a bug here as `raster` and `painter` are not initialized. Do you want to set this variables or should we remove them ?

maxcapodi78: I've fixed. Thanks

PR #4893, opened 2024-07-08
FEAT: Farfield refactoring

Samuelopez-ansys:


Samuelopez-ansys:
@SMoraisAnsys Any idea why documentation style is failing?

pyaedt/generic/farfield_visualization.py
```diff
@@ -0,0 +1,2187 @@
...
+            if incident_power:
+                for power_freq in incident_power:
+                    value = incident_power[power_freq]
+                    frequency = power_freq
+                    if isinstance(power_freq, str):
+                        frequency, units = decompose_variable_value(power_freq)
+                        if units:  # pragma: no cover
+                            frequency = unit_converter(frequency, "Freq", units, "Hz")
+
+                    new_incident_power[frequency] = value
```
SMoraisAnsys, line 193:
```suggestion
            new_incident_power = {}
            if incident_power:
                for power_freq in incident_power:
                    value = incident_power[power_freq]
                    frequency = power_freq
                    if isinstance(power_freq, str):
                        frequency, units = decompose_variable_value(power_freq)
                        if units:  # pragma: no cover
                            frequency = unit_converter(frequency, "Freq", units, "Hz")
                    new_incident_power[frequency] = value
```

pyaedt/generic/farfield_visualization.py
```diff
@@ -0,0 +1,2187 @@
...
+            self.frequency = self.frequencies[freq_index]
+        else:
+            self.frequency = self.frequencies[0]
+
+        # Update
+        for row, col in port_indices.values():
+            self.__a_min = min(self.__a_min, row - 1)
+            self.__a_max = max(self.__a_max, row - 1)
+            self.__b_min = min(self.__b_min, col - 1)
+            self.__b_max = max(self.__b_max, col - 1)
```
SMoraisAnsys, line 276: you might want to use numpy here to speed things up

pyaedt/generic/farfield_visualization.py
```diff
@@ -0,0 +1,2187 @@
...
+
+    @property
+    def s_parameters(self):
+        """Passive s-parameters"""
+        if self.touchstone_data:
+            touchstone_frequencies = self.touchstone_data.f
+            index = np.abs(touchstone_frequencies - self.frequency).argmin()
+            return self.touchstone_data.s[index]
+        else:
+            return None
```
SMoraisAnsys, line 316:
```suggestion
        return None  # pragma: no cover
```

pyaedt/generic/farfield_visualization.py
```diff
@@ -0,0 +1,2187 @@
...
+        return self.__incident_power_element
+
+    @property
+    def incident_power(self):
+        """Total incident power in watts"""
+        incident_power_element = self.incident_power_element
+        if incident_power_element:
+            return sum(incident_power_element.values())
+        else:  # pragma: no cover
+            return None
```
SMoraisAnsys, line 338:
```suggestion
        return None  # pragma: no cover
```

pyaedt/generic/farfield_visualization.py
```diff
@@ -0,0 +1,2187 @@
...
+        return power
+
+    @property
+    def accepted_power(self):
+        """Total accepted power in watts"""
+        power_element = self.accepted_power_element
+        if power_element:
+            return sum(power_element.values())
+        else:  # pragma: no cover
+            return None
```
SMoraisAnsys, line 359:
```suggestion
        return None  # pragma: no cover
```

pyaedt/generic/farfield_visualization.py
```diff
@@ -0,0 +1,2187 @@
...
+        return power
+
+    @property
+    def radiated_power(self):
+        """Total radiated power in watts"""
+        power_element = self.radiated_power_element
+        if power_element:
+            return sum(power_element.values())
+        else:  # pragma: no cover
+            return None
```
SMoraisAnsys, line 380:
```suggestion
        return None  # pragma: no cover
```

pyaedt/generic/farfield_visualization.py
```diff
@@ -0,0 +1,2187 @@
...
+
+        # Combine farfield of each port
+        initial_port = self.all_element_names[0]
+        freq_name_key = self.frequencies[self.__freq_index]
+        data = self.__raw_data[initial_port][freq_name_key]
+        length_of_ff_data = len(data["rETheta"])
+        theta_range = data["Theta"]
+        phi_range = data["Phi"]
+        Ntheta = len(theta_range)
+        Nphi = len(phi_range)
```
SMoraisAnsys, line 555: Can you rename those variables ?

pyaedt/generic/farfield_visualization.py
```diff
@@ -0,0 +1,2187 @@
...
+            for _, element in enumerate(self.all_element_names):
+                if self.active_s_parameters[element] is not None:
+                    operation = 1 - np.power(np.abs(self.active_s_parameters[element]), 2)
+                    accepted_power[element] = self.incident_power_element[element] * operation
+                else:
+                    accepted_power[element] = 0.0
+            total_accepted_power = sum(accepted_power.values())
+            return total_accepted_power
+        else:  # pragma: no cover
+            return None
```
SMoraisAnsys, line 658:
```suggestion
        return None  # pragma: no cover
```

pyaedt/generic/farfield_visualization.py
```diff
@@ -0,0 +1,2187 @@
...
+        edgeTaper_dB = -200
+
+        edgeTaper = 10 ** ((float(edgeTaper_dB)) / 20)
+
+        threshold = 1e-10
+
+        # find the distance between current cell and array center in terms of index
+        lattice_vector = self.__lattice_vector[self.__freq_index]
+        if not lattice_vector or not len(lattice_vector) == 6:
+            return 1
```
SMoraisAnsys, line 691: Not sure it makes sense to return 1 here, can you explain why ?

pyaedt/generic/farfield_visualization.py
```diff
@@ -0,0 +1,2187 @@
...
+        >>> data.plot_contour()
+
+        """
+        if not title:
+            title = quantity
+
+        data = self.combine_farfield(phi, theta)
+        if quantity not in data:  # pragma: no cover
+            self.__logger.error("Far field quantity is not available.")
+            return False
```
SMoraisAnsys, line 867: Function signature doesn't match `bool`. Should we raise an exception since we leverage the `pyaedt_function_handler` ?

pyaedt/generic/farfield_visualization.py
```diff
@@ -0,0 +1,2187 @@
...
+        select = np.abs(data["Theta"]) <= max_theta  # Limit theta range for plotting.
+
+        data_to_plot = data[quantity][select, :]
+        data_to_plot = conversion_function(data_to_plot, quantity_format)
+        if not isinstance(data_to_plot, np.ndarray):  # pragma: no cover
+            self.logger.error("Wrong format quantity")
+            return False
```
SMoraisAnsys, line 874: See previous title.

pyaedt/generic/farfield_visualization.py
```diff
@@ -0,0 +1,2187 @@
...
+        >>> setup_name = "Setup1 : LastAdaptive"
+        >>> frequencies = [77e9]
+        >>> sphere = "3D"
+        >>> data = app.get_antenna_data(frequencies,setup_name,sphere)
+        >>> data.plot_cut(theta=20)
+        """
+
+        data = self.combine_farfield(phi, theta)
+        if quantity not in data:  # pragma: no cover
+            self.logger.error("Far field quantity not available")
```
SMoraisAnsys, line 960: Same as before

pyaedt/generic/farfield_visualization.py
```diff
@@ -0,0 +1,2187 @@
...
+        >>> app = pyaedt.Hfss(version="2023.2", design="Antenna")
+        >>> setup_name = "Setup1 : LastAdaptive"
+        >>> frequencies = [77e9]
+        >>> sphere = "3D"
+        >>> data = app.get_antenna_data(frequencies,setup_name,sphere)
+        >>> data.polar_plot_3d(theta=10)
+        """
+        data = self.combine_farfield(phi, theta)
+        if quantity not in data:  # pragma: no cover
+            self.logger.error("Far field quantity is not available.")
```
SMoraisAnsys, line 1080: Again

pyaedt/generic/farfield_visualization.py
```diff
@@ -0,0 +1,2187 @@
...
+            return False
+
+        ff_data = conversion_function(data[quantity], quantity_format)
+        if not isinstance(ff_data, np.ndarray):  # pragma: no cover
+            self.logger.error("Format of the quantity is wrong.")
```
SMoraisAnsys, line 1085: Again :)

pyaedt/generic/farfield_visualization.py
```diff
@@ -0,0 +1,2187 @@
...
+        elif isinstance(rotation, (list, tuple)):  # pragma: no cover
+            rotation = np.array(rotation)
+        text_color = "white"
+        if background is None:
+            background = [255, 255, 255]
+            text_color = "black"
+
+        farfield_data = self.combine_farfield(phi_scan=0, theta_scan=0)
+        if quantity not in farfield_data:  # pragma: no cover
+            self.logger.error("Far field quantity is not available.")
```
SMoraisAnsys, line 1180: Again, i'll stop notifying it for this file

pyaedt/generic/farfield_visualization.py
```diff
@@ -0,0 +1,2187 @@
...
+                toggle_vis_cad(False)
+            else:
+                p.add_checkbox_button_widget(toggle_vis_cad, value=True, position=(10, 70), size=30)
+
+            p.add_text("Show Geometry", position=(70, 75), color=text_color, font_size=10)
+
+        if image_path:
+            p.show(screenshot=image_path)
+        if show:  # pragma: no cover
+            p.show()
```
SMoraisAnsys, line 1321: You should update this to avoid the problem of calling show without specifying autoclose, see https://github.com/ansys/pyaedt/pull/4912 and the issue associated 
Note: I might have missing this in other methods

pyaedt/generic/farfield_visualization.py
```diff
@@ -0,0 +1,2187 @@
...
+
+        theta = np.deg2rad(np.array(farfield_data["Theta"]))
+        phi = np.deg2rad(np.array(farfield_data["Phi"]))
+        mesh = get_structured_mesh(theta=theta, phi=phi, ff_data=ff_data)
+        return mesh
+
+    @pyaedt_function_handler()
+    def __get_geometry(self):
+        """Get 3D meshes."""
+        from pyaedt.generic.plot import ModelPlotter
```
SMoraisAnsys, line 1428: Can this be moved at the module level instead of the function level ?

pyaedt/generic/farfield_visualization.py
```diff
@@ -0,0 +1,2187 @@
...
+            object_name = os.path.basename(obj.path)
+            name = os.path.splitext(object_name)[0]
+            original_path = os.path.dirname(obj.path)
+            new_path = os.path.join(os.path.abspath(export_path), object_name)
+
+            if not os.path.exists(new_path):
+                new_path = shutil.move(obj.path, export_path)
+            if os.path.exists(os.path.join(original_path, name + ".mtl")):  # pragma: no cover
+                try:
+                    os.remove(os.path.join(original_path, name + ".mtl"))
```
SMoraisAnsys, line 2130: Try to use (shutil.rmtree)[https://docs.python.org/3/library/shutil.html#shutil.rmtree] instead, you can ignore errors with an argument

SMoraisAnsys (changes requested):
Thanks for this huge rework and great use of "scope" with variable/method names !
Warning: while this is a great change, you have to really rack your brain to ensure that the scope you defined won't change later on as this will be though to handle !

I left some minor comments. Among them is the fact of returning directly a value instead of using `else` statement before. JFYI, if we don't specify a value to be returned, Python returns `None` by default. So, we can discard `return None` if you want !

pyaedt/generic/farfield_visualization.py
Samuelopez-ansys: @SMoraisAnsys This method is computing a number, that is called the weight of the element, which later is multiplied to the element field. If there isn't lattice vector (if it not an array), the weight can not be computer, and it is returning 1.0, because multiplying the field by 1 will not have any effect, I modified to 1.0 , because it is returning a float.

PR #4898, opened 2024-07-09
FIX: replaced argument output_dir with output_file for method export_field_file

Alberto-DM:
replaced argument output_dir with output_file for method export_field_file as it requires to specify a file path, not a dir.


MaxJPRey (approved):
LGTM.

PR #4913, opened 2024-07-12
FIX: Export configuration and AEDT crash

maxcapodi78:
- export_configuration method was failing under certain circumstances
- aedt_objects were still referenced after a project was closed and could cause crash in AEDT during debug mode.


pyaedt/generic/configurations.py
```diff
@@ -1700,22 +1700,27 @@ def _export_mesh_operations(self, dict_out):
...
             for mesh in self._app.mesh.meshregions:
-                if mesh.name == "Settings":
+                if mesh.name in ["Settings", "Global"]:
                     args = ["NAME:Settings"]
                 else:
                     args = ["NAME:" + mesh.name, "Enable:=", mesh.Enable]
                 args += mesh.settings.parse_settings_as_args()
-                args += getattr(mesh, "_parse_assignment_value")()
+                if mesh.name not in ["Settings", "Global"]:
+                    args += getattr(mesh, "_parse_assignment_value")()
```
SMoraisAnsys, line 1709: Can you move that line in the previous `else` statement ? You would be in the case were `mesh.name not in ["Settings", "Global"]`. Unless you need this to be added after `mesh.settings.parse_settings_as_args()`

pyaedt/generic/configurations.py
```diff
@@ -1700,22 +1700,27 @@ def _export_mesh_operations(self, dict_out):
...
                 args += ["UserSpecifiedSettings:=", not mesh.manual_settings]
                 mop = OrderedDict({})
                 _arg2dict(args, mop)
-                if self._app.modeler[args[-3][0]].history().command == "CreateSubRegion":
+                if (
+                    mesh.name not in ["Settings", "Global"]
+                    and self._app.modeler[args[-3][0]].history().command == "CreateSubRegion"
+                ):
```
SMoraisAnsys, line 1716: You might also consider moving that above

SMoraisAnsys (approved):
LGTM, I left minor comments

pyaedt/generic/configurations.py
maxcapodi78: same comment as above

Alberto-DM (dismissed):
LGTM

SMoraisAnsys (dismissed):

SMoraisAnsys (approved):

PR #4929, opened 2024-07-17
REFACTOR: Refacto args

gmalinve:


_unittest/test_02_3D_modeler.py
```diff
@@ -1052,7 +1052,7 @@ def test_59a_region_property(self):
 
     def test_59b_region_property_failing(self):
         self.aedtapp.modeler.create_air_region()
-        assert not self.aedtapp.modeler.change_region_coordinate_system(region_cs="NoCS")
+        assert not self.aedtapp.modeler.change_region_coordinate_system(coordinate_system="NoCS")
```
maxcapodi78, line 1055: why not using simply "name"? the fact that is a coordinate system is already in method name

_unittest/test_08_Primitives3D.py
```diff
@@ -1215,38 +1221,35 @@ def test_64_create_3dcomponent(self):
...
-            self.component3d_file,
-            included_cs="Global",
-            is_encrypted=True,
-            password="password_test",
+            self.component3d_file, coordinate_systems="Global", is_encrypted=True, password="password_test"
         )
         assert self.aedtapp.modeler.create_3dcomponent(
             self.component3d_file,
-            included_cs="Global",
+            coordinate_systems="Global",
```
maxcapodi78, line 1238: in this sense instead it makes sense to use coordinate_system (singular)

examples/04-Icepak/Icepak_3DComponents_Example.py
```diff
@@ -175,13 +168,9 @@
...
-ipk.modeler.create_3dcomponent(
-    component_file=os.path.join(temp_folder, "componentLibrary", "PCBAssembly.a3dcomp"),
-    component_name="PCBAssembly",
-    auxiliary_dict=True,
-    included_cs=["Global", "HeatsinkCS", "PCB_Assembly"],
-    reference_cs="PCB_Assembly"
-)
+ipk.modeler.create_3dcomponent(input_file=os.path.join(temp_folder, "componentLibrary", "PCBAssembly.a3dcomp"),
+                               name="PCBAssembly", coordinate_systems=["Global", "HeatsinkCS", "PCB_Assembly"],
+                               reference_cs="PCB_Assembly", export_auxiliary=True)
```
maxcapodi78, line 173: should not be "reference_coordinate_system"

examples/05-Q3D/Q2D_Example_Stripline.py
```diff
@@ -212,9 +212,9 @@
 a.plot(snapshot_path=os.path.join(q.working_directory, "plot.jpg"))  # Save plot as jpg
 
 # Add a parametric sweep and analyze.
-parametric = q.parametrics.add(sweep_var="sig_bot_w", start_point=75, end_point=100, step=5,
+parametric = q.parametrics.add(sweep_variable="sig_bot_w", start_point=75, end_point=100, step=5,
```
maxcapodi78, line 215: maybe simply "variable"

maxcapodi78 (changes requested):
few comments

_unittest/test_02_3D_modeler.py
gmalinve: if you look at the implemantation I already used "name" for the region name:
`@pyaedt_function_handler(region_cs="coordinate_system", region_name="name")`
`def change_region_coordinate_system(self, coordinate_system="Global", name="Region")`
@maxcapodi78

examples/04-Icepak/Icepak_3DComponents_Example.py
gmalinve: @maxcapodi78 I kept it as is but I can change reference_cs to reference_coordinate_system

_unittest/test_02_3D_modeler.py
maxcapodi78: in such case better to use assignment.

maxcapodi78 (approved):

PR #4951, opened 2024-07-25
FEAT: added import and export method to Rmxprt class

maxcapodi78:


pyaedt/rmxprt.py
```diff
@@ -325,3 +329,69 @@ def create_setup(self, name="MySetupAuto", setup_type=None, **kwargs):
...
+        -------
+        bool
+            ``True`` when successful, ``False`` when failed.
+        """
+
+        from pyaedt.generic.general_methods import read_configuration_file
+
+        new_dict = read_configuration_file(input_file)
+        for k, v in new_dict.items():
+            dictin = self.stator
```
Alberto-DM, line 386: For readability, I would put the line `dictin = self.stator` in an `else:` at the end of the if... else chain.

pyaedt/rmxprt.py
```diff
@@ -325,3 +329,69 @@ def create_setup(self, name="MySetupAuto", setup_type=None, **kwargs):
...
+        str
+           Full path to json file created.
+        """
+
+        def jsonalize(dict_in, dict_out):
+            dict_out[dict_in.node] = {}
+            for k, v in dict_in.props.items():
+                if not k.endswith("/Choices"):
+                    dict_out[dict_in.node][k] = v
+            for _, c in dict_in.children.items():
```
Alberto-DM, line 353: why not taking directly the values?

```suggestion
            for c in dict_in.children.values():
```

Alberto-DM (dismissed):
I made two comments for readability of the code. Otherwise LGTM.

pyaedt/rmxprt.py
maxcapodi78: I didn't do because the list of options for stator is the biggest :-D

Alberto-DM (approved):
LGTM

PR #4957, opened 2024-07-26
FEAT: Symbolic link removed it

Samuelopez-ansys:


SMoraisAnsys (approved):
LGTM, one less thing related to Iron Python !

PR #4976, opened 2024-07-31
FEAT: implemented new method get_all_report_quantities

maxcapodi78:
to get all the report quantities organized by type, solution and category.


Samuelopez-ansys (dismissed):

pyaedt/modules/PostProcessor.py
```diff
@@ -1122,6 +1122,52 @@ def available_report_quantities(
...
+            A dictionary with primary key the report type, secondary key the solution type and
+            third key the report categories.
+        """
+        rep_quantities = {}
+        for rep in self.available_report_types:
+            rep_quantities[rep] = {}
+            if solution:
+                sols = [solution]
+            else:
+                sols = self.available_report_solutions(rep)
```
SMoraisAnsys, line 1159: Would it make sense to use an argument `solutions` instead of `solution` ? I mean, could a user want the report quantities for a set of solutions instead of one (or all if nothing is provided) ?

If yes, I would suggest to use `solutions` as an argument. That way, the behavior for one unique solution, multiples or none provided would be the same internally and we could get rid of `sols` and `sol` to use `solutions` and `solution`.

SMoraisAnsys (approved):
LGTM, I left a minor comment which might require some refactoring if you agree with ti

pyaedt/modules/PostProcessor.py
maxcapodi78: Even if I cannot figure out why, it could be the case

maxcapodi78: I've implemented it

pyaedt/modules/PostProcessor.py
```diff
@@ -1122,6 +1122,52 @@ def available_report_quantities(
...
+        for rep in self.available_report_types:
+            rep_quantities[rep] = {}
+            if isinstance(solutions, str):
+                solutions = [solutions]
+            elif solutions is None:
+                solutions = self.available_report_solutions(rep)
+            for sol in solutions:
+                rep_quantities[rep][sol] = {}
+                for quant in self.available_quantities_categories(
+                    rep, context=context, solution=sol, is_siwave_dc=is_siwave_dc
```
SMoraisAnsys, line 1163:
```suggestion
        solutions=None,
        context=None,
        is_siwave_dc=False,
    ):
        """Return all the possible report categories organized by report types, solution and categories.

        Parameters
        ----------
        solutions : list, optional
            Solutions to get the report quantities.
            The default is ``None``, in which case the all solutions are used.
        context : str, dict, optional
            Report Context.
            The default is ``None``, in which case the default context is used.
            For Maxwell 2D/3D Eddy Current solution types this can be provided as a dictionary
            where the key is the matrix name and value the reduced matrix.
        is_siwave_dc : bool, optional
            Whether if the setup is Siwave DCIR or not. Default is ``False``.

        Returns
        -------
        dict
            A dictionary with primary key the report type, secondary key the solution type and
            third key the report categories.
        """
        rep_quantities = {}
        for rep in self.available_report_types:
            rep_quantities[rep] = {}
            if solutions is None:
                solutions = self.available_report_solutions(rep)
            for solution in solutions:
                rep_quantities[rep][solution] = {}
                for quant in self.available_quantities_categories(
                    rep, context=context, solution=solution, is_siwave_dc=is_siwave_dc
```

pyaedt/modules/PostProcessor.py
```diff
@@ -1122,6 +1122,52 @@ def available_report_quantities(
...
+                ):
+                    rep_quantities[rep][sol][quant] = self.available_report_quantities(
+                        rep, quantities_category=quant, context=context, solution=sol, is_siwave_dc=is_siwave_dc
+                    )
```
SMoraisAnsys, line 1167:
```suggestion
                    rep_quantities[rep][solution][quant] = self.available_report_quantities(
                        rep, quantities_category=quant, context=context, solution=solution, is_siwave_dc=is_siwave_dc
                    )
```

SMoraisAnsys (commented):
I was thinking about constraining users to provide a list of strings. If no value is provided (default `None`) then we fetch all solutions.
This way, we allow users to switch between one solution (list with one element), multiple solutions (list with any number of elements) and the default behavior of retrieving all available solutions.

If it does not make that much sense then we can revert to your previous implementation. My idea was to couple the internal implementation with what the user can see from the method signature.

pyaedt/modules/PostProcessor.py
maxcapodi78: It doesn't make much sense. Most of the time (>90% of the cases) user will enter a single setup or nothing. The way I've implemented still the user can enter more than 1 setup but I would expect not much cases on that.

maxcapodi78: Furthermore the way I've implemented keeps the compatibility with other methods in post (available_quantities_categories, available_report_quantites...) which expects a solution as an input

SMoraisAnsys (approved):

Samuelopez-ansys (approved):

PR #4980, opened 2024-07-31
FEAT: Desktop launch

maxcapodi78:
Right now we manually start AEDT and then connect to the server through Grpc. With this refactoring we are gonna use AEDT API to launch AEDT in version greater than 2024R1 and leave the same behaviour in versions lower than 24R1.
@Samuelopez-ansys maybe some additional tests are needed before merging. In my tests this approach seemed more stable. I'd like to check how parallel UT works


MaxJPRey:
@maxcapodi78 So from now on, we will be able to start the Desktop with command line through gRPC?

Samuelopez-ansys (approved):

PR #4981, opened 2024-07-31
FEAT: Fields markers

lorenzovecchietti:


lorenzovecchietti:
@Samuelopez-ansys @SMoraisAnsys Codecov is not passing as the method works only in graphical mode (bug already reported)

pyaedt/modules/solutions.py
```diff
@@ -1797,6 +1801,112 @@ def plotsettings(self):
             ]
         return arg
 
+    @pyaedt_function_handler()
+    def get_points_value(self, points, filename=None, visibility=False):
```
Samuelopez-ansys, line 1805:
```suggestion
    def get_points_value(self, points, filename=None, visibility=False):  # pragma: no cover
```

Samuelopez-ansys (approved):

Samuelopez-ansys (dismissed):

pyaedt/modules/solutions.py
```diff
@@ -1797,6 +1801,112 @@ def plotsettings(self):
...
+            Each key is associated with a list with the same length of the argument points.
+            If pandas is installed, the output is a pandas DataFrame with point names as
+            index and coordinates and quantity as columns.
+        """
+        self.oField.ClearAllMarkers()
+
+        # Clean inputs
+        if isinstance(points, dict):
+            points_name = list(points.keys())
+            points_value = [points[pt_name] for pt_name in points_name]
```
SMoraisAnsys, line 1840:
```suggestion
            points_name, points_value = list(points.keys()), list(points.values())
```

pyaedt/modules/solutions.py
```diff
@@ -1797,6 +1801,112 @@ def plotsettings(self):
...
+            with open_file(temp_file.name, "r") as f:
+                reader = csv.DictReader(f)
+                out_dict = defaultdict(list)
+                for row in reader:
+                    for key in row.keys():
+                        if key == "Name":
+                            val = row[key]
+                        else:
+                            val = float(row[key].lstrip())
+                        out_dict[key.lstrip()].append(val)
```
SMoraisAnsys, line 1881: It is a bit awkward to using a context manager to open a file and closing it right away.
Could you create the temp_file through classical variable assignment, call `ExportMarkerTable` afterward and then use the context manager if needed ?

```python
temp_file = tempfile..NamedTemporaryFile(mode="w+", delete=False, suffix=".csv")
temp_file.close()
self.oField.ExportMarkerTable(temp_file.name)
...
```

SMoraisAnsys (commented):
@lorenzovecchietti I left some minor comments to ease future maintenance / readability

PR #4984, opened 2024-08-01
DOCS: pyvista issue in Array example

Samuelopez-ansys:


examples/02-HFSS/Array.py
```diff
@@ -148,10 +148,6 @@
...
-plot_3d = ffdata.plot_3d(quantity='RealizedGain', image_path=os.path.join(working_directory, "Image.jpg"),  show=False)
-
-##########################################################
-# Close all matplotlib figures
-# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
-
-plt.close('all')
+# plot_3d = ffdata.plot_3d(quantity='RealizedGain',
+#                          output_file=os.path.join(working_directory, "Image.jpg"),
+#                          show=False)
```
SMoraisAnsys, line 153: Since this is a temporary comment, can you create an issue to track this problem later ?

SMoraisAnsys (dismissed):
LGTM

_unittest/test_46_FarField.py
```diff
@@ -141,13 +141,13 @@ def test_04_far_field_data(self, local_scratch):
         assert ffdata.get_accepted_power()
 
         img1 = os.path.join(self.local_scratch.path, "ff_2d1.jpg")
-        ffdata.plot_cut(primary_sweep="Theta", secondary_sweep_value="all", image_path=img1, show=False)
+        ffdata.plot_cut(primary_sweep="Theta", secondary_sweep_value="all", output_file=img1, show=False)
```
gmalinve, line 144:
```suggestion
        ffdata.plot_cut(primary_sweep="FIX", secondary_sweep_value="all", output_file=img1, show=False)
```

SMoraisAnsys:
@Samuelopez-ansys @maxcapodi78 It seems that I had forgotten to remove a line of code but the changes are working. However, there seem to be something that was not expected as I obtain the following result

image

From my understanding, this is because of the plotter created in `__get_geometry`. We might want to change the current behavior of the method, for example deleting each plotter at the end of the method ?

Samuelopez-ansys:
@SMoraisAnsys we can try your suggestion, tomorrow I can try, if you know how to do it please commit the modification

SMoraisAnsys:
I added a custom `close` method to `ModelPlotter`. This method ensures that we are closing the `pyvista` plotter associated to the created model plotters AND ensures that it does not get rendered when building the documentation.
Previously the plotters were added because we are using `pvista.BUILDING_GALLERY = True`. Now our code should be working for any value of `pvista.BUILDING_GALLERY`.

PR #4990, opened 2024-08-02
FEAT: Add transient support

lorenzovecchietti:


Samuelopez-ansys (approved):

pyaedt/modules/AdvancedPostProcessing.py
```diff
@@ -1111,6 +1105,8 @@ def evaluate_faces_quantity(
             Dictionary of parameters defined for the specific setup with values. The default is ``{}``.
         ref_temperature: str, optional
             Reference temperature to use for heat transfer coefficient computation. The default is ``""``.
+        time : str
```
siva-krishnaswamy, line 1108:
```suggestion
        time : str, optional
```

pyaedt/modules/PostProcessor.py
```diff
@@ -5710,6 +5711,8 @@ def add_calculation(
         ref_temperature : str, optional
             Reference temperature to use in the calculation of the heat transfer
             coefficient. The default is ``"AmbientTemp"``.
+        time : str
```
siva-krishnaswamy, line 5714:
```suggestion
        time : str, optional
```

_unittest/test_98_Icepak.py
```diff
@@ -1801,3 +1802,11 @@ def test_80_global_mesh_region(self):
...
         g_m_r.global_region.object.material_name = "Carbon Monoxide"
         assert g_m_r.global_region.object.material_name == "Carbon Monoxide"
+
+    def test_81_transient_fs(self, add_app):
+        app = add_app(application=Icepak, project_name=transient_fs, subfolder=test_subfolder)
+        fs = app.post.create_field_summary()
+        for t in ["0s", "1s", "2s", "3s", "4s", "5s"]:
+            fs.add_calculation("Object", "Surface", "Box1", "Temperature", time=t)
+        df = fs.get_field_summary_data(pandas_output=True)
+        assert not df["Mean"].empty
```
siva-krishnaswamy, line 1812: can we add also assert len(df) == 6 to ensure that all timesteps are added in the calculation. Also what will happen if the timestep provided is not in the solution timestep ?

siva-krishnaswamy (commented):
looks good. Just few suggestions.  Also what will happen if the timestep provided is not in the solution timestep ?

pyaedt/modules/AdvancedPostProcessing.py
lorenzovecchietti: You're right. I should've added `optional`. I'll add that in another PR.

_unittest/test_98_Icepak.py
lorenzovecchietti: You'll get an error raised from AEDT. I didn't implement any sanity check (neither the possibility to add all time steps) as there is no way (from APIs) to get saved times. We could read this from setup props, but I don't think it's worth it. It can cause more problems than solutions.

PR #4992, opened 2024-08-02
FIX: improved _find_free_port method

maxcapodi78:
renamed _check_grpc_port to _is_port_occupied


Alberto-DM (approved):
LGTM

pyaedt/desktop.py
```diff
@@ -236,8 +236,9 @@ def _find():
 
     while True:
         new_port = _find()
-        if new_port not in list(active_sessions().values()):
+        if new_port not in list(active_sessions().values()) and new_port not in range(50051, 50070, 1):
```
MaxJPRey, line 239: For my own understanding, why can't the port be 50053 for instance?
I know 50051 is for gRPC but what about the following.

Alberto-DM: That port range is used by AEDT when it starts from Windows Start Menu. It starts using port 50051 and increases the port number for every new additional session started. When launched by PyAEDT, we wanted to avoid any potential conflict in case PyAEDT would select a port in that range (due to a race condition, because, if the port is already occupied, it is excluded). Actually, there is no predefined upper limit, but we felt that it is extremely unlikely for a user to have more that 20 concurrent sessions started manually.

MaxJPRey: Make complete sense. Thanks @Alberto-DM !
I agree, I don't anticipate any user to have more than 20 sessions started manually.

PR #5006, opened 2024-08-07
FIX: Using excitation type "winding" instead of "winding group"

IreneWoyna:


IreneWoyna:
@gmalinve Maxwell throws an error and doesn't create a circuit already when I'm calling the method: 
wdg_groups = self.excitations_by_type["Winding Group"]
(also when I'm actually not using wdg_groups)

That's why I commented it in the code.

pyaedt/maxwell.py
```diff
@@ -1982,7 +1982,8 @@ def create_external_circuit(self, circuit_design=None):
 
         circuit = MaxwellCircuit(design=circuit_design)
 
-        wdgs = self.excitations_by_type["Winding Group"]
+        wdgs = self.excitations_by_type["Winding"]
```
gmalinve, line 1985: @IreneWoyna you should check both cases :)

IreneWoyna: see my comment below, Maxwell throws an error already when I try to check the "winding group" and doesn't create the circuit...

gmalinve: @IreneWoyna yes but you should check that "Winding" or "Winding Group" are keys of the excitations_by_type dictionary. If yes you get you excitations as: self.excitation_by_type[excitation_type].

IreneWoyna: @gmalinve now it should work

Samuelopez-ansys (dismissed):

gmalinve (approved):

Samuelopez-ansys (approved):

PR #5008, opened 2024-08-07
FIX: Added separation of faces and objects in assign_voltage method

AlfVII:
I was implementing Electrostatic solver for Ansyas, and I got an error when assigning voltages to a list of more than one objects. I looked the code a bit and I saw (in maxwell.py, lines 980-986) that you distinguish between list of one element and more than one.
When there is only one you check if the elements are objects or faces and assign in correspondence. When there are more than one you always assign them as "Faces", which breaks the code when working with objects.


AlfVII:
@ansys-reviewer-bot review

Samuelopez-ansys:
@AlfVII I cherry-picked your PR here: https://github.com/ansys/pyaedt/pull/5010

I will close this PR. Thank you for your contribution. 

We need to cherry-pick to trigger the unit test (it uses Ansys internal licenses)

PR #5011, opened 2024-08-07
CHORE: Update pre-commit to v0.4.3

SMoraisAnsys:
The new pre-commit release has [improvement](https://github.com/ansys/pre-commit-hooks/pull/217) on the licenser-header hook which should GREATLY reduce the overhead associated to that hook when doing a commit.

@dcrawforAtAnsys This is not the enhancement I mentioned previously as we were waiting for https://github.com/fsfe/reuse-tool/pull/1055 to be merged. However, this is a complementary approach that will save us a lot of time !

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -396,6 +396,7 @@ model.index\+
 # local environment settings used by e.g. Visual Studio Code
 /.env
 /doc/source/local_config.json
+/.venv*
 
 # test coverage output
 /.cov/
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -396,6 +396,7 @@ model.index\+
 # local environment settings used by e.g. Visual Studio Code
 /.env
 /doc/source/local_config.json
+/.venv*
```
SMoraisAnsys, line 399: I'm adding this as .venv* is a common name pattern used to create local virtual environment.

Samuelopez-ansys (approved):

MaxJPRey (approved):

PR #5014, opened 2024-08-07
FIX: Use active design even if more than one of that design_type exists

A-Phamfam:
When making a new class instance such as Hfss() or any other design_type, the intended behavior is to get the active design, but if more than one design_types exist, it would always error with "No consistent unique design is present. Inserting a new design."

```
design: str, optional
Name of the design to select. The default is None, in which case an attempt is made to get an active design. If no designs are present, an empty design is created.
```

This is because it is simply checking if only one instance of that design_type exists in the project rather than actually getting the current active design.

### DIFF
--- a/pyaedt/application/Design.py
+++ b/pyaedt/application/Design.py
@@ -1037,39 +1037,8 @@ class Design(AedtObjects):
         warning_msg = ""
         names = self.get_oo_name(self.oproject)
         if names:
-            valids = []
-            for name in names:
-                des = self.get_oo_object(self.oproject, name)
-                des_type = None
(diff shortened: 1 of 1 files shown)


Samuelopez-ansys:
Hi @A-Phamfam Thanks for your PR. 

I think we need more checks in order to perform this refactoring. First, some users prefer creating a new design when the "design" parameter is not passed in the pyaedt.Hfss() call.

We will consider this idea, but for now, we can not merge it.

You can instead use pyaedt.Desktop(), and then find which are the projects and designs opened on this Electronics Desktop and then pick the one you want to connect.

PR #5019, opened 2024-08-08
FIX: Changed the default global log filename to use a unique id

Alberto-DM:
Changed the default global log filename to use a unique id to avoid conflicts in concurrent sessions.

### DIFF
--- a/pyaedt/generic/settings.py
+++ b/pyaedt/generic/settings.py
@@ -29,7 +29,8 @@ import uuid
 is_linux = os.name == "posix"
 
 
-def default_log_filename():
+def generate_log_filename():
+    """Generate a log filename."""
     base = "pyaedt"
     username = os.path.split(os.path.expanduser("~"))[-1]
     unique_id = uuid.uuid4()
--- a/pyaedt/generic/settings.py
+++ b/pyaedt/generic/settings.py
@@ -33,7 +33,7 @@ def default_log_filename():
     base = "pyaedt"
     username = os.path.split(os.path.expanduser("~"))[-1]
     unique_id = uuid.uuid4()
-    return f"{base}_{username}_{unique_id}.log"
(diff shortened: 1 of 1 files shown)


Samuelopez-ansys (approved):
Good first fix

pyaedt/generic/settings.py
```diff
@@ -24,10 +24,18 @@
 
 import os
 import time
+import uuid
 
 is_linux = os.name == "posix"
 
 
+def default_log_filename():
+    base = "pyaedt"
```
SMoraisAnsys, line 33:
```suggestion
def generate_log_filename():
    """Generate a log filename."""
    base = "pyaedt"
```

pyaedt/generic/settings.py
```diff
@@ -64,7 +72,7 @@ def __init__(self):
         self._force_error_on_missing_project = False
         self._enable_pandas_output = False
         self.time_tick = time.time()
-        self._global_log_file_name = "pyaedt_{}.log".format(os.path.split(os.path.expanduser("~"))[-1])
+        self._global_log_file_name = default_log_filename()
```
SMoraisAnsys, line 75:
```suggestion
        self._global_log_file_name = generate_log_filename()
```

SMoraisAnsys (dismissed):
LGTM, I left a minor comment about naming. This should avoid our tests to write in the same log file most of the time !

Alberto-DM:
> This should avoid our tests to write in the same log file most of the time !

"[...] In other words, only after generating 1 billion UUIDs every second for approximately 100 years would the probability of creating a single duplicate reach 50%"
I would accept the risk for our tests 😄

SMoraisAnsys (approved):

Samuelopez-ansys (approved):

PR #5036, opened 2024-08-12
FEAT: create equation based surface

rragona:
I am adding the function `create_equationbased_surface`. This feature was missing in `Primitives3D.py`.
The function definition and the unittest are inspired by the similar function `create_equationbased_curve`.


Samuelopez-ansys:
New PR created #5038

PR #5037, opened 2024-08-12
FEAT: Add a get_working_coordinate_system() function

A-Phamfam:
There exists a set_working_coordinate_system() function so there should be a corresponding get_working_coordinate_system() function.

The unit tests have also been updated so that a get check is done after each set.


SMoraisAnsys:
@A-Phamfam Hey and thanks for your contribution. You code has been cherry-picked into another branch and PR (see #5039) because tests are not passing if the PR has not the correct right access (being performed from a fork, this is the case).
Note that cherry-picking the commits keeps you as the original author of the commits !

PR #5041, opened 2024-08-13
FEAT: add mesh skin depth for maxwell 2d 

gmalinve:


Samuelopez-ansys:
@gmalinve One question, with this implementation is also possible to retrieve the skin depth from an existing design? I mean, this PR does the SET of this property, but I am wondering if the get is also implemented.

gmalinve:
@Samuelopez-ansys yes, it does :) I just checked

pyaedt/modules/Mesh.py
```diff
@@ -1101,35 +1101,44 @@ def assign_skin_depth(
...
             name = generate_unique_name("SkinDepth")
 
-        if maximum_elements is None:
-            restrictlength = False
-            maximum_elements = "1000"
+        if self._app.design_type == "Maxwell 2D":
+            seltype = "Edges"
+            props = OrderedDict(
+                {
+                    seltype: assignment,
```
SMoraisAnsys, line 1108:
```suggestion
            props = OrderedDict(
                {
                    "Edges": assignment,
```
There is no real need to create `seltype` in this scope as it is never used later on. Once leaving it, only `name` and `props` are used. Keeping such variable could make someone think that an error was introduced in this refactoring as it gives the impression that `seltype` should be used after the `if` ... `else` ... statements.

SMoraisAnsys (approved):
LGTM, I left a minor comment.

pyaedt/modules/Mesh.py
gmalinve: @SMoraisAnsys you are right! an oversight ;)

SMoraisAnsys (approved):

src/ansys/aedt/core/modules/mesh.py
```diff
@@ -1101,35 +1101,38 @@ def assign_skin_depth(
...
-            seltype = "Objects"
-        else:
-            seltype = None
-        if seltype is None:
-            self.logger.error("Error in Assignment")
-            return
+            if isinstance(assignment[0], int):
+                seltype = "Faces"
+            elif isinstance(assignment[0], str):
+                seltype = "Objects"
```
SMoraisAnsys, line 1122: Following the modified lines of code, shouldn't we log an error here is the assigment[0] is not int or str ?
Otherwise, the error that the user will see might be confusing (even if we asked for a list of string or int in the method description).

gmalinve: @SMoraisAnsys I removed it because assignment will always be either an int or a string. If you look previous lines, assignment is computed here:
`assignment = self.modeler.convert_to_selections(assignment, True)`
which returns always an int or a string. It's not the assignment passed by the user.
Hope this makes sense :)

SMoraisAnsys (approved):

PR #5043, opened 2024-08-13
Reduce run-time

lorenzovecchietti:


examples/04-Icepak/Sherlock_Example.py
```diff
@@ -92,49 +93,35 @@
...
-# ~~~~~~~~~~
-# Plot the model.
-
-ipk.plot(show=False, output_file=os.path.join(project_folder, "Sherlock_Example.jpg"), plot_air_objects=False)
-
-###############################################################################
-# Delete PCB objects
-# ~~~~~~~~~~~~~~~~~~
-# Delete the PCB objects.
-
```
SMoraisAnsys, line 122: Shouldn't you keep that section description ?

SMoraisAnsys (commented):
Could you let us know what was the reason for the increase in time ?
Also can you perform the same changes in the pyaedt-examples repo ?

Samuelopez-ansys:
@lorenzovecchietti In the post processing section, is not possible to use new_report.plot() ? 

This will reduce the size of the example.

lorenzovecchietti:
> Could you let us know what was the reason for the increase in time ?

Mostly thanks to a change in the data used for the example.

Then I noticed some error/thing that could be improved, so here the PR

> Also can you perform the same changes in the pyaedt-examples repo ?

I already worked extensively on the example on the pyaedt-examples repo months ago (and the updated examples are only there). Given that the speed-up is due to changes in the example data repo, pyaedt-examples repo should get the benefits as well.
As soon as the repo will be published, I'll put in more work to check any difference that should be moved there (but again examples in pyaedt-examples repo should be mostly ahead and more complete)

Samuelopez-ansys (approved):

PR #5051, opened 2024-08-14
FEAT: Plane wave

Samuelopez-ansys:


src/ansys/aedt/core/hfss.py
```diff
@@ -6308,6 +6308,183 @@ def wave_port(
...
+            ``["Ephi", "Etheta"]``.
+            The default is ``"Vertical"``.
+        propagation_vector : list, optional
+            Propagation vector.
+             If ``vector_format`` is ``"Cartesian"`` the type must be a ``list`` with three elements.
+             If `vector_format`` is ``"Spherical"`` the type must be a ``list`` with two elements, the first element
+             corresponds to the phi sweep, which must be ``list`` of three elements, start, stop, and number of points.
+             The second element has the same format, it corresponds to the theta sweep.
+             The default is ``[0.0, 0.0, 1.0]`` for ``"Cartesian"`` and
+             ``[["0.0deg", "0.0deg", 1], ["0.0deg", "0.0deg", 1]]`` for ``"Spherical"``.
```
SMoraisAnsys, line 6345:
```suggestion
            Propagation vector.
            If ``vector_format`` is ``"Cartesian"`` the type must be a ``list`` with three elements.
            If ``vector_format`` is ``"Spherical"`` the type must be a ``list`` with two elements. The first element
            corresponds to the phi sweep, which must be a ``list`` of three elements: start, stop, and number of points.
            The second element has the same format, it corresponds to the theta sweep.
            The default is ``[0.0, 0.0, 1.0]`` for ``"Cartesian"`` and
            ``[["0.0deg", "0.0deg", 1], ["0.0deg", "0.0deg", 1]]`` for ``"Spherical"``.
```

src/ansys/aedt/core/hfss.py
```diff
@@ -6308,6 +6308,183 @@ def wave_port(
...
+        wave_type : str, optional
+            Type of plane wave. Options are ``"Propagating"``, ``"Evanescent"``,  or ``"Elliptical"``.
+            The default is ``"Propagating"``.
+        wave_type_properties : list, optional
+            Properties of the plane wave type.
+            Not additional properties are needed if ``"Propagating"`` selected. The default is ``None``.
+            Propagation constant if ``"Evanescent"`` selected, real and imaginary part.
+            The default is ``[0.0, 1.0]``.
+            Polarization angle and ratio if ``"Elliptical"`` selected. The default is ``["0deg", 1.0]``.
```
SMoraisAnsys, line 6354:
```suggestion
            If ``"Propagating"`` is used, no additional properties are needed. The default is ``None``.
            If ``"Evanescent"`` is selected, the propagation constant is expressed as both real and
            imaginary components. The default is ``[0.0, 1.0]``.
            If ``"Elliptical"`` is used, the polarization angle and ratio are defined. The default is
            ``["0deg", 1.0]``.
```

src/ansys/aedt/core/hfss.py
```diff
@@ -6308,6 +6308,183 @@ def wave_port(
...
+                "IsCartesian": True,
+                "EoX": polarization[0],
+                "EoY": polarization[1],
+                "EoZ": polarization[2],
+                "kX": propagation_vector[0],
+                "kY": propagation_vector[1],
+                "kZ": propagation_vector[2],
+                "OriginX": x_origin,
+                "OriginY": y_origin,
+                "OriginZ": z_origin,
```
SMoraisAnsys, line 6448: I would create `inc_wave_args` before the `if` statement to set `"OriginX"`, `"OriginY"` and `"OriginZ"` which are shared for cartesian format and others.

src/ansys/aedt/core/hfss.py
```diff
@@ -6308,6 +6308,183 @@ def wave_port(
...
+                "PhiStop": propagation_vector[0][1],
+                "PhiPoints": propagation_vector[0][2],
+                "ThetaStart": propagation_vector[1][0],
+                "ThetaStop": propagation_vector[1][1],
+                "ThetaPoints": propagation_vector[1][2],
+                "EoPhi": polarization[0],
+                "EoTheta": polarization[1],
+                "OriginX": x_origin,
+                "OriginY": y_origin,
+                "OriginZ": z_origin,
```
SMoraisAnsys, line 6481: Cf my previous comment.

SMoraisAnsys (commented):
I would recommend updating the error message to follow the pattern "Invalid ..." and not "... invalid". Note that it is just a better (personal) feeling at reading!

SMoraisAnsys (approved):

PR #5056, opened 2024-08-16
FIX: gracefully handle exited processes

### LINKED ISSUE #5055: active_sessions issues spurious errors when processes exit while it is iterating over the machine's process list

### Description of the bug

It is possible that a process exits between the times that psutil.process_iter returns a Process object and we subsequently check the name on a following line in:

https://github.com/ansys/pyaedt/blob/ca17b806fffe6861974f23e68398144f5e12d482/pyaedt/generic/general_methods.py#L1578

In this case a psutil.NoSuchProcess exception is raised, which we should simply ignore: if a process no longer exists it cannot be an active session of AEDT.

### Steps To Reproduce

Launch many desktops using the below script on Windows (here I set processes=56 to run two AEDT instances per core to increase the probability of occurrence) using this command:

python test_many_desktops.py > test_many_desktops.log

Observe instances of "An error occurred while retrieving information for the active AEDT sessions" in test_many_desktops.log.  You may need to try more than once as this problem is intermittent.

```
import logging
import multiprocessing as mp
import os
import sys

from pyaedt import Desktop, Hfss3dLayout, settings as _pyaedtsettings

# _pyaedtsettings.use_multi_desktop = True


class AssertingHandler(logging.Handler):
    def emit(self, record: logging.LogRecord) -> None:
        if record.levelno >= logging.ERROR:
            print(f"ASSERT: {record.message}")
            print(record.exc_info)


version = "2024.2"


def the_work(id: int):
    logging.basicConfig(level=logging.DEBUG, stream=sys.stdout)
    logger = logging.getLogger()
    try:
        logger.addHandler(AssertingHandler())
        logger.info(f"Launching worker {id}")
        with Desktop(version=version, new_desktop=True, non_graphical=True, close_on_exit=True) as desktop:
            logger.debug(f"Worker {id} with PID {os.getpid()} has desktop with PID {desktop.aedt_process_id}")
    finally:
        pass


if __name__ == "__main__":
    processes = 56
    with mp.Pool(processes) as p:
        multiple_results = [p.apply_async(the_work, (i,)) for i in range(processes)]
        [r.wait(300) for r in multiple_results]
```

Windows

3.10

### Installed packages

```
annotated-types==0.7.0
ansys-pythonnet==3.1.0rc3
attrs==23.2.0
cffi==1.16.0
clr-loader==0.2.6
defusedxml==0.7.1
fonttools==4.53.0
fpdf2==2.7.9
jsonschema==4.22.0
numpy==1.26.4
pillow==10.3.0
plumbum==1.8.3
psutil==5.9.8
-e git+https://github.com/ansys/pyaedt@ca17b806fffe6861974f23e68398144f5e12d482#egg=pyaedt
pycparser==2.22
pydantic==2.7.4
pydantic_core==2.18.4
pyedb==0.14.1
pytomlpp==1.0.13
pywin32==306
referencing==0.35.1
rpds-py==0.18.1
rpyc==6.0.0
Rtree==1.2.0
toml==0.10.2
typing_extensions==4.12.2
```

isaacwaldron:
Ignore psutil.NoSuchProcess exceptions in
`pyaedt.generic.general_methods.active_sessions` to avoid issuing a spurious error when a process exits during the loop body.  Also improve reporting when another type of exception does occur to allow for easier debugging.

Fixes #5055

### DIFF
--- a/src/ansys/aedt/core/generic/general_methods.py
+++ b/src/ansys/aedt/core/generic/general_methods.py
@@ -1592,9 +1592,10 @@ def active_sessions(version=None, student_version=False, non_graphical=False):
                                 if i.pid == p.pid and (i.laddr.port > 50050 and i.laddr.port < 50200):
                                     return_dict[p.pid] = i.laddr.port
                                     break
-        except psutil.NoSuchProcess:  # pragma: no cover
-            # the process exited so it cannot be an active session
-            pass
+        except psutil.NoSuchProcess as e:  # pragma: no cover
+            pyaedt_logger.debug(
+                f"The process exited and cannot be an active session: {e}"
+            )
         except Exception as e:  # pragma: no cover
             pyaedt_logger.error(
                 f"A(n) {type(e)} error occurred while retrieving information for the active AEDT sessions: {e}"
--- a/src/ansys/aedt/core/generic/general_methods.py
+++ b/src/ansys/aedt/core/generic/general_methods.py
@@ -1592,8 +1592,13 @@ def active_sessions(version=None, student_version=False, non_graphical=False):
                                 if i.pid == p.pid and (i.laddr.port > 50050 and i.laddr.port < 50200):
                                     return_dict[p.pid] = i.laddr.port
                                     break
-        except Exception:
-            pyaedt_logger.error("An error occurred while retrieving information for the active AEDT sessions")
+        except psutil.NoSuchProcess:
+            # the process exited so it cannot be an active session
+            pass  # pragma: no cover
+        except Exception as e:
+            pyaedt_logger.error(
+                f"A(n) {type(e)} error occurred while retrieving information for the active AEDT sessions: {e}"
+            )  # pragma: no cover
     return return_dict
 
 
--- a/src/ansys/aedt/core/generic/general_methods.py
+++ b/src/ansys/aedt/core/generic/general_methods.py
@@ -1592,13 +1592,13 @@ def active_sessions(version=None, student_version=False, non_graphical=False):
                                 if i.pid == p.pid and (i.laddr.port > 50050 and i.laddr.port < 50200):
                                     return_dict[p.pid] = i.laddr.port
                                     break
-        except psutil.NoSuchProcess:
+        except psutil.NoSuchProcess:  # pragma: no cover
             # the process exited so it cannot be an active session
-            pass  # pragma: no cover
-        except Exception as e:
+            pass
+        except Exception as e:  # pragma: no cover
             pyaedt_logger.error(
                 f"A(n) {type(e)} error occurred while retrieving information for the active AEDT sessions: {e}"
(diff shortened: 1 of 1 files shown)


isaacwaldron:
To add a bit of reasoning for this fix, this issue was causing my downstream test suites to fail when they were checking that no errors were logged during certain tests.

SMoraisAnsys:
@isaacwaldron Sry but we just did a huge refactoring of the code base, could you update your changes in src/ansys/aedt/core/... ?

isaacwaldron:
@SMoraisAnsys I think a simple rebase handled the file move/rename for me, amazing!  Latest version of that file looks like it has my changes.

src/ansys/aedt/core/generic/general_methods.py
```diff
@@ -1592,8 +1592,13 @@ def active_sessions(version=None, student_version=False, non_graphical=False):
                                 if i.pid == p.pid and (i.laddr.port > 50050 and i.laddr.port < 50200):
                                     return_dict[p.pid] = i.laddr.port
                                     break
-        except Exception:
-            pyaedt_logger.error("An error occurred while retrieving information for the active AEDT sessions")
+        except psutil.NoSuchProcess:  # pragma: no cover
+            # the process exited so it cannot be an active session
+            pass
```
SMoraisAnsys, line 1597:
```suggestion
        except psutil.NoSuchProcess as e:  # pragma: no cover
            pyaedt_logger.error(
                f"The process exited and cannot be an active session: {e}"
            )
```

SMoraisAnsys (changes requested):
Nice handling ! This might help us with some failing tests 🤞 

I added a minor comment as we try to avoid try ... except pass pattern (considered a bad pattern), it's better to log something.

src/ansys/aedt/core/generic/general_methods.py
isaacwaldron: @SMoraisAnsys could we instead issue a debug message here? This is not an error in my estimation: we are iterating over all the processes on the machine and one happened to exit between the time `psutil.process_iter()` returned it and (I suspect) we checked the name (or even more rarely, an `ansysedt` process happened to exit after we check the name but later in the loop).

SMoraisAnsys: Fine by me !

SMoraisAnsys:
Btw, latest merge in main solved the documentation style and build failures. Feel free to force push again if u want

src/ansys/aedt/core/generic/general_methods.py
SMoraisAnsys: @isaacwaldron I updated the changes following the previous comment

Samuelopez-ansys (dismissed):

SMoraisAnsys (approved):

PR #5069, opened 2024-08-20
FIX: PyEDB Path class name refactor

Samuelopez-ansys:


src/ansys/aedt/core/modules/solve_setup.py
```diff
@@ -2031,7 +2031,7 @@ def _get_primitives_points_per_net(self):
                 if prim.layer_name not in layers_elevation:
                     continue
                 z = layers_elevation[prim.layer_name]
-                if "EdbPath" in str(prim):
+                if prim.__class__.__name__ in ["EdbPath", "Path"]:
```
SMoraisAnsys, line 2034: While I don't like testing the name of the class, I don't see another way to make things work unless we specify that `pyedb` version should be 0.24+ (where 0.24.0 is the version where `Path` appeared).

SMoraisAnsys (approved):

ring630 (approved):
LGTM.

Issue #5023, opened 2024-08-08
Rework or clarify the way we use settings


SMoraisAnsys:
There seem to be some misunderstanding with how the code base works and how we want the final result to be.

This observation comes from @Alberto-DM who noticed that even when testing and setting the `pyaedt.settings` to not log into a file, we end up with a file being written. This comes from how python works. For example, in the following code extracted from our conftest.py file

```python
from pyaedt.generic.settings import settings

settings.enable_local_log_file = False
settings.enable_global_log_file = False
settings.number_of_grpc_api_retries = 6
settings.retry_n_times_time_interval = 0.5
settings.enable_error_handler = False
settings.enable_desktop_logs = False
settings.desktop_launch_timeout = 180
settings.release_on_exception = False
```

To import `settings` we are loading `pyaedt.__init__.py`, then `pyaedt.generic.__init__.py` and then `pyaedt.generic.settings.py`.
However, doing so leads instantiating `settings = Settings()` since other modules loaded through `pyaedt.__init__.py` do so.
Since the default value of `enable_global_log_file` is `True` then, if for whatever reason one of the file triggers some logging - like logging that one has not `pyvista`, `numpy`, or whateer - we'll have a file with the logged content, while the user might think that by setting `settings.enable_global_log_file = False` it prevented it.

I'm not sure of the correct approach to handle this as making changes might break users code or might require deep changes.

I'll investigate on the coming days and let you know what approach I can think about. Of course, any feedback is much appreciated !

Samuelopez-ansys:
@SMoraisAnsys Could we close this issue right?

SMoraisAnsys:
Closed by #5092

PR #5092, opened 2024-08-26
REFACTOR: Settings from YAML configuration file

### LINKED ISSUE #5083: Defining the folder where the batch.log file is created
### Discussed in https://github.com/ansys/pyaedt/discussions/5060

<div type='discussions-op-text'>

<sup>Originally posted by **shizzlemaster** August 19, 2024</sup>
Hello community,

I am using pyaedt to write a simulation script that is called by another application and I have the following issue:

The batch.log is created in the folder of my external application e.g. .../my_application/

This is not wanted in my case and I want to specify the directory where the batch.log is created but I do not know how.

Can anyone help me with this issue?

Thanks in advance

Marcus</div>

SMoraisAnsys:
As stated in #5023, we currently have a problem as the settings are load when someone does `import pyaedt` and the default behavior can be the one being used even after a user stated that he wanted another behavior. This comes from how Python import works and the fact that we create a static `settings` leveraged in multiple files.

To avoid this problem, this PR consists in refactoring the way we are currently working with settings by adding a mechanism to load a YAML configuration file which modifies the settings. Also, the imports in `ansys.aedt.core.__init__` are modified to ensure that `settings` is the first thing imported before importing other PyAEDT modules.
The load configuration file method is triggered at `settings` creation and it tries to read:
- the content of a YAML file provided through the environment variable `PYAEDT_LOCAL_SETTINGS_PATH`;
- OR the content of a YAML file `pyaedt_settings.yaml` in the user's `APPDATA` folder for Windows and
`HOME` folder for Linux.

Unlike other PyAEDT piece of code, we leverage `pyyaml` and not `pytomlpp` as TOML file expect a value for every key. However, we have settings with default value being `None` and this is incompatible with TOML.

Also, some inner settings were accessible to users. I added an extra class to encapsulate such settings.

Closes #5083, #5023 
Related #5060


src/ansys/aedt/core/__init__.py
```diff
@@ -67,7 +67,13 @@ def custom_show_warning(message, category, filename, lineno, file=None, line=Non
 __version__ = "0.11.dev0"
 version = __version__
 
-#
+# isort: off
+# Settings have to be imported before importing other Pyaedt modules
```
MaxJPRey, line 71:
```suggestion
# Settings have to be imported before importing other PyAEDT modules
```

doc/source/Resources/pyaedt_settings.yaml
```diff
@@ -0,0 +1,117 @@
...
+  # RAM allocated for the LSF job
+  lsf_ram: 1000
+  # Timeout in seconds for trying to start the interactive session
+  lsf_timeout: 3600
+  # Value passed in the LSF 'select' string to the ui resource
+  lsf_ui: null
+  # Enable or disable use LSF Scheduler
+  use_lsf_scheduler: false
+
+# Settings related to environment variables thare are set before launching a new AEDT session
```
MaxJPRey, line 63:
```suggestion
# Settings related to environment variables are set before launching a new AEDT session
```

doc/source/User_guide/settings.rst
```diff
@@ -0,0 +1,141 @@
+Settings
+========
+
+The Settings class is designed to handle the configurations of PyAEDT and AEDT.
+This includes behavior for logging, LSF scheduler, environment variable and general
+settings. Most of the default values used can be modified using a YAML configuration file.
+The path to this YAML file should be defined through the environment variable
+``PYAEDT_LOCAL_SETTINGS_PATH``. If the environment variable is set and a file exists,
+the default configuration settings are updated according to the content of the file.
+If the environment variable is not defined, a check if performed to see if a file named
```
MaxJPRey, line 10:
```suggestion
If the environment variable is not defined, a check is performed to see if a file named
```

doc/source/User_guide/settings.rst
```diff
@@ -0,0 +1,141 @@
...
+``"pyaedt_settings.yaml"`` exist in the user's ``APPDATA`` folder for Windows and
+``HOME`` folder for Linux. If such file exists, it is then used to update the default
+configuration.
+
+Bellow  is the content that can be updated through the YAML file.
```
MaxJPRey, line 15:
```suggestion
Below  is the content that can be updated through the YAML file. 
```
I will let have a look at bellow verb definition. Most likely not what you meant 😃

doc/source/User_guide/settings.rst
```diff
@@ -0,0 +1,141 @@
...
+``HOME`` folder for Linux. If such file exists, it is then used to update the default
+configuration.
+
+Bellow  is the content that can be updated through the YAML file. 
+
+:download:`YAML configuration file <../Resources/pyaedt_settings.yaml>`
+
+.. note::
+    Not all settings from class ``Settings`` can be modified through this file
+    as some of them expect python objects or values obtained from code execution.
```
MaxJPRey, line 21:
```suggestion
    as some of them expect Python objects or values obtained from code execution.
```

doc/source/User_guide/settings.rst
```diff
@@ -0,0 +1,141 @@
...
+        # RAM allocated for the LSF job
+        lsf_ram: 1000
+        # Timeout in seconds for trying to start the interactive session
+        lsf_timeout: 3600
+        # Value passed in the LSF 'select' string to the ui resource
+        lsf_ui: null
+        # Enable or disable use LSF Scheduler
+        use_lsf_scheduler: false
+
+    # Settings related to environment variables thare are set before launching a new AEDT session
```
MaxJPRey, line 87:
```suggestion
    # Settings related to environment variables are set before launching a new AEDT session
```

MaxJPRey (approved):
Nice work

doc/source/Resources/pyaedt_settings.yaml
```diff
@@ -0,0 +1,117 @@
+# This file contains the settings used by default to set the PyAEDT environment variables and
+# global settings. If you want to have a different behavior you have to copy / paste this file
+# into your PersonalLib directory and restart your workflow.
```
Samuelopez-ansys, line 3: @SMoraisAnsys Could be in any location right? PersonalLib is only one option right?

SMoraisAnsys: Yeah, I'll update the doc

doc/source/Resources/pyaedt_settings.yaml
SMoraisAnsys: I wanted to specify that the following settings are needed before launching AEDT. I can switch from `"that"` to `"which"` but I feel like you miss a part from the message by removing any of them.

Samuelopez-ansys (approved):

PR #5094, opened 2024-08-27
FEAT: Circuit Netlist class

Samuelopez-ansys:
Circuit Netlist class was not part of PyAEDT. The API is limited because it only allows analyzing and performing postprocessing. And it is not working in non graphical mode.


maxcapodi78 (dismissed):

_unittest/conftest.py
```diff
@@ -197,13 +197,21 @@ def _method(
...
+                version=desktop_version,
+                non_graphical=NONGRAPHICAL,
+            )
+        else:
+            return application(
+                project=test_project,
+                design=design_name,
+                version=desktop_version,
+                non_graphical=NONGRAPHICAL,
+            )
```
SMoraisAnsys, line 214:
```suggestion
        args = {
            project=test_project,
            design=design_name,
            version=desktop_version,
            non_graphical=NONGRAPHICAL,
        }
        if solution_type:
            args["solution_type"]=solution_type
        return application(**args)
```
Could we refactor into something like that ?

SMoraisAnsys (approved):
LGTM, I left a minor comment

SMoraisAnsys (approved):

PR #5095, opened 2024-08-27
CHORE: Add git blame ignore revs file

SMoraisAnsys:
This PR consist in adding a `.git-blame-ignore-revs` file. The interest of this file is that it allows to ignore specific revisions. For example, I'm discarding a few old commits that were mainly focused on formatting such that when you try to see what was the last change on a line of code, you have more chance to retrieve data associated to the commit which added the line of code (or modified the logic in it) and not the last format commit which updated the line.

The file name is important as it is the standard one and it can be leveraged by your IDE and Github. For example, with Visual Studio Code, I didn't have to do any configuration to make that work correctly (you just need to restart you IDE though).
Also it can be leveraged through Github in a Web browser ! See [Viewing a file - Github docs](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view)

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,36 @@
+# One commit hash per line, and all commits in the file will be ignored by git blame.
+
+# Apply flake8 policy for unused imports (#449)
+f2ebb7a80978e1ea82328e9b6afd3f9484362552
+# Organize module imports (#448)
+98bb0a1bfd27f8174ef86a131b3f6eed09f562ad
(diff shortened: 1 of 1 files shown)


Samuelopez-ansys (approved):

MaxJPRey:
This would have saved me so much time @SMoraisAnsys back in the days.
Thanks.

MaxJPRey (approved):

PR #5104, opened 2024-08-29
FEAT: Added support to asc files

maxcapodi78:
Added converter from asc file to circuit schematic
Added extension to AEDT to import different schematic format supported by PyAEDT


Samuelopez-ansys (dismissed):

SMoraisAnsys (dismissed):
This should fix the CICD.

Samuelopez-ansys (approved):

PR #5122, opened 2024-09-03
FEAT: Arbitrary wave port extension

svandenb-dev:
This extension provides a simple UI to automate arbitrary wave port generation.


Samuelopez-ansys:
@svandenb-dev You need to add it to the extension_catalog.toml and add a logo for the extension :) Good job

Samuelopez-ansys (dismissed):

Samuelopez-ansys:
@svandenb-dev This extension does not follow the same structure as the others. I will refactor it, we can not accept all the backend parts inside the frontend call, because we should add  unit test. 

The front end should only take user inputs, and the main call, should perform the rest.

svandenb-dev (commented):
LGTM

Samuelopez-ansys (dismissed):

PR #5131, opened 2024-09-04
FIX: Allow scale_min and max to be set if not log scale

gmalinve:


src/ansys/aedt/core/modules/advanced_post_processing.py
```diff
@@ -447,7 +447,9 @@ def plot_field_from_fieldplot(
         if is_pcb:
             model.z_scale = 5
 
-        if scale_min and scale_max:
+        if log_scale and scale_min == 0:
+            self.logger.warning("Invalid scale minimum value for logarithm scale.")
+        elif not log_scale and scale_min and scale_min >= 0 and scale_max:
```
gmalinve, line 452: @Samuelopez-ansys @SMoraisAnsys I know you could argue something here but here we should cover the case when scale is not logarithmic.
If log_scale is not logarithmic scale_min could be None (default value) but if it's None the check `scale_min>=0` will fail so that's why I had to put `and scale_min` before. If you have better options I'm more than glad to listen :)

SMoraisAnsys: The `elif` condition will be a bit hard to read later on. However I got your point.
If I had to clarify this, I would add an extra logic to check if `scale_min` is not `None` while `scale_max` is `None` (and vice versa) to warn the user that both value should be provided together. Then I would ensure that `scale_min >= 0` and `scale_max > scale_min` if that makes sense.

SMoraisAnsys (approved):
LGTM. I left a comment though. Feel free to discard it if you want but it might help for readability and user understanding.

src/ansys/aedt/core/modules/advanced_post_processing.py
gmalinve: @SMoraisAnsys but if you implement a check like this one:
`  if scale_min and not scale_max or not scale_min and scale_max:`
      `self.logger.warning("Both scale_min and scale_max must be provided.")`
This will fail even if both scale_min and max are provided in the case scale_min = 0.
It's the same problem that I now have at line 452.

SMoraisAnsys: That's why I spoke about an extra check with `scale_min >= 0` and `scale_max > scale_min`. Would that make sense ?

SMoraisAnsys (changes requested):
There are holes in the current implementation. This should patch them if I'm correct.

SMoraisAnsys (dismissed):

SMoraisAnsys (approved):

PR #5146, opened 2024-09-09
DOCS: Improve advanced fields calc doc

gmalinve:


SMoraisAnsys (approved):
LGTM, great enhancement !

@Samuelopez-ansys @gmalinve @maxcapodi78 I've a question related to the changes and the Examples section. I know we are not leveraging `doctest` (yet or if we will ever do) but this is a tool that we could use at some point to ensure that our examples are running and make sense. Thus, would it make sense to provide a valid path instead of `my_toml = os.path.join("my_path_to_toml", "my_toml.toml")` ? For example, leveraging the embedded catalog.
I know this will complexify the mini example but wanted to bring that subject none the less :D

gmalinve:
@SMoraisAnsys @Samuelopez-ansys @maxcapodi78 I think it's a good idea using _doctest_ although I'm worried that this will entail the refactoring of many docstrings but it's worth it.
In the meanwhile we decide this matter I proceeded to close this PR :)

PR #5164, opened 2024-09-11
FEAT: ECAD import extension

lorenzovecchietti:


_unittest_solvers/example_models/T45/DevicePropertiesReduced.csv
siva-krishnaswamy, line 1: Hi @lorenzovecchietti , I was using the same thing in my toolkit and then I noticed issue from the exported csv file. there was a bug instead of ThetaJc(kel_per_w) you can notice ThetaJb(kel_per_w) twice. And also the exported csv has "Part " with additional space at the end when exported from the advanced device options

doc/source/_static/extensions/ecad_importer.png
siva-krishnaswamy, line 1: Hi @lorenzovecchietti , Instead of tk logo, we can use ansys logo. It is already there in workflow directory.

icon_path = os.path.join(pyaedt.workflows.__path__[0], "images", "large", "logo.png")

siva-krishnaswamy: Hi @lorenzovecchietti, Does the devices in CSV part of PCB component ? If not, Can we have a toggle button which will create the components outside the PCB 3D component ?

src/ansys/aedt/core/workflows/icepak/ecad_importer.py
siva-krishnaswamy, line 1: Can we add a method for changing color in pyaedt instead of native apis ?

src/ansys/aedt/core/workflows/icepak/images/large/ecad.png
siva-krishnaswamy, line 1: @lorenzovecchietti , 
Large icon works for this toolkit but I can notice that we have space for only one large icon in Icepak. I am not sure what will happen if we have multiple toolkits. Maybe this is something which we need to take care in future

src/ansys/aedt/core/workflows/icepak/ecad_importer.py
lorenzovecchietti: At the beginning, I was using the native object.color method. But with many objects it could become very slow.

src/ansys/aedt/core/workflows/icepak/images/large/ecad.png
lorenzovecchietti: I tried on my laptop and it seems that a button to scroll horizontally is appearing.

siva-krishnaswamy: Then it is great !

PR #5173, opened 2024-09-14
DOCS: Fix pyaedt logo in README.md

SMoraisAnsys:
As title says. Follows #5154 which was incomplete.


SMoraisAnsys:
Here is the resulting landing page 

image

Samuelopez-ansys (approved):

Issue #5100, opened 2024-08-28
Warning generated when setting the name of a 3D object in HFSS to the current name


adamgregory81:
### Before submitting the issue

- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### Description of the bug

This applies to the Object3d in ansys.aedt.core.modeler.cad.object_3d.py
If I set the name of the object to the current name of the object, the following warning is triggered:
"<NAME> is already used in current design"

It seems the problem is in the @name.setter function:
        if obj_name not in self._primitives.object_names:
            if obj_name != self._m_name:

Those 2 lines might work better if the order was swapped.

This isn't critical since it is only a warning, but it would be nice to remove the warning.

### Steps To Reproduce

create a 3d object in HFSS by creating a box or whatever:
my_3d_object = modeler.create_box(position, dimensions_list, name="Something")

Then set the name to the current name:
my_3d_object.name = "Something"

The warning will trigger because the new name value is checked to be a member of the current list of HFSS names before it is checked to the same name already set for the object.

### Which Operating System are you using?

Windows

### Which Python version are you using?

3.8

### Installed packages

aiohttp==3.8.3
aiosignal==1.2.0
altgraph==0.17.3
annotated-types==0.7.0
ansys-pythonnet==3.1.0rc3
appdirs==1.4.4
async-timeout==4.0.2
attrs==24.2.0
cachetools==5.3.0
cffi==1.17.0
charset-normalizer==2.1.1
clr-loader==0.2.6
contourpy==1.0.5
cycler==0.11.0
defusedxml==0.7.1
et-xmlfile==1.1.0
fonttools==4.37.3
fpdf2==2.7.9
frozenlist==1.3.1
idna==3.4
imageio==2.22.0
importlib_metadata==8.4.0
importlib_resources==6.4.4
jsonschema==4.23.0
jsonschema-specifications==2023.12.1
kiwisolver==1.4.4
matplotlib==3.6.0
multidict==6.0.2
numpy==1.23.3
openpyxl==3.0.10
packaging==24.1
pandas==1.5.1
pefile==2023.2.7
pillow==10.4.0
pkgutil_resolve_name==1.3.10
plumbum==1.8.3
psutil==6.0.0
py2exe==0.13.0.0
pyaedt==0.10.0
pycparser==2.22
pydantic==2.8.2
pydantic_core==2.20.1
pyedb==0.25.0
pyinstaller==6.10.0
[...]

PR #5174, opened 2024-09-16
FEAT: Create SBR array

Samuelopez-ansys:


src/ansys/aedt/core/modeler/cad/primitives_3d.py
```diff
@@ -1464,6 +1464,56 @@ def _create_reference_cs_from_3dcomp(self, assignment, password):
...
+        ipkapp_temp = Icepak(project=os.path.join(app.toolkit_directory, temp_proj_name))
+        ipkapp_temp.delete_design(ipkapp_temp.design_name)
+        app.oproject.CopyDesign(app.design_name)
+        ipkapp_temp.oproject.Paste()
+        temp_proj = ipkapp_temp.project_file
+        ipkapp_temp.close_project()
+        return temp_proj
+
+    @staticmethod
+    def __get_all_mapping(temp_proj):
```
SMoraisAnsys, line 1480: This method purpose seems a bit strange to me as it seems to return a mapping referencing itself `g[key_0][key_1] = g`.
Is this code valid ? If yes, what is its purpose ?

src/ansys/aedt/core/modules/boundary.py
```diff
@@ -245,6 +246,45 @@ def __init__(self, app, component_type, component_name, props):
...
+        str
+           Name of the object.
+
+        """
+        return self._name
+
+    @name.setter
+    def name(self, component_name):
+        if component_name not in self._app.native_component_names:
+            if component_name != self._name:
```
SMoraisAnsys, line 264:
```suggestion
        if component_name != self._name:
            if component_name not in self._app.native_component_names:
```
This makes me think about #5100 (which I should handle soon). I would recommend switching the order as mentioned by @adamgregory81.

src/ansys/aedt/core/hfss.py
```diff
@@ -1428,41 +1509,33 @@ def create_sbr_antenna(
...
-            parameters_defaults["Array Symmetry Type"] = 0
-            parameters_defaults["Array Weight Type"] = 3
-            parameters_defaults["Array Beam Angle Theta"] = "0deg"
-            parameters_defaults["Array Weight Edge TaperX"] = -200
-            parameters_defaults["Array Weight Edge TaperY"] = -200
-            parameters_defaults["Array Weight Cosine Exp"] = 1
-            parameters_defaults["Array Differential Pattern Type"] = 0
-            if is_array:
-                name = generate_unique_name("pAntArray")
+            parameters_defaults["Sheet"] = -1
```
SMoraisAnsys, line 1512: Just wondering why `-1` is used as default. Wouldn't it be better to set it to `None` ?

src/ansys/aedt/core/modeler/cad/primitives_3d.py
```diff
@@ -1464,6 +1464,56 @@ def _create_reference_cs_from_3dcomp(self, assignment, password):
             app.oproject.Close()
             return assignment.target_coordinate_system
 
+    @staticmethod
+    def __create_temp_project(app):
```
SMoraisAnsys, line 1468: 👍 for naming this static methods as "private". At first I though it was a mistake as the "usual" behavior of static method is to be public but here they are only used internally. Glad you did this the way you did it !

SMoraisAnsys (commented):
I left a couple of comments. Once answered, the changes seem fine to me.

src/ansys/aedt/core/hfss.py
Samuelopez-ansys: Because the AEDT API needs a -1, or at least this is what is recorded, and in the API documentation there is no more information.

src/ansys/aedt/core/modeler/cad/primitives_3d.py
Samuelopez-ansys: Indeed, we should start migrating some similar method to private, step by step :)

src/ansys/aedt/core/modeler/cad/primitives_3d.py
lorenzovecchietti: The method reads the aedt file and extract the relevant info. Also transform the extracted list in an easy-to-use dictionary `{obj_name: props}`.

SMoraisAnsys: Fine, thanks for the answer. It clarified the code content :)

SMoraisAnsys (approved):

SMoraisAnsys (approved):

PR #5183, opened 2024-09-19
FEAT: Close desktop at exit

SMoraisAnsys:
This PR adds the automatic desktop closure at python exit.

This should be useful for people not remembering about closing the desktop (all the more when using `grpc`) or when an exception is raised.


Samuelopez-ansys (approved):
LGTM

PR #5184, opened 2024-09-20
TESTING: pyedb main branch to fix 0.28.0

SMoraisAnsys:
> **Warning**: Dot not merge !

@ring630 Ping for visibility.


SMoraisAnsys:
@ring630 Seems like the current main branch solves the previous problem. However, we need to understand why this happened and if possible fix the problem instead of using the latest workaround added.

SMoraisAnsys:
Closing as this PR has fulfilled its purpose : checking pyaedt failure in linux and testing fixes from pyedb side.

PR #5190, opened 2024-09-20
FIX: pyedb dependency

hui-zhou-a:


SMoraisAnsys (changes requested):
It would be better to have a way to upgrade pyedb from pyaedt_installer_from_aedt (as well as toolkits). Could you handle that ?

ring630:
> It would be better to have a way to upgrade pyedb from pyaedt_installer_from_aedt (as well as toolkits). Could you handle that ?

If pyaedt_installer_from_aedt does pip install -U pyaedt

> It would be better to have a way to upgrade pyedb from pyaedt_installer_from_aedt (as well as toolkits). Could you handle that ?

Does the new commit do the job?

SMoraisAnsys:
For the current situation yes but the problem will happen again after a new release. What would be for the best is to have a way to directly indicate that one wants to upgrade to a particular version or the latest for a set of package that we control.

MaxJPRey:
> For the current situation yes but the problem will happen again after a new release. What would be for the best is to have a way to directly indicate that one wants to upgrade to a particular version or the latest for a set of package that we control.

Agreed, that would be better as a fix for long term.

pyproject.toml
```diff
@@ -136,6 +136,7 @@ installer = [
     "osmnx>=1.1.0,<1.10",
     "pandas>=1.1.0,<2.3",
     "pyvista[io]>=0.38.0,<0.45",
+    "pyedb>=0.27.0; python_version > '3.7'",
```
maxcapodi78, line 139: this may be not needed as pyedb is required already as base dependencies

pyproject.toml
```diff
@@ -37,7 +37,7 @@ dependencies = [
     "jsonschema",
     "psutil",
     "pyedb>=0.4.0; python_version == '3.7'",
-    "pyedb>=0.24.0; python_version > '3.7'",
+    "pyedb>=0.27.0; python_version > '3.7'",
     "pyedb!=0.28.0; python_version > '3.7'",
```
maxcapodi78, line 41: why we skip 0.28.0?

maxcapodi78 (changes requested):
@SMoraisAnsys please check my comments

pyproject.toml
hui-zhou-a: pyaedt doesn't pass unittest with pyedb 0.28.0.

pyproject.toml
SMoraisAnsys: @maxcapodi78 We can remove the python_version specification since we are now python3.8+ compatible (as `pyedb`).
@hui-zhou-a I agree that keeping `pyedb` is a required dependency for installer installation target. However, putting `>=0.27.0` is not enough to ensure that `pyedb` is updated when doing `pip install pyaedt[installer]` unless you are calling for an upgrade explicitly.

pyproject.toml
SMoraisAnsys: We had an issue with pyedb `0.28.0` which broke the tests in Linux. This has been a work in progress and should be fixed in `0.29.0`. It was working with pyedb `0.29.dev0` a few days ago but failed when we tried to release `0.29.0` ....

PR #5199, opened 2024-09-23
FIX: ACF file variable override bug

Samuelopez-ansys:


gmalinve (approved):

src/ansys/aedt/core/application/analysis.py
```diff
@@ -1801,17 +1801,17 @@ def analyze_setup(
...
+        if acf_file:  # pragma: no cover
             self._desktop.SetRegistryFromFile(acf_file)
-            name = ""
+            acf_name = ""
             with open_file(acf_file, "r") as f:
                 lines = f.readlines()
                 for line in lines:
                     if "ConfigName" in line:
-                        name = line.strip().split("=")[1]
+                        acf_name = line.strip().split("=")[1].strip("'")
```
SMoraisAnsys, line 1811: Could you add a comment / description to the PR to explain this change ?

SMoraisAnsys (approved):

src/ansys/aedt/core/application/analysis.py
Samuelopez-ansys: line.strip().split("=")[1] was returning a ¨double¨ string like:
" `hola` " , which was not accepted by the Desktop registry. So we remove it and then it works.

PR #5200, opened 2024-09-24
FIX: Swap object rename order

Samuelopez-ansys:


src/ansys/aedt/core/modeler/cad/object_3d.py
```diff
@@ -1096,15 +1096,11 @@ def name(self):
...
     @name.setter
     def name(self, obj_name):
-        if obj_name not in self._primitives.object_names:
-            if obj_name != self._m_name:
-                vName = []
-                vName.append("NAME:Name")
-                vName.append("Value:=")
-                vName.append(obj_name)
+        if obj_name != self._m_name:
+            if obj_name not in self._primitives.object_names:
```
Alberto-DM, line 1100:
```suggestion
        if obj_name != self._m_name and obj_name not in self._primitives.object_names:
```
Why not joining the two if?

Alberto-DM (dismissed):
Looks good to me, but it can be further simplified with a single `if` clause.

src/ansys/aedt/core/modeler/cad/object_3d.py
SMoraisAnsys: Agreeing with Alberto comment

SMoraisAnsys (dismissed):
LGTM, agreeing with Alberto though :)

Alberto-DM (approved):
LGTM

PR #5203, opened 2024-09-24
CI: Avoid using self-hosted runners in drafts

SMoraisAnsys:
Following a discussion with @Alberto-DM, this PR aims at reducing the use of our self hosted runners.

AFAIK, the best we can do with the current capabilities of Github is to avoid using our self hosted runners when one is working on a draft. This will allow other PRs, marked as ready to review, to use the self hosted runners instead of waiting for a draft PR finishing "wasting" resources. Since one could still want to trigger the jobs using our self hosted runners none the less, we should also allow people to run them by manually clicking on the run button.

### DIFF
--- a/.github/workflows/manual_draft.yml
+++ b/.github/workflows/manual_draft.yml
@@ -1,0 +1,16 @@
+name: Manual draft workflow
+
+on:
+  workflow_dispatch:
+
+jobs:
+
+  test-job-0:
+    if: github.event.pull_request.draft == true
+    runs-on: ubuntu-latest
+    steps:
+      - name: Tests step
+        run: |
+          echo "This job runs!"
+          echo "github.event_name: ${{ github.event_name}}"
+          echo "${{ toJson(github.event) }}"
--- a/.github/workflows/manual_draft.yml
+++ b/.github/workflows/manual_draft.yml
@@ -42,7 +42,7 @@ jobs:
 
   test-solvers-windows:
     name: Testing solvers and coverage (Windows)
-    if: github.event.pull_request.draft == true && github.event.inputs.test-solvers-windows == 'yes'
+    if: github.event.inputs.test-solvers-windows == 'yes'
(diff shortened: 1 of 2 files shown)


SMoraisAnsys:
Closing and reopening this PR as a draft to check if the changes work as expected.

Alberto-DM (dismissed):
Nicely done!

SMoraisAnsys:
@Alberto-DM This is not working atm as I can't run a job that is skipped by default, I'll check if it's related to the if statement.

SMoraisAnsys:
Here is the result I'm having on my fork

image

We should make this PR as ready for review after ensuring that everything is not running as expected :)

SMoraisAnsys:
In draft mode, some jobs are skipped as expected

image

Samuelopez-ansys (dismissed):
Great!

.github/workflows/manual_draft.yml
```diff
@@ -0,0 +1,288 @@
...
+        with:
+          name: pytest-results
+          path: junit/test-results.xml
+        if: ${{ always() }}
+
+# # =================================================================================================
+# # vvvvvvvvvvvvvvvvvvvvvvvvvvvvvv    RUNNING ON SELF-HOSTED RUNNER    vvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
+# # =================================================================================================
+
+  # TODO: Si if we can use ansys/actions
```
MaxJPRey, line 224:
```suggestion
  # TODO: See if we can use ansys/actions
```
? Not sure 😃

SMoraisAnsys: I'll remove that comment as this file should be modified accordingly to changes it ci_cd.yml file.
If you are willing to do this changes in the ci_cd.yml file, feel free to open a PR :D

Samuelopez-ansys (approved):

PR #5210, opened 2024-09-25
FEAT: Added method to plot nets in Hfss 3d Layout which uses Edb Plot

maxcapodi78:
Added new method in HFSS 3D Layout nets to plot using matplotlib the net.


MaxJPRey (dismissed):
LGTM.

Samuelopez-ansys (approved):

PR #5217, opened 2024-09-26
FIX: Issue error and skip material on parse error

### LINKED ISSUE #5215: Limited error handling in import_materials_from_file

### Description of the bug

Loading a JSON file for materials that fails to load due to some missing data generates just an error by the catch-all error handler, e.g.:

```
PyAEDT ERROR: **************************************************************
PyAEDT ERROR:   File "<frozen runpy>", line 198, in _run_module_as_main
PyAEDT ERROR:   File "<frozen runpy>", line 88, in _run_code
PyAEDT ERROR:   File "C:\git\Ansys\ic-setup-toolkit\venv\Scripts\pytest.exe\__main__.py", line 7, in <module>
PyAEDT ERROR:     sys.exit(console_main())
PyAEDT ERROR:   File "c:\git\Ansys\materials-avalanche\tests\test_apps.py", line 113, in test_json2amat_bad_json
PyAEDT ERROR:     assert json2amat_main(args) == 1
PyAEDT ERROR:   File "C:\git\Ansys\materials-avalanche\src\materials_avalanche\json2amat.py", line 70, in main
PyAEDT ERROR:     if not hfss.materials.import_materials_from_file(input_path.as_posix()):
PyAEDT ERROR: 'red' on import_materials_from_file
PyAEDT ERROR: Method arguments:
PyAEDT ERROR:     input_file = C:/git/Ansys/materials-avalanche/testdata/variations/process_materials_missing_rgb.json
PyAEDT ERROR: **************************************************************
```

While we can use this to find the actual problem, it would be nice to avoid the catch-all error handler and have a pointer at least to the material being parsed to narrow the search scope.

### Steps To Reproduce

1. Take any valid materials JSON with "appearance data" and remove the RGB color values (keys "red,", "green," "blue")
2. Pass above file into import_materials_from_file

Windows

3.10

### Installed packages

```
annotated-types==0.7.0
ansys-pythonnet==3.1.0rc3
asttokens==2.4.1
attrs==24.2.0
certifi==2024.8.30
cffi==1.17.1
charset-normalizer==3.3.2
click==8.1.7
click-plugins==1.1.1
cligj==0.7.2
clr-loader==0.2.6
colorama==0.4.6
contourpy==1.3.0
coverage==7.6.1
cycler==0.12.1
decorator==5.1.1
defusedxml==0.7.1
et-xmlfile==1.1.0
exceptiongroup==1.2.2
execnet==2.1.1
executing==2.1.0
fiona==1.10.1
fonttools==4.54.1
fpdf2==2.7.9
geopandas==0.14.4
idna==3.10
imageio==2.35.1
iniconfig==2.0.0
ipython==8.27.0
jedi==0.19.1
joblib==1.4.2
jsonschema==4.23.0
kiwisolver==1.4.7
markdown-it-py==3.0.0
matplotlib==3.9.2
matplotlib-inline==0.1.7
mdurl==0.1.2
meshio==5.3.5
mock==5.1.0
networkx==3.3
numpy==1.26.4
openpyxl==3.1.5
osmnx==1.9.4
packaging==24.1
pandas==2.2.3
parso==0.8.4
pillow==10.4.0
platformdirs==4.3.6
pluggy==1.5.0
plumbum==1.8.3
pooch==1.8.2
prompt_toolkit==3.0.48
psutil==6.0.0
pure_eval==0.2.3
-e git+https://github.com/ansys/pyaedt@44045940c750e23e6e409f28f972cfe865525263#egg=pyaedt
pycparser==2.22
pydantic==2.8.2
pydantic_core==2.20.1
pyedb==0.27.0
Pygments==2.18.0
pyparsing==3.1.4
pyproj==3.6.1
pytest==8.3.3
pytest-cov==5.0.0
pytest-xdist==3.6.1
python-dateutil==2.9.0.post0
pytomlpp==1.0.13
pytz==2024.2
pyvista==0.44.1
pywin32==306
PyYAML==6.0.2
referencing==0.35.1
requests==2.32.3
rich==13.8.1
rpds-py==0.20.0
rpyc==6.0.1
Rtree==1.3.0
scikit-learn==1.5.2
scikit-rf==1.3.0
scipy==1.14.1
scooby==0.10.0
shapely==2.0.6
six==1.16.0
SRTM.py==0.3.7
stack-data==0.6.3
threadpoolctl==3.5.0
toml==0.10.2
tomli==2.0.1
traitlets==5.14.3
typing_extensions==4.12.2
tzdata==2024.2
urllib3==2.2.3
utm==0.7.0
vtk==9.3.1
wcwidth==0.2.13
```

isaacwaldron:
Issue a specific error message when parsing a material fails in import_materials_from_file.  Also directly return False instead of relying on the catch-all exception handler.

Closes #5215


src/ansys/aedt/core/modules/material_lib.py
```diff
@@ -850,11 +850,15 @@ def import_materials_from_file(self, input_file=None, **kwargs):
...
-                materials_added.append(newmat)
+                try:
+                    newmat = Material(self, newname, val, material_update=False)
+                    newmat.update()
+                    newmat._material_update = True
+                    self.material_keys[newname] = newmat
+                    materials_added.append(newmat)
+                except KeyError as e:
+                    self.logger.error(f"Failed to import material {el!r} from {input_file!r}: key error on {e}")
+                    return False
```
SMoraisAnsys, line 861:
```suggestion
                    raise e
```

SMoraisAnsys: Users can chose to disable the error handler and leave error being raised. In any users shoes, I'd prefer knowing that something failed by see the raised exception instead of going through logging to check if an exception occurred. Since the changes you did would break that logic, could you raise the exception back ? If there is a good reason for not doing so, please let me know. You might want to remove `{e}` from the log message if you raise `e`.

isaacwaldron: @SMoraisAnsys thank you for pointing out the error handling setting.  Is there some reason other than historical precedent why that is enabled by default?  This often hides the true source of errors in my experience.

Re-raising the exception makes sense to me as then the user's setting for error handling will apply: if they have the catch-all turned on the method will return False and if not then they can catch the exception themselves.  I use `{e}` since for a `KeyError` it becomes the string expression for the bad key, e.g. `'Red'`.  I could be more explicit here and use `{e.args[0]}` in case this changed in a future version of Python, but for now this would allow an enterprising user to dig into the `Material` constructor and see where this key was used.

SMoraisAnsys: > @SMoraisAnsys thank you for pointing out the error handling setting, I was not aware of this. Is there some reason other than historical precedent why that is enabled by default? This often hides the true source of errors in my experience.
> 
> Re-raising the exception makes sense to me as then the user's setting for error handling will apply: if they have the catch-all turned on the method will return False and if not then they can catch the exception themselves. I use `{e}` since for a `KeyError` it becomes the string expression for the bad key, e.g. `'Red'`. I could be more explicit here and use `{e.args[0]}` in case this changed in a future version of Python, but for now this would allow an enterprising user to dig into the `Material` constructor and see where this key was used.

I think it was performed to allow users to have a more resilient experience when early users worked with `pyaedt`. This might be something we would like to update when release pyaedt `v1.0`.
@Samuelopez-ansys Would this make sense or should we keep it as before ?

src/ansys/aedt/core/modules/material_lib.py
```diff
@@ -770,7 +770,7 @@ def find_datasets(d, out_list):
         return write_configuration_file(json_dict, output_file)
 
     @pyaedt_function_handler(full_path="input_file")
-    def import_materials_from_file(self, input_file=None, **kwargs):
+    def import_materials_from_file(self, input_file=None, stop_on_error: bool = True, **kwargs):
```
maxcapodi78, line 773: it seems that stop_on_error is added but never used

maxcapodi78 (dismissed):
stop_on_error is added but never used

src/ansys/aedt/core/modules/material_lib.py
isaacwaldron: Good catch, initially I thought to add a flag to allow the user to say "I know what I'm doing" and continue/complete the import even if one or more materials failed to import.  I'll remove this.

_unittest/test_03_Materials.py
```diff
@@ -527,3 +528,14 @@ def test_16_import_materials_from_workbench(self):
             self.testapp2.materials.material_keys["84zn_12ag_4au_imp"].thermal_expansion_coefficient.thermalmodifier
             == "pwl($TM_84Zn_12Ag_4Au_imp_thermal_expansion_coefficient, Temp)"
         )
+
+    def test_17_json_missing_rgb(self, caplog: pytest.LogCaptureFixture):
```
SMoraisAnsys, line 532: I don't want to ask too much but could you mock the JSON content instead of adding data to the repo ?
I'd like to avoid adding files when the file content isn't really worth it. Here the goal is to trigger the warning so having only on material should be enough and the associated dictionary shouldn't be too huge.

isaacwaldron:
Converted to draft to accommodate feedback on test mocking.

_unittest/test_03_Materials.py
isaacwaldron: Done!

isaacwaldron:
@SMoraisAnsys @maxcapodi78 ready for review once again.

SMoraisAnsys (approved):
LGTM, thanks for the test upgrade :)

Alberto-DM (approved):
LGTM

Samuelopez-ansys (approved):

PR #5218, opened 2024-09-27
FEAT: LOG messages

maxcapodi78:
settings.enable_debug_grpc_api_logger = True 
will log all methods to the log as debug lines.

### DIFF
--- a/src/ansys/aedt/core/aedt_logger.py
+++ b/src/ansys/aedt/core/aedt_logger.py
@@ -629,7 +629,7 @@ class AedtLogger(object):
         self._messages.append([message_type, msg1, self.project_name, self.design_name])
         if not (self._log_on_file or self._log_on_screen) or not self._global:
             return
-        if len(message_text) > 250:
+        if len(message_text) > 250 and message_type < 3:
             message_text = message_text[:250] + "..."
         if message_type == 0:
             self._global.info(message_text, *args, **kwargs)
@@ -922,7 +922,7 @@ class AedtLogger(object):
 
     def debug(self, msg, *args, **kwargs):
         """Write a debug message to the global logger."""
-        if not settings.enable_debug_logger or not settings.enable_logger:
+        if not (settings.enable_debug_logger or settings.enable_debug_grpc_api_logger) or not settings.enable_logger:
             return
         if args:
             try:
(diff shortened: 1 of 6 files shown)


Samuelopez-ansys (approved):

src/ansys/aedt/core/aedt_logger.py
```diff
@@ -931,7 +931,8 @@ def debug(self, msg, *args, **kwargs):
                 msg1 = msg
         else:
             msg1 = msg
-        self._log_on_dekstop(0, msg1, "Global")
+        if not settings.enable_debug_grpc_api_logger:
```
SMoraisAnsys, line 934: Why is that if statement here ? If `settings.enable_debug_grpc_api_logger` is False, you would end up with a log on desktop right ?

src/ansys/aedt/core/application/design.py
```diff
@@ -684,7 +686,7 @@ def design_name(self, new_name):
                 timeout -= timestep
                 if timeout < 0:
                     raise RuntimeError("Timeout reached while checking design renaming.")
-        self._design_name = new_name
+            self._design_name = new_name
```
SMoraisAnsys, line 689: I'd leave it as it was before because if `new_name` is in `self.design_list`, you'll not update `self._design_name`. I've looked into `set_active_design` and I didn't find this update.

src/ansys/aedt/core/application/design.py
```diff
@@ -739,9 +741,12 @@ def project_name(self):
 
         >>> oProject.GetName
         """
+        if self._project_name:
+            return self._project_name
         if self.oproject:
             try:
                 self._project_name = self.oproject.GetName()
+                self._project_path = None
```
SMoraisAnsys, line 749: Why are you resetting `self._project_path` here ? This applies to other properties changes.

src/ansys/aedt/core/aedt_logger.py
maxcapodi78: yes and it would cause a huge slowdown. The debug will be only on the file

src/ansys/aedt/core/application/design.py
maxcapodi78: because if oproject changes the cached value has to be changed as well.
We may think to revert the three cached objects I've added into this PR. I mainly did to reduce the number of API calls

PR #5233, opened 2024-10-02
FIX: Defining Modelithics attributes change

ramin4667:
This PR updates the lumped export uni tests to pass the Modelithics attributes by changing the part library to Modelithics.


myoung301 (changes requested):
Can we test for the new error message somewhere? The message that gets issued when the part library is not set to Modelithics?

ramin4667:
> Can we test for the new error message somewhere? The message that gets issued when the part library is not set to Modelithics?

Unit tests are updated.

SMoraisAnsys (approved):
LGTM
@myoung301 I think the changes follow your request right ?

myoung301 (approved):

PR #5242, opened 2024-10-04
DOCS: Remove examples

Samuelopez-ansys:


pyproject.toml
```diff
@@ -117,6 +90,7 @@ doc-no-examples = [
     "scikit-rf>=0.30.0,<1.4",
     "pyvista[io]>=0.38.0,<0.45",
```
SMoraisAnsys, line 91: Not sure but can't we also get rid of matplotlib, scikit-rf and pyvista dependencies for doc build?

SMoraisAnsys (commented):
Two minor comments:
- can you update the doc dependencies to remove the one that are not used (pyvista, ...)?
- can you update the Makefile in the doc folder ?

pyproject.toml
Samuelopez-ansys: @SMoraisAnsys Done

SMoraisAnsys (approved):
LGTM

MaxJPRey (approved):

PR #5245, opened 2024-10-04
FEAT: Add SML component in Twin Builder

Samuelopez-ansys:


chrpetre (dismissed):
thanks @Samuelopez-ansys this looks great - just made a few minor changes

src/ansys/aedt/core/modeler/circuits/primitives_twin_builder.py
```diff
@@ -475,3 +475,111 @@ def create_periodic_waveform_source(
...
+                                "SDB:=",
+                                False,
+                            ],
+                        ],
+                    ],
+                ]
+            )
+            return True
+        except Exception:  # pragma: no cover
+            self.logger.warning("Property %s has not been edited. Check if readonly.", name)
```
SMoraisAnsys, line 584:
```suggestion
            self.logger.warning(f"Property {name} has not been edited. Check if readonly.")
```
We can use f-strings now !!! :D

SMoraisAnsys (approved):
LGTM

SMoraisAnsys (dismissed):

src/ansys/aedt/core/modeler/circuits/primitives_twin_builder.py
```diff
@@ -475,3 +475,111 @@ def create_periodic_waveform_source(
...
+    ):
+        """Create and place a new component based on a .sml file.
+
+        Parameters
+        ----------
+        input_file : str
+            Path to .sml file.
+        model : str
+            Model name to import.
+        pins_names : str
```
MaxJPRey, line 494: Should the type be a list?

SebastienMorais: There is a call to `pins_names.lower()` so I assume that's not the case :)

Samuelopez-ansys: Both of you are right, I like more a list. I will change it, because it is more pythonic.

I have implemented it as a string because it is what AEDT API expects, but we can do that conversion inside

MaxJPRey: Sounds good to me.

SMoraisAnsys (approved):

PR #5247, opened 2024-10-04
REFACTOR: Populate pyvista Plotter with initial data, enabling further customization before rendering

lorenzovecchietti:


SMoraisAnsys (approved):
Good idea, LGTM !
However, I'd use the "REFACTOR" prefix for your PR :)

PR #5248, opened 2024-10-04
FEAT: Prepare for EMIT Python 3.12 support.

myoung301:
Preparing EMIT to provide Python 3.12 starting with AEDT 2025.1.

### DIFF
--- a/_unittest_solvers/test_26_emit.py
+++ b/_unittest_solvers/test_26_emit.py
@@ -8,7 +8,7 @@ import pytest
 from ansys.aedt.core.generic.general_methods import is_linux
 from ansys.aedt.core.generic import constants as consts
 
-if (3,7) < sys.version_info <(3, 12):
+if (3, 7) < sys.version_info < (3, 12) or ( (3,9) < sys.version_info < (3,13) and config["desktopVersion"] > "2024.2" ):
     from ansys.aedt.core import Emit
     from ansys.aedt.core import generate_unique_project_name
     from ansys.aedt.core.emit_core.emit_constants import EmiCategoryFilter
@@ -30,7 +30,8 @@ def aedtapp(add_app):
 
 
 @pytest.mark.skipif(is_linux, reason="Emit API fails on linux.")
-@pytest.mark.skipif(sys.version_info < (3,8) or sys.version_info >=(3, 12), reason="Emit API is only available for Python 3.8+.")
+@pytest.mark.skipif(sys.version_info < (3,10), reason="Emit API is only available for Python 3.10+.")
+@pytest.mark.skipif(sys.version_info > (3,12) and config["desktopVersion"] < "2025.1", reason="Emit API is only available for Python 3.12 in AEDT version 2025.1 and later.")
 class TestClass:
 
     @pytest.fixture(autouse=True)
@@ -58,6 +59,9 @@ class TestClass:
                 assert self.aedtapp.results is not None
             elif sys.version_info.major == 3 and sys.version_info.minor == 11:
                 assert str(type(self.aedtapp._emit_api)) == "<class 'EmitApiPython311.EmitApi'>"
+                assert self.aedtapp.results is not None
+            elif sys.version_info.major == 3 and sys.version_info.minor == 12:
+                assert str(type(self.aedtapp._emit_api)) == "<class 'EmitApiPython312.EmitApi'>"
                 assert self.aedtapp.results is not None
 
     @pytest.mark.skipif(config["desktopVersion"] <= "2022.1", reason="Skipped on versions earlier than 2021.2")
@@ -694,16 +698,14 @@ class TestClass:
     )
     def test_13_static_type_generation(self):
         domain = self.aedtapp.results.interaction_domain()
-        if sys.version_info < (3, 8):
+        if sys.version_info < (3, 10):
             py_version = "EmitApiPython"
-        elif sys.version_info < (3, 9):
-            py_version = "EmitApiPython38"
-        elif sys.version_info < (3, 10):
-            py_version = "EmitApiPython39"
         elif sys.version_info < (3, 11):
             py_version = "EmitApiPython310"
         elif sys.version_info < (3, 12):
             py_version = "EmitApiPython311"
+        elif sys.version_info < (3, 13):
+            py_version = "EmitApiPython312"
         assert str(type(domain)) == "<class '{}.InteractionDomain'>".format(py_version)
 
         # assert str(type(TxRxMode)) == "<class '{}.tx_rx_mode'>".format(py_version)
@@ -937,8 +939,12 @@ class TestClass:
         assert len(radiosRX) == 2
 
     @pytest.mark.skipif(
-        config["desktopVersion"] <= "2023.1" or TEST_REVIEW_FLAG,
+        config["desktopVersion"] <= "2023.1",
         reason="Skipped on versions earlier than 2023.2",
+    )
+    @pytest.mark.skipif(
+        TEST_REVIEW_FLAG,
+        reason="Test under review",
     )
     def test_18_interference_scripts_no_filter(self, add_app):
         self.aedtapp = add_app(project_name="interference", application=Emit, subfolder=TEST_SUBFOLDER)
@@ -998,8 +1004,12 @@ class TestClass:
         assert protection_power_matrix == expected_protection_power
 
     @pytest.mark.skipif(
-        config["desktopVersion"] <= "2023.1" or TEST_REVIEW_FLAG,
+        config["desktopVersion"] <= "2023.1",
         reason="Skipped on versions earlier than 2023.2",
+    )
+    @pytest.mark.skipif(
+        TEST_REVIEW_FLAG,
+        reason="Test under review",
     )
     def test_20_interference_filtering(self, add_app):
         self.aedtapp = add_app(project_name="interference", application=Emit, subfolder=TEST_SUBFOLDER)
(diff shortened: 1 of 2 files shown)


_unittest_solvers/test_26_emit.py
```diff
@@ -30,7 +30,8 @@ def aedtapp(add_app):
 
 
 @pytest.mark.skipif(is_linux, reason="Emit API fails on linux.")
-@pytest.mark.skipif(sys.version_info < (3,8) or sys.version_info >=(3, 12), reason="Emit API is only available for Python 3.8+.")
+@pytest.mark.skipif(sys.version_info < (3,10), reason="Emit API is only available for Python 3.10+.")
+@pytest.mark.skipif(sys.version_info > (3,12) and config["desktopVersion"] < "2025.1", reason="Emit API is only available for Python 3.12 in AEDT version 2025.1 and later.")
```
SMoraisAnsys, line 34:
```suggestion
@pytest.mark.skipif(sys.version_info >= (3,12) and config["desktopVersion"] < "2025.1", reason="Emit API is only available for Python 3.12 in AEDT version 2025.1 and later.")
```

_unittest_solvers/test_26_emit.py
```diff
@@ -694,16 +698,14 @@ def test_12_radio_getters(self, add_app):
...
-        elif sys.version_info < (3, 9):
-            py_version = "EmitApiPython38"
-        elif sys.version_info < (3, 10):
-            py_version = "EmitApiPython39"
         elif sys.version_info < (3, 11):
             py_version = "EmitApiPython310"
         elif sys.version_info < (3, 12):
             py_version = "EmitApiPython311"
+        elif sys.version_info < (3, 13):
+            py_version = "EmitApiPython312"
```
SMoraisAnsys, line 708: Could we remove those if/elif and use `py_version = f"EmitApiPython{sys.version_info[0]}{sys.version_info[1]}"` instead ?

src/ansys/aedt/core/modeler/schematic.py
```diff
@@ -23,23 +23,17 @@
...
 from ansys.aedt.core.generic.general_methods import is_linux
 from ansys.aedt.core.generic.general_methods import pyaedt_function_handler
 from ansys.aedt.core.generic.settings import settings
 from ansys.aedt.core.modeler.cad.modeler import Modeler
-
-if (3, 8) < sys.version_info < (3, 12):
-    from ansys.aedt.core.modeler.circuits.primitives_emit import EmitComponent
-    from ansys.aedt.core.modeler.circuits.primitives_emit import EmitComponents
-else:  # pragma: no cover
-    warnings.warn("Emit API is only available for Python 3.8+,<3.12.")
```
SMoraisAnsys, line 40: Unless there is a version compatible with python3.13 (python version that should be available soon), I would like to keep the previous code and change (3.12) into (3.13) and "<3.12" into "3.13".

SMoraisAnsys (changes requested):
Thanks for the update. I left some minor comments.

_unittest_solvers/test_26_emit.py
myoung301: I see that this is incorrect for Python 3.12. I will make it `sys.version_info > (3,11)`. We can't use `>=` because `sys.version_info` has more than 2 elements so checking for equivalence with a 2-element tuple will always be false.

_unittest_solvers/test_26_emit.py
myoung301: Yeah, thanks for the idea. I also applied this to lines 48-64.

src/ansys/aedt/core/modeler/schematic.py
myoung301: Makes sense - I've made the change and corrected the warning message that was inaccurate before. Thank you.

myoung301:
@SMoraisAnsys Thank you for the review. I think I've addressed each of the comments. Please review the updates when you have a chance.

_unittest_solvers/test_26_emit.py
```diff
@@ -30,7 +30,8 @@ def aedtapp(add_app):
 
 
 @pytest.mark.skipif(is_linux, reason="Emit API fails on linux.")
-@pytest.mark.skipif(sys.version_info < (3,8) or sys.version_info >=(3, 12), reason="Emit API is only available for Python 3.8+.")
+@pytest.mark.skipif(sys.version_info < (3,10), reason="Emit API is only available for Python 3.10+.")
```
SMoraisAnsys, line 33:
```suggestion
@pytest.mark.skipif(sys.version_info < (3, 8), reason="Emit API is only available for Python 3.8+.")
```
For the time being we need to let people test `pyaedt` with Emit on python 3.8 (e.g. they use 2024.2). Once we get rid of the compatibility with python 3.8 in `pyaedt` (planned with `v1.0.0`), I suggest that we apply your changes.

_unittest_solvers/test_26_emit.py
```diff
@@ -30,7 +30,8 @@ def aedtapp(add_app):
...
+@pytest.mark.skipif(sys.version_info > (3,11) and config["desktopVersion"] < "2025.1", reason="Emit API is only available for Python 3.12 in AEDT version 2025.1 and later.")
```
SMoraisAnsys, line 34:
```suggestion
@pytest.mark.skipif(sys.version_info[0:2] == (3, 12) and config["desktopVersion"] < "2025.1", reason="Emit API is only available for Python 3.12 in AEDT version 2025.1 and later.")
```

_unittest_solvers/test_26_emit.py
```diff
@@ -694,16 +682,9 @@ def test_12_radio_getters(self, add_app):
...
-        elif sys.version_info < (3, 9):
-            py_version = "EmitApiPython38"
-        elif sys.version_info < (3, 10):
-            py_version = "EmitApiPython39"
-        elif sys.version_info < (3, 11):
-            py_version = "EmitApiPython310"
-        elif sys.version_info < (3, 12):
-            py_version = "EmitApiPython311"
+        py_version = "EmitApiPython"
+        if sys.version_info > (3, 9):
```
SMoraisAnsys, line 686:
```suggestion
        if sys.version_info > (3, 7):
```

src/ansys/aedt/core/modeler/schematic.py
```diff
@@ -32,14 +32,15 @@
 from ansys.aedt.core.generic.general_methods import pyaedt_function_handler
 from ansys.aedt.core.generic.settings import settings
 from ansys.aedt.core.modeler.cad.modeler import Modeler
+from ansys.aedt.core.modeler.circuits.object_3d_circuit import CircuitComponent
+from ansys.aedt.core.modeler.circuits.object_3d_circuit import Wire
 
-if (3, 8) < sys.version_info < (3, 12):
+if (3, 8) < sys.version_info < (3, 13):
```
SMoraisAnsys, line 38:
```suggestion
if (3, 7) < sys.version_info < (3, 13):
```

src/ansys/aedt/core/modeler/schematic.py
```diff
@@ -32,14 +32,15 @@
...
     from ansys.aedt.core.modeler.circuits.primitives_emit import EmitComponent
     from ansys.aedt.core.modeler.circuits.primitives_emit import EmitComponents
 else:  # pragma: no cover
-    warnings.warn("Emit API is only available for Python 3.8+,<3.12.")
-from ansys.aedt.core.modeler.circuits.object_3d_circuit import CircuitComponent
-from ansys.aedt.core.modeler.circuits.object_3d_circuit import Wire
+    warnings.warn("EMIT API is only available for Python 3.9-3.12.")
```
SMoraisAnsys, line 42:
```suggestion
    warnings.warn("EMIT API is only available for Python 3.8-3.12.")
```

SMoraisAnsys (changes requested):
Thanks for taking into account my previous comments and helping us with the changes available with 25.1.
Here are some extra requests as I noticed that there was a typo in our current code base. As far as I know, we still have to handle AEDT versions where EMIT was available with Python 3.8. Therefore, as long as `pyaedt` is compatible with it (support dropping planned in `v1.0.0`) we need to keep it.

_unittest_solvers/test_26_emit.py
myoung301: Since we are skipping these tests for versions that aren't supported, we won't encounter the unversioned type name and can remove this condition.

_unittest_solvers/test_26_emit.py
```diff
@@ -8,7 +8,7 @@
 from ansys.aedt.core.generic.general_methods import is_linux
 from ansys.aedt.core.generic import constants as consts
 
-if (3,7) < sys.version_info <(3, 12):
+if (3, 7) < sys.version_info < (3, 12) or ( (3,9) < sys.version_info < (3,13) and config["desktopVersion"] > "2024.2" ):
```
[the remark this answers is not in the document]
myoung301: I broke this apart into separate version checks for pre/post 2025R1 to make it clearer.

_unittest_solvers/test_26_emit.py
myoung301: I applied a similar change here, separating the logic for before/after 2025R1.

_unittest_solvers/test_26_emit.py
myoung301: This particular change is no longer applicable with the way the logic is broken into two separate checks based on AEDT version.

src/ansys/aedt/core/modeler/schematic.py
myoung301: Applied.

src/ansys/aedt/core/modeler/schematic.py
myoung301: Applied.

myoung301:
Thanks again. I think I've addressed each of the comments, but let me know your thoughts.

SMoraisAnsys (approved):
LGTM, thanks for all the changes and the contribution

PR #5254, opened 2024-10-07
CHORE: Update 0.12.dev

gmalinve:


SMoraisAnsys:
Update from wrong branch resulting in wrong changes (from 0.11.dev0 to 0.11.0 when we should move to 0.12.dev0 as title says)

PR #5256, opened 2024-10-07
FIX: check solve_inside in create_fieldplot_volume

gmalinve:


src/ansys/aedt/core/visualization/post/post_common_3d.py
```diff
@@ -1753,14 +1753,23 @@ def create_fieldplot_volume(
...
         """
+        assignment = self.modeler.convert_to_selections(assignment, True)
+
+        for a in assignment:
+            if a not in list(self.modeler.objects_by_name.keys()):
+                self.logger.error("{} does not exist in current design".format(a))
+                return False
+            elif self.modeler.objects_by_name[a].is_conductor and not self.modeler.objects_by_name[a].solve_inside:
+                self.logger.warning("Solve inside is unchecked for {} object.".format(a))
+                return False
```
SMoraisAnsys, line 1772:
```suggestion
        for element in assignment:
            if element not in list(self.modeler.objects_by_name.keys()):
                self.logger.error(f"{element} does not exist in current design")
                return False
            elif self.modeler.objects_by_name[a].is_conductor and not self.modeler.objects_by_name[a].solve_inside:
                self.logger.warning(f"Solve inside is unchecked for {element}.")
                return False
```
Can you explain to me why there is a need for the `elif` case ? IMO, it's a bit strange to see this checking in a method named `create_fieldplot_volume`. I might lack knowledge and if that's the case, please enlighten me.

_unittest/test_12_1_PostProcessing.py
```diff
@@ -102,7 +102,17 @@ def test_01B_Field_Plot(self):
         plot1.update()
         assert self.aedtapp.post.field_plots[plot1.name].IsoVal == "Tone"
         assert plot1.change_plot_scale(min_value, "30000", scale_levels=50)
-        assert self.aedtapp.post.create_fieldplot_volume("inner", "Vector_E", setup_name, intrinsic)
+        assert not self.aedtapp.post.create_fieldplot_volume("inner", "Vector_E", setup_name, intrinsic)
+        inner = self.aedtapp.modeler.objects_by_name["inner"]
+        inner.solve_inside = True
+        inner_plot = self.aedtapp.post.create_fieldplot_volume("inner", "Vector_E", setup_name, intrinsic)
+        self.aedtapp.analyze(self.aedtapp.active_setup)
+        assert inner_plot
```
SMoraisAnsys, line 110: Shouldn't that call me moved before analyze ? It makes me think that analyze is required for the assert to work. That's not the case right ?

Samuelopez-ansys (approved):

PR #5262, opened 2024-10-08
REFACTOR: Renaming filter into filter_order

ramin4667:
This PR renames order to filter_order to make it more meaningful.

### DIFF
--- a/_unittest/test_45_FilterSolutions/test_filter/test_attributes.py
+++ b/_unittest/test_45_FilterSolutions/test_filter/test_attributes.py
@@ -108,20 +108,20 @@ class TestClass:
             design.attributes.diplexer_type = diplexer_type
             assert design.attributes.diplexer_type == diplexer_type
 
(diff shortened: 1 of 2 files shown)


SMoraisAnsys:
Could you please avoid adding the link to tfs (at most refer the issue number).
If possible, could you provide a better explanation as to why you are proposing this changes ?

ramin4667:
> Could you please avoid adding the link to tfs (at most refer the issue number). If possible, could you provide a better explanation as to why you are proposing this changes ?

The link is removed.

SMoraisAnsys:
Thanks @ramin4667, can you explain why this change is more meaningful ? That is not clear to me

ramin4667:
> Member

@SMoraisAnsys The rest of the attributes start with filter_***, and the user could mistakenly use filter_order instead of order. This change makes attribute names consistent.

SMoraisAnsys (approved):

Issue #5258, opened 2024-10-07
ERROR with AEDT is not installed on your system when running unitest


anspraksaph:
### Before submitting the issue

- [X] I have searched among the existing issues
- [X] I am using a Python virtual environment

### Description of the bug

When using PyAedt With this commit hash: 99b3187517e463d15dfdfef1cf6491c56945ddeb, unable to run test C:\pyaedt\_unittest_solvers\test_26_emit.py with this error "AEDT is not installed on your system. Install AEDT version 2022 R2 or higher."

There is no problem with this commit hash: c9df21df4435f4e0b966727ec5d8cc24f8d984db (When using with the same AEDT version)(I'm using EDT 25.1)

### Steps To Reproduce

1. Clone PyAedt at "C:/pyaedt". 
2. With Python 3.11, set virtual environment, activate it and run these command:
- python -m pip install pip -U
- python -m pip install wheel setuptools -U
- python -m pip install C:/pyaedt\.[tests]
- python -m pip install pytest-azurepipelines
- pytest -v C:\pyaedt\\_unittest_solvers\test_26_emit.py

### Which Operating System are you using?

Windows

### Which Python version are you using?

3.11

### Installed packages

(test_env_3.10) C:\Users\Public\Desktop\PyAedt_Terminal>python -m pip freeze
annotated-types==0.7.0
ansys-pythonnet==3.1.0rc3
asttokens==2.4.1
attrs==24.2.0
certifi==2024.8.30
cffi==1.17.1
charset-normalizer==3.3.2
click==8.1.7
click-plugins==1.1.1
cligj==0.7.2
clr-loader==0.2.6
colorama==0.4.6
contourpy==1.3.0
coverage==7.6.1
cycler==0.12.1
decorator==5.1.1
defusedxml==0.7.1
et-xmlfile==1.1.0
exceptiongroup==1.2.2
execnet==2.1.1
executing==2.1.0
fiona==1.10.1
fonttools==4.54.1
fpdf2==2.8.1
geopandas==0.14.4
idna==3.10
imageio==2.35.1
importlib_resources==6.4.5
iniconfig==2.0.0
ipython==8.27.0
jedi==0.19.1
joblib==1.4.2
jsonschema==4.23.0
jsonschema-specifications==2023.12.1
kiwisolver==1.4.7
markdown-it-py==3.0.0
matplotlib==3.9.2
matplotlib-inline==0.1.7
mdurl==0.1.2
meshio==5.3.5
[...]

PR #5264, opened 2024-10-08
FIX: Case where only a prerelease is installed on system

Alberto-DM:
Solves a bug preventing AEDT from opening when only a pre-release is installed on the system.

Solves #5258


SMoraisAnsys (approved):
LGTM, nice catch !

Samuelopez-ansys (approved):

PR #5265, opened 2024-10-08
FEAT: add twinbuilder apis

chrpetre:
adding new Twin Builder APIs with corresponding unit testing to
- modify a property value
- create a subsheet


chrpetre:
@Samuelopez-ansys it looks like there is a pyaedt api to change quantity/property using : 
e.g. 
myComp = tb.modeler.schematic.create_component(....)
myComp.set_property(property/quantity name, property/quantiy value)
therefore I don't think the new methods
- update_quantity_value
- update_property_value
are needed at all. Please take a look and let me know (in that case, I can remove them/test cases, and you would need to modify the new pyaedt/tb example accordingly)

Samuelopez-ansys:
@chrpetre It Looks good to me. 

Once it is merge in the main, you can go to the PyAEDT examples repository, and create a new Pull request to update Twin Builder examples. The examples repository uses the current pyaedt main, so it means you do not need to wait until the PyaAEDT release.

Samuelopez-ansys (approved):

Samuelopez-ansys (dismissed):

Samuelopez-ansys (approved):

PR #5266, opened 2024-10-08
FIX: Trap topology condition update for multiband filters

ramin4667:
This PR updates the Trap topology condition to accept multiple band filters.

### DIFF
--- a/_unittest/test_45_FilterSolutions/test_filter/test_multiple_bands_table.py
+++ b/_unittest/test_45_FilterSolutions/test_filter/test_multiple_bands_table.py
@@ -91,7 +91,7 @@ class TestClass:
             == "Either no value is set for this band or the rowIndex must be greater than zero and less than row count"
         )
 
-    def test_clear_table():
+    def test_clear_table(self):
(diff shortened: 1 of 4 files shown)


_unittest/test_45_FilterSolutions/test_filter/test_multiple_bands_table.py
```diff
@@ -90,3 +90,16 @@ def test_remove_row(self):
             info.value.args[0]
             == "Either no value is set for this band or the rowIndex must be greater than zero and less than row count"
         )
+
+    def test_clear_table():
+        design = ansys.aedt.core.FilterSolutions(implementation_type=FilterImplementation.LUMPED)
```
myoung301, line 95: @ramin4667 In order to test the functionality of `clear_table()`, it seems like a good idea to confirm there is data in the table before we call `clear_table()`. Also, what is the significance of 7 - why test 7 rows? (maybe add a comment explaining the rationale - this may help with future test maintenance).

ramin4667: The table is filled with 2 rows by default. Comments are added to clarify the tests.

myoung301 (approved):
Looks good.

PR #5268, opened 2024-10-09
CI: Update artifact action on every workflows

SMoraisAnsys:
Follow up of #5260 where changes weren't applied on every workflow.

### DIFF
--- a/.github/workflows/manual_draft.yml
+++ b/.github/workflows/manual_draft.yml
@@ -88,7 +88,7 @@ jobs:
           flags: system,solver
 
       - name: Upload pytest test results
-        uses: actions/upload-artifact@v3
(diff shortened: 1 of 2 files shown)


Samuelopez-ansys (approved):

Samuelopez-ansys:
@SMoraisAnsys Do we still need nightly doc workflow? The documentation could be uploaded on every PR right?

MaxJPRey (approved):
LGTM.

SMoraisAnsys:
> @SMoraisAnsys Do we still need nightly doc workflow? The documentation could be uploaded on every PR right?

As long as we don't move to the next ansys actions, we'll have to do it nightly as the index upload is very long. See below
image

I'll update the repo to use new versions of ansys actions in another PR

PR #5271, opened 2024-10-09
FEAT: method to add notes on design 

anur7:


src/ansys/aedt/core/application/design.py
```diff
@@ -4203,6 +4203,26 @@ def set_temporary_directory(self, path):
...
+        text : str
+            Notes to be inserted.
+
+        Returns
+        -------
+        bool
+            ``True`` when successful, ``False`` when failed.
+        """
+        if not isinstance(text, str):
+            self.logger.error("Text is not valid.")
```
SMoraisAnsys, line 4221:
```suggestion
            self.logger.error("Input type of edit_notes is not valid.")
```

SMoraisAnsys (dismissed):
LGTM, I left a minor comment.

gmalinve (changes requested):

gmalinve (approved):
LGTM.

PR #5276, opened 2024-10-10
FIX: FilterSolutions graph min and max not working

ramin4667:
This PR fixes graph settings for the defined values set by the user. Also, it updates dll path and dll unit test to test the current version.

### DIFF
--- a/_unittest/test_45_FilterSolutions/test_filter/test_dll_interface.py
+++ b/_unittest/test_45_FilterSolutions/test_filter/test_dll_interface.py
@@ -35,7 +35,9 @@ import pytest
 @pytest.mark.skipif(config["desktopVersion"] < "2025.1", reason="Skipped on versions earlier than 2025.1")
 class TestClass:
     def test_version(self):
-        assert ansys.aedt.core.filtersolutions_core.api_version() == "FilterSolutions API Version 2025 R1 (Beta)"
+        version_string = config["desktopVersion"].replace(".", " R")
+        expected_version_string = "FilterSolutions API Version " + version_string + " (Beta)"
+        assert ansys.aedt.core.filtersolutions_core.api_version() == expected_version_string
 
     def test_string_to_enum(self):
         assert (
--- a/_unittest/test_45_FilterSolutions/test_filter/test_dll_interface.py
+++ b/_unittest/test_45_FilterSolutions/test_filter/test_dll_interface.py
@@ -36,7 +36,7 @@ import pytest
 class TestClass:
     def test_version(self):
         version_string = config["desktopVersion"].replace(".", " R")
-        expected_version_string = "FilterSolutions API Version " + version_string + " (Beta)"
+        expected_version_string = f"FilterSolutions API Version {version_string} (Beta)"
         assert ansys.aedt.core.filtersolutions_core.api_version() == expected_version_string
 
     def test_string_to_enum(self):
--- a/src/ansys/aedt/core/filtersolutions_core/dll_interface.py
+++ b/src/ansys/aedt/core/filtersolutions_core/dll_interface.py
@@ -59,7 +59,7 @@ class DllInterface:
             version + "CL" in aedt_versions.installed_versions
         ):
             raise ValueError("Specified version {} is not installed on your system".format(version[0:6]))
-        self.dll_path = os.path.join(aedt_versions.installed_versions[version], "nuhertz/FilterSolutionsAPI.dll")
+        self.dll_path = os.path.join(aedt_versions.installed_versions[version], "nuhertz", "FilterSolutionsAPI.dll")
         print("DLL Path:", self.dll_path)
         if not os.path.isfile(self.dll_path):
             raise RuntimeError(f"The 'FilterSolutions' API DLL was not found at {self.dll_path}.")  # pragma: no cover
--- a/src/ansys/aedt/core/filtersolutions_core/ideal_response.py
+++ b/src/ansys/aedt/core/filtersolutions_core/ideal_response.py
@@ -341,8 +341,10 @@ class IdealResponse:
         y_axis_parameter: `FrequencyResponseColumn`, optional
              Frequency response column to return. The default is the frequency response magnitude in dB.
         minimum_frequency: str, optional
+            Minimum frequency to set for the frequency response.
             The default is ``None``.
         maximum_frequency: str, optional
+            Maximum frequency to set for the frequency response.
             The default is ``None``.
         vsg_analysis_enabled: bool, optional
             The default is ``False``.
@@ -355,10 +357,10 @@ class IdealResponse:
             list of the requested parameters.
         """
 
-        if minimum_frequency is not None:
-            minimum_frequency = self.graph_setup.minimum_frequency
         if maximum_frequency is not None:
-            maximum_frequency = self.graph_setup.maximum_frequency
+            self.graph_setup.maximum_frequency = maximum_frequency
+        if minimum_frequency is not None:
+            self.graph_setup.minimum_frequency = minimum_frequency
         self.vsg_analysis_enabled = vsg_analysis_enabled
         frequency = self._frequency_response_getter(FrequencyResponseColumn.FREQUENCY)
         frequency_hz = []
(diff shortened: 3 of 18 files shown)


src/ansys/aedt/core/filtersolutions_core/ideal_response.py
```diff
@@ -341,9 +341,9 @@ def frequency_response(
         y_axis_parameter: `FrequencyResponseColumn`, optional
              Frequency response column to return. The default is the frequency response magnitude in dB.
         minimum_frequency: str, optional
-            The default is ``200 MHz``.
+            The default is ``None``.
```
myoung301, line 344: We should describe the behavior of ``None``

src/ansys/aedt/core/filtersolutions_core/ideal_response.py
```diff
@@ -341,9 +341,9 @@ def frequency_response(
...
         maximum_frequency: str, optional
-            The default is ``5 GHz``.
+            The default is ``None``.
```
myoung301, line 346: Desbribe meaning of ``None``

src/ansys/aedt/core/filtersolutions_core/ideal_response.py
```diff
@@ -354,8 +354,11 @@ def frequency_response(
             of the defined frequency ranges, and the second is a
             list of the requested parameters.
         """
-        self.graph_setup.minimum_frequency = minimum_frequency
-        self.graph_setup.maximum_frequency = maximum_frequency
+
+        if minimum_frequency is not None:
+            minimum_frequency = self.graph_setup.minimum_frequency
+        if maximum_frequency is not None:
+            maximum_frequency = self.graph_setup.maximum_frequency
```
myoung301, line 361: Is this backwards? Shouldn't we be assigning the new value to self.graph_setup.minimum_frequency when the new value is not none?

```suggestion
        if minimum_frequency is not None:
            self.graph_setup.minimum_frequency = minimum_frequency
        if maximum_frequency is not None:
            self.graph_setup.maximum_frequency = maximum_frequency
```

If this didn't cause a unit test to fail, we need to improve the frequency_response unit test.

src/ansys/aedt/core/filtersolutions_core/ideal_response.py
```diff
@@ -391,8 +394,10 @@ def time_response(
             of the defined time ranges, and the second is a
             list of the requested parameters.
         """
-        self.graph_setup.minimum_time = minimum_time
-        self.graph_setup.maximum_time = maximum_time
+        if minimum_time is not None:
```
myoung301, line 397: Same change here, and same question about the unit test.

src/ansys/aedt/core/filtersolutions_core/ideal_response.py
```diff
@@ -424,8 +429,10 @@ def s_parameters(
             of the defined frequency ranges, and the second is a
             list of the requested parameters.
         """
-        self.graph_setup.minimum_frequency = minimum_frequency
-        self.graph_setup.maximum_frequency = maximum_frequency
+        if minimum_frequency is not None:
```
myoung301, line 432: Same thing here and same question about the unit test.

myoung301 (changes requested):

src/ansys/aedt/core/filtersolutions_core/ideal_response.py
ramin4667: The unit tests are not failing. What error are you expecting to have the tests improved?

src/ansys/aedt/core/filtersolutions_core/ideal_response.py
ramin4667: Done

src/ansys/aedt/core/filtersolutions_core/ideal_response.py
ramin4667: Done

src/ansys/aedt/core/filtersolutions_core/ideal_response.py
myoung301: @ramin4667 I meant that we should describe what happens when the value is ``None``. It doesn't seem like this is obvious/unambiguous.

src/ansys/aedt/core/filtersolutions_core/ideal_response.py
myoung301: @ramin4667 These expressions were backwards and therefore the plot extents were not being applied. If the unit test verifies the plot extents are properly set, it should have been failing before you applied this correction.

src/ansys/aedt/core/filtersolutions_core/ideal_response.py
ramin4667: Done

myoung301: @ramin4667 The updated comment mentions that the orignal default (e.g. 200 MHz) will be used if the value is None. But is that only true if the limit has never been set or is that true in general? For example, if someone sets the limit to 100MHz, and then calls the function and leaves the limit as None, will it get set back to 200MHz or stay at the 100MHz value? If it is set back to 200MHz, then I think your comment is correct, but if the limit stays at whatever value is currently set, then I think the comment needs to be corrected to make that clear.

src/ansys/aedt/core/filtersolutions_core/ideal_response.py
myoung301: @ramin4667 I realize I didn't unresolve this conversation before so I'm unresolving it now until we figure out if the unit tests need to be fixed/enhanced.

src/ansys/aedt/core/filtersolutions_core/dll_interface.py
```diff
@@ -59,7 +59,7 @@ def _init_dll_path(self, version):
             version + "CL" in aedt_versions.installed_versions
         ):
             raise ValueError("Specified version {} is not installed on your system".format(version[0:6]))
-        self.dll_path = os.path.join(aedt_versions.installed_versions[version], "nuhertz/FilterSolutionsAPI.dll")
+        self.dll_path = os.path.join(aedt_versions.installed_versions[version], "nuhertz", "FilterSolutionsAPI.dll")
```
SMoraisAnsys, line 62: Good catch !

_unittest/test_45_FilterSolutions/test_filter/test_dll_interface.py
```diff
@@ -35,7 +35,9 @@
 @pytest.mark.skipif(config["desktopVersion"] < "2025.1", reason="Skipped on versions earlier than 2025.1")
 class TestClass:
     def test_version(self):
-        assert ansys.aedt.core.filtersolutions_core.api_version() == "FilterSolutions API Version 2025 R1 (Beta)"
+        version_string = config["desktopVersion"].replace(".", " R")
+        expected_version_string = "FilterSolutions API Version " + version_string + " (Beta)"
```
SMoraisAnsys, line 39:
```suggestion
        expected_version_string = f"FilterSolutions API Version {version_string} (Beta)"
```

SMoraisAnsys (commented):
I'm not sure to understand why are the changes required.

Are the previous default values not working even for 2024.1, 2024.2 and so on or is it only for 2025.1 ? Also, when you say that the min and max are not working, does that mean that we shouldn't change the value of any input unless the user provided it ? If so, do we have users leveraging the fact that not providing a value to `minimum_frequency` will set it to ``"200 MHz"` ? If that's the case we might want to add a temporary log to let users know about this change of behavior.

src/ansys/aedt/core/filtersolutions_core/dll_interface.py
ramin4667: The versions less than 2025.1 do not include the API dll. The dfault values are 200MHz, and 5 GHz. user can set them either using "graph_setup.minimum_frequency" and "graph_setup.maximum_frequency" or directly entering in the "frequency_response" function.

_unittest/test_45_FilterSolutions/test_filter/test_dll_interface.py
ramin4667: Applied.

src/ansys/aedt/core/filtersolutions_core/ideal_response.py
ramin4667: Updated.

myoung301 (approved):

PR #5283, opened 2024-10-11
VSin, ISin Sources added to Maxwell Circuit Primitives

DaveTwyman:
Create Page and Select Page added to Circuit Primitives

### DIFF
--- a/src/ansys/aedt/core/modeler/circuits/primitives_circuit.py
+++ b/src/ansys/aedt/core/modeler/circuits/primitives_circuit.py
@@ -1261,6 +1261,21 @@ class CircuitComponents(object):
         except Exception:
             return False
 
+    @pyaedt_function_handler()
+    def create_page(self,name):
+        """Adds a new page to circuit schematic.
+
+        Parameters
+        """
+        self.oeditor.CreatePage (name)
+
+    @pyaedt_function_handler()
+    def select_page(self,name):
+        """Adds a new page to circuit schematic.
+
+        Parameters
+        """
+        self.oeditor.SelectPage (name)
 
 class ComponentInfo(object):
     """Manages Circuit Catalog info."""
--- a/src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
+++ b/src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
@@ -284,3 +284,88 @@ class MaxwellCircuitComponents(CircuitComponents):
         )
         id.set_property("Name", name)
         return id
+
+
+    @pyaedt_function_handler(compname="name")
+    def create_ISin(self,name=None,value=1,location=None, angle=0, use_instance_id_netlist=False):
+        """Create a sinusoidal current source.
+
+        Parameters
+        ----------
+        name : str, optional
+            Name of the current source. The default is ``None``.
+        value : float, optional
+            Value for the amplitude of current. The default is ``1``.
+        location : list of float, optional
+            Position on the X axis and Y axis.
+        angle : float, optional
+            Angle of rotation in degrees. The default is ``0``.
+        use_instance_id_netlist : bool, optional
+            Whether to use the instance ID in the net list or not. The default is ``False``.
+
+        Returns
+        -------
+        :class:`ansys.aedt.core.modeler.cad.object_3dcircuit.CircuitComponent`
+            Circuit Component Object.
+
+        References
+        ----------
+
+        >>> oEditor.CreateComponent
+        """
+        if location is None:
+            location = []
+        id = self.create_component(
+            name,
+            component_library="Sources",
+            component_name="ISin",
+            location=location,
+            angle=angle,
+            use_instance_id_netlist=use_instance_id_netlist,
+        )
+
+        id.set_property("Ia", value)
+        id.set_property("Name", name)
+        return id
+
+    @pyaedt_function_handler(compname="name")
+    def create_VSin(self,name=None,value=1,location=None, angle=0, use_instance_id_netlist=False):
+        """Create a sinusoidal voltage source.
+
+        Parameters
+        ----------
+        name : str, optional
+            Name of the voltage source. The default is ``None``.
+        value : float, optional
+            Value for the amplitude of voltage. The default is ``1``.
+        location : list of float, optional
+            Position on the X axis and Y axis.
+        angle : float, optional
+            Angle of rotation in degrees. The default is ``0``.
+        use_instance_id_netlist : bool, optional
+            Whether to use the instance ID in the net list or not. The default is ``False``.
+
+        Returns
+        -------
+        :class:`ansys.aedt.core.modeler.cad.object_3dcircuit.CircuitComponent`
+            Circuit Component Object.
+
+        References
+        ----------
+
+        >>> oEditor.CreateComponent
+        """
+        if location is None:
+            location = []
+        id = self.create_component(
+            name,
+            component_library="Sources",
+            component_name="VSin",
+            location=location,
+            angle=angle,
+            use_instance_id_netlist=use_instance_id_netlist,
+        )
+
+        id.set_property("Va", value)
+        id.set_property("Name", name)
+        return id


src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
```diff
@@ -284,3 +284,87 @@ def create_winding(self, name=None, location=None, angle=0, use_instance_id_netl
         )
         id.set_property("Name", name)
         return id
+
+    @pyaedt_function_handler(compname="name")
+    def create_ISin(self, name=None, value=1, location=None, angle=0, use_instance_id_netlist=False):
```
SMoraisAnsys, line 289: Please avoid using capital letters in methods name, we follow the snake case convention.

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
```diff
@@ -284,3 +284,87 @@ def create_winding(self, name=None, location=None, angle=0, use_instance_id_netl
...
+            Circuit Component Object.
+
+        References
+        ----------
+
+        >>> oEditor.CreateComponent
+        """
+        if location is None:
+            location = []
+        id = self.create_component(
```
SMoraisAnsys, line 317: Please avoid using `id` as it is a built-in function.
image

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
```diff
@@ -284,3 +284,87 @@ def create_winding(self, name=None, location=None, angle=0, use_instance_id_netl
...
+            angle=angle,
+            use_instance_id_netlist=use_instance_id_netlist,
+        )
+
+        id.set_property("Ia", value)
+        id.set_property("Name", name)
+        return id
+
+    @pyaedt_function_handler(compname="name")
+    def create_VSin(self, name=None, value=1, location=None, angle=0, use_instance_id_netlist=False):
```
SMoraisAnsys, line 331: Same as previous comment

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
```diff
@@ -284,3 +284,87 @@ def create_winding(self, name=None, location=None, angle=0, use_instance_id_netl
...
+            Circuit Component Object.
+
+        References
+        ----------
+
+        >>> oEditor.CreateComponent
+        """
+        if location is None:
+            location = []
+        id = self.create_component(
```
SMoraisAnsys, line 359: Same as previous comment

SMoraisAnsys (commented):
Thanks for the changes, could you address the comments and add tests ?

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
DaveTwyman: OK, I'll adjust it to 'compid', I'll also change the other five existing uses in this file as well whilst I'm there

gmalinve: @DaveTwyman please also remember to adhere to python case convention and use underscores to separate words.

DaveTwyman: @gmalinve , Ok will do, just to check this is referring to the 'compid' suggestion, if so I'll use 'comp_id' instead

DaveTwyman: @SMoraisAnsys ,@gmalinve
Yikes, looking at this a bit deeper, 'id' is also the returned value from many of the existing circuit components in both 'primitives_circuit' and 'primitives_maxwell_circuit'. Probably best if I stick to using the existing convention for now as it works. We could raise another pull request to make all these changes another time?

gmalinve: @DaveTwyman we can think about this later :) first you need to fix your methods and add unit tests

DaveTwyman: Yes, I agree

gmalinve:
@DaveTwyman please install pre-commit in your venv because I see that every time you push a commit it automatically triggers the auto fixes pre-commit: 
image
This package will automatically run all the checks specified in .pre-commit-config.yaml when you push a commit

DaveTwyman:
> @DaveTwyman please install pre-commit in your venv because I see that every time you push a commit it automatically triggers the auto fixes pre-commit: image This package will automatically run all the checks specified in .pre-commit-config.yaml when you push a commit.
> 
> Also remember to update and merge main in your branch from time to time.

@gmalinve Good point about the pre-commit, I see it's installed on my machine. I've just updated it too. I guess the issue is I need to start using it before each push. 

I believe I'm updating main before each push from PyCharm, but let me know if you can see evidence something is wrong here. Thanks again as always

SMoraisAnsys:
@DaveTwyman If it is installed on your machine then it shouldn't let you commit. The precommit should trigger the problems locally before allowing you to make any commit.

DaveTwyman:
> @DaveTwyman If it is installed on your machine then it shouldn't let you commit. The precommit should trigger the problems locally before allowing you to make any commit.

Ok, understood, I'm just reading up on how to set this up now

gmalinve:
@DaveTwyman you need to add unit tests now :)

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
```diff
@@ -284,3 +284,87 @@ def create_winding(self, name=None, location=None, angle=0, use_instance_id_netl
...
```
gmalinve, line 317: @DaveTwyman create_component returns a component not an id so better to call it in another way. You stated it also in the "Returns" section.

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
```diff
@@ -284,3 +284,87 @@ def create_winding(self, name=None, location=None, angle=0, use_instance_id_netl
...
+        -------
+        :class:`ansys.aedt.core.modeler.cad.object_3dcircuit.CircuitComponent`
+            Circuit Component Object.
+
+        References
+        ----------
+
+        >>> oEditor.CreateComponent
+        """
+        if location is None:
```
gmalinve, line 315: @DaveTwyman please also handle the case when name=None because otherwise the method will file. Generate a random name if it is None.

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
```diff
@@ -284,3 +284,87 @@ def create_winding(self, name=None, location=None, angle=0, use_instance_id_netl
         )
         id.set_property("Name", name)
         return id
+
+    @pyaedt_function_handler(compname="name")
```
gmalinve, line 288: @DaveTwyman why are you using the function handler? you are not using compname. This was used by us to deprecate old arguments not following python conventions. This is a copy/paste error I guess.

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
```diff
@@ -284,3 +284,87 @@ def create_winding(self, name=None, location=None, angle=0, use_instance_id_netl
...
+    def create_isin(self, name=None, value=1, location=None, angle=0, use_instance_id_netlist=False):
+        """Create a sinusoidal current source.
+
+        Parameters
+        ----------
+        name : str, optional
+            Name of the current source. The default is ``None``.
```
gmalinve, line 295:
```suggestion
            Name of the current source. The default is ``None`` in which case a random name will be generated.
```

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
```diff
@@ -284,3 +284,87 @@ def create_winding(self, name=None, location=None, angle=0, use_instance_id_netl
...
+        id = self.create_component(
+            name,
+            component_library="Sources",
+            component_name="ISin",
+            location=location,
+            angle=angle,
+            use_instance_id_netlist=use_instance_id_netlist,
+        )
+
+        id.set_property("Ia", value)
```
gmalinve, line 326: @DaveTwyman this is not correct, please debug and see what happens when you set a property that doesn't exist. You access "Ia" through parameters. you don't have to use set_property.

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
```diff
@@ -284,3 +284,87 @@ def create_winding(self, name=None, location=None, angle=0, use_instance_id_netl
...
+        id.set_property("Name", name)
+        return id
+
+    @pyaedt_function_handler(compname="name")
+    def create_vsin(self, name=None, value=1, location=None, angle=0, use_instance_id_netlist=False):
```
gmalinve, line 331: all the comments in create_isin should be applied here too.

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
```diff
@@ -284,3 +284,87 @@ def create_winding(self, name=None, location=None, angle=0, use_instance_id_netl
         )
         id.set_property("Name", name)
         return id
+
+    @pyaedt_function_handler(compname="name")
+    def create_isin(self, name=None, value=1, location=None, angle=0, use_instance_id_netlist=False):
```
gmalinve, line 289: @DaveTwyman I'd change name: create_i_sin and create_v_sin. This is more pythonic.

gmalinve (changes requested):

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
DaveTwyman: @gmalinve , This is the usage of id in the rest of the Maxwell circuit components, for the create_component and in the returns section. @SMoraisAnsys made the same comment earlier on in the history of this PR. I went with consistency throughout the library beating having two library components written in a different way to the other five. 

I'll of course go with your judgement/experience here, but let me know which of these or another you think best
1) Change id usage in all seven components now
2) Implement it differently in these two new components only
3) Implement the id usage in the two new components to be consistent with the other five in the library (As per this PR) and raise another PR for the future to adjust this (And add the remaining missing components whilst we're at it)

Personally I would prefer 3) as it lets me get on with the circuit for the Transformer Toolkit.

gmalinve: @DaveTwyman good point. I'd go for the first one, it's just a matter of renaming a variable in all methods but if you don't feel comfortable you can do it in your PR for your methods and then rename the other variables in another PR. 
For me it's just a matter of changing from "id" to "comp" in all methods (or only in yours and remember to do it afterwards for all the others).

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
DaveTwyman: @gmalinve , I'm using the function handler because the other five components in the Maxwell library use it so I've followed the same structure.

gmalinve: @DaveTwyman sorry I may have typed it wrongly. You have to use the function handler but why do you need compname?

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
DaveTwyman: Hi @gmalinve, I just double checked and this works, I couldn't get a bug to appear.
Ia is one of the ten parameters that ISin can take and the most used parameter (it's the amplitude of the sinusoidal current), it's fed into PyAEDT 'create_isin' using the 'value' parameter. I did think of renaming 'value' to something like 'IPeak', 'Current' etc but since the rest of the Maxwell circuit components in the library have a 'value' parameter I've stuck with using 'value' for consistency. The usage of '.set_property' is throughout the rest of the 'primitives_maxwell_circuit' . 
Were you suggesting another approach to 'set_property' that has some advantages?

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
DaveTwyman: @gmalinve , Yes ok I think your suggestion makes it easier to read than 'create_isin' and 'create_vsin'. I'll change this

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
DaveTwyman: @gmalinve , Good point, whilst it creates the component as expected when 'name=None' it does generate several error messages in the Message Manager window. When 'name = Newname' there are no errors. 

I see all the other components (Resistor, Inductor etc) in the same library have the same behavior when None is passed for the name. Looking at it in a bit more depth, the circuit editor schematic defaults to allow elements to be created with no name (unlike say boundaries or geometry in the 3D modeler of AEDT). I think this behavior should be looked at in more detail either creating a new random name as you've suggested or assigning no name as is the default for new components being added to Maxwell schematic.

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
DaveTwyman: Either this or create the component without a name as per default Maxwell circuit editor usage.

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
gmalinve: @DaveTwyman this is not a bug. You simply have to access the property and change its value through the "parameters" property, without using the set_property method.

DaveTwyman: Hi @gmalinve , Sorry I'm not following this, could you tell me what you would replace line 326 with and why?

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
DaveTwyman: @gmalinve , Agreed the renaming of 'id' to 'comp' is trivial using ctrl-r in pyaedt. I'll go and do this then. I think it doesn't matter what the name is returned as outside the function but I'll need to go and check this now

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
DaveTwyman: @SMoraisAnsys ,@gmalinve. Got it, I'll delete this from the two new components.

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
DaveTwyman: Renamed 'id' as 'comp_name' throughout 'primitives_maxwell_circuit' file, so marking this as resolved.

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
DaveTwyman: Marked as resolved, I've changed 'id' to 'comp_name' throughout 'primitives_maxwell_circuit'

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
DaveTwyman: I'll mark this as resolved since I'm applying the same changes to 'create_v_sin' as to 'create_i_sin'

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
DaveTwyman: I've changed this so marked as resolved

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
SMoraisAnsys: Please push your changes before marking as resolved. Currently I'm still seeing the same behavior as the one we mentionned.

SMoraisAnsys (changes requested):
Could you please use `component` as variable name instead of `comp_name` ? This could be confusing for future maintenance as it gives the impression that the variable is a string when it is not.
Also can you add a test for `create_page` ?

Thanks again for taking the time to perform all the requested changes.

DaveTwyman:
> Could you please use `component` as variable name instead of `comp_name` ? This could be confusing for future maintenance as it gives the impression that the variable is a string when it is not. Also can you add a test for `create_page` ?
> 
> Thanks again for taking the time to perform all the requested changes.

@SMoraisAnsys , Not at all, thanks for your patience whilst I work through this. 

I could do with some guidance on the unit test for the 'create_page', there is no way to check on the pages name, index or other properties, the underlying IronPython doesn't have such a feature. Are you aware of another instance in PyAEDT where we create something that cannot then be queried. I could then copy that UT. In this case there isn't a useful check that can be added to the UT. 

Should we do something like add a dummy UT, that just holds a comment?

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
DaveTwyman: The None case is now supported for both new components and the others in this library. Support for names of type String, Integer and Float is now covered as per existing Maxwell Circuit Editor usage

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
DaveTwyman: Done, Thanks

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
DaveTwyman: Hi @gmalinve , I've made all components in 'primitives_maxwell_circuit' take name parameter in same way. So they can accept None (Now without an error) and changed doc to add support for String, Integer and Float. 
This is the same behavior as in the schematic.

SMoraisAnsys:
I went through `Circuit` help PDF files and you might find something usefull with `SelectPage` and `GetNumPages`.

SMoraisAnsys:
Hi @DaveTwyman, could you please add the last changes required for this PR ? :)

DaveTwyman:
Yes, I'm planning to get back to finish that shortly
________________________________

DaveTwyman:
> Hi @DaveTwyman, could you please add the last changes required for this PR ? :)

Hello @SMoraisAnsys , I've added the last unit test, and method needed to support that test (plus another unit test for the new method). So all the new methods now have supporting unit tests. 

It seems that the test coverage codecov is very low, do you think this is because of making multiple single line changes 'id' to 'component' or issues elsewhere?
image

_unittest/test_21_Circuit.py
```diff
@@ -985,3 +985,10 @@ def test_51_import_asc(self):
         self.aedtapp.insert_design("ASC")
         asc_file = os.path.join(local_path, "example_models", test_subfolder, "butter.asc")
         assert self.aedtapp.create_schematic_from_asc_file(asc_file)
+
+    def test_52_create_page(self):
+        self.aedtapp.create_page()
+        assert self.aedtapp.GetNumPages() == 2
```
gmalinve, line 991: @DaveTwyman you are not testing your method here. Your method is in the schematic. Why are you testing GetNumPages()? Also remember to assert the failure with assert not.

_unittest/test_35_MaxwellCircuit.py
```diff
@@ -114,3 +114,11 @@ def test_08_import_netlist(self):
         self.aedtapp.insert_design("SchematicImport")
         self.aedtapp.modeler.schematic.limits_mils = 5000
         assert self.aedtapp.create_schematic_from_netlist(self.netlist_file1)
+
+    def test_09_create_voltage_source(self):
+        id = self.aedtapp.modeler.schematic.create_vsin(name="voltage_source", value=3.14, angle=90)
```
gmalinve, line 119: @DaveTwyman this method does not exist, you created create_v_sin and this doesn't return an id but rather a component.

_unittest/test_35_MaxwellCircuit.py
```diff
@@ -114,3 +114,11 @@ def test_08_import_netlist(self):
...
+        assert id.parameters["Va"] == "3.14"
+
+    def test_10_create_current_source(self):
+        id = self.aedtapp.modeler.schematic.create_isin(name="current_source", value=2.72, angle=90)
```
gmalinve, line 123: @DaveTwyman same here.

_unittest/test_35_MaxwellCircuit.py
```diff
@@ -114,3 +114,11 @@ def test_08_import_netlist(self):
         self.aedtapp.insert_design("SchematicImport")
         self.aedtapp.modeler.schematic.limits_mils = 5000
         assert self.aedtapp.create_schematic_from_netlist(self.netlist_file1)
+
+    def test_09_create_voltage_source(self):
+        id = self.aedtapp.modeler.schematic.create_vsin(name="voltage_source", value=3.14, angle=90)
+        assert id.parameters["Va"] == "3.14"
```
gmalinve, line 120: @DaveTwyman this property doesn't exist. You create it with the set_property. If you use set_property and input a property name that doesn't exist, this will be created.

src/ansys/aedt/core/modeler/circuits/primitives_circuit.py
```diff
@@ -1262,6 +1262,55 @@ def create_wire(self, points, name=""):
...
+            ``True`` when successful, ``False`` when failed.
+        References
+        ----------
+        >>> oeditor.CreatePage ()
+        Examples
+        --------
+        >>>from ansys.aedt.core import MaxwellCircuit
+        >>>app=MaxwellCircuit ()
+        >>>schematic=app.modeler.schematic
+        >>>schematic.create_page (name="NewPageName")
```
gmalinve, line 1284:
```suggestion
        >>>component=app.modeler.schematic.create_page(name=page_name)
```

src/ansys/aedt/core/modeler/circuits/primitives_circuit.py
```diff
@@ -1262,6 +1262,55 @@ def create_wire(self, points, name=""):
...
+        name : str, int or float
+            Name to be used when creating the new circuit schematic.
+        Returns
+        -------
+        bool
+            ``True`` when successful, ``False`` when failed.
+        References
+        ----------
+        >>> oeditor.CreatePage ()
+        Examples
```
gmalinve, line 1279:
```suggestion
        >>> oEditor.CreatePage()
        
        Examples
```

src/ansys/aedt/core/modeler/circuits/primitives_circuit.py
```diff
@@ -1262,6 +1262,55 @@ def create_wire(self, points, name=""):
...
+
+        Parameters
+        ----------
+        name : str, int or float
+            Name to be used when creating the new circuit schematic.
+        Returns
+        -------
+        bool
+            ``True`` when successful, ``False`` when failed.
+        References
```
gmalinve, line 1276:
```suggestion
            ``True`` when successful, ``False`` when failed.
            
        References
```

src/ansys/aedt/core/modeler/circuits/primitives_circuit.py
```diff
@@ -1262,6 +1262,55 @@ def create_wire(self, points, name=""):
...
             return False
 
+    def create_page(self, name):
+        """Add a new circuit schematic page.
+
+        Parameters
+        ----------
+        name : str, int or float
+            Name to be used when creating the new circuit schematic.
+        Returns
```
gmalinve, line 1272:
```suggestion
            Name to be used when creating the new circuit schematic.
            
        Returns
```

src/ansys/aedt/core/modeler/circuits/primitives_circuit.py
```diff
@@ -1262,6 +1262,55 @@ def create_wire(self, points, name=""):
...
+        Examples
+        --------
+        >>>from ansys.aedt.core import MaxwellCircuit
+        >>>app=MaxwellCircuit ()
+        >>>schematic=app.modeler.schematic
+        >>>schematic.create_page (name="NewPageName")
+
+        """
+        if not isinstance(name, (str, int, float)):
+            self.logger.error("Argument 'name' to 'create_page' is not of data type String, Integer or Float.")
```
gmalinve, line 1288:
```suggestion
            self.logger.error("Wrong type for argument ``name``.")
```

src/ansys/aedt/core/modeler/circuits/primitives_circuit.py
```diff
@@ -1262,6 +1262,55 @@ def create_wire(self, points, name=""):
...
+            return False
+        self.oeditor.CreatePage(name)
+        return True
+
+    def get_num_pages(self):
+        """Gets the number of circuit schematic pages.
+
+        Returns
```
gmalinve, line 1296: @DaveTwyman same comments and suggestions as previous method.

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
```diff
@@ -85,8 +85,8 @@ def create_resistor(self, name=None, value=50, location=None, angle=0, use_insta
 
         Parameters
         ----------
-        name : str, optional
-            Name of the resistor. The default is ``None``.
+        name : str, int, float or optional
+            Name of the resistor. The default is ``None`` which adds a resistor without a name.
```
gmalinve, line 89: @DaveTwyman are you sure that by default it creates a resistor without a name?

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
```diff
@@ -109,27 +109,27 @@ def create_resistor(self, name=None, value=50, location=None, angle=0, use_insta
...
             component_name="Res",
             location=location,
             angle=angle,
             use_instance_id_netlist=use_instance_id_netlist,
         )
-
-        id.set_property("R", value)
-        id.set_property("Name", name)
-        return id
+        component.set_property("R", value)
```
gmalinve, line 120: @DaveTwyman do not use the set_property on a property that doesn't exist, otherwise this will be created. You can access this property through parameters.

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
```diff
@@ -109,27 +109,27 @@ def create_resistor(self, name=None, value=50, location=None, angle=0, use_insta
...
+        if isinstance(name, (str, int, float)):
```
gmalinve, line 121: @DaveTwyman do not set the name like that, you can access the name through the component object!
Please debug!

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
```diff
@@ -161,18 +161,19 @@ def create_inductor(self, name=None, value=50, location=None, angle=0, use_insta
             use_instance_id_netlist=use_instance_id_netlist,
         )
 
-        id.set_property("L", value)
-        id.set_property("Name", name)
-        return id
+        component.set_property("L", value)
```
gmalinve, line 164: @DaveTwyman please look at my previous comments

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
```diff
@@ -203,18 +204,19 @@ def create_capacitor(self, name=None, value=50, location=None, angle=0, use_inst
             use_instance_id_netlist=use_instance_id_netlist,
         )
 
-        id.set_property("C", value)
-        id.set_property("Name", name)
-        return id
+        component.set_property("C", value)
```
gmalinve, line 207: @DaveTwyman look at my previous comments

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
```diff
@@ -244,17 +246,18 @@ def create_diode(self, name=None, location=None, angle=0, use_instance_id_netlis
             use_instance_id_netlist=use_instance_id_netlist,
         )
 
-        id.set_property("Name", name)
-        return id
+        if isinstance(name, (str, int, float)):
```
gmalinve, line 249: @DaveTwyman look at my previous comments

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
```diff
@@ -274,13 +277,98 @@ def create_winding(self, name=None, location=None, angle=0, use_instance_id_netl
...
             name,
             component_library="Dedicated Elements",
             component_name="Winding",
             location=location,
             angle=angle,
             use_instance_id_netlist=use_instance_id_netlist,
         )
-        id.set_property("Name", name)
-        return id
+        if isinstance(name, (str, int, float)):
```
gmalinve, line 288: @DaveTwyman please look at my previous comments

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
```diff
@@ -274,13 +277,98 @@ def create_winding(self, name=None, location=None, angle=0, use_instance_id_netl
...
+        component = self.create_component(
+            name,
+            component_library="Sources",
+            component_name="ISin",
+            location=location,
+            angle=angle,
+            use_instance_id_netlist=use_instance_id_netlist,
+        )
+
+        component.set_property("Ia", value)
```
gmalinve, line 329: @DaveTwyman please look at my previous comments

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
```diff
@@ -274,13 +277,98 @@ def create_winding(self, name=None, location=None, angle=0, use_instance_id_netl
...
+        component = self.create_component(
+            name,
+            component_library="Sources",
+            component_name="VSin",
+            location=location,
+            angle=angle,
+            use_instance_id_netlist=use_instance_id_netlist,
+        )
+
+        component.set_property("Va", value)
```
gmalinve, line 371: @DaveTwyman please look at my previous comments

gmalinve (changes requested):
@DaveTwyman please review and fix all the comments.
You created unit tests that don't run at all. Please test, run and debug locally before pushing commits to your PR.

_unittest/test_35_MaxwellCircuit.py
DaveTwyman: @gmalinve , 'create_vsin' was an earlier iteration of the method name before 'create_v_sin' was used. We also decided to move from id to component. I'll adjust all these changes then mark as resolved after I've pushed changes out.

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
DaveTwyman: @gmalinve , Yes that is the default and intended behavior for all the Maxwell Circuit Editor Components.

_unittest/test_35_MaxwellCircuit.py
DaveTwyman: @gmalinve I'm sure all usages of id have already been replaced with component. Is this an older branch you've reviewed perhaps?

gmalinve: @DaveTwyman what do you mean an older branch? this is your branch.

src/ansys/aedt/core/modeler/circuits/primitives_circuit.py
```diff
@@ -1262,6 +1262,57 @@ def create_wire(self, points, name=""):
...
+        -------
+        int
+            The number of pages in the circuit schematic.
+        References
+        ----------
+        >>> oeditor.GetNumPages ()
+        Examples
+        --------
+        >>>from ansys.aedt.core import MaxwellCircuit
+        >>>app=MaxwellCircuit ()
```
gmalinve, line 1308: @DaveTwyman please remove all the whitespaces between the method name and the parenthesis. Please check all these occurrences in all your changes.

_unittest/test_35_MaxwellCircuit.py
DaveTwyman: @gmalinve , I should have wrote 'older commit' not 'older branch', I checked and id is replaced with component everywhere.

gmalinve: @DaveTwyman please look up, fix the tests where there is id and not component. You change from id to component only in the code, not in the tests.

_unittest/test_35_MaxwellCircuit.py
DaveTwyman: @gmalinve , Got it id is changed to component in the unit tests

src/ansys/aedt/core/modeler/circuits/primitives_circuit.py
DaveTwyman: @gmalinve , Yes this is clearer.

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
DaveTwyman: @gmalinve set_property replaced with parameters

_unittest/test_21_Circuit.py
DaveTwyman: Changed as requested with more testing of name types added

src/ansys/aedt/core/maxwellcircuit.py
```diff
@@ -242,3 +242,53 @@ def export_netlist_from_schematic(self, output_file):
...
+        >>>app=MaxwellCircuit()
+        >>>component=app.create_page(name="page_name")
+        """
+        if not isinstance(name, (str, int, float)):
+            self.logger.error("Wrong type for argument ``name``.")
+            return False
+        self.oeditor.CreatePage(name)
+        return True
+
+    def get_num_pages(self):
```
SMoraisAnsys, line 275: I think it would be better to add `nb_pages` as a property instead of adding this feature as a method.

src/ansys/aedt/core/modeler/circuits/primitives_circuit.py
```diff
@@ -1262,6 +1262,57 @@ def create_wire(self, points, name=""):
...
+        >>>component=app.modeler.schematic.create_page(name=page_name)
+
+        """
+        if not isinstance(name, (str, int, float)):
+            self.logger.error("Wrong type for argument ``name``.")
+            return False
+        self.oeditor.CreatePage(name)
+        return True
+
+    def get_num_pages(self):
```
SMoraisAnsys, line 1295: I think it would be better to add `nb_pages` as a property instead of adding this feature as a method.

tests/system/general/test_35_MaxwellCircuit.py
```diff
@@ -115,3 +115,27 @@ def test_08_import_netlist(self):
...
+            assert self.component.parameters["Name"] == item
+        assert self.component.parameters["Ia"] == 2.72
+
+    def test_11_create_page(self):
+        assert self.aedtapp.create_page("string_test") == True
+        assert self.aedtapp.create_page(123) == True
+        assert self.aedtapp.create_page(3.14) == True
+        assert not self.aedtapp.create_page(["create_page_test"])
+
+    def test_12_get_num_pages(self):
```
SMoraisAnsys, line 139: Could you please add a test on the return value and not only on the value type ?

SMoraisAnsys (changes requested):
I think it would be better to add the get number pages feature as a property, could you update the code ?
Also, the added code in `primitives_circuit.py` is not tested, could you test it ?
I left an extra minor comment.

src/ansys/aedt/core/maxwellcircuit.py
DaveTwyman: @SMoraisAnsys , That would be a better implementation. I'll make that switch to a parameter

SMoraisAnsys: Not exactly, I had in mind:

```python
@property
def nb_pages(self):
    return self.oeditor.CreatePage(name)
```

tests/system/general/test_35_MaxwellCircuit.py
DaveTwyman: The IronPython method for 'CreatePage' returns only the default None when run.
I don't think it's possible to return other attributes such as the created page name from a method that doesn't have these attributes.(Or it may well be something I've not figured out yet in python...)

Similar methods such as 'Selectpage' and 'DeletePage' similary return None and not the page name.
I've added the nb_pages property after every page is created in the unit test.
So we are at least checking these pages are created whilst not being able to check the names of them.

SMoraisAnsys:
Hey @DaveTwyman, I'm not sure about the status of this PR. Do you want to add VSin and ISin on top of the pages features or should this be splited into two PRs and some commits got added inadvertently ?
About the pages features, can you add a test to `create_page` on `CircuitComponents` class ? Also, would it make sense to add `nb_pages` to this class ?

tests/system/general/test_35_MaxwellCircuit.py
```diff
@@ -115,3 +115,26 @@ def test_08_import_netlist(self):
         self.aedtapp.insert_design("SchematicImport")
         self.aedtapp.modeler.schematic.limits_mils = 5000
         assert self.aedtapp.create_schematic_from_netlist(self.netlist_file1)
+
+    def test_09_create_voltage_source(self):
+        name_type = ["voltage_source", 123, 1.23]
+        for item in name_type:
+            self.component = self.aedtapp.modeler.schematic.create_v_sin(name=item, value=3.14, angle=90)
```
gmalinve, line 122:
```suggestion
            component = self.aedtapp.modeler.schematic.create_v_sin(name=item, value=3.14, angle=90)
```
@DaveTwyman why self.component? it's not a property of the test class. Please remove all the other occurances.

gmalinve (changes requested):

src/ansys/aedt/core/circuit.py
```diff
@@ -205,6 +205,11 @@ def _get_number_from_string(self, stringval):
         except Exception:
             return from_rkm_to_aedt(value)
 
+    @property
+    def nb_pages(self):
```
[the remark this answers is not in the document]
DaveTwyman: @gmalinve , Thanks for reviewing this PR. I'll move these two to the schematic.

tests/system/general/test_35_MaxwellCircuit.py
DaveTwyman: @gmalinve , Good point, with hindsight there was a lot wrong with test_35.

DaveTwyman:
@gmalinve, @SMoraisAnsys , I've made all the requested changes to this PR. 
Can you let me know what is now needed to merge it to the main.
All checks have passed apart from 2skipped (Release Project and Upload Release Documentation)

I've just pressed the Update branch button as this branch was not out of date

Thanks

DaveTwyman:
> @gmalinve, @SMoraisAnsys , I've made all the requested changes to this PR. Can you let me know what is now needed to merge it to the main. All checks have passed apart from 2skipped (Release Project and Upload Release Documentation)
> 
> I've just pressed the Update branch button as this branch was not out of date
> 
> Thanks

@gmalinve, @SMoraisAnsys
As per this mornings Office Hours call, creation of a new class in the Schematic.py would be needed to keep track of the page number. I've added this work to this PR

tests/system/general/test_35_MaxwellCircuit.py
```diff
@@ -115,3 +115,26 @@ def test_08_import_netlist(self):
...
+        assert component.parameters["Ia"] == 2.72
+
+    def test_11_create_page(self):
+        assert self.aedtapp.modeler.create_page("string_test")
+        assert self.aedtapp.modeler.nb_pages == 2
+        assert self.aedtapp.modeler.create_page(123)
+        assert self.aedtapp.modeler.nb_pages == 3
+        assert self.aedtapp.modeler.create_page(3.14)
+        assert self.aedtapp.modeler.nb_pages == 4
+        assert not self.aedtapp.modeler.create_page(["create_page_test"])
```
anur7, line 140: @DaveTwyman is there any reason why are you testing create_page giving as argument different type of data types? I think the code lines 134 and 140 are enough to test create_page.

DaveTwyman: @anur7 , The other lines are there to test that passing other data types works correctly. As there is no AEDT documentation on page name data types, we've restricted it to String, Integer and Float and just test these types.

src/ansys/aedt/core/modeler/schematic.py
```diff
@@ -477,6 +477,42 @@ def _arg_with_dim(self, value, units=None):
...
+
+        """
+        if not isinstance(name, (str, int, float)):
+            self.logger.error("Wrong type for argument ``name``.")
+            return False
+        self.oeditor.CreatePage(name)
+        return True
+
+    @property
+    def nb_pages(self):
```
SMoraisAnsys, line 512: This is not tested, could you add a test ?

src/ansys/aedt/core/modeler/circuits/primitives_maxwell_circuit.py
SMoraisAnsys, line 121: Just wondering, why is this change required ? Was `id.set_property("Name", name)` not working as expected ?

PR #5272, opened 2024-10-10
CHORE: Use f-string instead of format

gmalinve:

PR #5297, opened 2024-10-15
FEAT: New class ReportPlotter to handle all matplotlib plot

maxcapodi78:
New class to handle all matplotlib methods and options


gmalinve:
@maxcapodi78 @Samuelopez-ansys please avoid using .format() in strings. I just refactor all the occurrences in order to be compatible with f-string. (#5272).

Samuelopez-ansys (approved):
This is a game changer

PR #5315, opened 2024-10-21
CI: Bump ansys actions version

SMoraisAnsys:
As title say.

We've been left behind with ansys/actions due to various reasons - the biggest one being the artifact compatibility issue.
Now that all the issues that slowed us down are solved, we can update our CICD !

Notable changes:

- To be compatible with `ansys/actions/doc-build` I'm removing the name of the pdf in `conf.py`. This should lead to the creation of a pdf named `pyaedt.pdf`. In the process we lose the version in the pdf file name but this is a requirement for the `doc-build` action. I don't think this is a problem though.
- Meilisearch is being deprecated so I dropped the indexing jobs. I'll update our documentation to remove meilisearch and leverage static search in another PR.


MaxJPRey:
`In the process we lose the version in the pdf file name but this is a requirement for the doc-build action. I don't think this is a problem though.`
This is absolutely fine. Thanks @SMoraisAnsys

Samuelopez-ansys (approved):

PR #5319, opened 2024-10-21
REFACTOR: Regular code improvement

SMoraisAnsys:
Changes fixing issues raised by bandit.

### DIFF
--- a/src/ansys/aedt/core/visualization/report/common.py
+++ b/src/ansys/aedt/core/visualization/report/common.py
@@ -578,7 +578,7 @@ class CommonReport(object):
             try:
                 self._props["context"]["polyline"] = self.traces[0].properties["Geometry"]
             except Exception:
-                pass
+                self._post._app.logger.debug("Something went wrong while processing polyline.")
         return self._props["context"].get("polyline", None)
 
     @polyline.setter
@@ -1133,7 +1133,7 @@ class CommonReport(object):
                         variations[tr.properties["Secondary Sweep"]] = ["All"]
                 self._props["context"]["variations"] = variations
             except Exception:
-                pass
+                self._post._app.logger.debug("Something went wrong while processing variations.")
         return self._props["context"]["variations"]
 
     @variations.setter
@@ -1292,7 +1292,7 @@ class CommonReport(object):
             try:
                 return self.traces[0].properties["Domain"]
             except Exception:
-                pass
+                self._post._app.logger.debug("Something went wrong while accessing trace's Domain property.")
         return self._props["context"]["domain"]
 
     @domain.setter
--- a/src/ansys/aedt/core/visualization/advanced/misc.py
+++ b/src/ansys/aedt/core/visualization/advanced/misc.py
@@ -132,7 +132,9 @@ def convert_nearfield_data(dat_folder, frequency=6, invert_phase_for_lower_faces
                     real.append(line[3])
                     imag.append(line[4])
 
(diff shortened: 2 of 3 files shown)


src/ansys/aedt/core/visualization/report/common.py
```diff
@@ -578,7 +578,7 @@ def polyline(self):
             try:
                 self._props["context"]["polyline"] = self.traces[0].properties["Geometry"]
             except Exception:
-                pass
+                self._post._app.logger.debug("Something went wrong while processing polyline.")
```
Alberto-DM, line 581: Following the EAFP principle, some parts of the code have this `try: except: pass` clause instead of an `if else` check, because is faster. If we put a log entry for every Exception hit, we might flood the logger. 
Although I haven't check in detail this part of the code, I rise this concern: are you sure that this is an Exception worth to be logged?

src/ansys/aedt/core/visualization/report/common.py
```diff
@@ -1133,7 +1133,7 @@ def variations(self):
                         variations[tr.properties["Secondary Sweep"]] = ["All"]
                 self._props["context"]["variations"] = variations
             except Exception:
-                pass
+                self._post._app.logger.debug("Something went wrong while processing variations.")
```
Alberto-DM, line 1136: Same comment as before about EAFP principle.

src/ansys/aedt/core/visualization/report/common.py
```diff
@@ -1292,7 +1292,7 @@ def domain(self):
             try:
                 return self.traces[0].properties["Domain"]
             except Exception:
-                pass
+                self._post._app.logger.debug("Something went wrong while accessing trace's Domain property.")
```
Alberto-DM, line 1295: Same comment as before about EAFP principle.

Alberto-DM (dismissed):

src/ansys/aedt/core/visualization/report/common.py
SMoraisAnsys: I understand your point but the pattern `try: ... except: pass` is generally not recommended. Indeed, this pattern:
- swallows all exceptions, unless you add the type of exception you want to handle but that's not the case here;
- hides failures as there is no indication or feedback on what went wrong, which makes it harder to detect and fix a problem.

Here, I'm just logging at debug level so this shouldn't hinder too much unless you want to log at debug level. And, in that case, I feel like having information would help solving the problem even if the log are more verbose.

Finally, we have to use `bandit` and `safety` to upgrade our vulnerability handling. The pattern `try: ... except: pass` is reported by `bandit` as an issue of "low" severity but still an issue.

@Samuelopez-ansys @maxcapodi78 @gmalinve What are your thoughts on this ?

gmalinve: @Alberto-DM @SMoraisAnsys 
I agree with Sebastien. 
If we know upfront that this part of the code is not intended to be repeated until it suceeds, where passing is good enough, then I agree on the logging regardless.

MaxJPRey: That is correct. We should avoid the try/except/pass when possible especially as it can have serious side effects.
With try/except we have to catch and handle specific exceptions as much as possible.

@Alberto-DM's concerns about performance makes total sense and is a very valid point.
So here, it is a good practice to only use the debug mode for the logger to not overload the logger as mentioned.
It could be a pattern that we keep in mind for other similar case.

PR #5321, opened 2024-10-21
FIX: typo in output_dir argument of method hfss\validate_full_design

Alberto-DM:
Fixed typo in 'output_dir' argument of method hfss\validate_full_design


SMoraisAnsys (approved):
Nice catch !

Samuelopez-ansys (approved):

PR #5330, opened 2024-10-22
FIX:  Eigenmode Edit source

### LINKED ISSUE #5322: Add `edit_source` / `edit_sources` functionality for Eigenmode solutions

Include the ability to edit the source distribution for Eigenmode-type solutions, as is possible in the GUI via Field Overlays -> Edit Sources... option for a simulation with this solution type

Under `hfss.py`, modify `edit_sources` to have particular functionality given the `hfss.solution_type` parameter is "Eigenmode". Specifically, format the `setting` object for the desired eigenmode setup (I would do this myself but I am struggling with finding/guessing the proper format for this). 

_No response_

Samuelopez-ansys:
Close #5322 

Edit sources in Eigenmode was not working.

I also merged functionalities of edit_source/edit_sources and edit_source_from_file, which were similar, and we should not maintain both, so I added a deprecation.


src/ansys/aedt/core/hfss.py
```diff
@@ -3803,59 +3811,93 @@ def edit_sources(
...
+                        phase.append("0deg")
+            if eigenmode_stored_energy:
+                setting = ["Name:=", "Modes", "Magnitudes:=", power]
+            else:
+                setting = ["Name:=", "Modes", "Magnitudes:=", power, "Phases:=", phase]
 
-        args = [argument]
-        args.extend(setting)
+            args = [argument]
+            args.extend([setting])
```
SMoraisAnsys, line 3890: Could settings be defined as a list of list of every cases, i.e. for "Eigenmode" and non "Eigenmode"?
This way we could move `args = [argument]` and `args.extend(settings)` at the same level as `self.osolution.EditSources(args)`. I think this would avoid code duplicate and help for future maintenance / readability.

src/ansys/aedt/core/hfss.py
```diff
@@ -3894,6 +3936,7 @@ def edit_source(self, assignment=None, power="1W", phase="0deg"):
         True
 
         """
+        warnings.warn("`edit_sources` is deprecated. Use `edit_sources` method instead.", DeprecationWarning)
```
SMoraisAnsys, line 3940: Since it is deprecated, shouldn't this method be implemented using `edit_sources` instead ? Like you did for `edit_sources_from_file`

SMoraisAnsys (commented):
Great job here. I left a two minor comments.

src/ansys/aedt/core/hfss.py
Samuelopez-ansys: Done, the problem of this method is that it was not covering all the combination of edit sources

SMoraisAnsys (approved):
LGTM

PR #5331, opened 2024-10-22
FIX: Remove the clipboard isolation env variable in the kernel_converter script

Alberto-DM:
Remove the clipboard isolation env variable if present. 
Setting the env variable "ANS_USE_ISOLATED_CLIPBOARD" prevents the converter from working.

### DIFF
--- a/src/ansys/aedt/core/workflows/project/kernel_converter.py
+++ b/src/ansys/aedt/core/workflows/project/kernel_converter.py
@@ -283,6 +283,13 @@ def convert(args):
         files_path += search_files(args["file_path"], "*.aedt")
     else:
         files_path = [args["file_path"]]
+    # Remove the clipboard isolation env variable if present
+    env_var_name = "ANS_USE_ISOLATED_CLIPBOARD"
+    saved_env_var = None
+    if env_var_name in os.environ:
+        saved_env_var = os.environ[env_var_name]
+        del os.environ[env_var_name]
+
     output_desktop = Desktop(
         new_desktop=True,
         version=version,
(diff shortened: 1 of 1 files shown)


src/ansys/aedt/core/workflows/project/kernel_converter.py
```diff
@@ -283,6 +283,13 @@ def convert(args):
         files_path += search_files(args["file_path"], "*.aedt")
     else:
         files_path = [args["file_path"]]
+    # Remove the clipboard isolation env variable if present
+    env_var_name = "ANS_USE_ISOLATED_CLIPBOARD"
+    saved_env_var = None
+    if env_var_name in os.environ:
+        saved_env_var = os.environ[env_var_name]
+        del os.environ[env_var_name]
```
MaxJPRey, line 291:
```suggestion
    saved_env_var = os.environ.pop(env_var_name, None) 
```

MaxJPRey: This should be equivalent. To double check.

SMoraisAnsys (commented):
IMO we shouldn't change the user's environment variable to make something work.
What we could do is log a warning / error to notify the user instead right ?
This way, he'll be able to know the impact of such environment variable on the workflow and deactivate it if needed.

SMoraisAnsys:
> > IMO we shouldn't change the user's environment variable to make something work. What we could do is log a warning / error to notify the user instead right ? This way, he'll be able to know the impact of such environment variable on the workflow and deactivate it if needed.
> 
> We set env variables on multiple occasions to make PyAEDT work. We don't change the global system environment variables. We are going to change only the variables of that specific python session where the script is run. In fact, restoring it is an extra precaution, that possibly makes a difference only if the script is used in a larger program. But your suggestion made me realize that what I did will not solve the problem. Since this code is an extension, it is executed from an existing AEDT session, opened manually by the user, under the current system environment variables. So, changing the variable like I did, is too late. I will follow your suggestion, but for a different reason ;)

Glad to have asked then haha

Samuelopez-ansys (approved):

PR #5338, opened 2024-10-23
CHORE: remove random int generators

gmalinve:


gmalinve:
@SMoraisAnsys @Samuelopez-ansys I didn't find any more occurances of random.randint() except something in rpyc_services that I didn't touch because it's in a `if is_ironpython` clause, so it will be removed anyway.
There are some occurances in python_optimizers.py but the method is called from numpy, for example:
`np.random.randint(self.var_bound[i][0], self.var_bound[i][1] + 1)`
so I don't know if this one needs to be refactored too. Any comment?

SMoraisAnsys (approved):
LGTM.
@gmalinve I think `numpy.random.randit` is not a problem because AFAIK it is cryptographically secure (which is not the case for `random.randint`)

PR #5348, opened 2024-10-28
CI: Remove needs to doc-build

SMoraisAnsys:
Fix error introduced while refactoring the CICD to use recent ansys/actions.


Samuelopez-ansys (approved):

MaxJPRey (approved):
LGTM.

PR #5352, opened 2024-10-28
FEAT: Circuit current probe

boyang2022:


Samuelopez-ansys:
Hi @boyang2022 , you missed to test when you do not define the location. Thank you for the pull request!

SMoraisAnsys (changes requested):
Hey @boyang2022 , thanks for the contribution !

This changes seem to be copy pasted from `create_voltage_probe` (which isn't a problem!). Could you refactor that to have a single method with an additional argument to distinguish between voltage and current and which would contain the logic, e.g. `__create_probe` ?

That way, both `create_voltage_probe` and `create_current_probe` could call that method behind the scene and that would help us to maintain both at the same time.

src/ansys/aedt/core/modeler/circuits/primitives_nexxim.py
```diff
@@ -850,6 +850,49 @@ def create_voltage_probe(self, name=None, location=None, angle=0, use_instance_i
...
         return cmpid
 
+    @pyaedt_function_handler(probe_name="name")
+    def create_current_probe(self, name=None, location=None, angle=0, use_instance_id_netlist=False):
+        """Create a current probe.
+
+        Parameters
+        ----------
+        name :
+            Name of the current probe. The default is ``None``.
```
MaxJPRey, line 860: Not sure there is a default value when looking at the code below.
```suggestion
            Name of the current probe.
```

boyang2022: There could be probes, similar to sources. It's better to keep them separately.

SMoraisAnsys: @boyang2022 I'm not sure about which comment you are answering to. Could you explain this statement ?

Samuelopez-ansys (approved):
LGTM

Samuelopez-ansys (approved):

SMoraisAnsys (approved):
LGTM

Samuelopez-ansys (approved):

PR #5363, opened 2024-10-30
FIX: The open export page fix for unit tests

ramin4667:
This PR updates the export page unit tests to skip opening the export page before individual tests. Also, updates the part library file to default values to pass the tests.


myoung301 (approved):
Looks good!

PR #5367, opened 2024-11-01
FIX: close UI by click X should not trigger parametrization

hui-zhou-a:
- Release desktop as soon as necessary information is extracted from active design
- The old implementation creates a new parametrized design even the user click X to close the window


SMoraisAnsys (approved):
LGTM

PR #5372, opened 2024-11-04
BUILD: Update ansys theme version

### LINKED ISSUE #5316: DOCS: Remove meilisearch feature
Meilisearch is being deprecated due to multiple reasons. One of the biggest concern was how the service it could break for an unexpected amount of time and break our CICD workflows.

We should move to the new approach based on static search. See https://github.com/ansys/ansys-sphinx-theme/pull/487

SMoraisAnsys:
Changes include removing deprecated options (meiliserach related) and adding new options related to static search.

Close #5316

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -76,9 +76,8 @@ dotnet = [
     "dotnetcore2==3.1.23; platform_system=='Linux'",
     "pywin32>=303; platform_system=='Windows'",
 ]
-
 doc = [
-    "ansys-sphinx-theme>=0.10.0,<1.1",
+    "ansys-sphinx-theme>=1.0.0,<1.3",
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -76,9 +76,8 @@ dotnet = [
     "dotnetcore2==3.1.23; platform_system=='Linux'",
     "pywin32>=303; platform_system=='Windows'",
 ]
-
 doc = [
-    "ansys-sphinx-theme>=0.10.0,<1.1",
+    "ansys-sphinx-theme>=1.0.0,<1.3",
```
MaxJPRey, line 80: Thanks for updating the range.

SMoraisAnsys: That was on my TODO list since a few weeks :')

PR #5374, opened 2024-11-04
FIX: Assign impedance to sheet in Eigenmode

### LINKED ISSUE #5373: Bug: Duplicated method and not working in Eigenmode

Samuelopez-ansys:
Close #5373


src/ansys/aedt/core/hfss.py
```diff
@@ -3508,75 +3508,6 @@ def assign_lumped_rlc_to_sheet(
             return self._create_boundary(name, props, "Lumped RLC")
         return False
 
-    @pyaedt_function_handler(sheet_name="assignment", sourcename="name", is_infground="is_inifinite_ground")
-    def assign_impedance_to_sheet(self, assignment, name=None, resistance=50, reactance=0, is_inifinite_ground=False):
```
MaxJPRey, line 3512: Just to be sure. We won't create any breaking changes with this.
Maybe it was long time due.

Samuelopez-ansys: @MaxJPRey  We had the method name duplicated, I reviewed both methods, and this one is like an old version of the remaining one. In addition, it was missed Eigenmode HFSS Solution type

PR #5376, opened 2024-11-04
FEAT: Expose port properties on Components3dLayout

### LINKED ISSUE #5375: Expose reference offset and PEC reference sheet size properties for relevant component types

Currently, the "port properties" on a component (reference offset height, reference auto creation, and reference X/Y dimensions) can only be set using `Components3DLayout.set_die_type` but this is only applicable to components of type IC while components of types Other and IO also have these properties.

Note from end user:

> We’re using PyAEDT to create circuit ports on solderballs in Hfss3dlayout. The component is either of type “IC” or “Other”. Either way, since it’s a circuit port, we do not want PEC on top of it. The way to remove PEC as I understand to uncheck Reference size Auto then set both X/Y to 0.
>
> I want to do this in HFSS 3D Layout, so I’m looking into PyAEDT documentation. Components3DLayout.set_solderball() doesn’t work as it only has reference_offset as argument, and I do use this function to create solderballs (but it creates the PEC). Components3DLayout.set_die_type() seems to be able to do what I want, but it only works for type “IC” but not type “Other”. I’m running with AEDT 24R2.

_No response_

_No response_

isaacwaldron:
1. Expose getters for die_properties (die type, orientation, height, and a reserved property) and port_properties (reference offset height, reference size auto size, reference size X dimension, reference size Y dimension)
2. Expose setter for port_properties
3. Update ANSYS-HSD_V1 test project to include die properties set on IC component U10 to support testing of port_properties setter

Closes #5375


tests/system/general/test_01_3dlayout_edb.py
```diff
@@ -420,3 +421,24 @@ def test_23_dissolve_element(self):
     @pytest.mark.skipif(config["desktopVersion"] <= "2024.1", reason="Introduced in 2024R1")
     def test_24_open_ic_mode_design(self):
         assert self.ic_mode_design.ic_mode
+
```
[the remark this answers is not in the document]
isaacwaldron: This seems to have resolved itself after all the workflow steps completed.

src/ansys/aedt/core/modeler/pcb/object_3d_layout.py
```diff
@@ -534,6 +537,78 @@ def enabled(self, status):
...
+        Tuple[int, int, str, int]
+            Tuple of die type (``0`` for None, ``1``, for FlipChip, or ``2`` for WireBond), die orientation (``0`` for
+            Chip Top or ``1`` for Chip Bottom), die height as a string, and a reserved property as an integer.
+        """
+        if self._part_type_id != 4:
+            return None
+        return _get_die_properties(self._get_model_info())
+
+    @property
+    def _has_port_properties(self) -> bool:
```
SMoraisAnsys, line 555: Could you please make this a private method (rename into `__has_port_properties`) instead of a property ? I don't see the point of making it a "protected" property. If there is, please let me know.

src/ansys/aedt/core/modeler/pcb/object_3d_layout.py
```diff
@@ -784,6 +833,15 @@ def set_solderball(
...
 
+    def _get_model_info(self):
+        props = self._oeditor.GetComponentInfo(self.name)
+        model_info = ""
+        for p in props:
+            if "PortProp(" in p:
+                model_info = p
+                break
+        return model_info
+
```
SMoraisAnsys, line 844: I like that you extracted lines of codes into sub methods ! Could you make this method private ?
Could you make this method a static methods of the class though ? I don't think they will be used in a different context than this class. Also, could you make them private if possible ?

src/ansys/aedt/core/modeler/pcb/object_3d_layout.py
```diff
@@ -2301,3 +2359,34 @@ def remove(self):
...
+    rsy = "0"
+    rsa = True
+    rh = "0"
+    if m:
+        rh = m.group(1)
+        rsx = m.group(3)
+        rsy = m.group(4)
+        if m.group(2) == "false":
+            rsa = False
+    return rh, rsa, rsx, rsy
```
SMoraisAnsys, line 2392: Following previous comment, could you make these methods static methods of the class  and make them private if possible ?

tests/system/general/test_01_3dlayout_edb.py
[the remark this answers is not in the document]
SMoraisAnsys: > This seems to have resolved itself after all the workflow steps completed.

Indeed, codecov leverages the feedback from multiple tests jobs so the coverage might not be accurate as long as the tests are not passer :)

SMoraisAnsys (commented):
Great enhancement ! I can't state if the inner logic should be updated but I left minor comments on the code. Thanks for the PR

src/ansys/aedt/core/modeler/pcb/object_3d_layout.py
isaacwaldron: I can make this private but I don't think it can be static as it does depend the instance's `self.name` property.

isaacwaldron:
@SMoraisAnsys I have incorporated your feedback.  I did leave `__has_port_properties` as a property and left `__get_model_info` as non-static as noted in my comments.  I also made the existing protected property private instead.

src/ansys/aedt/core/modeler/pcb/object_3d_layout.py
SMoraisAnsys: Yeah my bad, I did the comment while writting the following one (where the methods can be made static).
Just make it private :)

src/ansys/aedt/core/modeler/pcb/object_3d_layout.py
```diff
@@ -500,7 +534,7 @@ def part_type(self):
         return self._oeditor.GetPropertyValue("BaseElementTab", self.name, "Part Type")
 
     @property
-    def _part_type_id(self):
+    def __part_type_id(self):
```
SMoraisAnsys, line 537: Can you remove the property decorator and make calls with parentheses as you mentioned before ?

src/ansys/aedt/core/modeler/pcb/object_3d_layout.py
isaacwaldron: Per your other comment I will convert both of these properties (new and existing) to methods.

src/ansys/aedt/core/modeler/pcb/object_3d_layout.py
isaacwaldron: Done.

src/ansys/aedt/core/modeler/pcb/object_3d_layout.py
isaacwaldron: Done.

src/ansys/aedt/core/modeler/pcb/object_3d_layout.py
isaacwaldron: Done.

src/ansys/aedt/core/modeler/pcb/object_3d_layout.py
isaacwaldron: Done.

SMoraisAnsys (approved):
LGTM but I'd like to have the feedback of someone else before merging.
@Samuelopez-ansys @gmalinve please, can you have a look ? Or could you ping someone else that cool do it ? :)

maxcapodi78 (approved):

PR #5378, opened 2024-11-05
REFACTOR: General code improvements

SMoraisAnsys:
Changes include:
- use of `defusedxml` to safely parse an XML file;
- use of `secrets` to generate random values between 0 and 255;
- raise `RuntimeError` instead of using `assert` in code.


Samuelopez-ansys (approved):
LGTM

PR #5386, opened 2024-11-06
FEAT: Update wireframe method

Samuelopez-ansys:
Useful method to update the same property of multiple objects. If you have multiple objects, this is much more efficient that the current method, which is looping every object and change the property value one by one.


src/ansys/aedt/core/modeler/cad/primitives.py
```diff
@@ -6280,6 +6281,82 @@ def update_object(self, assignment):
...
+            self.logger.error("Invalid property name.")
+            return False
+
+        # Retrieve property settings
+        property_name = property_mapping[property_key]["property_name"]
+        reset_attr = property_mapping[property_key]["reset_attr"]
+
+        # Handle special cases for material
+        if property_key == "material_name" and isinstance(value, str):
+            matobj = self._materials.checkifmaterialexists(value)
```
SMoraisAnsys, line 6329: This method name is horrible (`checkifmaterialexist`) we should rename it into something like `exists_material` and deprecated the old name.

src/ansys/aedt/core/modeler/cad/primitives.py
```diff
@@ -6280,6 +6281,82 @@ def update_object(self, assignment):
...
+            if isinstance(value, tuple) or isinstance(value, list):
+                R = clamp(value[0], 0, 255)
+                G = clamp(value[1], 0, 255)
+                B = clamp(value[2], 0, 255)
+                value_command = ["R:=", str(R), "G:=", str(G), "B:=", str(B)]
+            else:
+                self.logger.error("Invalid color.")
+                return False
+
+        # Reset property values
```
SMoraisAnsys, line 6349: Is there a reason for resetting the property values before changing them with `_change_geometry_property` ?

tests/system/general/test_08_Primitives3D.py
```diff
@@ -2035,3 +2035,39 @@ def test_94_create_equationbased_surface(self):
             x_uv="(sin(_v*2*pi)^2+1.2)*cos(_u*2*pi)", y_uv="(sin(_v*2*pi)^2+1.2)*sin(_u*2*pi)", z_uv="_v*2"
         )
         assert surf.name in self.aedtapp.modeler.sheet_names
+
+    def test_95_update_geometry_property(self):
```
SMoraisAnsys, line 2039: Can you extend code coverage ?

SMoraisAnsys (approved):
LGTM It will be usefull for sure !
I left some minor comments.

maxcapodi78 (approved):
Looks great

PR #5392, opened 2024-11-08
FIX: Pyaedt installer in linux

### LINKED ISSUE #5390: Cannot install with pyaedt_installer_from_aedt.py on Linux

### Description of the bug

While the script works correctly in Windows, the added disclaimer does not behave correctly. Instead of forwarding an `input` in the shell, there are no interaction allowed and the script simply fails.

### Steps To Reproduce

Install in Linux environment.

Linux

3.10

### Installed packages

Not relevant

SMoraisAnsys:
As title says.

Changes include:
- using IronPython to create the disclaimer and abort actions if users do not approve the disclaimer;
- strengthening the script by using `subprocess.call` or `subprocessdotnet.call` instead of `os.system` to avoid the vulnerability associated to the latter.

@Samuelopez-ansys I tested it with AEDT 24.2 on Linux and Windows. The installation through the GUI is working as expected.
Do you want to use the same kind of windows when working with the installation of extensions ?

Close #5390


Samuelopez-ansys:
@SMoraisAnsys Good job! Maybe it is a good idea to add it to the extension manager too

PR #5394, opened 2024-11-08
FIX: Allow modeler object chamfer type "Left Distance-Angle"

cathaychris:
There was a typo in the chamfer operation that assigned the angle to the "LeftDistance" property instead of the "Angle" property when chamfer type 2 (Left Distance-Angle) was used. There was also a typo that prevented type 2 from being accessed through the API.

### DIFF
--- a/src/ansys/aedt/core/modeler/cad/object_3d.py
+++ b/src/ansys/aedt/core/modeler/cad/object_3d.py
@@ -2042,9 +2042,9 @@ class Object3d(object):
             vArg2.append("ChamferType:="), vArg2.append("Left Distance-Right Distance")
         elif chamfer_type == 2:
             vArg2.append("Angle:="), vArg2.append(str(angle) + "deg")
-            vArg2.append("ChamferType:="), vArg2.append("Left Distance-Right Distance")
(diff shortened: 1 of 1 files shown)


SMoraisAnsys:
Thanks for catching this issue @cathaychris. Due to some restrictions we have for testing, we won't be able to use the current PR to add your changes in the main branch. We'll create a new PR and cherry-pick your commits so that the CICD can run with the correct credentials. Note that you'll still be the authors of those changes, we'll only by the commiters.

SMoraisAnsys:
While the changes make sense on the paper, they seem to behave strangely in practice. I'm trying things and will let you know when I've got something working as expected.
Thanks again @cathaychris for your contribution !

cathaychris:
Hi @SMoraisAnsys 
Thanks for having a look. I agree it's strange. I tested it once and it seemed to behave, but sometimes does not (the property fields are not populated with the supplied arguments), seems to depend on the object history or something.
My workarounds for accessing full chamfer functionality are currently:
(1) manually calculate properties for the `Left Distance - Right Distance` configuration from any of the other cases (some basic trigonometry), and call that configuration,
(2) set any value and then access the node properties and manually set `Left Distance`, `Right Distance`, `Angle`.
Both are somewhat inelegant.

Issue #5362, opened 2024-10-30
Error Messages when 'sweep_along_path' and 'sweep_around_axis' are passed more than one object


DaveTwyman:
### Before submitting the issue

- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### Description of the bug

PyAEDT 'sweep_along_path' and 'sweep_around_axis' create 3D object(s) correctly but return below errors when passed a list of sweep objects (more than one object to be swept at a time)

Errors originate from the 'update_object' method called at end of 'sweep_along_path' and 'sweep_around_axis' when it is passed a list of assignments.

### Steps To Reproduce

These errors are produced:
![Image](https://github.com/user-attachments/assets/ab45354e-3f3c-4389-8be1-30b3fd2bdb9d)
When the below is run:

          from ansys.aedt.core import Maxwell3d
          M3D = Maxwell3d()
          
          linear_path=M3D.modeler.create_polyline(points=[[0,0,0],[0,100,0]],name="line")
          
          circle1 = M3D.modeler.create_circle(orientation='Z', origin=[5,0,0],radius=2, num_sides=8,name="circle1")
          circle2 = M3D.modeler.create_circle(orientation='Z', origin=[15,0,0],radius=2, num_sides=8,name="circle2")
          
          M3D.modeler.sweep_along_path(assignment=[circle1,circle2] ,sweep_object=linear_path) 
          ##### M3D.modeler.sweep_around_axis(assignment=[circle1,circle2] ,axis='Z') 
          
          M3D.modeler.fit_all()
          M3D.release_desktop(close_projects=False,close_desktop=False)

This can be corrected by replacing:

          return self.update_object(assignment)
                                                  
with:

          if isinstance(assignment, list):
              updated_obj = []
              for sel_obj in assignment:
                  updated_obj.append(self.update_object(sel_obj))
              return updated_obj
          else:
              return self.update_object(assignment)

in two affected methods 'sweep_along_path' and 'sweep_around_axis'

Alternatively we can modify 'update_object' to handle lists and update the following for consistency 'sweep_along_path', 'sweep_around_axis', "thicken_sheet' and "sweep_along_vector"

### Which Operating System are you using?

Windows

### Which Python version are you using?

3.11

### Installed packages

annotated-types==0.7.0
ansys-aedt-toolkits-antenna==0.3.2
ansys-aedt-toolkits-common==0.6.3
ansys-pythonnet==3.1.0rc3
attrs==23.2.0
[...]

PR #5415, opened 2024-11-12
Update 'sweep_along_path' and 'sweep_around_axis' #5362

DaveTwyman:
Update 'sweep_along_path' and 'sweep_around_axis' #5362


DaveTwyman:
Opened in error

PR #5422, opened 2024-11-13
Add `material_appearance` property to `Modeler3D`

atlanswer:
**Changes**
- Add property `material_appearance` to the `Modeler3D` class
- Modify one test

**Detail**
- The implementation is the same as `display_wireframe`

**Questions**
- Does `update_geometry_property` in `GeometryModeler` need an update?
- It seems like `display_wireframe` does not have a standalone unit test so I didn't include one for `material_appearance`

### DIFF
--- a/src/ansys/aedt/core/modeler/cad/object_3d.py
+++ b/src/ansys/aedt/core/modeler/cad/object_3d.py
@@ -98,6 +98,7 @@ class Object3d(object):
         self._surface_material = None
         self._color = None
         self._wireframe = None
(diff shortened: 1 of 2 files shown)


Samuelopez-ansys:
Thanks for catching this issue @atlanswer . Due to some restrictions we have for testing, we won't be able to use the current PR to add your changes in the main branch. We'll create a new PR and cherry-pick your commits so that the CICD can run with the correct credentials. Note that you'll still be the authors of those changes, we'll only be the commiters.

PR #5437, opened 2024-11-15
DOCS: Fix python code error in visualization report

SMoraisAnsys:
Changes include:
- fixing the documentation error;
- refactoring the `Fields`'s `__init__` method to use `post_app` as input argument name instead of `app`.

The second point would have helped to understand that one must provide a post application and not an app like `Hfss`, ...


MaxJPRey (approved):
LGTM.

PR #4848, opened 2024-06-25
FEAT: Enable harmonic force

JuliusSaitz:

PR #5440, opened 2024-11-18
FEAT: enable harmonic force 

anur7:
With this new PR:
- Added missing arguments for Transient solver (already done by @JuliusSaitz with the old not merged PR #4848)
- Implemented the method for Eddy Current solver.
- Still not complete, for Transient solver it is needed other improvements.


src/ansys/aedt/core/maxwell.py
```diff
@@ -1787,70 +1787,200 @@ def assign_radiation(self, assignment, radiation=None):
...
+                        "non-rotational transient solutions. Only Harmonic Force will be calculated."
+                    )
+                if not self.is3d and self.geometry_mode == "about Z":
+                    if use_number_of_cycles_from_stop_time or use_number_of_cycles_for_stop_time:
+                        self.logger.warning(
+                            " ``number_of_cycles_from_stop_time´´ and ``number_of_cycles_for_stop_time´´"
+                            "are not available for TransientZ."
+                        )
+                        use_number_of_cycles_from_stop_time = False
+                        use_number_of_cycles_for_stop_time = False
```
gmalinve, line 1944: @anur7 this part is very confusing. I'd simplify it. I also add an extra check on 2d and transient because use_number_of_cycles_from_stop_time and use_number_of_cycles_for_stop_time can be True in 3d Transient and Transient Aphi designs.
But this part has to be reviewed.

gmalinve:
@anur7 I pushed a change to fix the failing test but there are some parts in the method that need to be reviewed and simplified.
Also the test has to be improved, it would be bettere to test more arguments, not just providing the assignment.

src/ansys/aedt/core/maxwell.py
anur7: @gmalinve if I'm not wrong the Maxwell project used in UT contains only a transient 3d design. To increase the code coverage we need also a 2d design for testing.

PR #5442, opened 2024-11-18
FEAT: fix create_polyline for angular arc + update docs

gmalinve:


SMoraisAnsys (dismissed):
LGTM, could you add a test to ensure that show the fix effect ?
This would avoid future regression and explain why tests were not failing despite that bug.

src/ansys/aedt/core/modeler/cad/polylines.py
```diff
@@ -615,6 +615,7 @@ def _point_segment_string_array(self):
                     if current_segment.type == "AngularArc":
                         points_str.append(self._pl_point(current_segment.extra_points[i - 1]))
                         index_count += 1
+                        pos_count_incr += 1
```
SMoraisAnsys, line 618: Can you add a test to show why previous implementation should have been failing?

gmalinve: sure!

Samuelopez-ansys (approved):
Good job!

PR #5444, opened 2024-11-18
FIX: Export model obj and rename argument

### LINKED ISSUE #5443: Bug located in export_model_obj

### Description of the bug

Recent changes to `export_model_obj` (see https://github.com/ansys/pyaedt/pull/5432) inverted the logic associated to the export into one or multiple objects. From my understanding, the argument `export_as_single_objects` had the right meaning : "exporting the model into multiple objects" and we should revert the changes that got merged. If the name is confusing then we might want to rename the argument, for example using `export_as_multiple_objects` to export multiple objects or `export_as_single_object` to export into one object.

>NOTE: The changes make the pyaedt-examples repo failing. Specificly, this example https://github.com/ansys/pyaedt-examples/blob/main/examples/electrothermal/components_csv.py

If it is not clear

### Steps To Reproduce

See above

Windows

3.10

### Installed packages

Doesn't matter

SMoraisAnsys:
Revert part of the changes in #5432 and rename the method argument to avoid miss reading.

Close #5443

### DIFF
--- a/src/ansys/aedt/core/hfss.py
+++ b/src/ansys/aedt/core/hfss.py
@@ -6132,7 +6132,7 @@ class Hfss(FieldAnalysis3D, ScatteringMethods):
         from ansys.aedt.core.visualization.advanced.hdm_plot import HDMPlotter
 
         hdm = HDMPlotter()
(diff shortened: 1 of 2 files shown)


Samuelopez-ansys (approved):
LGTM

PR #5449, opened 2024-11-19
FEAT: Import data table in HFSS solution

### LINKED ISSUE #5387: Add import table data for circuit and hfss.

Import Table Data for both applications is only possible via ironpython methods currently.

_No response_

_No response_

### LINKED ISSUE #5418: Add DeleteImportData to PyAEDT

To delete solution data which came from table imports I currently use the following in HFSS:

osolution = pyaedt_app.osolution

solution = "solution_name"
osolution.DeleteImportData(solution)

This is neccessary to prevent the programm from crash when try to import solution data with the same name again. Therefore if the "same" solution is about to get imported again first the "old" one which got the identical name needs to be deleted.

Currently solved via IronPython command. But major drawback that for updatng the internal registry needs a save_project() which can take time for larger projects. The need to save to refresh internal entries after a deletion is the major problem.

_No response_

_No response_

Samuelopez-ansys:
Close #5418
Close #5387 

Next step is include it in Circuit and HFSS 3DLayout, but they have a different implementation
Include the postprocessing part


maxcapodi78 (approved):
LGTM

PR #5460, opened 2024-11-20
FIX: Adding EMIT functions to PyAEDT docs

ramin4667:
This PR adds EMIT functions to the PyAEDT document page.


Samuelopez-ansys:
@ramin4667 What is difference compared to this page? https://aedt.docs.pyansys.com/version/stable/API/_autosummary/ansys.aedt.core.emit.Emit.html

All applications are there.

PR #5464, opened 2024-11-21
FIX: generic empty initial library

gkorompi:


Samuelopez-ansys:
@gkorompi If the user just call create_component(), will it work? I am just wondering if component_library is needed for the default case.

gkorompi:
> @gkorompi If the user just call create_component(), will it work? I am just wondering if component_library is needed for the default case.

@Samuelopez-ansys Yes i tested it and it works. For an empty component_library ="" it searches in all libraries, hence it looks like a better default choice. (Nevertheless he has to call the component_name argument)

Samuelopez-ansys (dismissed):

Samuelopez-ansys (approved):
LGTM

PR #5471, opened 2024-11-22
FIX: Modeler objects were not correctly updated in some cases.

maxcapodi78:


svandenb-dev (approved):
LGTM

Samuelopez-ansys (dismissed):

Alberto-DM (approved):
LGTM

Samuelopez-ansys (approved):

src/ansys/aedt/core/modeler/cad/primitives.py
```diff
@@ -198,9 +198,7 @@ def __init__(self, parent, obj_type="o", props=None):
                 self.__obj_names[value.name] = value
                 if self.__obj_type == "o":
                     self.__parent._object_names_to_ids[value.name] = key
-            self.__refreshed = True
-        else:
-            self.__refreshed = False
+        self.__refreshed = False
```
SMoraisAnsys, line 201: I'm not sure if this change is enough, can we talk about this briefly ?
I'll comment the PR afterward.

maxcapodi78: Yes it will. The Objects class is initialized with the modeler. When we parse the info from the aedt file the __refreshed value is set to True and this prevent the objects dictionary to refresh later on. I've removed this possibility, so even in case of information parsed from the AEDT, the objects keep the possibility to refresh itself and recover missed objects

Alberto-DM: Just explained with a call.

PR #5473, opened 2024-11-22
FEAT: implement resistive sheet

gmalinve:


src/ansys/aedt/core/maxwell.py
```diff
@@ -2961,6 +2961,132 @@ def assign_zero_tangential_h_field(self, assignment, boundary=None):
...
+        """Assign a resistive sheet boundary between two conductors.
+
+        Available for Maxwell 3D Magnetostatic, Eddy Current and Transient designs.
+        For 3D Magnetostatic designs, the user can specify the nonlinear anode and cathode coefficients.
+        To understand the nonlinear relationship used by AEDT between the conductivity and current density,
+        please refer to Maxwell Help guide.
+
+        Parameters
+        ----------
+        assignment : list of int or :class:`ansys.aedt.core.modeler.cad.object_3d.Object3d`
```
SMoraisAnsys, line 2989:
```suggestion
        assignment : list of int, str or :class:`ansys.aedt.core.modeler.cad.object_3d.Object3d`
```
Just wondering, can't assignment be an `str` ?

src/ansys/aedt/core/maxwell.py
```diff
@@ -2961,6 +2961,132 @@ def assign_zero_tangential_h_field(self, assignment, boundary=None):
...
+        assignment = self.modeler.convert_to_selections(assignment, True)
+
+        if not name:
+            boundary = generate_unique_name("ResistiveSheet")
+
+        props = dict(
+            {
+                "Faces": assignment,
+            }
+        )
```
SMoraisAnsys, line 3069:
```suggestion
        props = {
            "Faces": assignment,
        }
```
`{ .. }` is already a dict

src/ansys/aedt/core/maxwell.py
```diff
@@ -2961,6 +2961,132 @@ def assign_zero_tangential_h_field(self, assignment, boundary=None):
...
+            props["CathodeParA"] = cathode_a
+            props["CathodeParB"] = cathode_b
+            props["CathodeParC"] = cathode_c
+            props["CathodeParD"] = cathode_d
+
+        bound = BoundaryObject(self, boundary, props, "ResistiveSheet")
+        if bound.create():
+            self._boundaries[bound.name] = bound
+            return bound
+        return False
```
SMoraisAnsys, line 3088: @gmalinve Just wondering, do you want to try adding a unit/integration test for `return False` ? If yes, I can help you with that !

SMoraisAnsys (approved):
LGTM, I left minor comments

src/ansys/aedt/core/maxwell.py
gmalinve: @SMoraisAnsys The native mathod doesn't accept the names because face objects don't have names but only ids.

Samuelopez-ansys (approved):
LGTM

Issue #5438, opened 2024-11-18
Bug located in 'duplicate' function in modeler_pcb.py


abhandare21:
### Before submitting the issue

- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### Description of the bug

User gets an error message when trying to duplicate a list of objects.

Current line 761 in https://github.com/ansys/pyaedt/blob/v0.11.3/src/ansys/aedt/core/modeler/modeler_pcb.py

`self.oeditor.Duplicate(["NAME:options", "count:=", count], ["NAME:elements", ",".join(assignment)], vector)`

Suggested change to fix bug

`self.oeditor.Duplicate(["NAME:options", "count:=", count], ["NAME:elements"]+assignment, vector)`

### Steps To Reproduce

```
# Imports
from pyaedt import Hfss3dLayout

# Variables
project_name = 'Duplicate function'
design_name= 'test'
aedt_version = '2024.2'
layer_name = 's'
pitch = 2e-3

# Create Hfss3dLayout instance
h3dl = Hfss3dLayout(projectname=project_name, designname=design_name, specified_version=aedt_version, close_on_exit=False)

# Create a new layer in the modeler
h3dl.modeler.layers.add_layer(layer_name, 'signal', '0mm', '0mm', 'copper')

# Create a rectangle
rect = [h3dl.modeler.create_rectangle(layer_name, origin=['0mm', '0mm'], sizes=['1mm', '1mm'])]

# Duplicate the rectangle 5 times in the x direction
rect.extend(h3dl.modeler.duplicate(rect, 4, [pitch, 0])[0])

# Duplicate the rectangle 5 times in the y direction
h3dl.modeler.duplicate(rect, 4, [0, pitch])

# Exit AEDT
h3dl.release_desktop(close_desktop=False, close_projects=False)
```

### Which Operating System are you using?

Windows

### Which Python version are you using?

3.11

### Installed packages

python 3.11.0 installed using conda

pip install list
annotated-types==0.7.0
ansys-pythonnet==3.1.0rc4
attrs==24.2.0
cffi==1.17.1
clr-loader==0.2.6
defusedxml==0.7.1
fonttools==4.54.1
fpdf2==2.8.1
jsonschema==4.23.0
jsonschema-specifications==2024.10.1
numpy==1.26.4
pandas==2.2.3
pillow==11.0.0
plumbum==1.9.0
psutil==6.1.0
pyaedt==0.11.3
pycparser==2.22
[...]

PR #5476, opened 2024-11-23
FIX: Duplicate function by @abhandare21 in #5438

abhandare21:
Updated the duplicate function in the modeler_pcb.py file line 761, to allow a list of items to be passed as argument for hfss3dlayout to duplicate.

### DIFF
--- a/src/ansys/aedt/core/modeler/modeler_pcb.py
+++ b/src/ansys/aedt/core/modeler/modeler_pcb.py
@@ -758,7 +758,7 @@ class Modeler3DLayout(Modeler, Primitives3DLayout):
         self.cleanup_objects()
         if isinstance(assignment, str):
             assignment = [assignment]
-        self.oeditor.Duplicate(["NAME:options", "count:=", count], ["NAME:elements", ",".join(assignment)], vector)
+        self.oeditor.Duplicate(["NAME:options", "count:=", count], ["NAME:elements"] + assignment, vector)
         return self.cleanup_objects()
 
     @pyaedt_function_handler(objects="assignment")


gmalinve:
Hi @abhandare21,

Thanks for your contribution. I have one question, why do you need this change? does it bring any benefit to your workflow? do you have any example where this can be useful?

Thanks again,

Giulia

abhandare21:
Hi @gmalinve, I am building a big array of elements in an x-y grid, which requires me to duplicate multiple items in both directions. It is much faster when I pass a list versus running in a for loop.

gmalinve:
@abhandare21 why do you need a for loop if you can duplicate a geometry along a direction with the _duplicate_ method?
I saw your issue, why don't you just create a geometry (without saving it into a list) and then simply duplicate it along x and y direction? like that:
```
# Create a rectangle
rect = h3dl.modeler.create_rectangle(layer_name, origin=['0mm', '0mm'], sizes=['1mm', '1mm'])

# Duplicate the rectangle 5 times in the x direction
rect_x = h3dl.modeler.duplicate(rect, 4, [pitch, 0])

# Duplicate the rectangle 5 times in the y direction
rect_y = h3dl.modeler.duplicate(rect, 4, [0, pitch])
```
With this code snippet you get something like this:
image

I tested your code and it doesn't duplicate along the y direction, I end up having something like this:
image

Samuelopez-ansys:
In addition, did you try 3D Component array? This technique is much efficient for computation:

https://examples.aedt.docs.pyansys.com/version/dev/examples/high_frequency/antenna/array.html

abhandare21:
@gmalinve, the code I have shared in the issue won't work unless you change line 761 in the modeler_pcb.py file

What I want the duplicate function to do is this
image

And if I have to do that with the current code, I need to iterate over a list of items in the x array to duplicate them in the y direction.

If you do a record script through Tools>Record script to file, then the duplicate line is as such
`oEditor.Duplicate(["NAME:options","count:=",4 ],["NAME:elements","rect__3","rect_0","rect__4","rect__5","rect__6"],[0,0.002])`

In order to get this using the duplicate function, you would need to change the existing line 761 from
`self.oeditor.Duplicate(["NAME:options", "count:=", count], ["NAME:elements", ",".join(assignment)], vector)`
to
`self.oeditor.Duplicate(["NAME:options", "count:=", count], ["NAME:elements"]+assignment, vector)`

This is the output of the two lines respectively.
`["NAME:elements", ",".join(assignment)]`

> ['NAME:elements', 'NewObject_RTAS5M,rect__1,rect__3,rect__4,rect__2']

`["NAME:elements"] + assignment`

> ['NAME:elements', 'NewObject_RTAS5M', 'rect__1', 'rect__3', 'rect__4', 'rect__2']

This change should get the code to align with what HFSS is doing.

abhandare21:
@Samuelopez-ansys would this work with 3Dlayout?

Samuelopez-ansys:
It does not work with HFSS 3DLayout right. But a 3D Component in HFSS using pure layout geometry, will still use "Phi mesher". 

If you need more information about how to improve your workflow, please reach the Ansys support team, from my experience, I see more benefits using 3D Component array in HFSS, but maybe I am wrong because I do not know your application

abhandare21:
@Samuelopez-ansys I have about 100 signal and dielectric layer each with about 150ish object on each layer. Hence the ask to update the duplicate function so 3Dlayout users can effectively use the code.

Samuelopez-ansys:
@abhandare21 Thanks for your contribution. We have cherry picked your Pull request, so I will proceed to close this PR.

The new PR is: #5476

abhandare21:
Thanks !

PR #5484, opened 2024-11-27
FIX: corrected the way we identify the nets automatically from the sources.

maxcapodi78:
## Description
corrected the way we identify the nets automatically from the sources.


hui-zhou-a (approved):
LGTM.

Issue #5497, opened 2024-11-28
Set tuning range for HFSS derivatives


Samuelopez-ansys:
### Description of the feature

Parameters in HFSS could be used for derivatives. It would be useful to add the set of these derivatives.

PR #5499, opened 2024-11-29
FIX: Apply value to derivative variable

Samuelopez-ansys:
## Description
Set tuning offset was not setting the value of the variable correctly

## Issue linked
#5497 

## Checklist

### DIFF
--- a/src/ansys/aedt/core/modules/solve_setup.py
+++ b/src/ansys/aedt/core/modules/solve_setup.py
@@ -2483,27 +2483,50 @@ class SetupHFSS(Setup, object):
     def set_tuning_offset(self, offsets):
         """Set derivative variable to a specific offset value.
 
+        This method adjusts the tuning ranges for derivative variables in the design, allowing for specific offset
+        values to be applied. If a variable is not specified in the ``offsets`` dictionary,
+        its offset is set to ``0`` by default. Each value must be within ±10% of the nominal
+        value of the corresponding variable.
+
         Parameters
         ----------
         offsets : dict
-            Dictionary containing the variable name and it's offset value.
+            Dictionary where keys are variable names and values are the corresponding offset values to be applied.
 
         Returns
         -------
         bool
+            ``True`` when successful, ``False`` when failed.
+
+        References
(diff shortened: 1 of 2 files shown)


src/ansys/aedt/core/modules/solve_setup.py
```diff
@@ -2483,27 +2483,51 @@ def add_derivatives(self, derivative_list):
...
             if v not in offsets:
                 offsets[v] = 0
-        arg = []
+        arg = ["NAME:TuningRanges"]
         for k, v in offsets.items():
-            arg.append(f"DeltaOffset({k})")
-            arg.append(f"{abs(self._app.variable_manager[k].numeric_value) * (-0.1)}")
-            arg.append(f"{v}")
-            arg.append(f"{abs(self._app.variable_manager[k].numeric_value) * 0.1}")
+            arg.append(f"Range:=")
```
SMoraisAnsys, line 2521:
```suggestion
            arg.append("Range:=")
```
Only use f-string when there is a variable :)

src/ansys/aedt/core/modules/solve_setup.py
```diff
@@ -3200,27 +3224,51 @@ def add_derivatives(self, derivative_list):
...
```
SMoraisAnsys, line 3262:
```suggestion
            arg.append("Range:=")
```

SMoraisAnsys (approved):
LGTM, I left minor comments.
The two methods `set_tuning_offset` seem to be the exact same code. We should really think about refactoring a bit the code to avoid code duplicates.
Do we have an issue to track all those duplicates and see if we can't come up with new classes encapsulating all those duplicates ?

Samuelopez-ansys:
I agree regarding the duplicities. In this case this method can be used only for SetupHFSS and SetupAuto, but not for the others, so we can add it to Setup class, and internally check if the Setups are of one of these 2 types. We can do It in another PR.

gmalinve (approved):

gmalinve (approved):

PR #5527, opened 2024-12-04
BUILD: Update requirement of osmnx to greater than 1.1.0 and lower than 2.1

SMoraisAnsys:
## Description
As title says.

## Checklist


MaxJPRey (approved):
LGTM.

SMoraisAnsys:
@Samuelopez-ansys Would it make sense to extend the testing of features related to osmnx or should I do that in another PR?

Samuelopez-ansys:
@

SMoraisAnsys:
> @SMoraisAnsys Why do we need compatibility with previous versions?
> 
> If the output is the same, why not just use the new API?

We could do that but that means that people using older version of osmnx won't be able to use new versions of pyaedt.
This is not a problem in my eyes but I'd rather clean up the dependencies once we move to version `1.0` than doing it right now.
It's an open discussion though :)

Samuelopez-ansys (approved):
LGTM

PR #5531, opened 2024-12-05
FIX: Loading default settings value

### LINKED ISSUE #5536: Improve settings to work with default YAML file

### Description of the bug

While showing the use of `Settings` with a configuration file to update the default settings of `pyaedt`, I discovered that the default YAML file in doc/source/Resources is not correctly loaded. Indeed, by default some parameters are cast into other types, e.g. `int`, but the default value is None. This leads to a `TypeError`.

Pinging @kruglovIFX for visibility.

### Steps To Reproduce

Set `PYAEDT_LOCAL_SETTINGS_PATH` to the documentation file mentioned above or call `load_yaml_configuration` on it.

Windows

3.10

### Installed packages

Not relevant

SMoraisAnsys:
## Description
This changes should allow one to load the default pyaedt_settings.yaml file.
Previously, if a user tried to use the file as it is and only modified one setting, e.g. `enable_screen_logs`, then loading the YAML file failed because some fields are casted / tested on the fly.

Also, while working on this and testing, I found that we were not correctly hanling the update of the AEDT environment variables stored in `Settings`. This PR also fixes it.

Close #5536

Pinging @kruglovIFX for visibility. Note that you'll have to wait for this to be merged and a new release to published in order to leverage this fix.


SMoraisAnsys:
@Samuelopez-ansys @Alberto-DM Would it make sense to add `ANS_NODEPCHECK=1` by default even if we are not in Windows environment ?

I'm asking that because if a user tries to come up with his own default settings file, he might start from the one we provide as an example. However, this file does not contain ANS_NODEPCHECK. The reason for that is that this environment variable seems to be leveraged in Linux environment but not in Windows. Therefore, users might end up with a configuration were ANS_NODEPCHECK is not set :/

Alberto-DM:
> @Samuelopez-ansys @Alberto-DM Would it make sense to add `ANS_NODEPCHECK=1` by default even if we are not in Windows environment ?
> 
> I'm asking that because if a user tries to come up with his own default settings file, he might start from the one we provide as an example. However, this file does not contain ANS_NODEPCHECK. The reason for that is that this environment variable seems to be leveraged in Linux environment but not in Windows. Therefore, users might end up with a configuration were ANS_NODEPCHECK is not set :/

If the env variable ANS_NODEPCHECK is ignored in Windows, we can keep it always set to 1.

SMoraisAnsys:
> > @Samuelopez-ansys @Alberto-DM Would it make sense to add `ANS_NODEPCHECK=1` by default even if we are not in Windows environment ?
> > I'm asking that because if a user tries to come up with his own default settings file, he might start from the one we provide as an example. However, this file does not contain ANS_NODEPCHECK. The reason for that is that this environment variable seems to be leveraged in Linux environment but not in Windows. Therefore, users might end up with a configuration were ANS_NODEPCHECK is not set :/
> 
> If the env variable ANS_NODEPCHECK is ignored in Windows, we can keep it always set to 1.

The problem is that I don't know the impact of `ANS_NODEPCHECK` in Windows environment. Since it was only set for Linux, I assumed it would have some impact on Windows. @Samuelopez-ansys Do we have a way to check if there is any impact ? If that's the case, please let me know and I'll do the test myself.

Samuelopez-ansys:
@SMoraisAnsys I will check internally if it has any impact in Windows.

Samuelopez-ansys:
@SMoraisAnsys Confirmed. ANS_NODEPCHECK has only impact in Linux

SMoraisAnsys:
@Samuelopez-ansys @gmalinve While working on improving code coverage, I discovered that `block_figure_plot` was not handled with YAML loading / dumping. It got through https://github.com/ansys/pyaedt/pull/5297 without being noticed.
This ping is just a reminder that if you ever add a new setting or see a PR doing so, we need to update the loading of YAML files.
Note that I added a test that should fail if one doesn't do it :D

Samuelopez-ansys (approved):

Issue #5543, opened 2024-12-09
Bug located in documentation of some object


maxcapodi78:
### Before submitting the issue

- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### Description of the bug

In some cases docstrings are not generated correctly for Returns and this cause autocomplete to not being able to identify the correct object. An example is post object.

### Steps To Reproduce

Check documentation

### Which Operating System are you using?

Windows

### Which Python version are you using?

3.8

### Installed packages

.

PR #5544, opened 2024-12-09
FIX: Readibility improvements

maxcapodi78:
Readibility improvements

## Description
Fixed many docstrings errors!
modeler of PostCommon class has been made internal.

## Issue linked
#5543 

## Checklist


src/ansys/aedt/core/visualization/post/common.py
```diff
@@ -555,9 +552,16 @@ def _oproject(self):
         return self._app._oproject
 
     @property
-    def modeler(self):
+    def _modeler(self):
```
SMoraisAnsys, line 555: Could you explain why you want to rename this property as `_modeler` ? I think property should be used to mask implementation details and provide a clean interface so I would prefer to stick with `modeler`

SMoraisAnsys (commented):
If the goal is to make `modeler` internal, could we make `_modeler` an attribute of the class instead of a property ? See my comment

src/ansys/aedt/core/application/analysis.py
```diff
@@ -233,7 +233,7 @@ def materials(self):
         return self._materials
 
     @property
-    def setups(self):
+    def setups(self) -> list[Setup]:
```
SMoraisAnsys, line 236:
```suggestion
    def setups(self) -> List[Setup]:
```
I think you need to import List from typing and change `list` into `List`.
Once we move to Python3.9+ we will be able to use `list[Setup]` but that's not the case yet (waiting for v1.0 ? :p)

SMoraisAnsys (changes requested):
You need to modify a signature to use `List` instead of `list`. To make that work, you also need to add `from typing import List`

src/ansys/aedt/core/visualization/post/common.py
maxcapodi78: modeler is an object of the app:
hfss.modeler

having it internal in post has 2 main purposes:
- avoiding to load modeler when the post class is initialized (before it was initialized in the __init__)
- make it hidden. The purpose of this property is to make coding simpler to methods in Post that requires access to the modeler. It's not intended to be used by users

src/ansys/aedt/core/application/analysis.py
maxcapodi78: gotcha. I'll modify accordingly

src/ansys/aedt/core/visualization/post/common.py
SMoraisAnsys: I see but it's a bit strange to hide something using a Python property.
Would it be possible to remove the Python property and add `self._modeler = self._app._modeler` in the `__init__` ? I know it would revert your changes in the `__init__` but it would follow the logic of only providing public stuff in Python properties. If required, we can add a mecanism to load the modeler lazily with an extra attribute like `self.__modeler_loaded`. Also, what's the value of not loading the modeler by default ? Does is make the code much faster ?

maxcapodi78: modeler initialization takes time on big problems. people interested in post processing only may not need at all the modeler. Is all about efficiency. BTW you cannot initialize self._app._modeler because also _modeler is not initialized when you launch pyaedt but first time user uses modeler property

SMoraisAnsys: Gotcha. I'm still not really happy with this `_modeler` property since property should be things accessible to our users. However, I understand your point. If others are fine with that, I won't oppose firmly but... still feeling we shouldn't use a property here. Would it be possible to refactor the code to avoid that behavior ? It seems to be about encapsulating data so we should be able to reach a good trade off :)

Samuelopez-ansys (approved):

SMoraisAnsys (approved):

PR #5547, opened 2024-12-09
CHORE: Clean up template file

SMoraisAnsys:
## Description
A template file got introduced by mistake while doing the changes with AEDT 24.2
Probably a file resulting from example / extension testing.


Samuelopez-ansys (approved):
Great!

MaxJPRey (approved):
Out of curiosity, was it creating any problem with our CI?
Probably no, but just checking some workflows behavior.

SMoraisAnsys:
> Out of curiosity, was it creating any problem with our CI? Probably no, but just checking some workflows behavior.

It does not create any problem, I was looking at our pyproject.toml file and fell on the file by mistake. It seems that I added it months ago while doing the migration to AEDT 24.2 and it's probable a left over of running some examples or extensions

PR #5548, opened 2024-12-09
BUILD: Modify TOML dependency

SMoraisAnsys:
## Description
Currently we are using a TOML dependency that seem to be faster than others. However, I don't think we are impacted by read / write performance on TOML files and would prefer to leverage projects that are used / maintained by the Python community.

Changes consist in using the recommended packages to [read](https://pypi.org/project/tomli/) and [write](https://pypi.org/project/tomli-w/) TOML files. Note that the read package is as light as possible and doesn't contain write capability. It is also the package built-in in Python3.11+. Both packages are referenced in Python official documentation, see https://docs.python.org/3/library/tomllib.html


Samuelopez-ansys (dismissed):
LGTM

pyproject.toml
```diff
@@ -39,7 +39,8 @@ dependencies = [
     "pyedb>=0.4.0; python_version == '3.7'",
     "pyedb>=0.24.0; python_version > '3.7'",
     "pyedb!=0.28.0; python_version > '3.7'",
-    "pytomlpp; python_version < '3.12'",
+    "tomli; python_version < '3.11'",
+    "tomli-w; python_version < '3.11'",
```
SMoraisAnsys, line 43:
```suggestion
    "tomli-w",
```

Samuelopez-ansys (dismissed):

PR #5552, opened 2024-12-09
DOCS: Fix numpy docstyle errors.

MaxJPRey:
The goal is to fix some errors raised when applying numpy docstyle.

It should fix couple of issues but not all of them.


SMoraisAnsys (approved):
LGTM, thanks for those changes @MaxJPRey they will help us in improving our documentation and also ease things when we move to `sphinx-autoapi`

SMoraisAnsys (approved):

Issue #5554, opened 2024-12-10
Improve .NET handling in Linux environment


SMoraisAnsys:
Currently the .NET runtime used with PyAEDT is the one provided by `dotnetcore2`. However there are incompatibilities with recent system libraries and we faced that issue when working with default system libraries of Ubuntu-22.04

Since we need to be compatible with such OS, I'd recommend allowing users to provide their own .NET runtime (that they install) and fallback to using `dotnetcore2` if no such runtime is found. This way, current users shouldn't be impacted by any changes in pyaedt and users facing the bug I just mentionned should be able to keep working with PyAEDT.

I think we should remove `dotnetcore2` from PyAEDT's dotnet install target (`pyaedt[dotnet]`) when reaching `v1.0`.

PR #5555, opened 2024-12-10
REFACTOR: Loading dotnet runtime

SMoraisAnsys:
## Description
To improve compatibility with system libraries in Linux, the Python package `dotnetcore2` should be removed from PyAEDT's dotnet installation target. Indeed, the embedded version of `.NET` associated to `dotnetcore2` is old and has **incompatibilities** with recent versions of `openssl` such as the one installed by default in Ubuntu 22.04.

This PR propose to add this modification to the breaking changes of release v1.0 and, in the mean time, to allow users to:
- use a .NET runtime that they have installed instead of the one embedded in `dotnetcore2`
- fallback to the `dotnetcore2` approach if no .NET runtime is detected / provided through environment variable (DOTNET_ROOT).

## Issue linked
See #5554

---

> [!NOTE]  
Tests implementation differs from pyedb since pyaedt has a `filterwarnings` option for pytest.

---

> [!WARNING]  
**DO NOT MERGE until feedback is provided**

### DIFF
--- a/doc/source/release_1_0.rst
+++ b/doc/source/release_1_0.rst
@@ -186,9 +186,9 @@ The following table list the name changes with the old and new paths:
 +----------------------------------------------------------------+--------------------------------------------------------------------------+
 
 Dotnet changes in Linux
----------------------
+-----------------------
 
-To improve compatibility with system libraries, the Python package `dotnetcore2` is being removed from PyAEDT's dotnet installation target.
+To improve compatibility with system libraries in Linux, the Python package `dotnetcore2` is being removed from PyAEDT's dotnet installation target.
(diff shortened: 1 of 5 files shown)


SMoraisAnsys:
Hi @AlfVII @clguenthe I'm pinging you because, following a previous issue, I had the feeling that you might still be using PyAEDT with dotnet. If that's the case, could you give a try at this PR to check that it suits you well ? If something is wrong, please let me know !

clguenthe:
HI @SMoraisAnsys, I checked this change with the my solution which is still using .NET communication, and it works fine. Thanks a lot for the heads-up :)

PR #5576, opened 2024-12-16
FIX: Temporary avoid attrs v24.3.0

SMoraisAnsys:
## Description
Temporary solution for the packaging issue in CICD. This PR should be reverted at some point once `attrs` gets a new release with license available on PyPI.

## Issue linked
Here is the issue associated to our CICD failure  https://github.com/python-attrs/attrs/issues/1386


Samuelopez-ansys (approved):
Thanks!

Issue #5597, opened 2024-12-19
Get Component Definition Path for Circuit Components


gkorompi:
### Description of the feature

-

PR #5598, opened 2024-12-19
FEAT: get the path of a circuit component and add import for .sss files

gkorompi:
## Description
Retrieve the path of a Circuit component, specifically spice imported spice models etc. 

## Issue linked
#5804 
#5597 

## Checklist


src/ansys/aedt/core/modeler/circuits/primitives_circuit.py
```diff
@@ -1239,6 +1239,23 @@ def create_wire(self, points, name=""):
...
+            Name of the component.
+        """
+        for component in self.components.values():
+            if component.component_info["InstanceName"] == component_name:
+                component_definition = component.component_info["Info"]
+                if not self.o_component_manager.GetData(component_definition):
+                    self.logger.warning("Component has no path")
+                    return False
+                else:
+                    return (self.o_component_manager.GetData(component_definition)[24][1][12][1].split(" ")[1])[1:-1]
```
Samuelopez-ansys, line 1257: Hi @gkorompi , THis is nice, maybe it would be better to have the component path at component level, so I would do it as a property like>

component.component_path 

What do you think?

If you like it, I think it should be at CircuitComponent level.

Samuelopez-ansys (dismissed):

src/ansys/aedt/core/modeler/circuits/primitives_circuit.py
gkorompi: @Samuelopez-ansys I agree! That was my initial thought, but when in CircuitComponent level the function o_component_manager is not there (it's in the level before that) that is why i implemented it there. Let me know what do you think. Thanks!

Samuelopez-ansys: If I am not wrong you can access: self._circuit_components.o_component_manager

SMoraisAnsys:
Please provide a description to your PR

image

SMoraisAnsys (commented):
Hi @gkorompi thanks for the contribution !

I'm not sure about the specific location that you are hard coding.

Are you sure this information will be here for every cases ? When going around pyaedt code sources, I found cases where `"NAME:CosimDefinition"` was first in the list but the 12th element wasn't a list (so couldn't be caught through `[1]`).

If I'm wrong and the 12th element is correct, is this place the same for every version of AEDT? If not, could you update the code ?

gmalinve (dismissed):

gmalinve:
@gkorompi I left a couple of comments but I'd handle the property in a different way. Something like: 
```
component_definition = self.component_info["Info"]
    component_data = self._circuit_components.o_component_manager.GetData(component_definition)
    
    if not component_data:
        self._circuit_components._app.logger.warning("Component has no path")
        return False

    for data in component_data:
        if isinstance(data, list) and isinstance(data[0], str) and data[0] == "NAME:CosimDefinitions":
            return (data[1][12][1].split(" ")[1])[1:-1]
    
    return False
```
be sure that what you are lloking for is always in this same location: data[1][12][1].

src/ansys/aedt/core/modeler/circuits/object_3d_circuit.py
```diff
@@ -1084,6 +1084,18 @@ def create_pin_def(pin_name, x, y, angle):
         self._circuit_components.oeditor.MovePins(self.composed_name, -0, -0, 0, 0, ["NAME:PinMoveData"])
         return True
 
+    @property
+    def component_path(self):
+        """Component definition path."""
+        component_definition = self.component_info["Info"]
+        if not self._circuit_components.o_component_manager.GetData(component_definition):
```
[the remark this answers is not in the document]
gkorompi: Thanks @gmalinve  for the feedback, will fix accordingly!

src/ansys/aedt/core/modeler/circuits/object_3d_circuit.py
```diff
@@ -1084,6 +1084,18 @@ def create_pin_def(pin_name, x, y, angle):
...
+            self._circuit_components._app.logger.warning("Component has no path")
+            return False
+        for i in self._circuit_components.o_component_manager.GetData(component_definition):
+            if type(i) == list and type(i[0]) == str:
+                if i[0] == "NAME:CosimDefinitions":
+                    return (i[1][12][1].split(" ")[1])[1:-1]
```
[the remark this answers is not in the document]
gkorompi: Yes, only the first index may vary, that's why i added a for loop and the if ="NAME:CosimDefinitions" statement. The [1][12][1] is stable as far as i have tested.

src/ansys/aedt/core/modeler/circuits/object_3d_circuit.py
```diff
@@ -1084,6 +1084,18 @@ def create_pin_def(pin_name, x, y, angle):
         self._circuit_components.oeditor.MovePins(self.composed_name, -0, -0, 0, 0, ["NAME:PinMoveData"])
         return True
 
+    @property
+    def component_path(self):
+        """Component definition path."""
+        component_definition = self.component_info["Info"]
+        component_data = self._circuit_components.o_component_manager.GetData(component_definition)
+        if not component_data:
+            self._circuit_components._app.logger.warning("Component has no path")
```
MaxJPRey, line 1093: Maybe here we could provide the component name to help the user identify the origin of the issue.

gkorompi: Hey @MaxJPRey, Yes i will added it thanks for the note!

src/ansys/aedt/core/modeler/circuits/object_3d_circuit.py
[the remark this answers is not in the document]
gkorompi: You are right @gmalinve  i will implement it for all cases. Thanks

src/ansys/aedt/core/modeler/circuits/object_3d_circuit.py
MaxJPRey: Great. Thanks @gkorompi .

maxcapodi78 (approved):
LGTM

PR #5622, opened 2024-12-31
FIX: Remove access to members before their definition.

MaxJPRey:
- `n_pin` was raising a warning of type: access-member-before-definition

- `coupling_limit_value` was used before assignment.

- The test `test_minimum_order_stop_band_freq` is already defined.

### DIFF
--- a/tests/system/general/test_45_FilterSolutions/test_filter/test_attributes.py
+++ b/tests/system/general/test_45_FilterSolutions/test_filter/test_attributes.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2021 - 2024 ANSYS, Inc. and/or its affiliates.
+# Copyright (C) 2021 - 2025 ANSYS, Inc. and/or its affiliates.
 # SPDX-License-Identifier: MIT
 #
 #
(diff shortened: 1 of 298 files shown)


tests/system/general/test_45_FilterSolutions/test_filter/test_attributes.py
```diff
@@ -151,7 +151,7 @@ def test_minimum_order_group_delay_cutoff(self):
         design.attributes.minimum_order_group_delay_cutoff = "500 MHz"
         assert design.attributes.minimum_order_group_delay_cutoff == "500 MHz"
 
-    def test_minimum_order_stop_band_freq(self):
+    def test_minimum_order_stop_band_freq_second_time(self):
```
MaxJPRey, line 154: @ramin4667 I just renamed the test. In this file it is defined twice if you look few lines above.
I was not sure if it was done on purpose to maybe set up again some parameters.

SMoraisAnsys (approved):
LGTM

SMoraisAnsys (changes requested):
Previous review was performed on my cell phone and I missed the test duplication (though it was only test names)
I think that one of the test should be removed since they are duplicates.

tests/system/general/test_45_FilterSolutions/test_filter/test_attributes.py
ramin4667: @MaxJPRey The duplicated test is removed.

MaxJPRey: Thanks @ramin4667 .

MaxJPRey:
> Previous review was performed on my cell phone and I missed the test duplication (though it was only test names) I think that one of the test should be removed since they are duplicates.

It is done @SMoraisAnsys .

SMoraisAnsys (approved):

PR #5626, opened 2025-01-03
FIX: Fixes to get_object_material_properties method

ansys-satyajeet:
## Description
The `get_object_material_properties` method gives an error when the material name is in mixed case or upper case. 

## Issue linked
Issue # 5625 - Bug in `get_object_material_properties` method

## Checklist

### DIFF
--- a/src/ansys/aedt/core/modules/material_lib.py
+++ b/src/ansys/aedt/core/modules/material_lib.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright (C) 2021 - 2024 ANSYS, Inc. and/or its affiliates.
+# Copyright (C) 2021 - 2025 ANSYS, Inc. and/or its affiliates.
 # SPDX-License-Identifier: MIT
 #
 #
--- a/src/ansys/aedt/core/modules/material_lib.py
+++ b/src/ansys/aedt/core/modules/material_lib.py
@@ -149,7 +149,7 @@ class Materials(object):
     @property
     def _mat_names_aedt_lower(self):
         if len(self._mats_lower) < len(self._mat_names_aedt):
-            self._mats_lower = [i.lower() for i in self._mat_names_aedt]
+            self._mats_lower = [i.casefold() for i in self._mat_names_aedt]
         return self._mats_lower
 
     @property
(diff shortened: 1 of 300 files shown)


SMoraisAnsys (changes requested):
Hi @ansys-satyajeet, thanks for the contribution !

I didn't know about the `casefold` method and only used `lower` for string objects. Could you extend your contribution to replace `lower` with `casefold` in the file you touched ? I just looked and the line of code `return self.surface_material_keys[item]` of method `__getitem__` of `Materials` should also be changed to handle lower cases.

ansys-satyajeet:
@SMoraisAnsys Hi Sebastian, I will check the method and implement the changes to handle lower and mixed cases for material names. 
Sorry about the novice question, if I make a new change, do I need to create a new pull request? I haven't made any pull requests on pyaedt code for a long time.

SMoraisAnsys:
> @SMoraisAnsys Hi Sebastian, I will check the method and implement the changes to handle lower and mixed cases for material names. Sorry about the novice question, if I make a new change, do I need to create a new pull request? I haven't made any pull requests on pyaedt code for a long time.

You can commit into your branch and publish the changes (`push`). This will update the current PR :)

src/ansys/aedt/core/modules/material_lib.py
```diff
@@ -605,12 +605,12 @@ def duplicate_surface_material(self, material, name):
...
         """
-        if not material.lower() in list(self.surface_material_keys.keys()):
+        if material.lower() not in list(self.surface_material_keys.keys()):
             self.logger.error(f"Material {material} is not present")
             return False
-        newmat = SurfaceMaterial(
-            self, name.lower(), self.surface_material_keys[material.lower()]._props, material_update=True
-        )
+        if not name:
+            name = material + "_clone"
```
lorenzovecchietti, line 612: Can you add coverage for this line?
Just add a line to `test_06_surface_material` in `test_03_Materials.py`. Something like:
`assert ipk.materials.duplicate_surface_material("Steel")`

lorenzovecchietti (changes requested):
I think @SMoraisAnsys suggested replacing all `lower()` calls with `casefold()` calls in the file, not the other way around.

I was unaware of `casefold()` until now, but it seems to be the recommended approach. Thanks to @ansys-satyajeet for introducing me to it.

SMoraisAnsys:
> I think @SMoraisAnsys suggested replacing all `lower()` calls with `casefold()` calls in the file, not the other way around.
> 
> I was unaware of `casefold()` until now, but it seems to be the recommended approach. Thanks to @ansys-satyajeet for introducing me to it.

Indeed, please use `casefold()` instead of `lower()` since it covers more use cases (which we should be handling AFAIK).

src/ansys/aedt/core/modules/material_lib.py
ansys-satyajeet: Sure. I will add the coverage for this line.

SMoraisAnsys (commented):
LGTM

@Samuelopez-ansys That's the PR I mentioned in #5639. What do you think about using `casefold` instead of `lower` ? If you also agree with that idea we could add it.
Note that this will lead to allowing users to leverage their own alphabet / characters and that might not be wanted. If you want we can also discuss about it in tomorrow's meeting.

Samuelopez-ansys:
@SMoraisAnsys @lorenzovecchietti I think this PR can be merged

SMoraisAnsys (approved):
LGTM

lorenzovecchietti (approved):

Issue #5504, opened 2024-11-29
Improve code statistics


Samuelopez-ansys:
**Code Quality Metrics**

- Improve Code Coverage: Increase the percentage of code covered by automated tests (unit, integration, and system tests).

- Reduce Cyclomatic Complexity: Break down complex functions into smaller, simpler ones. High complexity can lead to harder-to-maintain code.

**Performance Metrics**

- Optimize Execution Time: Profile the code using tools like cProfile or line_profiler to identify and optimize bottlenecks.

- Minimize Memory Usage: Use memory profiling tools like memory_profiler to detect inefficient memory usage and optimize data structures.

**Maintainability Metrics**

- Reduce Technical Debt: Address warnings, code smells, and deprecated functionality

- Improve Documentation: Ensure that all functions, classes, and modules are well-documented with clear examples. This helps developers understand and extend the codebase.

**Scalability Metrics**

- Improve Modularity: Break the code into smaller, reusable modules or libraries. Modular code is easier to test and extend.

PR #5631, opened 2025-01-07
FEAT: Add min version decorator

SMoraisAnsys:
## Description
This feature adds the ability to decorate a method with a check on AEDT version.
This way, check the version inside the method is no longer required and one can see right away that there is a limitation on the method.

The decorator should be compatible in many classes of PyAEDT. However, there is a limitation which requires the decorated object to be able to access the AEDT desktop.
Current ways to achieve such check is to leverage any existing attribute among `odesktop`, `_odesktop` and `_desktop`.

## Issue linked
Associated to #5504 

## Checklist

## Note

The test are performed on a mock object to simplify things but I also tested it on a fake Desktop and it worked just fine, here is the class I used to test with real AEDT object

```python
class DummyDesktop(Desktop):
    """Dummy class to test min version."""

    def __new__(cls, *args, **kwargs):
        instance = super(Desktop, cls).__new__(cls, *args, **kwargs)
        return instance

    def __init__(self):
        super().__init__()

    @min_aedt_version(PREVIOUS_YEAR_VERSION)
    def old_method(self):
        pass

    @min_aedt_version(NEXT_YEAR_VERSION)
    def future_method(self):
        pass
```

### DIFF
--- a/src/ansys/aedt/core/generic/checks.py
+++ b/src/ansys/aedt/core/generic/checks.py
@@ -67,11 +67,11 @@ def min_aedt_version(min_version: str):
 
     def aedt_version_decorator(method):
         def wrapper(self, *args, **kwargs):
-            odesktop = fetch_odesktop_from_common_attributes_names(self)
-            if odesktop is None:
-                odesktop = fetch_odesktop_from_private_app_attribute(self)
-            if odesktop is None:
-                odesktop = fetch_odesktop_from_desktop_class(self)
+            odesktop = (
+                fetch_odesktop_from_common_attributes_names(self)
+                or fetch_odesktop_from_private_app_attribute(self)
+                or fetch_odesktop_from_desktop_class(self)
+            )
             if odesktop is None:
                 raise AEDTRuntimeError("The AEDT desktop object is not available.")
 
--- a/src/ansys/aedt/core/generic/checks.py
+++ b/src/ansys/aedt/core/generic/checks.py
@@ -1,0 +1,71 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2021 - 2025 ANSYS, Inc. and/or its affiliates.
+# SPDX-License-Identifier: MIT
+#
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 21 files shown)


MaxJPRey (approved):
This is clearly a must have feature. Thanks @SMoraisAnsys . Let's make sure we promote its extensive usage.

tests/unit/test_utils.py
```diff
@@ -266,3 +269,58 @@ def test_write_toml(tmp_path):
...
+
+        odesktop = MagicMock()
+        odesktop.GetVersion.return_value = CURRENT_YEAR_VERSION
+
+        @min_aedt_version(PREVIOUS_YEAR_VERSION)
+        def old_method(self):
+            pass
+
+    dummy = Dummy()
+
```
MaxJPRey, line 292:
```suggestion
```

Samuelopez-ansys (dismissed):

src/ansys/aedt/core/generic/checks.py
```diff
@@ -46,25 +46,43 @@ def min_aedt_version(min_version: str):
...
-                    )
-                else:
-                    return method(self, *args, **kwargs)
+            odesktop = fetch_odesktop_from_common_attributes_names(self)
+            if odesktop is None:
+                odesktop = fetch_odesktop_from_private_app_attribute(self)
+            if odesktop is None:
+                odesktop = fetch_odesktop_from_desktop_class(self)
+            if odesktop is None:
+                raise AEDTRuntimeError("The AEDT desktop object is not available.")
```
gmalinve, line 76: @SMoraisAnsys what do you think of this instead?
```
odesktop = (
    fetch_odesktop_from_common_attributes_names(self)
    or fetch_odesktop_from_private_app_attribute(self)
    or fetch_odesktop_from_desktop_class(self)
)
if odesktop is None:
    raise AEDTRuntimeError("The AEDT desktop object is not available.")
```

SMoraisAnsys: Indeed, it would be better !

Samuelopez-ansys (approved):
Great job!

PR #5633, opened 2025-01-07
REFACTOR: Add possibility to specify python venv directory

hui-zhou-a:
## Description

## Issue linked

## Checklist

### DIFF
--- a/doc/source/Resources/pyaedt_installer_from_aedt.py
+++ b/doc/source/Resources/pyaedt_installer_from_aedt.py
@@ -40,10 +40,14 @@ is_windows = not is_linux
 
 VENV_DIR_PREFIX = ".pyaedt_env"
 
-VENV_DIR = None  # "e:/pyaedt_env"
-if VENV_DIR:
-    pass
-elif is_windows:
+"""
+It is possible create Python virtual environment in a specific directory by setting variable VENV_DIR. 
+For example,
+VENV_DIR = "e:/pyaedt_env"
+"""
+VENV_DIR = None
+
+if not VENV_DIR and is_windows:
     VENV_DIR = os.path.join(os.environ["APPDATA"], VENV_DIR_PREFIX)
 else:
     VENV_DIR = os.path.join(os.environ["HOME"], VENV_DIR_PREFIX)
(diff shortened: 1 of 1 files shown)


hui-zhou-a:
@Samuelopez-ansys I made minor change in the installer file. Just to make it possible to specify python venv directory. Please share your thought.

doc/source/Resources/pyaedt_installer_from_aedt.py
```diff
@@ -39,6 +39,16 @@
 
 
 VENV_DIR_PREFIX = ".pyaedt_env"
+
+VENV_DIR = None  # "e:/pyaedt_env"
+if VENV_DIR:
+    pass
+elif is_windows:
```
Samuelopez-ansys, line 46: Just to avoid the first "if", I would do:

```suggestion
if not VENV_DIR and is_windows:
```

doc/source/Resources/pyaedt_installer_from_aedt.py
```diff
@@ -39,6 +39,16 @@
 
 
 VENV_DIR_PREFIX = ".pyaedt_env"
+
+VENV_DIR = None  # "e:/pyaedt_env"
```
Samuelopez-ansys, line 43: Please remove the comment and add a brief explanation in the previous line (as a comment), to explain that the user can define a specific virtual environment.

Samuelopez-ansys (changes requested):

doc/source/Resources/pyaedt_installer_from_aedt.py
hui-zhou-a: > Just to avoid the first "if", I would do:

The first if is needed. Otherwise it will fail into Linux path.

Samuelopez-ansys:
@hui-zhou-a I think the VENV_DIR should include also the python_version, what I mean is that now VENV_DIR is just e:\venv

But if you analyze the code, the script will try to access something like:

e:\venv\310\Scripts

Which is not correct, because if the user pass a custom venv, I think the path passed will be directly e:\venv\310, also considering that if you create the venv manually, very likely the 310 will not exist.

doc/source/Resources/pyaedt_installer_from_aedt.py
```diff
@@ -207,11 +221,11 @@ def install_pyaedt():
         python_version = "3_7"
 
     if is_windows:
-        venv_dir = Path(os.environ["APPDATA"], VENV_DIR_PREFIX, python_version)
+        venv_dir = Path(VENV_DIR, python_version)
```
Samuelopez-ansys, line 224: This is what I mean

Samuelopez-ansys (approved):

Issue #5585, opened 2024-12-17
Circuit EMI receiver plotting function does not have a property for RBW and RBW Factor for 25R1


cmesibov:
### Before submitting the issue

- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### Description of the bug

Aedt release 25R1, pyaedt for Circuit 2D reporting does not have a property for RBW and RBW Factor.

Here is a code snippet the works for 24R1 and 24R2:

    emi_receiverPlot = circuitObj.post.reports_by_category.emi_receiver(emi_receiverList, solutionName)
    emi_receiverPlot.emission = 'CE'
    emi_receiverPlot.window = "Rectangular"
    emi_receiverPlot.max_frequency = "108MHz"
    emi_receiverPlot.time_start = "0ps"
    emi_receiverPlot.time_stop = Cispr25CemiTransientSetup[1]
    emi_receiverPlot.band = "4" # 150KHz - 108MHz
    emi_receiverPlot.create(emi_receiverPlotName)
    emi_receiverPlot.edit_x_axis_scaling(min_scale="150KHz", max_scale="108MHz")
    emi_receiverPlot.edit_x_axis_scaling(linear_scaling=False)
    emi_receiverPlot.edit_y_axis_scaling(linear_scaling=True, min_scale="-125", max_scale="100")

It needs the following to be added, for example:
    emi_receiverPlot.RBW = "1"   # 120KHz
    emi_receiverPlot.RBW_Factor = "2"   # 1.0

### Steps To Reproduce

Function properties missing

### Which Operating System are you using?

Windows

### Which Python version are you using?

3.10

### Installed packages

annotated-types==0.7.0
ansys-pythonnet==3.1.0rc4
attrs==24.2.0
certifi==2024.8.30
cffi==1.17.1
charset-normalizer==3.4.0
click==8.1.7
click-plugins==1.1.1
cligj==0.7.2
clr-loader==0.2.6
colorama==0.4.6
contourpy==1.3.0
cycler==0.12.1
defusedxml==0.7.1
docutils==0.21.2
-e git+https://github.com/ansys-internal/EMI_automation.git@4d5120c3c3cd15f74c1bff4ced8c594c7349fce6#egg=emi_utilities
et_xmlfile==2.0.0
exceptiongroup==1.2.2
fast_simplification==0.1.7
fiona==1.10.1
flit==3.10.1
flit_core==3.10.1
fonttools==4.54.1
fpdf2==2.8.1
[...]

PR #5640, opened 2025-01-08
FEAT: Added support for 25R1 EMI Receiver component RBW and RBW Factor feat…

cmesibov:
## Description
This PR adds support for 25R1 EMI Receiver component RBW and RBW Factor settings.

## Issue linked
Issue #5585

## Checklist
- [X] I have tested my changes locally.
- [X] I have added necessary documentation or updated existing documentation.
- [X] I have followed the coding style guidelines of this project.
- [X] I have reviewed my changes before submitting this pull request.
- [X] I have agreed with the Contributor License Agreement ([CLA](https://developer.ansys.com/form/cla-acceptance)).

### DIFF
--- a/src/ansys/aedt/core/visualization/report/emi.py
+++ b/src/ansys/aedt/core/visualization/report/emi.py
@@ -115,7 +115,7 @@ class EMIReceiver(CommonReport):
         str
             RBW Factor setting.
         """
-        return self._legacy_props["context"].get("RBW_factpr", None)
+        return self._legacy_props["context"].get("RBW_factor", None)
 
     @rbw_factor.setter
     def rbw_factor(self, value):
--- a/src/ansys/aedt/core/visualization/report/emi.py
+++ b/src/ansys/aedt/core/visualization/report/emi.py
@@ -29,6 +29,7 @@ This module provides all functionalities for creating and editing reports.
 
 """
 
+
 from ansys.aedt.core.generic.general_methods import generate_unique_name
 from ansys.aedt.core.generic.general_methods import pyaedt_function_handler
 from ansys.aedt.core.visualization.report.common import CommonReport
@@ -53,6 +54,8 @@ class EMIReceiver(CommonReport):
         self._emission = "CE"
         self.overlap_rate = 95
         self.band = "0"
+        self.rbw = "0"
+        self.rbw_factor = "0"
         self.primary_sweep = "Freq"
 
     @property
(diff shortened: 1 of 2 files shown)


SMoraisAnsys (changes requested):
Hi @cmesibov thanks for the contribution. There are a few things that should be modified in your contribution. First, could you add a description to your PR ?

Now more dev oriented:
- Since you added attributes `rbw` and `rbw_factor` with default values, it doesn't make sense to create the property getters (or the behavior might not be the one you expect). 
- The previous default value for RBW was 9000Hz, could you leave that same default value for backward compatibility ?
- From the previous behavior, I imagine that RBW must be a string so I'm not sure it's needed to cast it into a string in `emission`
- Can you add some tests to ensure that things work as expected ?

Finally, you told me that you would test your changes on previous AEDT versions. Did you do that ?

cmesibov:
@SMoraisAnsys , Thanks for the review.  The RBW and RBW Factor settings are set by index, not by value.  Therefore the  text setting of "900MHz" was in error to begin with.  The new init defaults of zero should be the default setting.  I have tested the changes on my own project for 24R2 and 25R1.  Would adding RBW and RWB Factor settings to test_76_emi_receiver be adequate?

SMoraisAnsys:
> test_76_emi_receiver

I don't know about that. If you have a simple reproducer that would be great. Otherwise, modifying an existing one is also a solution. From the PR title, some changes (most likely the RBW Factor) seem to be related to a specific version - I mean that it requires a minimum version. If that's the case, would you consider having a look at https://github.com/ansys/pyaedt/pull/5631 ?

Devin-Crawford:
@cmesibov : I updated the title - otherwise it will fail in the checks.

Devin-Crawford:
@cmesibov : It looks like this change fixes an error, but there was not defect submitted. For future reference it will be easier to understand what you did if you submit an issue describing what's wrong and then refer to that issue when you do the PR. Otherwise, this looks good.

SMoraisAnsys:
> @cmesibov : It looks like this change fixes an error, but there was not defect submitted. For future reference it will be easier to understand what you did if you submit an issue describing what's wrong and then refer to that issue when you do the PR. Otherwise, this looks good.

From my understanding, it seems that RBW was not something we could use before and setting whatever value did not cause an issue.
Here the changes are introducing the ability to work with RBW and RBW Factor.
@cmesibov Can you add tests as previously asked ?
Also, is the change on RBW Factor backward compatible ? Should we add a decorator to state that one shouldn't update the RBW value and RBW factor value with previous version of AEDT (see the changes introduced in https://github.com/ansys/pyaedt/pull/5631) ?

cmesibov:
@Devin-Crawford HI Devin,  I open a pyaedt git hub issue #5585 for the EMI receiver update for 25R1.

cmesibov:
@SMoraisAnsys I will be pushing the update for the added tests today.  Sorry for the delay.

cmesibov:
"Also, is the change on RBW Factor backward compatible ? Should we add a decorator to state that one shouldn't update the RBW value and RBW factor value with previous version of AEDT (see the changes introduced in #5631) ?"

Testing has shown that the 25R1 targeted feature doesn't cause a desktop error in 24R1 and 24R2.  But having the decorator sounds like a good idea.

cmesibov:
@SMoraisAnsys I have pushed updates with RBW and RBW factor values added to test_76_emi_receiver().  I think this will satisfy your requested change.

SMoraisAnsys:
I'll look at the changes when the test are passing

Samuelopez-ansys:
I have tested it in 2024R2 which is the 1st compatible version of the feature, and as @cmesibov said, it is not impacting

Samuelopez-ansys (approved):

PR #5666, opened 2025-01-16
FIX: Convert near field list name order

### LINKED ISSUE #5664: Bug located in test_47_convert_near_field

### Description of the bug

Currently when one runs the test `test_47_convert_near_field` there might be an error depending on the order the DAT files are explored.

For example, with the current data test, the implementation of `convert_nearfield_data` can start with `data_Ey_xmax.dat` or with `data_Ez_xmax.dat`.
However, this has a huge impact on the implementation since one file has the former contains 3 lines of data and the second 4.

I tested a dumb workaround consisting in sorting the file exploration `for data_file in sorted(file_names):` instead of `for data_file in file_names:` and the test is passing as expected.

However, this either mean that the data used for testing isn't correct or that we should change our implementation of `convert_nearfield_data`.

### Steps To Reproduce

As explain above

Linux

3.10

### Installed packages

Not relevant

Samuelopez-ansys:
## Description
Depeding on OS, the file list order could change, and the files could contain different size data. With this PR, if the field does not contain information, it is filled with a 0 value, because it means that there is no field value at this field component.

## Issue linked
Close #5664 

## Checklist

### DIFF
--- a/src/ansys/aedt/core/visualization/advanced/misc.py
+++ b/src/ansys/aedt/core/visualization/advanced/misc.py
@@ -154,14 +154,15 @@ def convert_nearfield_data(dat_folder, frequency=6, invert_phase_for_lower_faces
     index = 1
     for el in list(file_keys):
         for k in range(index, index + len(components[el].x)):
-            row = []
-            row.append(k)
-            row.append(components[el].x[k - index])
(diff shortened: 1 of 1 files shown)


svandenb-dev (approved):
LGTM

PR #5670, opened 2025-01-17
CHORE: SetupMaxwell refactoring to save fields and new SweepMaxwell for EC solvers

gmalinve:
## Description
**In this PR I want to:
- Improve the SetupMaxwell class in order to be able to save fields for all solvers.
- Add a new class in solve_sweep to handle EC freq. sweeps.
- Add/improve the already exisiting method to add frequency sweeps for EC in order to return the sweep object.**

## Issue linked

## Checklist


src/ansys/aedt/core/modules/setup_templates.py
```diff
@@ -292,13 +292,12 @@ def HFSS3DLayout_AdaptiveFrequencyData(freq):
...
-        "StartValue": "1e-08GHz",
-        "StopValue": "1e-06GHz",
-        "StepSize": "1e-08GHz",
+        "RangeType": "LinearStep",
+        "RangeStart": "1e-08GHz",
+        "RangeEnd": "1e-06GHz",
     }
 )
-subranges = dict({"Subrange": subrange})
+SweepEddyCurrent = subrange
```
jvela018, line 300: If "subrange" not used anywhere else, I'd just define 
`SweepEddyCurrent = dict({"RangeType": "LinearStep",
        "RangeStart": "1e-08GHz",
        "RangeEnd": "1e-06GHz"})`

src/ansys/aedt/core/modules/solve_setup.py
```diff
@@ -3717,6 +3716,104 @@ def enable_control_program(self, control_program_path, control_program_args=" ",
...
+                        range_props["RangeStep"] = f"{count}{units}"
+                    elif subrange_type == "LinearCount":
+                        range_props["RangeCount"] = count
+                    elif subrange_type == "SinglePoints":
+                        range_props["RangeEnd"] = f"{start}{units}"
+                    if "SweepRanges" in self.props.keys():
+                        if isinstance(self.props["SweepRanges"]["Subrange"], dict):
+                            temp = self.props["SweepRanges"]["Subrange"]
+                            self.props["SweepRanges"].pop("Subrange", None)
+                            self.props["SweepRanges"]["Subrange"] = [temp]
```
jvela018, line 3796: Can you explain lines 3792 downwards? 
1. Check that the keys are in the dictionary, 
2. Check that the dictionary from that key is an actual dictionary data type. 
3. Store the value of Subrange in a temporary variable
4. Remove Subrange from the initial entry
5. Replace Subrange with the one stored in the temp variable

Is this to add a Subrange as a List data type? I'm asking because it looks like you're removing something that you're just putting back in. Do I understand this logic correctly?

src/ansys/aedt/core/modules/solve_setup.py
```diff
@@ -3717,6 +3716,104 @@ def enable_control_program(self, control_program_path, control_program_args=" ",
...
+                else:
+                    if self.props["SaveFieldsType"] == "Custom":
+                        self.props.pop("SweepRanges", None)
+                    self.auto_update = False
+                    self.props["SaveFieldsType"] = "Every N Steps"
+                    self.props["N Steps"] = f"{count}"
+                    self.props["Steps From"] = f"{start}{units}"
+                    self.props["Steps To"] = f"{stop}{units}"
+                    self.update()
+                return True
```
jvela018, line 3810: The else in line 3801 only has one conditional. It could be replace by elif? This is just my personal choice but I don't mind it the way it is.

src/ansys/aedt/core/modules/solve_sweeps.py
```diff
@@ -827,6 +827,151 @@ def _get_args(self, props=None):
         return arg
 
 
+class SweepMaxwellEC(object):
```
jvela018, line 830: Do you think it'd be advantageous to have it's own methods to set_initial_freq(), get_initial_freq(), get_nsteps() etc. It doesn't have to be those names but in terms of what can be extracted, and set without going through the dictionaries. It seems that you're doing this by using props (e.g. sweep.props), but this is just an idea that maybe you could consider?

jvela018 (commented):
@gmalinve , 

I made some general comments that I'd like to iterate over.

src/ansys/aedt/core/modules/solve_setup.py
```diff
@@ -49,6 +49,7 @@
 from ansys.aedt.core.modules.solve_sweeps import SweepHFSS
 from ansys.aedt.core.modules.solve_sweeps import SweepHFSS3DLayout
 from ansys.aedt.core.modules.solve_sweeps import SweepMatrix
+from ansys.aedt.core.modules.solve_sweeps import SweepMaxwellEC
```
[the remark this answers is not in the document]
gmalinve: @jvela018 umm yes, good point here. We can discuss about that.
There are also methods that recall the Eddy current solver so not sure.

src/ansys/aedt/core/modules/solve_setup.py
gmalinve: @jvela018 yes, makes sense!

maxcapodi78 (approved):
LGTM

PR #5695, opened 2025-01-26
REFACTOR: General improvements

SMoraisAnsys:
## Description
General changes in the code base. The ones associated to:
- raising an `AEDTRuntimeError` instead of the `assert ..., ....` pattern is not up for discussion;
- handling a typo in argument name which was introduced when moving from old names to new ones, here the old argument was still used in `assign_thermal_map` method signature instead of `assignment`.

But the other modifications are up for discussion and could be generalized to other modules:
- use of `SOLUTIONS.Maxwell3d.SOMETHING` instead of `"SOMETHING"`;
- raise of an exception (mainly `AEDTRuntimeError`) instead of logging an error and returning False.

Note the I would like to open a discussion about `assign_matrix` whose current behavior is strange: multiple `solution_type` cases are handled inside of the implementation BUT if the `solution_type` is not in `(SOLUTIONS.Maxwell3d.Transient, SOLUTIONS.Maxwell3d.ElectricTransient)` then the method was logging an error and returning False.

@Samuelopez-ansys @maxcapodi78 @gmalinve What do you think about the transition of "logging and error and returning False" to "raising an exception" ? We already had a discussion on that point and the fact that methods are decorated with the pyaedt_handler decorator still ends up with a False value being returned when an error is raised. The additional value of the proposed approach is that it would allow:
- users to handle exceptions themselves;
- make it clear that a critical error has occurred;
- help to debug as exceptions provide a complete stack trace of the error.

## Issue linked
Associated to #5504 

## Checklist


src/ansys/aedt/core/maxwell.py
```diff
@@ -245,14 +242,16 @@ def set_core_losses(self, assignment, core_loss_on_field=False):
...
         """
-        if self.solution_type in ["EddyCurrent", "Transient"]:
-            assignment = self.modeler.convert_to_selections(assignment, True)
-            self.oboundary.SetCoreLoss(assignment, core_loss_on_field)
-            return True
-        else:
-            raise Exception("Core losses is only available with `EddyCurrent` and `Transient` solutions.")
-        return False
+        if self.solution_type not in (SOLUTIONS.Maxwell3d.EddyCurrent, SOLUTIONS.Maxwell3d.Transient):
+            raise AEDTRuntimeError("Solution type does not have matrix parameters")(
```
Samuelopez-ansys, line 246:
```suggestion
            raise AEDTRuntimeError(
```

Samuelopez-ansys (approved):
LGTM. Now we "only" need to do it for the other core modules

maxcapodi78 (approved):

maxcapodi78:
@Samuelopez-ansys @SMoraisAnsys  I like the idea of raising exception instead of returning False. It's correct and it's not breaking customer code.

MaxJPRey:
> @Samuelopez-ansys @SMoraisAnsys I like the idea of raising exception instead of returning False. It's correct and it's not breaking customer code.

I can only agree. Great work.

src/ansys/aedt/core/mechanical.py
```diff
@@ -334,11 +338,11 @@ def assign_thermal_map(
         #
         # Generate a list of model objects from the lists made previously and use to map the HFSS losses into Icepak.
         #
-        object_list = self.modeler.convert_to_selections(object_list, True)
-        if not object_list:
-            allObjects = self.modeler.object_names
+        assignment = self.modeler.convert_to_selections(assignment, True)
```
MaxJPRey, line 341: thanks for the renaming. 😃

MaxJPRey (approved):

PR #5702, opened 2025-01-29
FEAT: New extension version manager

### LINKED ISSUE #5732: Need an extension to manage PyAEDT/PyEDB version

To an average hardware engineer perspective, managing pyaedt/pyedb versions is trivial. With limited terminal and Python knowledge, simple tasks can be challenging. 

The intension is to implement an GUI for easy version management. 

Implement an built-in extension

_No response_

hui-zhou-a:
## Description
New extension Version manager.

## Issue linked
Close #5732 

## Checklist


doc/source/User_guide/extensions.rst
```diff
@@ -55,6 +55,13 @@ They are small automated workflows with a simple GUI.
 
             Lear how to convert projects from 2022R2 to newer versions.
 
+   .. grid-item-card:: Version manager
+            :link: pyaedt_extensions_doc/project/version_manager
+            :link-type: doc
+            :margin: 2 2 0 0
+
+            Manager pyaedt and pyedb versions.
```
MaxJPRey, line 63:
```suggestion
            Manage pyaedt and pyedb versions.
```

src/ansys/aedt/core/workflows/project/version_manager.py
```diff
@@ -0,0 +1,527 @@
...
+    if response.status_code == 200:
+        data = response.json()
+        return data["info"]["version"]
+    else:
+        return "Unknown"
+
+
+class VersionManager:
+    TITLE = "Version Manager {}".format(VERSION)
+    USER_GUIDE = "https://github.com/ansys-internal/pyaedt_version_manager"
```
Samuelopez-ansys, line 87: This is an internal website. I think you could point to the pyaedt documentation (once it is built)

Samuelopez-ansys:
@hui-zhou-a This extension is great! But could you please use the same style than the others? I think you can see the choke designer one which has already multi tabs like this one and probably it will be easier for you to adapt. 

The icon of the app when it is opened, is not the PyAnsys logo, and to respect consistency with the other extension you should also import it.

In summary, I think you could check the choke_designer toolkit and try to adapt the style

src/ansys/aedt/core/workflows/project/version_manager.py
MaxJPRey: Yes, please remove references to this org.

doc/source/User_guide/pyaedt_extensions_doc/project/version_manager.rst
```diff
@@ -0,0 +1,22 @@
...
+- Show Python version
+- Show the current PyAEDT&PyEDB versions
+- Show the latest PyAEDT&PyEDB release
+- Update PyAEDT&PyEDB to the latest version
+- Replace PyAEDT&PyEDB with specific version
+    - Default is main, which is the developer version
+    - Accepted inputs are branch name and version (0.34.0 for example)
+- Update PyAEDT&PyEDB from wheelhouse
+  - Check compatibility
+- Reset and update PyAEDT buttons in AEDT
```
SMoraisAnsys, line 22:
```suggestion
- Show the current PyAEDT and PyEDB versions
- Show the latest PyAEDT and PyEDB release available (if connected)
- Update PyAEDT and PyEDB to the latest version
- Install PyAEDT and PyEDB with a specific version
    - Default is main, which is the developer version
    - Accepted inputs are branch name and version (0.34.0 for example)
- Update PyAEDT and PyEDB from wheelhouse
  - Check compatibility
- Reset and update PyAEDT buttons in AEDT
```
Just wondering, are you sure that the accepted input is a version (e.g. 0.34.0) and not a tag (e.g. v0.34.0) ?

src/ansys/aedt/core/workflows/project/version_manager.py
```diff
@@ -0,0 +1,562 @@
...
+                xml_file = product / "TabConfig.xml"
+                if xml_file.exists():
+                    tree = ET.parse(xml_file)
+                    root = tree.getroot()
+                    panel_label = "Panel_PyAEDT_Extensions"
+                    for panel in root.findall("panel"):
+                        if panel.get("label") == panel_label:
+                            for button in panel.findall("button"):
+                                name = button.get("label")
+                                temp.append([product, name])
```
SMoraisAnsys, line 328: Can you have a look at what is done in customize_automation_tab.py, specifically

- the imports, see for example https://github.com/ansys/pyaedt/blob/main/src/ansys/aedt/core/workflows/customize_automation_tab.py#L32
- the parsing, see
https://github.com/ansys/pyaedt/blob/main/src/ansys/aedt/core/workflows/customize_automation_tab.py#L85-L96

src/ansys/aedt/core/workflows/project/version_manager.py
```diff
@@ -0,0 +1,562 @@
...
+    "Ansys software license agreement. Ansys does not warrant or support such "
+    "Third-Party Software.\n"
+    "Do you want to proceed ?"
+)
+
+
+def is_git_installed():
+    try:
+        # Run the command `git --version`
+        result = subprocess.run(["git", "--version"], stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
```
SMoraisAnsys, line 64:
```suggestion
        result = subprocess.run(["git", "--version"], stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
```
Add `check=True` ?

src/ansys/aedt/core/workflows/project/version_manager.py
```diff
@@ -0,0 +1,562 @@
...
+            if response.status_code == 200:
+                data = response.json()
+                released_version = data["info"]["version"]
+            else:
+                released_version = 0
+
+            if self.pyaedt_version > released_version:
+                subprocess.run([self.python_exe, "-m", "pip", "install", f"pyaedt=={released_version}"], check=True)
+            else:
+                subprocess.run([self.python_exe, "-m", "pip", "install", "-U", "pyaedt"], check=True)
```
SMoraisAnsys, line 373: I'm not sure about the behavior here. What's supposed to happen if `released_version == 0` ? Also, does it make sense to have a `pyaedt_version` that is higher than the last released version ?

src/ansys/aedt/core/workflows/project/version_manager.py
```diff
@@ -0,0 +1,562 @@
...
+            if response.status_code == 200:
+                data = response.json()
+                released_version = data["info"]["version"]
+            else:
+                released_version = 0
+
+            if self.pyedb_version > released_version:
+                subprocess.run([self.python_exe, "-m", "pip", "install", f"pyedb=={released_version}"], check=True)
+            else:
+                subprocess.run([self.python_exe, "-m", "pip", "install", "-U", "pyedb"], check=True)
```
SMoraisAnsys, line 429: Same as previous comment

src/ansys/aedt/core/workflows/project/version_manager.py
```diff
@@ -0,0 +1,562 @@
...
+
+            subprocess.run(
+                [
+                    self.python_exe,
+                    "-m",
+                    "pip",
+                    "install",
+                    "--no-cache-dir",
+                    "--no-index",
+                    f"--find-links=file:///{str(unzipped_path)}",
```
SMoraisAnsys, line 481: I'm not sure that f strings are working correctly with extensions, can you double check ?

SMoraisAnsys (changes requested):
I left some minor comments. Thanks for the contribution @hui-zhou-a !

doc/source/User_guide/pyaedt_extensions_doc/project/version_manager.rst
hui-zhou-a: I am sure it is a bug >_<

src/ansys/aedt/core/workflows/project/version_manager.py
hui-zhou-a: For your first question, it is a bug. Fixed.

self.pyedb_version can be dev version.which is higher than the release. In this case, install the release version.

src/ansys/aedt/core/workflows/project/version_manager.py
hui-zhou-a: same fix

src/ansys/aedt/core/workflows/project/version_manager.py
hui-zhou-a: I checked it is working on my machine. Is there a more reliable way?

src/ansys/aedt/core/workflows/project/version_manager.py
hui-zhou-a: added

hui-zhou-a:
@SMoraisAnsys I fixed what I can, please review again.

src/ansys/aedt/core/workflows/project/version_manager.py
Samuelopez-ansys: It works, because the extension will be launched in CPython. What you are thinking @SMoraisAnsys is about the PyAEDT installer, which is called directly in Ironpython, and the f-strings do not work. But again, this extension is a tkinter CPython extension.

src/ansys/aedt/core/workflows/project/version_manager.py
```diff
@@ -0,0 +1,544 @@
...
+UNKNOWN_VERSION = "Unknown"
+
+
+def get_latest_version(package_name):
+    response = requests.get(f"https://pypi.org/pypi/{package_name}/json")
+    if response.status_code == 200:
+        data = response.json()
+        return data["info"]["version"]
+    else:
+        return UNKNOWN_VERSION
```
SMoraisAnsys, line 70: @hui-zhou-a I'm not 100% sure that the method will always behave as planned. Indeed, it might return the latest version released by date and not in term of version. For example, let us say that I make two release in the following order 1.30.2 and a patch release 1.15.3. Will the code return 1.30.2 (version comparison) or 1.15.3 (date comparison).

Here is a snippet of code if the unwanted behavior ever happen. It consist in only working with stable versions, aka X.Y.Z and removes any X.Y.dev0 or X.Y.Z-whatever.

```python
def parse_version(version_str):
    match = re.match(r"^(\d+)\.(\d+)\.(\d+)$", version_str)
    if match:
        version_parts = tuple(map(int, match.groups()))
        return version_parts

def get_latest_version(package_name):
    response = requests.get(f"https://pypi.org/pypi/{package_name}/json")
    if response.status_code == 200:
        data = response.json()
        versions = list(data["releases"].keys())
        # Keep only stable version, i.e. version of the form X.Y.Z
        filtered_versions = [
            version for version in versions 
            if re.match(r"^\d+\.\d+\.\d+$", version)
        ]
        latest_version = max(filtered_versions, key=parse_version)
        return latest_version
    else:
        return UNKNOWN_VERSION
```

SMoraisAnsys (dismissed):
LGTM but have a look at my concern please.
Just in case we ever face that behavior !

SMoraisAnsys (approved):

src/ansys/aedt/core/workflows/project/version_manager.py
hui-zhou-a: @SMoraisAnsys I have not considered the scenario you described. I don't know if data["info"]["version"] returns the highest version or latest update. I intension is to get the highest version. It is potentially a bug. I like your proposal. I saw you approved this PR, so let's have this fix in a patch.

PR #5717, opened 2025-02-02
REFACTOR: Add Hfss improvements

SMoraisAnsys:
## Description
Follow up of #5695 applied to `Hfss`

## Issue linked
#5504 

## Checklist

### DIFF
--- a/src/ansys/aedt/core/hfss.py
+++ b/src/ansys/aedt/core/hfss.py
@@ -36,8 +36,10 @@ import warnings
 from ansys.aedt.core.application.analysis_3d import FieldAnalysis3D
 from ansys.aedt.core.application.analysis_hf import ScatteringMethods
 from ansys.aedt.core.generic.constants import INFINITE_SPHERE_TYPE
+from ansys.aedt.core.generic.constants import SOLUTIONS
 from ansys.aedt.core.generic.data_handlers import _dict2arg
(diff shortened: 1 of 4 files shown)


Samuelopez-ansys (approved):
LGTM

SMoraisAnsys:
@Samuelopez-ansys Just wondering if, for each of the following cases, I shouldn't raise an exception:
- `_create_boundary` is currently logging an error when the boundary is not created correctly and returning `result` (aka `False`)
- `_create_native_component` is returning `None` and logging an error

Also, in multiple places, there is this `self.logger.error("Radiation Field not available in this solution.")` which does not prevent the code from running. Should I raise an exception ? Log a warning ? Leave it as it is (that might be the most suitable approach)

Samuelopez-ansys:
Hi @SMoraisAnsys ,

About "Radiation Field not available in this solution." We can raise an exception because it is a bad "usage" of PyAEDT. 

Regarding _create_boundary, this is a good question, I would expand the raise Exception as much as possible, so I would also add the Exception to create_boundary.

About create_component the same, I would raise  an exception too, because it is not expected

Samuelopez-ansys (approved):

MaxJPRey (approved):
Definetly a lot of quality in this PR. Thanks @SMoraisAnsys .

PR #5755, opened 2025-02-10
REFACTOR: Use Mixin to create boundary

SMoraisAnsys:
## Description
Using Mixin is a good practice, we should try to refactor the code and follow this kind of approach. This PR is a starting point in that direction and is currently only adding a CreateBoundaryMixin class.

A Mixin is used to add reusable functionality to multiple classes without creating deep inheritance hierarchies. Typically, such class are small, contain additional methods and **should not be instantiated on its own.**

> [!NOTE]
Currently the changes are not applied to all main classes of PyAEDT as it is more a proof of concept. I'm only applying it in classes where the idea of a method used to create a boundary was already implemented.

## Issue linked
Associated to #5504 

## Checklist

### DIFF
--- a/src/ansys/aedt/core/mixins.py
+++ b/src/ansys/aedt/core/mixins.py
@@ -70,10 +70,11 @@ class CreateBoundaryMixin:
         """
         try:
             bound = BoundaryObject(self, name, props, boundary_type)
-            result = bound.create()
-            if result:
-                self._boundaries[bound.name] = bound
-                self.logger.info(f"Boundary {boundary_type} {name} has been created.")
-                return bound
+            if not bound.create():
+                raise AEDTRuntimeError(f"Failed to create boundary {boundary_type} {name}")
+
+            self._boundaries[bound.name] = bound
+            self.logger.info(f"Boundary {boundary_type} {name} has been created.")
+            return bound
         except GrpcApiError as e:
             raise AEDTRuntimeError(f"Failed to create boundary {boundary_type} {name}") from e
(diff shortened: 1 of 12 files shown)


src/ansys/aedt/core/mixins.py
```diff
@@ -0,0 +1,77 @@
...
+
+class CreateBoundaryMixin:
+    """Mixin that provides methods to apply boundary conditions.
+
+    This mixin is designed to be used in classes that require boundary condition.
+    Methods provided by this mixin should be implemented or extended by subclasses
+    to define specific boundary condition behaviors.
+
+    For example:
+        ```python
```
MaxJPRey, line 38: Just to be sure, do we need to specify the python type of code?

MaxJPRey (dismissed):
Great idea those mixin.

gmalinve:
@SMoraisAnsys thanks for taking care of this!
As I already mentioned I can provide many of these cases when you need.
I have a question for you though, what's the difference of using mixins and multiple inheritance?

SMoraisAnsys:
> @SMoraisAnsys thanks for taking care of this! As I already mentioned I can provide many of these cases when you need. I have a question for you though, what's the difference of using mixins and multiple inheritance?

Both involve multiple base classes so they are not too far from each other. The main difference is that Mixin is a lightweight class that should provide a small piece of functionality to be combined with other classes. Ofc course the default behavior can be extended :)

I would wait to extend this new file with new content as I need to apply this Mixin to multiple base module in PyAEDT (e.g. mechanical, ...)

Samuelopez-ansys (approved):

PR #5828, opened 2025-02-25
REFACTOR: Pathlib refactor: aedt_logger.py, common_rpc.py, circuit_netlist.py and test_47_CircuitNetlist.py 

jvela018:
Issue #5180 cont.

### DIFF
--- a/src/ansys/aedt/core/circuit_netlist.py
+++ b/src/ansys/aedt/core/circuit_netlist.py
@@ -158,7 +158,7 @@ class CircuitNetlist(AnalysisCircuitNetlist, object):
 
         Parameters
         ----------
-        input_file : str, class, optional
+        input_file : str or :class:'pathlib.Path', optional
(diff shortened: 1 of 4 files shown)


src/ansys/aedt/core/circuit_netlist.py
```diff
@@ -158,38 +159,40 @@ def browse_log_file(self, input_file=None):  # pragma: no cover
 
         Parameters
         ----------
-        input_file : str, optional
+        input_file : str, class, optional
```
Samuelopez-ansys, line 161: Why class type?

Samuelopez-ansys (approved):
LGTM

src/ansys/aedt/core/circuit_netlist.py
jvela018: Made the change accordingly to str or :class:'pathlib.Path', optional .

PR #5840, opened 2025-02-26
FEAT: Method to get max temp location

lorenzovecchietti:
Icepak users seldom need the field calculator due to the availability of the field summary. However, it proves efficient for determining the location of an object's maximum temperature.


Samuelopez-ansys (approved):
Good job! Good new feature

PR #5843, opened 2025-02-26
FIX: Unstable test fix due to FilterSolutions DLL

ramin4667:
## Description
**This PR updates the FilteSolutions API for testing the updated stable dll **

## Issue linked
**The 25.1 release tests are unstable because of the not checked-in license after the end of the test.**

## Checklist


src/ansys/aedt/core/filtersolutions_core/dll_interface.py
```diff
@@ -178,7 +179,10 @@ def api_version(self) -> str:
             API version.
         """
         version = self.get_string(self._dll.getVersion)
-        return version
+        match = re.search(r"Version (\d{4}) R(\d+)", version)
+        version_year = int(match.group(1))
+        version_release = int(match.group(2))
+        return version, float(f"{version_year}.{version_release}")
```
myoung301, line 185: Let's return a string to be consistent with other version usage in PyAEDT. If we reformat it to be "YYYY.VV" then we can just return that string, no need to return a tuple.

.github/workflows/ci_cd.yml
```diff
@@ -197,7 +197,7 @@ jobs:
           PYTHONMALLOC: malloc
         run: |
           .venv\Scripts\Activate.ps1
-          pytest ${{ env.PYTEST_ARGUMENTS }} --timeout=600 -m solvers --deselect=tests/system/solvers/test_45_FilterSolutions
+          pytest ${{ env.PYTEST_ARGUMENTS }} -m solvers
```
myoung301, line 200: For this PR, remove the changes to this file.

myoung301 (changes requested):
A couple changes requested.

.github/workflows/ci_cd.yml
ramin4667: Applied.

src/ansys/aedt/core/filtersolutions_core/dll_interface.py
ramin4667: Changed to string.

src/ansys/aedt/core/filtersolutions_core/dll_interface.py
```diff
@@ -177,8 +183,12 @@ def api_version(self) -> str:
         str
             API version.
         """
-        version = self.get_string(self._dll.getVersion)
-        return version
+        api_version = self.get_string(self._dll.getVersion)
+        match = re.search(r"Version (\d{4}) R(\d+)", api_version)
+        api_version_year = int(match.group(1))
```
myoung301, line 188: @ramin4667 I think match.group(1/2) are strings so no need to convert them to ints and back to strings again. We should be able to use them directly in the format string.

ramin4667: Integer cast is removed.

myoung301 (approved):

PR #5846, opened 2025-02-27
REFACTOR: refactoring of analysis.py

gmalinve:
## Description

## Issue linked

## Checklist

### DIFF
--- a/src/ansys/aedt/core/application/analysis.py
+++ b/src/ansys/aedt/core/application/analysis.py
@@ -49,13 +49,13 @@ from ansys.aedt.core.generic.constants import PLANE
 from ansys.aedt.core.generic.constants import SETUPS
 from ansys.aedt.core.generic.constants import SOLUTIONS
 from ansys.aedt.core.generic.constants import VIEW
-from ansys.aedt.core.generic.general_methods import _arg_with_dim
 from ansys.aedt.core.generic.general_methods import filter_tuple
 from ansys.aedt.core.generic.general_methods import generate_unique_name
 from ansys.aedt.core.generic.general_methods import is_linux
 from ansys.aedt.core.generic.general_methods import is_windows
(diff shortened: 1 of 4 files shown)


Samuelopez-ansys (approved):
Good catch

jvela018 (dismissed):
LGTM

tests/unit/test_quantity_numbers.py
```diff
@@ -101,3 +101,8 @@ def test_quantity_comparison():
     assert q2 < q1
     assert q1 >= q2
     assert q2 <= q1
+
+
+def test_quantity_str():
+    q = Quantity("A + Bmm")
+    assert str(q) == "A + Bmm"
```
jvela018, line 108: @gmalinve I don't think this test works. When Quantity does the decomposition, it seems (at least from a quick read) that it doesn't handle the addition operator in the string. 
image

gmalinve: @jvela018, if you debug you'll see that the case Quantity("A + Bmm") doesn't go inside the check if loc because there's no a valid units string found so it simply return: 
float_value = "A + Bmm"
units = ""

jvela018: @gmalinve , but if that's the case, float value is casted as float(), considering that your value is float_value = float("A + Bmm"), wouldn't that fail as your casting to float, and if it doesn't, is it just taking the ASCII numerical value of it? /Jon

gmalinve: in the case of "A + Bmm" the code doesn't go inside the if loc check so float_value = float("A + Bmm") is never hit and thus its returned value is "A + Bmm"

Samuelopez-ansys (approved):

Samuelopez-ansys (approved):

PR #5854, opened 2025-02-27
FIX: Licensingclient path update for version 252

ramin4667:
## Description
**This PR updates PyAEDT for the changed licensingclient folder in release 25.2.**

## Issue linked
**Error for not finding licensingclient folder.**

## Checklist


src/ansys/aedt/core/generic/general_methods.py
```diff
@@ -903,11 +903,15 @@ def available_license_feature(
 
     if not input_dir:
         input_dir = list(aedt_versions.installed_versions.values())[0]
-
+        ansys_version = float(list(aedt_versions.installed_versions.keys())[0])
```
myoung301, line 906: Is the `[0]` element the vesion of AEDT being used or the newest? I think this has to be the version used, not just the latest version. Also, I don't think you want to convert to float because float comparisons are problematic. I have seen string comparisons in many of the other version checks in the PyAEDT code. 

It might be easier here to check not check the version but rather check if the directory exists. If it exists in the first location then continue with that directory. Otherwise, set it to the second location. If you do it, be sure to add a comment about the reason for checking for a second location (location changed in 25R2). What do you think of this approach instead?

myoung301 (changes requested):

src/ansys/aedt/core/generic/general_methods.py
ramin4667: Applied.

src/ansys/aedt/core/generic/general_methods.py
```diff
@@ -905,10 +905,15 @@ def available_license_feature(
     if not input_dir:
         input_dir = list(aedt_versions.installed_versions.values())[0]
 
+    # The licensing client location has changed in the Windows 25R2 AEDT installation.
+    # Verification is required for Linux.
```
myoung301, line 909: I reviewed the code change and informational e-mail to confirm it applies to both OSs.
```suggestion
    # The licensing client location has changed in the 25R2 AEDT installation.
```

src/ansys/aedt/core/generic/general_methods.py
```diff
@@ -905,10 +905,15 @@ def available_license_feature(
...
+    if not check_if_path_exists(os.path.join(input_dir, "licensingclient")):
+        input_dir = os.path.dirname(input_dir)
+
     if is_linux:
         ansysli_util_path = os.path.join(input_dir, "licensingclient", "linx64", "lmutil")
     else:
-        ansysli_util_path = os.path.join(input_dir, "licensingclient", "winx64", "lmutil")
+        ansysli_util_path = os.path.join(input_dir, "licensingclient", "winx64", "lmutil.exe")
```
myoung301, line 916: I want to confirm this change was necessary. Did you get an error from the lack of the extension? On Windows, I think that a command can exclude the extension (doesn't have to include `.exe`).

myoung301 (changes requested):
A couple minor questions.

src/ansys/aedt/core/generic/general_methods.py
ramin4667: Removed.

src/ansys/aedt/core/generic/general_methods.py
ramin4667: The extension is removed.

src/ansys/aedt/core/generic/file_utils.py
```diff
@@ -948,11 +948,17 @@ def available_license_feature(
             port = name_env[0]
             name = name_env[1]
 
-    if not input_dir:
+    if not input_dir and aedt_versions.latest_version:
```
myoung301, line 951: @maxcapodi78 This logic is different than what you proposed. This appears to work for our use case, but is it ok to prefer the latest_version over current_version as we do here?

myoung301: I think it's best to favor the latest_version for licensing so this makes sense to me. The tests pass and this fixed the issue we were seeing locally for 25.2.

myoung301 (approved):

src/ansys/aedt/core/generic/file_utils.py
```diff
@@ -948,11 +948,17 @@ def available_license_feature(
...
+        input_dir = Path(aedt_versions.installed_versions[aedt_versions.latest_version])
+    elif not input_dir:
         input_dir = Path(aedt_versions.installed_versions[aedt_versions.current_version])
     else:
         input_dir = Path(input_dir)
 
+    # The licensing client location has changed in the Windows 25R2 AEDT installation.
+    if not check_if_path_exists(os.path.join(input_dir, "licensingclient")):
+        input_dir = Path(os.path.dirname(input_dir))
```
SMoraisAnsys, line 960: I'm not sure to understand how this behave with the following code defining `ansysli_util_path`. Can you example why, if the path does not exist, it is still used afterward ?

SMoraisAnsys (changes requested):

src/ansys/aedt/core/generic/file_utils.py
ramin4667: It checks if the `licensing client` folder exists in the input_dir. In V251, the input_dir refers to `C:/Program Files/ANSYS Inc/v251/AnsysEM`, which contains the `licensing client` folder. For V252, I was getting an error about not finding the folder since it was moved one level up to `C: /Program Files/ANSYS Inc/v252`. In line 960 the `input_dir` is moved one level up where the folder exists. The next line generates `ansys_util_path by referring to the updated `input_dir`.

src/ansys/aedt/core/generic/file_utils.py
```diff
@@ -948,11 +948,18 @@ def available_license_feature(
...
-    if not input_dir:
+    if not input_dir and aedt_versions.current_version:
         input_dir = Path(aedt_versions.installed_versions[aedt_versions.current_version])
+    elif not input_dir:
+        input_dir = Path(aedt_versions.installed_versions[aedt_versions.latest_version])
     else:
         input_dir = Path(input_dir)
 
+    # Checks if the licensingclient folder exists in input_dir. In V252, the folder was moved up one level,
+    # so input_dir is updated accordingly before generating ansys_util_path.
```
myoung301, line 959:
```suggestion
    # Starting with 25R2, the licensingclient directory is located one level higher in the installation
    # path. If the folder isn't found in the legacy location, use the parent directory.
```

ramin4667: Updated.

Samuelopez-ansys (approved):

Samuelopez-ansys:
@ramin4667 I approved this PR, but also I changed the default to check the license by default, hopefully the licenseclient in the future will be more consistent

src/ansys/aedt/core/generic/file_utils.py
SMoraisAnsys: Ah, I didn't understood the code correctly. Thanks !

src/ansys/aedt/core/generic/settings.py
```diff
@@ -204,7 +204,7 @@ def __init__(self):
         self.__wait_for_license: bool = False
         self.__lazy_load: bool = True
         self.__objects_lazy_load: bool = True
-        self.__skip_license_check: bool = False
+        self.__skip_license_check: bool = True
```
SMoraisAnsys, line 207: Why should we skip the license check ? @Samuelopez-ansys I think it was defined at False for a good reason no ?

Samuelopez-ansys: We thought that it was useful but some users had issues with the license, so by default the license check will be skipped like it was before.

Samuelopez-ansys: image
Like this one from another user, so it seems that the process sometimes returns this kind of exception.

Samuelopez-ansys (approved):

PR #5916, opened 2025-03-14
REFACTOR: Delete ML patch class

Samuelopez-ansys:
## Description
This class is an advanced usage of Stackup 3D class. It should not be part of PyAEDT, and if for some reason, any user needs it, we can think about creating an example.

## Issue linked

## Checklist


pyproject.toml
```diff
@@ -78,8 +78,6 @@ tests = [
     "pyvista[io]>=0.38.0,<0.45",
     "ansys-tools-visualization-interface; python_version >= '3.10'",
     "tables; python_version >= '3.10'",
-    # Never directly imported but required when loading ML related file see #4713
```
SMoraisAnsys, line 81: Can't we also remove joblib from the dependencies list?

SMoraisAnsys (changes requested):
I agree with creating an example instead of hosting this lines of code.
On top of `joblib` you should also remove the following files:
- patch_svr_model_100MHz_1GHz.joblib
- patch_svr_model_1GHz_10GHz.joblib

pyproject.toml
Samuelopez-ansys: Done

PR #6051, opened 2025-04-15
REFACTOR: 12_post_processing refactoring

nunesanderson:
## Description
**Post processing test refactor**

## Issue linked

## Checklist


tests/system/visualization/test_12_PostProcessing.py
```diff
@@ -131,32 +131,32 @@ def icepak_post(add_app):
...
+    app.close_project(save=False)
 
 
 class TestClass:
-    @pytest.fixture(autouse=True)
-    def init(self, local_scratch):
-        self.local_scratch = local_scratch
+    # @pytest.fixture(autouse=True)
+    # def init(self, local_scratch):
+    #     self.local_scratch = local_scratch
```
gmalinve, line 157: @nunesanderson remove.

gmalinve (changes requested):
@nunesanderson thanks for taking care of this! :)
Here some suggestions:
- I'd improve the test method names (test_manipulate_report_A, B, C etc.)
- All tests don't have a single purpose, they assert multiple methods.
- There are some methods that are asserted more than once in different tests (just as an example you can look at export_config and create_report_from_configuration just to mention a couple). We could check if there are repetitions without decreasing codecov.
- There are also some tests (test_parse_vector) where we test protected methods (is it really needed?). We can check.
- There are some tests, just as an example test_markers that creates multiple field plots to test markers. Maybe we could avoid some cases and open an aedtz where the field plot is already created.

PR #6068, opened 2025-04-22
FEAT: Support new EmitCom API for 25R2, add node classes for all EMIT node types

bryankaylor:
## Description
Updates the Revision class to use the new EmitCom API for accessing results and manipulating the EMIT project using EMIT nodes. 

## Issue linked
TFS Story F932911

## Checklist
- [X] I have tested my changes locally.
- [] I have added necessary documentation or updated existing documentation.
- [X] I have followed the coding style guidelines of this project.
- [X] I have added appropriate tests (unit, integration, system).
- [X] I have reviewed my changes before submitting this pull request.
- [X] I have agreed with the Contributor License Agreement ([CLA](https://developer.ansys.com/form/cla-acceptance)).


jsalant22:
@Samuelopez-ansys how can we test this with a 25.2 build?

src/ansys/aedt/core/emit_core/nodes/EmitNode.py
```diff
@@ -0,0 +1,407 @@
+# -*- coding: utf-8 -*-
```
SMoraisAnsys, line 1: Can you rename that file and the other to follow python convention of using lower caps for python files ? That helps to distinguish between module names and class names. Otherwise it can be confusing.
Note that this apply to the other new python files.

src/ansys/aedt/core/emit_core/nodes/EmitNode.py
```diff
@@ -0,0 +1,407 @@
...
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+import warnings
+
+import ansys.aedt.core.generic.constants as consts
+
+from ..emit_constants import EMIT_INTERNAL_UNITS
```
SMoraisAnsys, line 29: Avoid using relative imports if possible, could you use `ansys.aedt.core.emit_core.emit_constants` instead ?
This also applies to other modules

src/ansys/aedt/core/emit_core/nodes/EmitNode.py
```diff
@@ -0,0 +1,407 @@
...
+        Returns
+        -------
+        node: EmitNode
+            The node.
+
+        Examples
+        --------
+        >>> new_node = node._get_node(node_id)
+        """
+        from . import generated
```
SMoraisAnsys, line 147: Same comment on imports at method level

src/ansys/aedt/core/emit_core/nodes/generated/Amplifier.py
```diff
@@ -0,0 +1,255 @@
...
+        return self._duplicate(new_name)
+
+    def delete(self):
+        """Delete this node"""
+        self._delete()
+
+    @property
+    def filename(self) -> str:
+        """Filename
+        Name of file defining the outboard component
```
SMoraisAnsys, line 51:
```suggestion
        """Name of the file defining the outboard component.
```
Could you perform the same kind of changes in other methods to have a simple and short title description ending with a dot ?

src/ansys/aedt/core/emit_core/nodes/generated/Amplifier.py
```diff
@@ -0,0 +1,255 @@
...
+
+    @property
+    def noise_temperature(self) -> float:
+        """Noise Temperature
+        System Noise temperature (K) of the component
+
+        Value should be between 0 and 1000.
+        """
+        val = self._get_property("Noise Temperature")
+        return val  # type: ignore
```
SMoraisAnsys, line 70: Shouldn't you cast `val` into `float` here ? AFAIK, `_get_property` returns a `str` or list of `str`

src/ansys/aedt/core/emit_core/nodes/generated/Amplifier.py
```diff
@@ -0,0 +1,255 @@
...
+        self._oRevisionData.SetEmitNodeProperties(self._result_id, self._node_id, [f"Amplifier Type={value.value}"])
+
+    @property
+    def gain(self) -> float:
+        """Gain
+        Amplifier in-band gain
+
+        Value should be between 0 and 100.
+        """
+        val = self._get_property("Gain")
```
SMoraisAnsys, line 114: Same comments here and it applies to other methods to come

tests/system/solvers/test_26_emit.py
```diff
@@ -275,8 +283,9 @@ def test_04_radio_component(self, add_app):
                 assert "Frequency should be within 1Hz to 100 GHz." in str(e)
 
             # test power unit conversions
-            band_power = radio.band_tx_power(band)
-            assert band_power == 40.0
+            # TODO(bkaylor): What should the unit here be?
+            # band_power = radio.band_tx_power(band)
+            # assert band_power == 40.0
```
SMoraisAnsys, line 288: Needs to be handled

tests/system/solvers/test_26_emit.py
```diff
@@ -397,8 +409,8 @@ def test_07_antenna_component(self, add_app):
         assert position == (0.0, 0.0, 0.0)
 
     @pytest.mark.skipif(
-        config["desktopVersion"] <= "2023.1",
-        reason="Skipped on versions earlier than 2023.2",
+        (config["desktopVersion"] <= "2023.1") or (config["desktopVersion"] > "2025.1"),
```
SMoraisAnsys, line 412: Why are you skipping some tests with version 2025.1 ?

SMoraisAnsys (changes requested):
Thanks for the changes @bryankaylor 
I left some comments about your changes, could you address them or answer when they are not needed ?

Also, your changes involve a lot of new classes. Could you test them ? Here, you are introducing a lot of changes but the coverage isn't following.

src/ansys/aedt/core/emit_core/emit_constants.py
```diff
@@ -76,14 +76,111 @@ class UnitType(MutableEnum):
...
+        if units == "bps":
+            mult = 1.0
+        elif units == "kbps":
+            mult = 1e3
+        elif units == "Mbps":
+            mult = 1e6
+        elif units == "Gbps":
+            mult = 1e9
+        else:
+            raise ValueError(f"{units} are not valid units for data rate.")
```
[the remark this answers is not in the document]
Samuelopez-ansys: One additional comment, did you check from ansys.aedt.core.generic.constants import AEDT_UNITS ?

We should try to centralize everything related to constants in AEDT.

src/ansys/aedt/core/emit_core/nodes/generated/MultiplexerBand.py
```diff
@@ -0,0 +1,252 @@
...
+    def delete(self):
+        """Delete this node"""
+        self._delete()
+
+    class TypeOption(Enum):
+        BY_FILE = "By File"  # eslint-disable-line no-eval
+        LOW_PASS = "Low Pass"  # eslint-disable-line no-eval
+        HIGH_PASS = "High Pass"  # eslint-disable-line no-eval
+        BAND_PASS = "Band Pass"  # eslint-disable-line no-eval
+
```
SMoraisAnsys, line 58: Any reason for not addding `BAND_STOP` ?

src/ansys/aedt/core/emit_core/nodes/generated/MultiplexerBand.py
```diff
@@ -0,0 +1,252 @@
...
+    def duplicate(self, new_name: str):
+        """Duplicate this node"""
+        return self._duplicate(new_name)
+
+    def delete(self):
+        """Delete this node"""
+        self._delete()
+
+    class TypeOption(Enum):
+        BY_FILE = "By File"  # eslint-disable-line no-eval
```
SMoraisAnsys, line 54: What's the use of `eslint-disable-line` ? Isn't that for javascript code ? This applies everywhere it's used

src/ansys/aedt/core/emit_core/nodes/generated/MultiplexerBand.py
jsalant22: That's not currently an option for our multiplexer model. I'll add an enhancement request for it though.

src/ansys/aedt/core/emit_core/results/revision.py
```diff
@@ -22,15 +22,35 @@
...
 from ansys.aedt.core.emit_core.emit_constants import TxRxMode
+from ansys.aedt.core.emit_core.nodes import generated
+from ansys.aedt.core.emit_core.nodes.EmitNode import EmitNode
+from ansys.aedt.core.emit_core.nodes.generated import CouplingsNode
+from ansys.aedt.core.emit_core.nodes.generated import EmitSceneNode
+from ansys.aedt.core.emit_core.nodes.generated import ResultPlotNode
 from ansys.aedt.core.generic.general_methods import pyaedt_function_handler
 
 
+def error_if_below_aedt_version(version: int):
```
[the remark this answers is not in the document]
bryankaylor: Have now replaced all instances of this decorator with `@min_aedt_version()`, thanks!

src/ansys/aedt/core/emit_core/emit_constants.py
```diff
@@ -76,14 +76,111 @@ class UnitType(MutableEnum):
...
+    "Length": ["pm", "nm", "um", "mm", "cm", "dm", "meter", "meters", "km", "mil", "in", "ft", "yd", "mile"],
     "Time": ["ps", "ns", "us", "ms", "s"],
-    "Voltage": ["mV", "V"],
+    "Voltage": ["nV", "uV", "mV", "V", "kV", "MegV"],
     "Data Rate": ["bps", "kbps", "Mbps", "Gbps"],
     "Resistance": ["uOhm", "mOhm", "Ohm", "kOhm", "megOhm", "GOhm"],
 }
 """Valid units for each unit type."""
 
+EMIT_INTERNAL_UNITS = {
```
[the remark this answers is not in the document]
bryankaylor: We could use an Enum instead, but I think in this case the dict might be a bit more convenient (and this dict should be used internally in private underscored methods only.)

When we write the classes in the `generated` folder, most get/set properties have a unit type that we get from EMIT as a string. For example, the `amplifier.bandwidth` unit type is "Freq", and the `amplifier.saturation_level` unit type is "Power". These unit type strings are written into the second argument of `_convert_to_internal_units` and `_convert_from_internal_units`. Inside those two methods, this dict lets us directly use that string written during code generation to map to the internal EMIT units. 

If we used an Enum instead, we'd have to do an extra step in the conversion (either at code generation time or at runtime), like "Power" -> EmitInternalUnit.POWER -> "dBm" instead of "Power" -> "dBm", since our starting point will always be the string key.

tests/system/solvers/test_26_emit.py
bryankaylor: This test (`test_08_revision_generation`) is skipped for versions later than 2025R1, starting at 2025R2, because we changed the way revisions are managed. Previously, anytime you did an analysis with `results.analyze()`, a new revision used to access results would be saved. But now in 2025R2, if you call `results.analyze()` many times, you will only have one revision (called "current") until it is manually saved. 

This test still properly validates the behavior and passes in 2025R1 and earlier, so I have it skipped in 2025R2 and later. There's a new version of the test that runs in 2025R2 and later.

In a case like this where the behavior that a test validates changes, would you suggest a different strategy? I see a few different options: 

a) Have two different tests, one that runs on 2025R1 and earlier, and one that runs on 2025R2 and later
b) Update the test so it runs on 2025R2 and later, but is skipped on 2025R1 and earlier (e.g. "delete/replace the old test") 
c) Combine the two tests into one, do a version check in the test and branch to either validate 2025R1 behavior or 2025R2 behavior

I went with a) in this case, but don't have a strong preference and would be happy to do either of the other options, or something else if you have a better idea.

src/ansys/aedt/core/emit_core/nodes/generated/Amplifier.py
bryankaylor: @SMoraisAnsys Have now reformatted all the docstrings, every property's first sentence should either be a title or short description, both ending in a period.

tests/system/solvers/test_26_emit.py
bryankaylor: @SMoraisAnsys - Sorry, didn't tag you in this comment, did you have any guidance for which strategy is preferable here?

bryankaylor:
@SMoraisAnsys @Samuelopez-ansys - We've made some improvements to this PR by removing a lot of the generated code and improving our solver unit test to improve coverage. Now, when run locally with 25R2, the coverage of the new `src\ansys\aedt\core\emit_core\nodes\generated` folder is 78.45%. Could we get another review of this PR? Thank you!

jsalant22 (dismissed):

Samuelopez-ansys (commented):
@bryankaylor Why do not use the PyAEDT constants? We try as much as possible use common classes in PyAEDT to avoid duplicated code.

I think you could use with minimum refactoring: ansys.aedt.core.generic.constants

src/ansys/aedt/core/emit_core/nodes/emit_node.py
```diff
@@ -0,0 +1,474 @@
...
+            dec_val = float(vals[0])
+            units = vals[1].strip()
+            return dec_val, units
+        # no space before units, so find the first letter
+        for i, char in enumerate(value):
+            if char.isalpha():
+                dec_val = float(value[:i])
+                units = value[i:]
+                return dec_val, units
+        raise ValueError(f"{value} are not valid units for this property.")
```
SMoraisAnsys:
```suggestion
        raise ValueError(f"{value} is not valid for this property.")
```

src/ansys/aedt/core/emit_core/nodes/emit_node.py
```diff
@@ -0,0 +1,474 @@
...
+            self._emit_obj.oeditor.RenameComponent(self.name, requested_name)
+            new_name = requested_name
+        else:
+            new_name = self._oRevisionData.RenameEmitNode(self._result_id, self._node_id, requested_name)
+
+        return new_name
+
+    def _duplicate(self, new_name):
+        # TODO (maybe needs to be custom?)
+        pass
```
SMoraisAnsys:
```suggestion
        raise NotImplementedError("This method is not implemented yet.")
```

src/ansys/aedt/core/emit_core/results/results.py
```diff
@@ -79,9 +79,25 @@ def _add_revision(self, name=None):
...
+            # Do we need to delete the existing Current revisions? Or can we just return it?
+            for revision in current_revisions:
+                self.delete_revision("Current")
+            # if len(current_revisions) > 0:
+            #     current_revision = current_revisions[0]
+            #     current_revision._load_revision()
+            #     # current_revision.revision_loaded = True
+            current_revision = Revision(self, self.emit_project, name)
+            self.revisions.append(current_revision)
+            return current_revision
```
SMoraisAnsys: Can you clarify the comment here ?

src/ansys/aedt/core/emit_core/results/revision.py
```diff
@@ -60,40 +67,85 @@ class Revision:
...
+                """Index of the result for this revision."""
+
+                self.path = emit_obj.odesign.GetResultDirectory(name)
+                """Path to the EMIT result folder for the revision."""
+
+                self.name = name
+                """Name of the revision."""
+
+            # Get the SimulationNodeID for the specified result
+            # self._sim_node_id = self._emit_com.GetTopLevelNodeID(self.results_index, "Simulation")
```
SMoraisAnsys:
```suggestion
```

src/ansys/aedt/core/emit_core/results/revision.py
```diff
@@ -60,40 +67,85 @@ class Revision:
...
+        self.odesktop = emit_obj.odesktop
+        """Desktop object."""
+
+        self.parent_results = parent_results
+        """Parent Results object."""
+
+        self.aedt_version = int(parent_results.emit_project.aedt_version_id[-3:])
+        """AEDT version."""
+
+        if self.aedt_version > 251:
```
SMoraisAnsys, line 82: Is that feature only available for AEDT > 251 ? If yes, could you add the decorator for AEDT minimal version ? If no, could you extend the implementation to handle the previous AEDT version ? In the previous version, attributes like `path` were available.

bryankaylor: This object is available in 2025R2 and later. This constructor sets up all the properties we need for any version of AEDT, here's which properties are left out when moving to 2025R2 and why:

All versions:
-   emit_project
-   odesktop
-   parent_results
-   aedt_version  
-   name
-   path

2025R2 and later:
- results_index

2025R1 and earlier:
-   revision_number
-   timestamp
-   revision_loaded


`revision_number` has been replaced with results_index
`timestamp` is unnecessary now, because the Current revision is always up-to-date
`revision_loaded` is used internally, but is set elsewhere (in results.py) in 2025R2.

bryankaylor:
> @bryankaylor Why do not use the PyAEDT constants? We try as much as possible use common classes in PyAEDT to avoid duplicated code.
> 
> I think you could use with minimum refactoring: ansys.aedt.core.generic.constants

We have a separate file right now because a lot of those constants aren't relevant to AEDT, only to EMIT. We have a separate subprocess, iemit.exe, with its own units that we need to keep track of separately from the users' units set in AEDT, or convert from AEDT units to EMIT units. There's some other non-unit related code that helps us manage EMIT-specific Enums, which could maybe be moved to the AEDT constants module later on if needed, but it might be cleaner to leave it in emit_constants for now.

bryankaylor:
> LGTM, I left minor comments.
> 
> I know that some part of the code is generated but would it be possible (maybe later) to rework the code generation logic to avoid a lot of code duplicate ? For example, multiple `property` are defined in the same way in mutliple files and could be gathered below `EmitNode`. Would it also be possible to leverage `Enum` as discussed in https://github.com/ansys/pyaedt/pull/6068/files#r2101256934 ? I think some of the problem that may arise could be handled by defining classes like
> 
> ```python
> class MyEnum(str, Enum):
>     FIRST = "first"
>     SECOND = "second"
> ```
> 
> The defined enums could then be defined in the already existing /src/ansys/aedt/core/generic/constants.py and extend or leverage its content.

Thanks, I've applied your suggested changes. Sure, I can add a task to our backlog for later to see if we can gather more universal properties for the generated code and move it to the parent `emit_node.py` class. 

And good idea for the `Enum` classes to extend strings, maybe we could make use of that. I'll note that in our backlog as well to be considered later on.

Samuelopez-ansys (approved):

Issue #6100, opened 2025-05-01
Add support for UncoverFaces


Devin-Crawford:
### Description of the feature

Please add support for the method ``UncoverFaces``.

### Steps for implementing the feature

Native API:
``` python
oEditor = oDesign.SetActiveEditor("3D Modeler")
oEditor.UncoverFaces(
	[
		"NAME:Selections",
		"Selections:="		, "Circle1",
		"NewPartsModelFlag:="	, "Model"
	], 
	[
		"NAME:Parameters",
		[
			"NAME:UncoverFacesParameters",
			"FacesToUncover:="	, [19]
		]
	])
```

Add a method ``uncover_face`` to the ``GeometryModeler`` class. 

### Useful links and references

Use the ``cover_faces()`` method as a starting point. See line 1447 of primitives.py.

PR #6122, opened 2025-05-07
FEAT: Uncover face

jvela018:
## Description
Creation of a method uncover_face in the modeler class. 

## Issue linked
#6100


Samuelopez-ansys:
@jvela018 Is it not possible to uncover multiple faces in one single call? Please test it, because if the API allows multi-face selection, the PyAEDT method should also allow it. Thanks!

jvela018:
> @jvela018 Is it not possible to uncover multiple faces in one single call? Please test it, because if the API allows multi-face selection, the PyAEDT method should also allow it. Thanks!

@Samuelopez-ansys I added multi face selection. Is this implementation ok? Thanks

src/ansys/aedt/core/modeler/cad/primitives.py
```diff
@@ -1481,6 +1481,58 @@ def cover_faces(self, assignment):
...
+        References
+        ----------
+        >>> oEditor.UncoverFaces
+        """
+
+        faces = {}
+        flat_assignment = []
+
+        # create a flat list from assignment
+        for item in assignment:
```
Samuelopez-ansys, line 1507: I would just use one level:

flat_assignment = [subitem for item in assignment for subitem in (item if isinstance(item, list) else [item])]

jvela018: @Samuelopez-ansys , although this way is more compact, I think the current implementation is easier to read and debug if needed. If you don't mind I rather leave it as is.

src/ansys/aedt/core/modeler/cad/primitives.py
```diff
@@ -1481,6 +1481,58 @@ def cover_faces(self, assignment):
         self.oeditor.CoverSurfaces(["NAME:Selections", "Selections:=", obj_to_cover, "NewPartsModelFlag:=", "Model"])
         return True
 
+    @pyaedt_function_handler(selection="assignment")
+    def uncover_faces(self, assignment):
+        """Uncover surfaces/faces.
```
Samuelopez-ansys, line 1486:
```suggestion
        """Uncover faces.
```
Should not be only faces? In AEDT I do not think surfaces has a meaning

src/ansys/aedt/core/modeler/cad/primitives.py
```diff
@@ -1481,6 +1481,58 @@ def cover_faces(self, assignment):
...
+
+        Parameters
+        ----------
+        assignment : list
```
Samuelopez-ansys, line 1490:
```suggestion
        assignment : str, int, list
```

src/ansys/aedt/core/modeler/cad/primitives.py
```diff
@@ -1481,6 +1481,58 @@ def cover_faces(self, assignment):
...
+            List of Sheet object to uncover.
```
Samuelopez-ansys, line 1491:
```suggestion
            Sheet objects to uncover.
```

If I am not wrong, this method support both, faces IDs or face object

src/ansys/aedt/core/modeler/cad/primitives.py
```diff
@@ -1481,6 +1481,58 @@ def cover_faces(self, assignment):
         self.oeditor.CoverSurfaces(["NAME:Selections", "Selections:=", obj_to_cover, "NewPartsModelFlag:=", "Model"])
         return True
 
+    @pyaedt_function_handler(selection="assignment")
```
Samuelopez-ansys, line 1484: This is not needed because it is a new method, so you do not need to deprecate selection.

```suggestion
    @pyaedt_function_handler()
```

src/ansys/aedt/core/modeler/cad/primitives.py
```diff
@@ -1481,6 +1481,58 @@ def cover_faces(self, assignment):
...
+
+        Returns
+        -------
+        bool
+            ``True`` when successful, ``False`` when failed.
+
+        References
+        ----------
+        >>> oEditor.UncoverFaces
+        """
```
Samuelopez-ansys, line 1501: You could add an example in the docstring

tests/system/general/test_08_Primitives3D.py
```diff
@@ -2110,3 +2110,15 @@ def test_96_sweep_around_axis(self):
...
+        o1 = self.aedtapp.modeler.create_circle(cs_plane=0, position=[0, 0, 0], radius=10)
+        assert self.aedtapp.modeler.uncover_faces([o1.faces[0]])
+        c1 = self.aedtapp.modeler.create_circle(
+            orientation=ansys.aedt.core.constants.AXIS.X,
+            origin=[0, 10, 20],
+            radius="3",
+            name="Circle1",
+        )
+        b1 = self.aedtapp.modeler.create_box(origin=[-13.9, 0, 0], sizes=[27.8, -40, 25.4], name="Box1")
+        assert self.aedtapp.modeler.uncover_faces([c1.faces[0], b1.faces[0], b1.faces[2]])
```
Samuelopez-ansys, line 2124: What I meant here, is that after this operation, you could check that the number of faces of c1,  is different than at the beginning...

src/ansys/aedt/core/modeler/cad/primitives.py
jvela018: Example added.

Samuelopez-ansys (approved):
LGTM

PR #6156, opened 2025-05-14
FIX: Correct unit for h-field in set_non_linear() for bh curve definition

jvela018:
## Issue linked
 #6154


Samuelopez-ansys (commented):
Good catch! I fixed a bug last week because they were changed, but I introduced another bug.

Please could you add a test to avoid this? Maybe you can add something similar to what it triggered you to find the bug

jvela018:
@Samuelopez-ansys 
I added an asser to check that b and h have defined units. However, I noticed because it defaulted to none. I didn't add the cases where they're none because before defining them they could be None and I don't want them to fail. However, the units can probably be changed so maybe at some point they could be generalized so that they take for example uTesla and so forth. Then again, I guess we can modify the units after defining the material. In this case this wouldn't be affected at all.

Samuelopez-ansys (approved):

MaxJPRey (approved):
LGTM.

PR #5108, opened 2024-08-30
DOCS: Fix release_1_0.rst

gmalinve:

Issue #4775, opened 2024-06-06
Revisiting project files


jorgepiloto:
PyAEDT has been now public for more than two years. Since its public release, new tools and workflows have been discovered. Some of them have been adopted by the PyAnsys ecosystem.

Therefore, the goal of this meta-issue is to collect some issues that can be addressed within this project.

Please, open a pull-request per issue (if possible), to speed up reviewing process.

- [ ] **Verify proper metadata in pyproject.toml**
  Ensure authors and maintainers point to the `pyansys.core@ansys.com` email and that Ansys, Inc. is the main author. Update minimum Python version. Expand project classifiers. Migrate project tools configuration to this file.

- [ ] **Consider using Ruff for linting and formatting**
  [Ruff](https://docs.astral.sh/ruff/) collects all popular tools (black, isort, flake8, codespell, and more!) under a unique project. In addition, it is extremely fast as it operates on top of Rust. This would allow to remove files like `.flake8` from the root of the project.

- [ ] **Use ansys-pre-commit-hooks**
  The [Ansys pre-commit-hooks](https://github.com/ansys/pre-commit-hooks) ensures that a project is compatible with the latest PyAnsys guidelines.

- [ ] **Unify ignore words with Vale's accept.txt file**
  The `accept.txt` file from Vale can be feed into codespell as the `ignore_words.txt` file.

- [ ] **Revisit doc build**
  There is a `*.bat` file inside of the `doc/source/` dir. This file is not required. The `make.bat` and `Makefiles` install Python libraries. This should take place on the `pyproject.toml`. Both `make.bat` and `Makefile` are generated by the `sphinx-quickstart` command. However, modern tools like Tox avoid having these files. Finally, sphinx-autoapi can be used to render a class per page in the documentation, removing unnecessary RST files.

- [ ] **Dynamic doc build**
  Consider using `sphinx-jinja` to conditionally build documentation sections via environment variables. This speeds up the development of the docs by allowing developers to exclude sections like API and Examples.

- [ ] **Implement a tox.ini file**
  [Tox](https://tox.wiki/en/latest) is a tool for automating tasks. It is platform aware and executes Python code in a dedicated venv, protecting your own Python environment.

- [ ] **Dynamic CI/CD**
  By playing with labels, labeler and events on the CI/CD, it is possible to alleviate the workload of the self-hosted runner. This way, the machine only runs the right workflows, avoiding unnecessary computations.

jorgepiloto:
[...]

PR #6157, opened 2025-05-15
CHORE: Setting up ruff

ecoussoux-ansys:
## Description
This PR introduces the use of ``ruff`` for linting and formatting, replacing ``isort``, ``black`` and ``flake8``.
``ruff`` is an extremely fast Python linter and code formatter, that integrates the functionalities of several tools behind a single, common interface (https://docs.astral.sh/ruff/)
Its implementation in pyaedt was suggested in issue #4775.
This PR can also be seen as contributing to the efforts described in issue #5504.

It should be noted that ``ruff`` does not replace the following tools, whose use in the project is therefore not affected by the changes made in this PR:
* ``blacken-docs`` (used for formatting Python code blocks in documentation files): ``ruff`` does not support formatting/linting of embedded code (particularly in ``.rst`` or ``.md`` files, see astral-sh/ruff#8237). Similarly, ``blacken-docs`` does not support the ``ruff`` formatter as an alternative for ``black`` (see adamchainz/blacken-docs#352). This tool is therefore retained (as a pre-commit hook), despite the remaining dependence on ``black``,
* ``codespell`` (fixes common misspellings in source code and other files): Is not integrated into ``ruff`` (see https://github.com/astral-sh/ruff/issues/1628),
* ``numpydoc`` (used for formatting docstrings): Is also not integrated into ``ruff`` (see https://github.com/astral-sh/ruff/issues/8425). 

The changes in this PR are limited to the definition of a minimal configuration for ``ruff`` (in the ``pyproject.toml`` file) with only a few rules applied to the code base, together with the introduction of pre-commit hooks for linting and formatting with ``ruff``. Python files fixed by ``ruff`` are updated. Configurations for ``isort``, ``black`` and ``flake8`` are removed and the associated pre-commit hooks are also dropped.
The currently ignored rules shall be gradually applied to the project in subsequent PRs. For this task, a useful command-line instruction can be used, which will return the number of errors flagged for each selected rule as well as the availability of automatic fixes : ``ruff check . --statistics``

It is also worth mentioning that the list of rules selected until now for linting in the ``.flake8`` file is not regarded as a reference for defining the ``ruff`` configuration implemented. Instead, this config is based on the recommendations from the pyansys-dev-guide, as introduced in https://github.com/ansys/pyansys-dev-guide/pull/592.
Furthermore, the enforcement of ‘PTH’ type rules is not a priority as part of the introduction of ``ruff`` to the project, as an effort in this direction is currently underway as described in issue #5108.

## Issue linked
Related to #4775 and #5504.


pyproject.toml
```diff
@@ -142,15 +142,44 @@ Discussions = "https://github.com/ansys/pyaedt/discussions"
...
 
-[tool.black]
+[tool.ruff]
 line-length = 120
+exclude = [
+    "venv", 
+    "__init__.py", 
+    "doc/_build", 
+    "doc/source/examples"
+]
```
SMoraisAnsys, line 152:
```suggestion
```
I'd avoid using `exclude` here as the file you mention are ones that we want to ignore in Git. Therefore, could you extend the .gitignore file instead ? That way, we would leverage https://docs.astral.sh/ruff/settings/#respect-gitignore

Btw, why do you exclude init files ?

pyproject.toml
```diff
@@ -142,15 +142,44 @@ Discussions = "https://github.com/ansys/pyaedt/discussions"
 Releases = "https://github.com/ansys/pyaedt/releases"
 Changelog = "https://github.com/ansys/pyaedt/blob/main/doc/source/changelog.rst"
 
-[tool.black]
+[tool.ruff]
 line-length = 120
```
SMoraisAnsys, line 146:
```suggestion
line-length = 120
fix = true
```

pyproject.toml
```diff
@@ -142,15 +142,44 @@ Discussions = "https://github.com/ansys/pyaedt/discussions"
...
+exclude = [
+    "venv", 
+    "__init__.py", 
+    "doc/_build", 
+    "doc/source/examples"
+]
+
+[tool.ruff.format]
+quote-style = "double"
+indent-style = "space"
```
SMoraisAnsys, line 156:
```suggestion
indent-style = "space"
docstring-code-format = true
```
The added configuration is something close to what `blacken-docs` was for `black`. I think we should use it and remove `blacken-docs`.
Also, we might want to use `docstring-code-line-length = "dynamic"` later on. Maybe in another PR?

pyproject.toml
```diff
@@ -142,15 +142,44 @@ Discussions = "https://github.com/ansys/pyaedt/discussions"
...
+    "W", # pycodestyle, see https://docs.astral.sh/ruff/rules/#pycodestyle-e-w
+]
+ignore = [
+    "D",   # pydocstyle, see https://docs.astral.sh/ruff/rules/#pydocstyle-d
+    "E",   # pycodestyle, see https://docs.astral.sh/ruff/rules/#pycodestyle-e-w
+    "F",   # pyflakes, see https://docs.astral.sh/ruff/rules/#pyflakes-f
+    "N",   # pep8-naming, see https://docs.astral.sh/ruff/rules/#pep8-naming-n
+    "PTH", # flake8-use-pathlib, https://docs.astral.sh/ruff/rules/#flake8-use-pathlib-pth
+    "TD",  # flake8-todos, https://docs.astral.sh/ruff/rules/#flake8-todos-td
+]
```
SMoraisAnsys, line 170:
```suggestion
select = [
    "I", # isort, see https://docs.astral.sh/ruff/rules/#isort-i
    "W", # pycodestyle, see https://docs.astral.sh/ruff/rules/#pycodestyle-e-w
    "D",   # pydocstyle, see https://docs.astral.sh/ruff/rules/#pydocstyle-d
    "E",   # pycodestyle, see https://docs.astral.sh/ruff/rules/#pycodestyle-e-w
    "F",   # pyflakes, see https://docs.astral.sh/ruff/rules/#pyflakes-f
    "N",   # pep8-naming, see https://docs.astral.sh/ruff/rules/#pep8-naming-n
    "PTH", # flake8-use-pathlib, https://docs.astral.sh/ruff/rules/#flake8-use-pathlib-pth
    "TD",  # flake8-todos, https://docs.astral.sh/ruff/rules/#flake8-todos-td
]
ignore = [fill this list]
```
Everything should be tracked and you should specify all the rules that we need to ignore. This will provide an exhaustive list that we can deal with one item at a time.

pyproject.toml
```diff
@@ -142,15 +142,44 @@ Discussions = "https://github.com/ansys/pyaedt/discussions"
...
+
+[tool.ruff.lint.pydocstyle]
+# Use Numpy-style docstrings.
+convention = "numpy"
+
+[tool.ruff.lint.isort]
```
SMoraisAnsys, line 176:
```suggestion
[tool.ruff.lint.isort]
combine-as-imports = true
```

SMoraisAnsys (changes requested):
Thanks for the contribution @ecoussoux-ansys ! I left some minor changes that need to be taken care of before merging

pyproject.toml
ecoussoux-ansys: @SMoraisAnsys I understand that you are asking to add this option as it is part of the recommended setup provided in the pyansys-dev-guide.
However, here this would be conflicting with the option ``force-single-line = true``, that I introduced to reproduce a particular argument of the pre-commit hook previously used for ``isort`` in the project.
Which of the two options do you think we should go for?

pyproject.toml
ecoussoux-ansys: OK for adding this setting.
I did not include it at first to avoid unexpected auto-fixes during my first tests.

pyproject.toml
ecoussoux-ansys: As per our discussion, I will apply the suggested change.
The ``.gitignore`` file does not need to be extended. The ``__init__.py`` files can be removed from the exclude list.

pyproject.toml
ecoussoux-ansys: As per our discussion, I will accept the suggested change.
The setting ``docstring-code-line-length = "dynamic"`` can be applied immediately as this is the default configuration for ``ruff`` when using ``docstring-code-format = true`` (see https://docs.astral.sh/ruff/settings/#format_docstring-code-line-length).
However, the pre-commit hook for ``blacken-docs`` will be kept, as it provides linting/formatting of Python code blocks in ``.rst`` files and should not create conflicts for docstring formatting with ``ruff``.

pyproject.toml
ecoussoux-ansys: Exhaustive list of excluded linting rules is provided in the ``pyproject.toml`` file updated in commit XXX

pyproject.toml
ecoussoux-ansys: As per our discussion, the suggested change will *not* be applied, as the setting ``combine-as-imports = true`` is not compatible with ``force-single-line = true``, which was introduced to reproduce a particular argument of the pre-commit hook previously used for isort in the project.

SMoraisAnsys (approved):

PR #6198, opened 2025-05-26
CHORE: Do not check AEDT/EDB binary files with Ruff

SMoraisAnsys:
## Description
During a recent PR it was identified that Ruff was checking files that he shouldn't. This should ensure that AEDT/EDB files are not changed as updating them make the binary file unusable.

## Issue linked
None

## Checklist

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -33,7 +33,7 @@ jobs:
       contents: write
       pull-requests: write
     steps:
-      - uses: ansys/actions/doc-deploy-changelog@2cf9a9c43235a000d613c2b13e64c954232a4553 # v9.0.9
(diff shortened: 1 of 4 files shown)


ecoussoux-ansys (approved):
The change looks absolutely fine to me. However, I am surprised that such an issue was observed, since the documentation on ``ruff`` is pretty clear about the fact that both the linter and the formatter will only check for Python files in the project (https://docs.astral.sh/ruff/linter/ and https://docs.astral.sh/ruff/formatter/). Also this actually did not happen to me so far when working with the tool.
Nevertheless, I guess it's a sensible safety measure to introduce.

Samuelopez-ansys (dismissed):

Samuelopez-ansys:
Right, I copied and pasted an edb file, and the pre-commit modified the file. Not sure why, but I think this is safer because the edb file is very "sensitive"

PR #6203, opened 2025-05-27
CHORE: Enforce ``ruff`` pycodestyle E rules

ecoussoux-ansys:
## Description
This PR is a follow-up to #6157, which introduced a minimal configuration for ``ruff`` to pyaedt.
Here, the "E" rules (related to pycodestyle, see https://docs.astral.sh/ruff/rules/#error-e) are enforced on the entire codebase, allowing their removal from the list of ignored linting rules for ``ruff``.

This effort is part of the migration of pyaedt towards using ``ruff`` for code formatting and linting, as requested in issue #4775.
This PR is also contributing to the efforts described in issue #5504.

"E" rules were incrementally applied (one rule per commit) using the two following useful command-line instructions to guide this work:
* ``ruff check . --statistics``, returning the number of errors flagged for each selected linting rule, as well as the availability of automatic fixes,
* ``ruff check . --select <rule code>``, restricting the analysis of the project files by ``ruff`` to a specific linting rule, so that reported errors can be reviewed on a rule by rule basis.

*Note: A minor change is also made to the ``.pre-commit-config.yaml`` file in this PR, updating the repo url for the ``blacken-docs`` pre-commit hook.*

## Issue linked
Related to #4775 and #5504.


.pre-commit-config.yaml
```diff
@@ -48,7 +48,7 @@ repos:
   hooks:
     - id: check-github-workflows
 
-- repo: https://github.com/asottile/blacken-docs
+- repo: https://github.com/adamchainz/blacken-docs
```
SMoraisAnsys, line 51: Was that a typo introduced previously or did the repo got transferred ?

src/ansys/aedt/core/emit_core/results/revision.py
```diff
@@ -79,8 +79,13 @@ def __init__(self, parent_results, emit_obj, name=None):
...
-        key = lambda s: s.split("=", 1)[0]
-        val = lambda s: s.split("=", 1)[1]
+
+        def key(s):
+            return s.split("=", 1)[0]
+
+        def val(s):
+            return s.split("=", 1)[1]
+
         props = {key(s): val(s) for s in raw_props}
```
SMoraisAnsys: Can you refactor that ? The complexity is very low and it is too verbose now IMO

src/ansys/aedt/core/generic/configurations.py
```diff
@@ -2319,9 +2319,9 @@ def export_config(self, config_file=None, overwrite=False):
 
         if "" in pin_mapping:
             del pin_mapping[""]
-        for k, l in pin_mapping.items():
+        for k, L in pin_mapping.items():
             temp_dict3 = {}
-            for i in l:
+            for i in L:
```
SMoraisAnsys: Please rename the variables and don't a capital letter here. Capital letters often refer to "constant" values and here this is used to iterate over a dict so clearly the value will change

src/ansys/aedt/core/generic/configurations.py
```diff
@@ -2330,12 +2330,12 @@ def export_config(self, config_file=None, overwrite=False):
 
         port_dict = {}
         temp = pin_mapping.copy()
-        for k, l in temp.items():
-            if k not in ["gnd", "ports"] and len(l) == 1:
+        for k, L in temp.items():
```
SMoraisAnsys: Same for here and the following

src/ansys/aedt/core/modules/design_xploration.py
```diff
@@ -1368,8 +1368,8 @@ def add(
                     }
                     setup.props["Sweeps"]["SweepDefinition"].append(arg)
             else:
-                for l, k in dx_variables.items():
-                    arg = {"Variable": l, "Data": k, "OffsetF1": False, "Synchronize": 0}
+                for L, k in dx_variables.items():
```
SMoraisAnsys: I know this isn't performed now but could you try to use the idea of "key" and "value" when iterating on a dictionary ? That will help a lot with maintenance as it will avoid to double check what a variable should contain.
In some cases, other names might fit even better so feel free to use them :)

src/ansys/aedt/core/visualization/post/common.py
```diff
@@ -2172,7 +2172,7 @@ def far_field(self, expressions=None, setup=None, sphere_name=None, source_conte
             rep.source_context = source_context
             rep.report_type = "Radiation Pattern"
             if expressions:
-                if type(expressions) == list:
+                if isinstance(expressions, list):
```
SMoraisAnsys, line 2187: Before, there were tests of type equality (float and int for example). Why is it `isinstance` here ?

.pre-commit-config.yaml
ecoussoux-ansys: It was not a typo, the repo got transferred when maintenance was taken over by another developer. 
Also the previous link ("asottile") was anyway redirecting to the new one ("adamchainz"), so the pre-commit hook was still functional. I just thought it would be better to introduce the updated repo url.

src/ansys/aedt/core/emit_core/results/revision.py
ecoussoux-ansys: Any suggestion on how to reduce verbosity here?
I realise these two function definitions are quite lengthy for a very low complexity, but the changes were introduced in order to comply with rule E731, stating that "per PEP 8, you should always use a def statement instead of an assignment statement that binds a lambda expression directly to an identifier".
The doc on the ruff website for that rule is also illustrated by an example which to me seems quite similar to what is obtained here: https://docs.astral.sh/ruff/rules/lambda-assignment/

In that context, nothing really came to my mind to improve this..

SMoraisAnsys: `props = {raw_prop.split("=", 1)[0]: raw_prop.split("=", 1)[1] for raw_prop in raw_props}` should do the trick

If you find it inconvenient, just keep your current implementation

ecoussoux-ansys: Of course, it makes total sense, I didn't think of simply getting rid of the definition of these two functions altogether. 
I've implemented your suggestion in e86f9c2.

src/ansys/aedt/core/visualization/post/common.py
ecoussoux-ansys: This change (together with the other fixes from commit 178c3d76) aims at enforcing rule E721: https://docs.astral.sh/ruff/rules/type-comparison/
Do you think that writing ``if type(expressions) is list`` here would be better to resolve the error reported by ``ruff`` for this line?

src/ansys/aedt/core/modules/design_xploration.py
ecoussoux-ansys: Just to be sure I fully understand what you mean here: You would like this line to read ``for key, value in dx_variables.items():``, is that correct?
In that particular case, it could also be written as ``for var, data in dx_variables.items():``, right? 

I could definitely refactor quite some occurrences of that pattern this way and a good start would probably be to revisit the changes I made in commit 3ef24eb6, where I was enforcing rule E741 on the code base.

src/ansys/aedt/core/generic/configurations.py
ecoussoux-ansys: Alright, I will use more explicit variable names and avoid capital letters for non constant values. Also when iterating over dict (or other similar constructs), I think I could use syntaxes like ``for key, value in ...`` or even better names when possible (as per your other comment).

src/ansys/aedt/core/modules/design_xploration.py
SMoraisAnsys: This comment was a follow up of the previous. Feel free to rename variables to meaningful names whenever you want/can

src/ansys/aedt/core/visualization/post/common.py
SMoraisAnsys: Not really, I would use `isinstance` whenever it makes sense and is appropriate.
I did misread one of the changes and though you were now using `== int` instead of `is int`, my bad

src/ansys/aedt/core/generic/configurations.py
ecoussoux-ansys: Potential fix implemented in 558b1b5.

src/ansys/aedt/core/generic/configurations.py
ecoussoux-ansys: Potential fix implemented in 558b1b5.

src/ansys/aedt/core/modules/design_xploration.py
ecoussoux-ansys: Changes made in 558b1b5 attempt to use more relevant variable names (by default ``key, value/val``) when iterating on dict (or in other similar constructs) wherever variables named "L" were introduced in commit 3ef24eb.
Patterns like the one you highlighted (``for l, k in dict.items():``) are of course also found in many other parts of the code base, but were not revisited here.

src/ansys/aedt/core/generic/configurations.py
```diff
@@ -2319,24 +2319,24 @@ def export_config(self, config_file=None, overwrite=False):
 
         if "" in pin_mapping:
             del pin_mapping[""]
-        for k, l in pin_mapping.items():
+        for key, value in pin_mapping.items():
```
SMoraisAnsys:
```suggestion
        for key, values in pin_mapping.items():
```
Since you are going to iterate over "value", could you use a plural form so that we know it's a collection of things ?
Please, propagate that to the following lines and use "value" instead of "val"

src/ansys/aedt/core/icepak.py
```diff
@@ -2054,15 +2054,15 @@ def export_summary(
         if not solution_name:
             solution_name = self.nominal_sweep
         if variation:
-            for l in variation_list:
+            for v in variation_list:
```
SMoraisAnsys:
```suggestion
            for variation in variation_list:
```

src/ansys/aedt/core/modeler/advanced_cad/oms.py
```diff
@@ -173,7 +173,7 @@ def generate_buildings(self, center_lat_lon, terrain_mesh, max_radius=500):
 
                     xpos = np.array(outer.xy[0])
                     ypos = np.array(outer.xy[1])
-                    l = levels[n]
+                    lvl = levels[n]
```
SMoraisAnsys:
```suggestion
                    level = levels[n]
```

src/ansys/aedt/core/modeler/cad/modeler.py
```diff
@@ -1732,7 +1732,7 @@ def update(self):
         """
         # self._change_property(self.name, ["NAME:ChangedProps", ["NAME:Reference CS", "Value:=", self.ref_cs]])
         object_list_new = self._list_verification(self.props["List"], self.props["Type"])
-        if type == "Object":
+        if self.props["Type"] == "Object":
```
SMoraisAnsys, line 1735: @Samuelopez-ansys How was that even working before ? oO

src/ansys/aedt/core/visualization/advanced/sbrplus/hdm_parser.py
```diff
@@ -147,16 +147,16 @@ def _parse_list(self, type=None, base=None, size=1):
     def _parse_object(self, name):
         """Parser for an object message."""
         namesdict = {}
-        for l in self.parser_types[name]["layout"]:
-            type_to_parse = l["type"]
-            fields = l["field_names"]
+        for lay in self.parser_types[name]["layout"]:
```
SMoraisAnsys: Use "data" or "layout" ?

SMoraisAnsys (changes requested):
Thanks for the previous changes, here are new minor changes requests :D

src/ansys/aedt/core/icepak.py
ecoussoux-ansys: ``variation`` is the name of one of the input variables to the function defined where this line is located and due to the statement on line 2063, I believe we cannot use this name in the ``for`` loop here.
What about using ``var`` here instead?

src/ansys/aedt/core/visualization/advanced/sbrplus/hdm_parser.py
ecoussoux-ansys: Suggested change applied in c12693c.

src/ansys/aedt/core/modeler/advanced_cad/oms.py
ecoussoux-ansys: Suggested change applied in c12693c.

src/ansys/aedt/core/generic/configurations.py
ecoussoux-ansys: Suggested change applied in c12693c.

src/ansys/aedt/core/icepak.py
SMoraisAnsys: Or you could go with `iter_variation`. Just make the variable more meaningful than `v` please :)

ecoussoux-ansys: Suggested change applied in bfb2cb7.

SMoraisAnsys (approved):
LGTM. @Samuelopez-ansys can you have a look at the last discussion open ? The fix @ecoussoux-ansys is proposing seems to make sense but I'd prefer to have your opinion.

src/ansys/aedt/core/modeler/cad/modeler.py
Samuelopez-ansys: You are right! It was not tested, so no one complained. I have added 2 new tests.

Samuelopez-ansys (dismissed):

Samuelopez-ansys (dismissed):

PR #6220, opened 2025-06-02
REFACTOR: Improve primitives connect

### LINKED ISSUE #6207: Speed up Connect function

I found that calling connect function is very slow.  This is from core/modeler/cad/primitives.py.

The issue has nothing to do with the connect operation.  I tracked it down to the return value: 
 objects_list_after_connection.

The intersection is checked for each element of self.object_list.  But the intersection is slow and is not dependent on self.object_list.  I tested swapping the order, and the speed was acceptable.

For now, I have a workaround by fixing the problem on my end, but it would be nice to use the connect function out of the box.  (In my case, I can have 100s of objects in self.object_list, so the speed up is dramatic)

I was able to fix by changing this block: (Line 3982 of primitives.py in the latest src)

objects_list_after_connection = [
                obj
                for obj in self.object_list
                for sel in set(selections_list).intersection(self.object_names)
                if obj.name == sel
            ]

Just swapped the order of the loops to be this:

objects_list_after_connection = [
                obj
                for sel in set(selections_list).intersection(self.object_names)
                for obj in self.object_list
                if obj.name == sel
            ]

_No response_

SMoraisAnsys:
## Description
Improve primitievs `connect` method by removing unnecessary calls.

## Issue linked
Close #6207 

## Checklist

### DIFF
--- a/doc/changelog.d/6220.miscellaneous.md
+++ b/doc/changelog.d/6220.miscellaneous.md
@@ -1,0 +1,1 @@
+Improve primitives connect
--- a/src/ansys/aedt/core/modeler/cad/primitives.py
+++ b/src/ansys/aedt/core/modeler/cad/primitives.py
@@ -3979,12 +3979,9 @@ class GeometryModeler(Modeler):
             self.logger.info("Connection Correctly created")
 
             self.refresh_all_ids()
-            objects_list_after_connection = [
-                obj
-                for obj in self.object_list
-                for sel in set(selections_list).intersection(self.object_names)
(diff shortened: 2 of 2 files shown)


SMoraisAnsys:
@adamgregory81 I checked your issue (#6207) and the changes you proposed. Thanks for pointing out the code that could be improved ! Could you give a try at this branch to see if the changes are improving your case ? You just have to clone the project, switch to the branch `refactor/improve-primitives` and install the repo.

Let me know how it works for you !

PR #6227, opened 2025-06-04
REFACTOR: Import nastran extension and tests

SMoraisAnsys:
## Description
Refactoring of the import nastran extension where the extension has been updated to allow testing with tkinter direct interaction.

## Issue linked
Associated to #6228 

## Checklist

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -162,6 +162,7 @@ jobs:
           pytest-extra-args: ${{ env.PYTEST_ARGUMENTS }}
           python-version: ${{ env.MAIN_PYTHON_VERSION }}
           optional-dependencies-name: unit-tests
+          requires-xvfb: true
 
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@18283e04ce6e62d37312384ff67231eb8fd56d24 # v5.4.3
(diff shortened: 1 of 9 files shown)


SMoraisAnsys:
> Maybe instead of using os.path, we could use pathlib?

Sure

Samuelopez-ansys:
@SMoraisAnsys Great job! I like it a lot, should we also test 

root.nametowidget("preview_button").invoke()?

Or the toogle theme? Maybe we could add another test for this common funcionalities?

I was thinking that maybe in a next PR, we could create a class to define the default UI, because then maybe we can start creating method to building common stuff like multitabs, buttons... following the same structure, what do you think?

SMoraisAnsys:
> @SMoraisAnsys Great job! I like it a lot, should we also test
> 
> root.nametowidget("preview_button").invoke()?
> 
> Or the toogle theme? Maybe we could add another test for this common funcionalities?
> 
> I was thinking that maybe in a next PR, we could create a class to define the default UI, because then maybe we can start creating method to building common stuff like multitabs, buttons... following the same structure, what do you think?

Maybe I'm missing something but the preview button and toggle button are tested in https://github.com/ansys/pyaedt/pull/6227/files#diff-ac52ccb11fa1ba48938746d6dbe8a83b26b7d455bb64077eee93ad579dd83161
Do you have something else in mind ?

I do agree with a follow-up PR where we create a defaut UI class. I'll handle it if that's ok with you

PR #6232, opened 2025-06-04
CI: Update labeler permissions

MaxJPRey:
This pull request updates the permissions for the labeler workflow to ensure proper functionality.


.github/workflows/label.yml
```diff
@@ -30,7 +30,7 @@ jobs:
     name: Set labels
     needs: [label-syncer]
     permissions:
-      contents: read
+      contents: write
```
SMoraisAnsys: Why are you changing the permissions to `write` ? It doesn't seem required from the readme, see https://github.com/actions/labeler?tab=readme-ov-file#create-workflow

SMoraisAnsys (commented):
Thanks for the PR @MaxJPRey 
I just left a minor comment. Changes to `commenter` changes seem fine.

.github/workflows/label.yml
MaxJPRey: You are correct, I will revert it.

MaxJPRey:
```suggestion
      contents: read
```

SMoraisAnsys (approved):

PR #6235, opened 2025-06-05
FEAT: Configure layout

hui-zhou-a:
- [x] Refactor frontend to comply with the common extension UI
- [x] Add unittest for backend
- [x] Add unittest for frontend
- [x] Add documentation


Samuelopez-ansys:
@hui-zhou-a You also need to add the reference to the documentation here: 

doc\source\User_guide\extensions.rst

PR #6239, opened 2025-06-06
CHORE: Enforce ``ruff`` pyflakes F rules

ecoussoux-ansys:
## Description
Work in progress.
Description still to be added. 

## Issue linked
Related to https://github.com/ansys/pyaedt/issues/4775 and https://github.com/ansys/pyaedt/issues/5504.


doc/source/conf.py
```diff
@@ -275,10 +270,8 @@ def setup(app):
 html_theme_options = {
     "logo": "pyansys",
     "github_url": "https://github.com/ansys/pyaedt",
-    "navigation_with_keys": False,
     "show_prev_next": False,
     "show_breadcrumbs": True,
-    "collapse_navigation": True,
```
SMoraisAnsys, line 281: Just wondering, why are these options removed ?

src/ansys/aedt/core/application/design.py
```diff
@@ -472,14 +472,6 @@ def boundaries(self) -> List[BoundaryObject]:
                                 bb.append(component_boundary)
                                 bb.append(boundarytype)
 
-        if self.design_type == "Q3D Extractor" and self._aedt_version >= "2025.2":
-            net_object = self.get_oo_object(self.odesign, "Nets")
-            for net in self.get_oo_name(self.odesign, "Nets"):
-                if net not in bb:
-                    bb.append(net)
-                    net_type = self.get_oo_property_value(net_object, net, "Type")
-                    bb.append(net_type)
```
SMoraisAnsys: Why are you removing that ?

src/ansys/aedt/core/extensions/installer/jupyter_template.ipynb
```diff
@@ -1,20 +1,5 @@
 {
  "cells": [
-  {
```
SMoraisAnsys, line 3: I think there is a need to configure `ruff` so that it doesn't modifies IPYNB files. I'm not sure we should perform those changes

src/ansys/aedt/core/hfss3dlayout.py
```diff
@@ -886,40 +885,6 @@ def create_scattering(
             traces, sweep_name, variations=variations, report_category=solution_data, plot_name=plot
         )
 
-    @pyaedt_function_handler(activate="export", export_dir="output_dir")
-    def export_touchstone_on_completion(self, export=True, output_dir=""):
```
SMoraisAnsys, line 890: Why is that function removed ?

src/ansys/aedt/core/modeler/cad/primitives.py
```diff
@@ -7569,7 +7567,6 @@ def find_closest_edges(self, start_object, end_object, direction=0):
                     vert_dist_sum = GeometryOperators.arrays_positions_sum(
                         [vertex1_i, vertex2_i], [vertex1_j, vertex2_j]
                     )
-                    vect = GeometryOperators.distance_vector(start_midpoint, vertex1_j, vertex2_j)
                 else:
                     vert_dist_sum = GeometryOperators.arrays_positions_sum([start_midpoint], [end_midpoint])
```
SMoraisAnsys, line 7574: Can you remove the comment using `vect` below ?

src/pyaedt/generic/near_field_import.py
```diff
@@ -1 +0,0 @@
-from ansys.aedt.core.visualization.advanced.misc import convert_nearfield_data
```
SMoraisAnsys, line 1: Why this file removal ?

tests/integration/test_touchstone_data.py
```diff
@@ -31,7 +31,7 @@ def test_plot_insertion_losses(touchstone_file, patch_graphics_modules):
     ts = TouchstoneData(touchstone_file=touchstone_file)
     res = ts.plot_insertion_losses()
 
-    assert res is not []
+    assert res == []
```
SMoraisAnsys, line 34: There is a strange behavior here and it needs to be handled. Can you have a look at what `plot_insertion_losses` returns and why `res is not []` is the same as `res == []` ? I can help if needed

tests/system/general/test_08_Primitives3D.py
```diff
@@ -270,9 +270,8 @@ def test_05_center_and_centroid(self):
         assert GeometryOperators.v_norm(o.faces[0].center_from_aedt) - GeometryOperators.v_norm(o.faces[0].center) < tol
 
     def test_06_position(self):
-        udp = self.aedtapp.modeler.Position(0, 0, 0)
         with pytest.raises(IndexError) as execinfo:
```
SMoraisAnsys, line 273: Remember to update this with a `match=...` if needed

doc/source/conf.py
ecoussoux-ansys: This change is indeed due to ``ruff``. These 2 keys were in fact defined twice within a few lines here, which triggered F601 (multi-value-repeated-key-literal) errors that got resolved by deleting the duplicates.

src/ansys/aedt/core/application/design.py
ecoussoux-ansys: Thanks much for catching this!! 
This is absolutely unintentional: I somehow did not notice I was removing that when fixing (in 945cd09f) an import error I had introduced in the file (in a commit that predated the recent addition of this part in dfb7c1b9).
I reverted this is 3f4ef3a.

src/ansys/aedt/core/extensions/installer/jupyter_template.ipynb
ecoussoux-ansys: I really tried to be as cautious as possible when introducing changes to ``.ipynb`` files. In this particular case, I am removing a cell that was performing two unused imports. 
I understand the concern though and I may be missing the big picture here on the actual use of these files.
Do you think that all changes made to ``.ipynb`` files in this PR should be reverted? I could definitely configure ``ruff`` so that it ignores these files, by adding ``"*.ipynb"`` to the list of excluded files for instance.

src/ansys/aedt/core/hfss3dlayout.py
ecoussoux-ansys: This change is indeed due to ``ruff``. That function was in fact defined twice within the class ``Hfss3dLayout``, which triggered an F811 (redefined-while-unused) error. The two definitions for this function were very similar, but I decided to retain the second one (which anyway must have been the one used, right?) as it implements the function in the exact same way as in ``hfss.py`` for the ``Hfss`` class.
Do you agree with this change?

src/ansys/aedt/core/modeler/cad/primitives.py
ecoussoux-ansys: Sure, I deleted it in 3f4ef3a.

src/pyaedt/generic/near_field_import.py
ecoussoux-ansys: This change was indeed motivated by ``ruff``. This file and also ``src/pyaedt/generic/report_file_parser.py`` each contained only 1 line with an import that, if I am not mistaken, was never used, since I could find nowhere in the codebase instructions of the form ``from near_field_import import ...`` or ``from report_file_parser import ...``. These unused imports triggered F401 (unused-import) errors which are now fixed. On top of that, the two files were part of the "pyaedt side" of the ``src`` folder and had no equivalent in the ``ansys/aedt/core`` folder.
Do you agree with the deletion of these two files?

tests/integration/test_touchstone_data.py
ecoussoux-ansys: Again a change motivated by ``ruff`` here.

The original syntax with ``is not`` trigerred an F632 (is-literal) error, which I originally fixed by updating that part to ``res != []``, which seemed the most logical thing to do.
However, as the test subsequently started to fail, I checked locally and figured out that ``plot_insertion_losses()`` returns ``[]`` here! 
The ``ruff`` documentation on that topic (https://docs.astral.sh/ruff/rules/is-literal/) is quite informative and I believe provides a good explanation of what happens. Quoting from it: The ``is not`` comparator actually operates on identity and checks whether two objects are effectively the same, thus comparing the identities of the objects instead of their values. If the objects are not actually the same object, the comparison will always be ``False``, so that using ``is not`` (or ``is``) in such cases often works "by accident". This kind of comparisons against lists (or integers, dictionaries, sets, strings) is therefore not recommended.

In the end, this change makes the test pass again, but I am not sure if that was the right thing to do, as it seems that the intended purpose was to confirm that the output of ``plot_insertion_losses()`` was not an empty list..! I am obvisouly not very knowledgeable on that particular topic, but looking at the dummy ``touchstone_file`` used for this test, I am not that surprised by the output of ``plot_insertion_losses()``.
What is your take on this? Should we take further action and attempt to rewrite the test / document this somewhere, or simply leave it like this?

tests/system/general/test_08_Primitives3D.py
ecoussoux-ansys: I believe this is eventually not necessary, as the issue flagged by codacy was resolved by assigning the expression to ``_`` as you suggested. Do you agree?

src/ansys/aedt/core/extensions/installer/jupyter_template.ipynb
SMoraisAnsys: I though it was performed automatically on the project and wasn't intentional. If the cell wasn't needed then fine by me

src/ansys/aedt/core/hfss3dlayout.py
SMoraisAnsys: @Samuelopez-ansys Do you have any feedback on that ? I don't know which one should be kept.

src/pyaedt/generic/near_field_import.py
SMoraisAnsys: For back compatibility, I think we should be keeping that line of code :/ The thing is that a user could stil be importing the project with pyaedt and importing `convert_nearfield_data` from `near_field_import` :/
Could you please add this import as an exception. It will be removed once we get rid of the src/pyaedt folder.

tests/integration/test_touchstone_data.py
SMoraisAnsys: Leave it be, thanks for the extra explanation

tests/system/general/test_08_Primitives3D.py
SMoraisAnsys: I'm fine with it. It would just strengthen the test to be sure that the exception's message is the one expected but no worries here.

src/ansys/aedt/core/extensions/installer/jupyter_template.ipynb
ecoussoux-ansys: To be honest, I can't remember if that got auto-fixed by ``ruff`` or if I had to perform this manually, but in any case I double checked that the cell was indeed unnecessary.

src/pyaedt/generic/near_field_import.py
ecoussoux-ansys: Alright I understand. I reintroduced the files ``src/pyaedt/generic/near_field_import.py`` and ``src/pyaedt/generic/report_file_parser.py`` and deactivated the check for rule F401 on the performed imports in 4b00de5.

src/ansys/aedt/core/extensions/hfss/move_it.py
```diff
@@ -402,7 +401,7 @@ def main(data: MoveItExtensionData):
     hfss[index_var_name] = index_at_time
 
     if "PYTEST_CURRENT_TEST" not in os.environ:  # pragma: no cover
-        self.desktop.release_desktop(False, False)
+        app.release_desktop(False, False)
```
SMoraisAnsys, line 404: Nice catch !!!

src/ansys/aedt/core/hfss3dlayout.py
Samuelopez-ansys: Hi @ecoussoux-ansys @SMoraisAnsys 
Indeed this one was duplicated, and it is equivalent to the other one.

SMoraisAnsys (approved):

PR #6276, opened 2025-06-17
FEAT: Create coil extension

gmalinve:
## Description
**This new extension allows to create vertical and flat coils and parametrizes the model.**

## Issue linked

## Checklist


Samuelopez-ansys (approved):

Samuelopez-ansys (dismissed):

doc/source/User_guide/pyaedt_extensions_doc/maxwell/create_coil.rst
```diff
@@ -0,0 +1,29 @@
+Create coil design geometries
+=============================
+
+This extension aims  to create vertical or flat coil geometries in Maxwell 3D using specific parameters.
```
SMoraisAnsys:
```suggestion
This extension aims to create vertical or flat coil geometries in Maxwell 3D using specific parameters.
```

src/ansys/aedt/core/modeler/advanced_cad/coil.py
```diff
@@ -0,0 +1,388 @@
...
+            for attr in ["arc_segmentation", "section_segmentation", "wire_radius"]:
+                if hasattr(self, attr):
+                    if attr != "wire_radius":
+                        value = int(getattr(self, attr))
+                        if value == 1:
+                            self._app[attr] = 0
+                        else:
+                            self._app[attr] = value
+                    else:
+                        self._app[attr] = Quantity(getattr(self, attr), self._app.modeler.model_units)
```
SMoraisAnsys, line 98:
```suggestion
                    if attr == "wire_radius":
                        self._app[attr] = Quantity(getattr(self, attr), self._app.modeler.model_units)
                    else:
                        value = int(getattr(self, attr))
                        if value == 1:
                            self._app[attr] = 0
                        else:
                            self._app[attr] = value
```
This proposition is just for readability and highly subjective, feel free to discard.

Also, I'm wondering about the implementation of `sefl._app[attr]`. It feels like you are expecting the attributes to be `str` that need to be converted into `int` but not always with the direct value. Why is that ?

tests/system/extensions/test_create_coil.py
```diff
@@ -0,0 +1,150 @@
...
+        turns="5",
+        inner_width="12mm",
+        inner_length="6mm",
+        wire_radius="1mm",
+        inner_distance="2mm",
+        arc_segmentation="1",
+        section_segmentation="1",
+        distance_turns="5mm",
+        looping_position="0.5",
+    )
```
SMoraisAnsys: Unneeded here, only define the name and use it later in `name_text`.
Otherwise, do as the following test and create the extension data then use it with main directly.

tests/system/extensions/test_create_coil.py
```diff
@@ -0,0 +1,150 @@
...
+    assert main(data)
+    assert len(m3d_app.modeler.solid_objects) == 1
```
SMoraisAnsys, line 82: Can you check the there are no solid objects before the call to `main(data)` ?

tests/system/extensions/test_create_coil.py
```diff
@@ -0,0 +1,150 @@
...
+        wire_radius="1mm",
+        inner_distance="2mm",
+        direction="1",
+        pitch="3mm",
+        arc_segmentation="1",
+        section_segmentation="1",
+    )
+    assert main(data)
+    assert len(m3d_app.modeler.solid_objects) == 1
+    assert {"arc_segmentation", "section_segmentation"}.issubset(m3d_app.variable_manager.design_variable_names)
```
SMoraisAnsys: Same comment as before.
Also can you add the assert used in the previous test ? `assert m3d_app.modeler.solid_objects[0].name == "my_coil"`

tests/unit/extensions/test_create_coil.py
```diff
@@ -0,0 +1,97 @@
...
+
+@patch("ansys.aedt.core.extensions.misc.Desktop", new_callable=PropertyMock)
+def test_extension_default(mock_desktop, mock_aedt_app):
+    """Test instantiation of the Advanced Fields Calculator extension."""
+    mock_desktop.return_value = MagicMock()
+
+    extension = CoilExtension(withdraw=False)
+
+    assert EXTENSION_TITLE == extension.root.title()
+    assert "light" == extension.root.theme
```
SMoraisAnsys, line 55: Maybe you could also check here that every textlabel is empty ?

tests/unit/extensions/test_create_coil.py
```diff
@@ -0,0 +1,97 @@
...
+
+    extension.root.destroy()
+
+
+@patch("ansys.aedt.core.extensions.misc.Desktop", new_callable=PropertyMock)
+def test_create_button(mock_desktop, mock_aedt_app):
```
SMoraisAnsys, line 64: Shouldn't you have an exception raised if your user doesn't provide information for all the required data ?
I though that would be the default behavior following our previous discussions.

tests/unit/extensions/test_move_it.py
```diff
@@ -70,7 +70,7 @@ def test_move_it_extension_generate_button(mock_desktop, mock_aedt_app):
     """Test instantiation of the Move It extension."""
     mock_desktop.return_value = MagicMock()
 
-    extension = MoveItExtension(withdraw=True)
+    extension = MoveItExtension(withdraw=False)
```
SMoraisAnsys:
```suggestion
    extension = MoveItExtension(withdraw=True)
```
Otherwise the extension will keep being open

SMoraisAnsys (commented):
I left some comments

src/ansys/aedt/core/extensions/maxwell3d/create_coil.py
```diff
@@ -0,0 +1,510 @@
...
+from ansys.aedt.core.generic.numbers import Quantity
+from ansys.aedt.core.internal.errors import AEDTRuntimeError
+from ansys.aedt.core.modeler.advanced_cad.coil import Coil
+
+PORT = get_port()
+VERSION = get_aedt_version()
+AEDT_PROCESS_ID = get_process_id()
+IS_STUDENT = is_student()
+EXTENSION_TITLE = "Create coil"
+EXTENSION_DEFAULT_ARGUMENTS = {
```
SMoraisAnsys, line 49: Second pass of review, sry for the echo but I'm not sure why you are defining a default extension value. AFAIK there shouldn't be any for your repo. This has also impact on the dataclass definition.

src/ansys/aedt/core/extensions/maxwell3d/create_coil.py
```diff
@@ -0,0 +1,510 @@
...
+        self.inner_length_text = None
+        self.wire_radius_text = None
+        self.inner_distance_text = None
+        self.arc_segmentation_text = None
+        self.section_segmentation_text = None
+        self.looping_position_text = None
+        self.distance_text = None
+        self.z_pos_text = None
+        self.direction_text = None
+        self.pitch_text = None
```
SMoraisAnsys: @Samuelopez-ansys @gmalinve Would you mind if we gathered all thos widgets indides of a dict (e.g. self.__widgets = {}) which is filled while we populate the GUI ? This would hide a lot of attributes in the __init__ method.

tests/unit/extensions/test_create_coil.py
gmalinve: @SMoraisAnsys things have changed and I now have default values :)

src/ansys/aedt/core/extensions/maxwell3d/create_coil.py
```diff
@@ -0,0 +1,510 @@
...
+                self.pitch_text.delete("1.0", tk.END)
+                self.pitch_text.insert(tk.END, "3")
+
+        is_vertical_label = ttk.Label(self.root, text="Vertical Coil", style="PyAEDT.TLabel", width=20)
+        is_vertical_label.grid(row=0, column=0, pady=5, padx=5)
+        is_vertical = tk.IntVar(self.root, name="is_vertical", value=1)
+        self.check = ttk.Checkbutton(
+            self.root, variable=is_vertical, style="PyAEDT.TCheckbutton", name="is_vertical", command=on_checkbox_toggle
+        )
+        self.check.grid(row=0, column=1, pady=5, padx=5)
```
SMoraisAnsys: Would you be open to using the common padding used in multiple extensions ? See https://github.com/ansys/pyaedt/blob/main/src/ansys/aedt/core/extensions/misc.py#L56 used in https://github.com/ansys/pyaedt/blob/eeec7477e3da58dfc4e34b111adfae41cce9ac4e/src/ansys/aedt/core/extensions/project/via_design.py#L136 ?
If not, it might be worth to save that into you own constant variable and use it everywhere if possible.

src/ansys/aedt/core/extensions/maxwell3d/create_coil.py
```diff
@@ -0,0 +1,510 @@
...
+                inner_distance=self.inner_distance_text.get("1.0", tk.END).strip(),
+                direction=self.direction_text.get("1.0", tk.END).strip(),
+                pitch=self.pitch_text.get("1.0", tk.END).strip(),
+                arc_segmentation=self.arc_segmentation_text.get("1.0", tk.END).strip(),
+                section_segmentation=self.section_segmentation_text.get("1.0", tk.END).strip(),
+                looping_position=self.looping_position_text.get("1.0", tk.END).strip(),
+                distance_turns=self.distance_text.get("1.0", tk.END).strip(),
+            )
+            try:
+                self.coil.validate_coil_arguments(asdict(data), coil_type=data.coil_type)
```
SMoraisAnsys, line 403: Great to check here ! Some of my previous comment might be unworthy (not all of them though)

src/ansys/aedt/core/extensions/maxwell3d/create_coil.py
```diff
@@ -0,0 +1,510 @@
...
+        start_position = Quantity(data_dict["looping_position"]).value
+        start_point = [
+            centre_x + 0.25 * inner_x,
+            centre_y - (start_position - 0.5) * inner_y,
+            0,
+        ]
+    # Create coil profile
+    coil.create_sweep_profile(start_point, polyline)
+    # Replace 3D Component
+    # aedtapp.modeler.replace_3dcomponent(name=data.name)
```
SMoraisAnsys, line 474: Clean that line ?

SMoraisAnsys (commented):
Second pass of review

tests/unit/extensions/test_create_coil.py
gmalinve: @SMoraisAnsys entries have changed but I'll check that some entries have the expected value.

src/ansys/aedt/core/modeler/advanced_cad/coil.py
gmalinve: @SMoraisAnsys because I need to check if the segmentation (both arc and section) is 1. If this is the case then the value must be 0.

src/ansys/aedt/core/extensions/maxwell3d/create_coil.py
gmalinve: @SMoraisAnsys actually I left it intentionally because this will be the second step of implementation :)

src/ansys/aedt/core/extensions/maxwell3d/create_coil.py
gmalinve: @SMoraisAnsys do you mean something like that? self.__widgets = {
    "check": None,
    "name_text": None,
    "x_pos_text": None,
    "y_pos_text": None,
    "turns_text": None,
    "inner_width_text": None,
    "inner_length_text": None,
    "wire_radius_text": None,
    "inner_distance_text": None,
    "arc_segmentation_text": None,
    "section_segmentation_text": None,
    "looping_position_text": None,
    "distance_text": None,
    "z_pos_text": None,
    "direction_text": None,
    "pitch_text": None,
}

src/ansys/aedt/core/extensions/maxwell3d/create_coil.py
gmalinve: sure but I have to change it to my own values. @SMoraisAnsys

src/ansys/aedt/core/extensions/maxwell3d/create_coil.py
SMoraisAnsys: More like `self.__widgets = {}` in `__init__` and then, this gets new key,value while the UI is built.

Samuelopez-ansys (approved):
Good job @gmalinve 

In the second round, let's create the 3D Component, and it will be a very powerful extension.

Issue #6278, opened 2025-06-18
Improve Hfss.create_setup() to parse a frequency list.


Devin-Crawford:
### Description of the current limitations

If a user passes a list of frequencies to the ``Hfss.create_setup()`` method the list should be interpreted as a request for multiple frequency adaptive refinement. 

### Code sample expected

If a list is passed to the keyword ``Frequency``, this should be interpreted as the argument to ``"MultipleAdaptiveFreqSetup"`` which will be handled correctly in ``Hfss.create_setup()``.

### Useful links and references

See this example: https://examples.aedt.docs.pyansys.com/version/dev/examples/high_frequency/antenna/dipole.html

PR #6279, opened 2025-06-18
FEAT: Update create_setup method

Devin-Crawford:
Allow a list to be passed
for the "Frequency" argument to ``Hfss.create_setup()`` and automatically apply the multi-frequency adapt setup.

## Description
Resolve issue #6278. Add unit test to ``/test/system/general/test_11_Setup.py``

## Issue linked

## Checklist

### DIFF
--- a/src/ansys/aedt/core/hfss.py
+++ b/src/ansys/aedt/core/hfss.py
@@ -1795,6 +1795,10 @@ class Hfss(FieldAnalysis3D, ScatteringMethods, CreateBoundaryMixin):
         name = self.generate_unique_setup_name(name)
         setup = self._create_setup(name=name, setup_type=setup_type)
         setup.auto_update = False
+        if "Frequency" in kwargs.keys():
+            if type(kwargs["Frequency"] ) is list:
+                if not "MultipleAdaptiveFreqsSetup" in kwargs.keys():
+                    kwargs["MultipleAdaptiveFreqsSetup"] = kwargs["Frequency"]
         for arg_name, arg_value in kwargs.items():
             if setup[arg_name] is not None:
                 if arg_name == "MultipleAdaptiveFreqsSetup":  # A list of frequency values is passed if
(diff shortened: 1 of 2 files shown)


src/ansys/aedt/core/hfss.py
```diff
@@ -1795,6 +1795,10 @@ def create_setup(self, name="MySetupAuto", setup_type=None, **kwargs):
         name = self.generate_unique_setup_name(name)
         setup = self._create_setup(name=name, setup_type=setup_type)
         setup.auto_update = False
+        if "Frequency" in kwargs.keys():
+            if type(kwargs["Frequency"]) is list:
+                if "MultipleAdaptiveFreqsSetup" not in kwargs.keys():
+                    kwargs["MultipleAdaptiveFreqsSetup"] = kwargs["Frequency"]
```
SMoraisAnsys, line 1801:
```suggestion
        if "Frequency" in kwargs.keys() and type(kwargs["Frequency"]) is list:
            kwargs["MultipleAdaptiveFreqsSetup"] = kwargs["Frequency"]
```
I'm not sure if there is a need to check if `"MultipleAdaptiveFreqsSetup"` is in `kwargs`. Do we have a usecase where one defines `kwargs["Frequency"]` with a list and `kwargs["MultipleAdaptiveFreqsSetup"]` with a different one ?
Btw, what happens if the list is of length 1 ? Does the logic change from providing a single frequency value ?

Devin-Crawford: @SMoraisAnsys : Yes. This is a built-in keyword in the native API so we should proceed with caution. Passing ''**kwargs'' allows the user to pass any keywords that are valid in the native API.

SMoraisAnsys: Ok, feel free to resolve the discussion if the list of length 1 is not an issue

Devin-Crawford: Will close. The proposed solution will throw an error if "Frequency" is not a key in kwargs.

Samuelopez-ansys (approved):

PR #6287, opened 2025-06-20
REFACTOR: Configure layout

hui-zhou-a:
- Add comments in toml
- Add test for batch mode

### DIFF
--- a/tests/system/extensions/test_configure_layout.py
+++ b/tests/system/extensions/test_configure_layout.py
@@ -80,12 +80,13 @@ def test_configure_layout_export(mock_askdirectory, local_scratch, add_app):
 
 @patch("tkinter.filedialog.askdirectory")
 def test_configure_layout_batch(mock_askdirectory, local_scratch):
+    from ansys.aedt.core.extensions.project.configure_layout import main
+
     test_dir = Path(local_scratch.path)
     mock_askdirectory.return_value = str(test_dir)
     extension = ConfigureLayoutExtension(withdraw=False)
     extension.root.nametowidget("notebook").nametowidget("load").nametowidget("generate_template").invoke()
 
-    from ansys.aedt.core.extensions.project.configure_layout import main
-
     main(str(test_dir / "output"), str(Path(test_dir) / "example_serdes.toml"))
+
     assert (test_dir / "output" / "ANSYS_SVP_V1_1.aedb").exists()


src/ansys/aedt/core/extensions/project/configure_layout.py
```diff
@@ -423,9 +425,9 @@ def export_config_from_design():
 def main(
     working_directory,
     config_file,
-):  # pragma: no cover
-    ExtensionDataLoad.fpath_config = config_file
-    ExtensionDataLoad.working_directory = working_directory
+):
+    ExtensionDataLoad.fpath_config = Path(config_file)
+    ExtensionDataLoad.working_directory = Path(working_directory)
```
SMoraisAnsys, line 430: @hui-zhou-a Why are you working with `ExtensionDataLoad` in such a way instead of creating an instance and leveraging that instance ?

src/ansys/aedt/core/extensions/project/resources/configure_layout/example_serdes.toml
```diff
@@ -1,12 +1,12 @@
 title = "Demo Signal Integrity"
 
-version = "2025.1"
-layout_file = "D:/_pycharm_project/example-data/pyaedt/edb/ANSYS_SVP_V1_1.aedb"
+version = "2025.1"  # Only applicable in batch mode
+layout_file = "D:/_pycharm_project/example-data/pyaedt/edb/ANSYS_SVP_V1_1.aedb"  # Source layout. Available formats are aedt and aedb
```
SMoraisAnsys: To help with user's reading, shouldn't we add the comments above the assignment ? For example
```
# Only applicable in batch mode
version = "2025.1"
```

tests/system/extensions/test_configure_layout.py
```diff
@@ -76,3 +76,16 @@ def test_configure_layout_export(mock_askdirectory, local_scratch, add_app):
...
+def test_configure_layout_batch(mock_askdirectory, local_scratch):
+    test_dir = Path(local_scratch.path)
+    mock_askdirectory.return_value = str(test_dir)
+    extension = ConfigureLayoutExtension(withdraw=False)
+    extension.root.nametowidget("notebook").nametowidget("load").nametowidget("generate_template").invoke()
+
+    from ansys.aedt.core.extensions.project.configure_layout import main
+
+    main(str(test_dir / "output"), str(Path(test_dir) / "example_serdes.toml"))
+    assert (test_dir / "output" / "ANSYS_SVP_V1_1.aedb").exists()
```
SMoraisAnsys:
```suggestion
@patch("tkinter.filedialog.askdirectory")
def test_configure_layout_batch(mock_askdirectory, local_scratch):
    from ansys.aedt.core.extensions.project.configure_layout import main

    test_dir = Path(local_scratch.path)
    mock_askdirectory.return_value = str(test_dir)
    extension = ConfigureLayoutExtension(withdraw=False)
    extension.root.nametowidget("notebook").nametowidget("load").nametowidget("generate_template").invoke()

    main(str(test_dir / "output"), str(Path(test_dir) / "example_serdes.toml"))

    assert (test_dir / "output" / "ANSYS_SVP_V1_1.aedb").exists()
```

src/ansys/aedt/core/extensions/project/resources/configure_layout/example_serdes.toml
hui-zhou-a: Good suggestion.

src/ansys/aedt/core/extensions/project/configure_layout.py
hui-zhou-a: @SMoraisAnsys class and instance work in the same way when only one instance is needed, which is the case. I found using class is easier.

Samuelopez-ansys (approved):

src/ansys/aedt/core/extensions/project/configure_layout.py
SMoraisAnsys: I agree but this isn't a good habit to have. Better work with an instance attribute than with the class attributes as you're sure to avoid any side effect.
Pinging @Samuelopez-ansys for visibility

PR #6291, opened 2025-06-20
REFACTOR: refactored settings.py to use pathlib

Alberto-DM:
## Description
Refactrored the settings.py to leverage pathlib instead of os.path

## Issue linked
#5180

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,9 +23,9 @@ exclude: |
 repos:
 
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: v0.11.9
+  rev: v0.12.0
   hooks:
-    - id: ruff
+    - id: ruff-check
     - id: ruff-format
     
 - repo: https://github.com/codespell-project/codespell
--- a/README.md
+++ b/README.md
@@ -16,7 +16,9 @@
 [![PythonVersion](https://img.shields.io/badge/python-3.8+-blue.svg)](https://www.python.org/downloads/)
 [![GH-CI](https://github.com/ansys/pyaedt/actions/workflows/unit_tests.yml/badge.svg)](https://github.com/ansys/pyaedt/actions/workflows/unit_tests.yml)
 [![codecov](https://codecov.io/gh/ansys/pyaedt/branch/main/graph/badge.svg)](https://codecov.io/gh/ansys/pyaedt)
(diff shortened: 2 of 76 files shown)


src/ansys/aedt/core/generic/settings.py
```diff
@@ -825,17 +832,26 @@ def filter_settings_with_raise(settings: dict, allowed_keys: List[str]):
                     raise KeyError("An environment variable key is not part of the allowed keys.")
                 self.aedt_environment_variables = settings
 
-    def writte_yaml_configuration(self, path: str):
+    def write_yaml_configuration(self, path: Path | str):
```
SMoraisAnsys: Not sure if that 
```suggestion
    def write_yaml_configuration(self, path: Union[Path, str]):
```
The `|` syntax was introduced in python 3.10. Since we should be compatible, to some extent, with python 3.8, I wouldn't use `|` in settings.

src/ansys/aedt/core/generic/settings.py
```diff
@@ -825,17 +832,26 @@ def filter_settings_with_raise(settings: dict, allowed_keys: List[str]):
...
         """Write the current settings into a YAML configuration file."""
         import yaml
 
+        configuration_file = Path(path)
+
         data = {}
-        data["log"] = {key: getattr(self, key) for key in ALLOWED_LOG_SETTINGS}
-        data["lsf"] = {key: getattr(self, key) for key in ALLOWED_LSF_SETTINGS}
+        data["log"] = {
+            key: str(value) if isinstance(value := getattr(self, key), Path) else value for key in ALLOWED_LOG_SETTINGS
```
SMoraisAnsys: Same here, I'm not sure that this syntax (`:=`) is available in python 3.8

src/ansys/aedt/core/generic/settings.py
Alberto-DM: Done

src/ansys/aedt/core/generic/settings.py
Alberto-DM: you are right. I checked and this syntax works with python >= 3.9.
Fixed

tests/system/general/test_utils.py
```diff
@@ -52,9 +52,9 @@ def test_settings_load_default_yaml():
     assert default_settings_attributes == local_settings_attributes
 
 
-def test_settings_writte_default_yaml(tmp_path):
```
SMoraisAnsys, line 55: Nice typo catching !

SMoraisAnsys (approved):
LGTM

tests/system/general/test_utils.py
Alberto-DM: thanks ;)

PR #6310, opened 2025-06-25
FIX: Notify VTK for changes in the animation loop

eblanco-ansys:
## Description
Fixed animation issues in pyvista. It was necessary to notify VTK for the changes and refresh the view

## Issue linked
#6309 

## Checklist

### DIFF
--- a/doc/changelog.d/6332.fixed.md
+++ b/doc/changelog.d/6332.fixed.md
@@ -1,0 +1,1 @@
+Skip move on circuit if it is running on linux in non-graphical mode
--- a/src/ansys/aedt/core/modeler/circuits/object_3d_circuit.py
+++ b/src/ansys/aedt/core/modeler/circuits/object_3d_circuit.py
@@ -667,7 +667,7 @@ class CircuitComponent(object):
                 elif pin == "VAL" or pin not in self.parameters.keys():
                     self._pins.append(CircuitPins(self, pin, idx))
                 idx += 1
-        except AttributeError:
+        except (TypeError, AttributeError):
             self._pins.append(CircuitPins(self, self.composed_name, idx))
         return self._pins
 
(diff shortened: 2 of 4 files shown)


src/ansys/aedt/core/visualization/plot/pyvista.py
```diff
@@ -1587,8 +1587,12 @@ def p_callback():
                 i = 0
                 first_loop = False
             mesh_i = self.frames[i]._cached_polydata
-            scalars = mesh_i.point_data[self.frames[i].scalar_name]
-            mesh_i.point_data[self.frames[i].scalar_name] = scalars
+            new_scalars = mesh_i.point_data[self.frames[i].scalar_name]
+            # Update the existing mesh data directly
+            displayed_mesh = self.frames[0]._cached_polydata
+            displayed_mesh.point_data[self.frames[0].scalar_name] = new_scalars
```
SMoraisAnsys: What is the reason to directly use `frames[0]` instead of `frames[i]` ? Doesn't that break the logic associated to the use of `i` in this loop ?

eblanco-ansys: When pointing to `frames[0]` what is done is to change the data of the active frame, basically refreshing the animation with the new data. `i` is just used to obtain the point data associated with the current frame of the animation

Samuelopez-ansys: Hi @eblanco-ansys and @SMoraisAnsys I think I fixed the issue and I added tests. Both you were right, the animate method was created to reuse the same mesh and just update the scalar values, but I have added one extra step to check the if the mesh has changed, in that case, we clean the mesh and add the new one.

Samuelopez-ansys (dismissed):

Samuelopez-ansys (approved):

PR #6328, opened 2025-06-28
REFACTOR: configure layout

hui-zhou-a:
- Add support for relative path
- remove the test on help links as it leads to chicken and egg problem
- Fix working directory problem when launch from AEDT
- Add support for loading configuration to active design


src/ansys/aedt/core/extensions/project/configure_layout.py
```diff
@@ -437,6 +438,9 @@ def main(
     args = get_arguments(EXTENSION_TITLE)
 
     if not args["is_batch"]:
+        temp = Path(tempfile.TemporaryDirectory(suffix=".ansys").name)
+        temp.mkdir()
+        ExtensionDataLoad.working_directory = temp
         extension: ExtensionCommon = ConfigureLayoutExtension(withdraw=False)
```
SMoraisAnsys:
```suggestion
        extension: ExtensionCommon = ConfigureLayoutExtension(withdraw=False)
        extension.working_directory = temp
```
As mentioned in the previous PR please avoid modifying the class directly and update the instance instead.

src/ansys/aedt/core/extensions/project/configure_layout.py
```diff
@@ -379,12 +384,9 @@ def export_template_config():
...
 
         try:
-            dst_path = download_file(source="edb/ANSYS_SVP_V1_1.aedb", local_path=ExtensionDataLoad.working_directory)
+            download_file(source="edb/ANSYS_SVP_V1_1.aedb", local_path=ExtensionDataLoad.working_directory)
         except Exception:  # pragma: no cover
-            dst_path = ""
-
-        if dst_path:
-            content = re.sub(r'(layout_file\s*=\s*)".+?"', lambda m: m.group(1) + f"'{dst_path}'", content)
+            print("Failed to download example board.")
```
SMoraisAnsys:
```suggestion
            raise AEDTRuntimeError("Failed to download example board.")
```
This might need an import

src/ansys/aedt/core/extensions/project/configure_layout.py
```diff
@@ -348,10 +348,15 @@ def add_extension_content(self):
...
     def load_config():
-        config = CfgConfigureLayout(ExtensionDataLoad.fpath_config)
+        fpath_config = Path(ExtensionDataLoad.fpath_config)
+        config = CfgConfigureLayout(fpath_config)
         config.version = VERSION
+        layout_file = Path(config.layout_file)
 
-        app = Edb(edbpath=str(config.layout_file), edbversion=config.version)
+        if not bool(layout_file.drive):
+            layout_file = fpath_config.parent / layout_file
```
SMoraisAnsys: Could you exaplain those changes ?

src/ansys/aedt/core/extensions/project/resources/configure_layout/example_serdes.toml
```diff
@@ -3,8 +3,8 @@ title = "Demo Signal Integrity"
 # Specify AEDT version. Only applicable in batch mode
 version = "2025.1"
 
-# Source layout. Available formats are aedt and aedb
-layout_file = "D:/_pycharm_project/example-data/pyaedt/edb/ANSYS_SVP_V1_1.aedb"
+# Source layout. Available formats are aedt and aedb. Support both absolute and relative path
```
SMoraisAnsys:
```suggestion
# Source layout. Available formats are AEDT and AEDB. Support both absolute and relative path
```

tests/system/extensions/test_configure_layout.py
```diff
@@ -25,27 +25,11 @@
...
-import requests
-
 import ansys.aedt.core
-from ansys.aedt.core.extensions.project.configure_layout import GUIDE_LINK
-from ansys.aedt.core.extensions.project.configure_layout import INTRO_LINK
 from ansys.aedt.core.extensions.project.configure_layout import ConfigureLayoutExtension
 from ansys.aedt.core.extensions.project.configure_layout import ExtensionDataLoad
 
 
-def test_links():
```
SMoraisAnsys, line 37: Why are you removing that test ? If needed, you could just link to the dev doc instead of the stable one.

SMoraisAnsys (commented):
Thanks for the improvements, I've got a few comments on the changes.

src/ansys/aedt/core/extensions/project/configure_layout.py
hui-zhou-a: It is not intended to break here. Refactored to use msg instead.

src/ansys/aedt/core/extensions/project/configure_layout.py
hui-zhou-a: This is to support both absolute and relative path. 
c:edb/layout.aedb is a absolute path. 
edb/layout.aedb is a relative path. It will look for aedb directory in the same directory of toml file.

tests/system/extensions/test_configure_layout.py
hui-zhou-a: Brought test back and use dev instead.

src/ansys/aedt/core/extensions/project/configure_layout.py
SMoraisAnsys: When used through the UI, raisins an errer opens a dialogue Windows and doesnt break AFAIK

hui-zhou-a: The logic is like this
if there is internet connection, it copies the config files and downloads the example Edb.
if there is no internet connection, it copies the config files but skip downloading the example Edb. 

I refactored to add message instead of raising error or print.

src/ansys/aedt/core/extensions/project/configure_layout.py
```diff
@@ -378,23 +391,24 @@ def export_template_config():
         with open(example_master_config, "r", encoding="utf-8") as file:
             content = file.read()
 
-        try:
-            dst_path = download_file(source="edb/ANSYS_SVP_V1_1.aedb", local_path=ExtensionDataLoad.working_directory)
-        except Exception:  # pragma: no cover
-            dst_path = ""
+        example_edb = download_file(source="edb/ANSYS_SVP_V1_1.aedb", local_path=ExtensionDataLoad.working_directory)
```
SMoraisAnsys: Why are you removing the try except ? Wouldn't an exception be triggered when one tries to download a file without internet connection ? Or, are you exploiting the fact that by default the settings are set to handle the exception ?

hui-zhou-a: There is a use case that the user doesn't have internet. In this case, the user is allowed to get the configures as reference. However, he/she cannot run the example right away.

hui-zhou-a: The exception is handled by pyaedt.logger, and there will never be an exception in the extension.

Samuelopez-ansys (dismissed):
LGTM

Samuelopez-ansys (approved):

PR #6342, opened 2025-07-01
REFACTOR: Configure layout rlc on cap

hui-zhou-a:
Use point terminals to form ports instead of padstack terminals. This is a workaround to support converting ports to RLC boundary in MCAD when ECAD is imported as a 3dlayout component.

### DIFF
--- a/doc/changelog.d/6342.miscellaneous.md
+++ b/doc/changelog.d/6342.miscellaneous.md
@@ -1,0 +1,1 @@
+Configure layout rlc on cap


src/ansys/aedt/core/extensions/project/configure_layout.py
```diff
@@ -121,13 +121,13 @@ def get_edb_config_dict(self, edb: Edb):
         cfg_ports = []
         for i in self.rlc_to_ports:
             comp = edb.components[i]
-            p1, p2 = list(comp.pins.keys())
+            layer = comp.placement_layer
```
Samuelopez-ansys, line 124: Are these lines covered by the test?

hui-zhou-a: It is covered. Please refresh codacy.

Samuelopez-ansys (approved):

Issue #5328, opened 2024-10-22
Add method to order coil terminals in Transient A-Phi Formulation


AlfVII:
### Description of the current limitations

When working with coil terminals and windings in the Transient A-Phi Formulation Ansys requires that you order the coils after adding them. This can be done in the GUI by right-clicking on the winding, and in IronPython by the `oModule.OrderCoilTerminals`. I found no way of doing this with PyAedt.

### Code sample expected

Maybe something like 
`project.order_coil_terminals(winding, list_of_terminals)` ?

nunesanderson:
The API was fixed.
@gmalinve just letting you know.

gmalinve:
@nunesanderson I don't see any implementation. Where is it? 
No PR has ever been linked to this issue so I guess none took care of it.

nunesanderson:
@jvela018 Could you take a look on it?

PR #6360, opened 2025-07-07
FEAT: Control order connection between coil terminals in Maxwell3D TransientAPhiFormulation

jvela018:
## Description
This feature enables the user to create custom connection order between different coil terminals in the same winding definition

## Issue linked
#5328 

## Checklist

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -50,6 +50,7 @@ jobs:
           python-package-name: ${{ env.PACKAGE_NAME }}
           token: ${{ secrets.PYANSYS_CI_BOT_TOKEN }}
           dev-mode: ${{ github.ref != 'refs/heads/main' }}
+          extra-targets: 'all'
 
   # NOTE: We do not allow dependabot to trigger the CI/CD pipeline automatically.
   # This is to mitigate supply chain attacks, where a malicious dependency update
--- a/doc/changelog.d/6363.maintenance.md
+++ b/doc/changelog.d/6363.maintenance.md
@@ -1,0 +1,1 @@
+Add dependency check on all target
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -95,7 +95,6 @@ graphics = [
 jupyter = [
     "jupyterlab>=3.6.0,<4.5",
     "ipython>=7.30.0,<9.1",
-    "ipyvtklink>=0.2.0,<0.2.4",
 ]
 all = [
     "matplotlib>=3.5.0,<3.11",


tests/system/general/test_28_Maxwell3D.py
```diff
@@ -993,3 +993,61 @@ def test_enable_harmonic_force_layout(self, layout_comp):
...
+        terminal_list_order = [
+            "CoilTerminal2",
+            "CoilTerminal3",
+            "CoilTerminal5",
+            "CoilTerminal4",
+            "CoilTerminal1",
+            "CoilTerminal6",
+        ]
+
+        assert not m3d_app.order_coil_terminals(winding_name="Winding1", list_of_terminals=terminal_list_order)
```
nunesanderson: @jvela018 Sorry, but I did not get how this line would test if the order set is the correct one.

nunesanderson: Do we have a method to get the order and compare?

nunesanderson (dismissed):
Hi @jvela018 . I left one quick comment.
Thank you!
Anderson

src/ansys/aedt/core/maxwell.py
```diff
@@ -3179,6 +3179,50 @@ def assign_resistive_sheet(
 
         return self._create_boundary(name, props, "ResistiveSheet")
 
+    def order_coil_terminals(self, winding_name=None, list_of_terminals=None):
+        """Order Coil Terminals
+        This method enables the user to create custom connection order amongst different turns in a Winding definition.
```
SMoraisAnsys:
```suggestion
        """Order coil terminals

        Create custom connection order amongst different turns in a Winding definition.
```

src/ansys/aedt/core/maxwell.py
```diff
@@ -3179,6 +3179,50 @@ def assign_resistive_sheet(
...
+        Note that this feature is only available in the A-Phi formulation
```
SMoraisAnsys:
```suggestion
        Note that this feature is only available in the A-Phi formulation.
```

src/ansys/aedt/core/maxwell.py
```diff
@@ -3179,6 +3179,50 @@ def assign_resistive_sheet(
...
+
+        Examples
+        --------
+        >>> from ansys.aedt.core import Maxwell3d
+        >>> app = Maxwell3d(solution_type="TransientAPhiFormulation")
+        >>> app.order_coil_terminals(winding_name, list_of_terminals)
+        """
+        if self.solution_type != "TransientAPhiFormulation":
+            self.logger.warning("order_coil_terminals is only available in Transient A-Phi Formulation solution type")
+            return False
```
SMoraisAnsys: Raise an exception instead of returning False

src/ansys/aedt/core/maxwell.py
```diff
@@ -3179,6 +3179,50 @@ def assign_resistive_sheet(
...
+            winding = ["Name:" + winding_name]
+            args = []
+            args.extend(winding)
+            args.extend(list_of_terminals)
+            self.oboundary.OrderCoilTerminals(args)
+        except Exception:
+            raise AttributeError(
+                "Winding name must be a valid string, and list of terminals a valid list with valid names"
+            )
+        return True
```
SMoraisAnsys:
```suggestion
        except Exception as e:
            raise AttributeError(
                "Winding name must be a valid string, and list of terminals a valid list with valid names"
            ) from e
        return True
```

tests/system/general/test_28_Maxwell3D.py
jvela018: > @jvela018 Sorry, but I did not get how this line would test if the order set is the correct one.

@nunesanderson, you're right. The point of this test is not about the terminal list. That one should be reused. The point of this test is that it'll fail due to an attribute exception because this feature is only available on TransientAPhiFormulation. I'll modify the test and add a couple more. I'll request to review once I push the changes. Thanks

src/ansys/aedt/core/maxwell.py
```diff
@@ -3179,6 +3179,50 @@ def assign_resistive_sheet(
 
         return self._create_boundary(name, props, "ResistiveSheet")
 
+    def order_coil_terminals(self, winding_name=None, list_of_terminals=None):
```
gmalinve: the function handler is missing.

src/ansys/aedt/core/maxwell.py
```diff
@@ -3179,6 +3179,50 @@ def assign_resistive_sheet(
...
+        """Order coil terminals
+
+        Create custom connection order amongst different turns in a Winding definition.
+        Note that this feature is only available in the A-Phi formulation.
+
+        Parameters
+        ----------
+        winding_name : str
+            Name of Winding in AEDT
```
gmalinve, line 3192:
```suggestion
        winding_name : str, optional
            Name of winding in AEDT.
            If no name is provided a random name will be generated.
```
In the implementation you are not handling the case when winding_name = None. If it's none you should use the generate_unique_name("Winding") or whichever prefix you prefer.

src/ansys/aedt/core/maxwell.py
```diff
@@ -3179,6 +3179,50 @@ def assign_resistive_sheet(
...
+
+        list_of_terminals: lst of str
```
gmalinve, line 3194:
```suggestion
        list_of_terminals: lst of str, optional.
```
Why did you put it to None? and if this is the case you are not handling in the code implementation the value None. The method will fail.

src/ansys/aedt/core/maxwell.py
```diff
@@ -3179,6 +3179,50 @@ def assign_resistive_sheet(
...
+
+        References
+        ----------
+        >>> oModule.OrderCoilTerminals
+
+        Examples
+        --------
+        >>> from ansys.aedt.core import Maxwell3d
+        >>> app = Maxwell3d(solution_type="TransientAPhiFormulation")
+        >>> app.order_coil_terminals(winding_name, list_of_terminals)
```
gmalinve: This example doesn't tell much. Could you show how to create coils and then use this method to reorder them?

src/ansys/aedt/core/maxwell.py
```diff
@@ -3179,6 +3179,50 @@ def assign_resistive_sheet(
...
+        """
+        if self.solution_type != "TransientAPhiFormulation":
+            raise AttributeError("order_coil_terminals is only available in Transient A-Phi Formulation solution type")
```
gmalinve: I think is more an AEDTRuntimeError exception rather than attribute exception

src/ansys/aedt/core/maxwell.py
jvela018: @gmalinve, to use the feature both parameters winding and coil terminals need to be defined. Furthermore, the winding needs to be already populated with the terminals to be ordered. Otherwise, you can't use the feature. It would be possible to generate random names internally, but then the user wouldn't know what terminals or windings are available to them.

src/ansys/aedt/core/maxwell.py
```diff
@@ -3179,6 +3179,50 @@ def assign_resistive_sheet(
...
+
+        try:
+            winding = ["Name:" + winding_name]
+            args = []
+            args.extend(winding)
+            args.extend(list_of_terminals)
```
```suggestion
            args = ["Name:" + winding_name] + list_of_terminals
```
suggestion. The code in try/except can be as simple as this

src/ansys/aedt/core/maxwell.py
jvela018: @gmalinve the terminals can't be optional, and indeed it will raise an error. The method needs to get the list of terminals to be ordered.

src/ansys/aedt/core/maxwell.py
```diff
@@ -3179,6 +3179,50 @@ def assign_resistive_sheet(
...
+            self.oboundary.OrderCoilTerminals(args)
+        except Exception as e:
```
gmalinve: I don't think you need a try/ecept clause if you add the pyaedt function handler. If it fails than it's the function handler that handles it

src/ansys/aedt/core/maxwell.py
gmalinve: yes, but if you put winding_name=None and list_of_terminals=None it means that you want them to be optional.

src/ansys/aedt/core/maxwell.py
gmalinve: see the comment above

src/ansys/aedt/core/maxwell.py
jvela018: removed optional arguments.

src/ansys/aedt/core/maxwell.py
jvela018: removed optional arguments

src/ansys/aedt/core/maxwell.py
jvela018: function handler was added

src/ansys/aedt/core/maxwell.py
jvela018: extensive example added

src/ansys/aedt/core/maxwell.py
jvela018: AEDTRuntimeError added

src/ansys/aedt/core/maxwell.py
jvela018: modified to more compact form
self.oboundary.OrderCoilTerminals(["Name:" + winding_name, *list_of_terminals])

src/ansys/aedt/core/maxwell.py
jvela018: try/except has been removed and pyaedt_function_handler introduced

jvela018 (commented):
@gmalinve , I've addressed all your comments and pushed changes. Thanks

src/ansys/aedt/core/maxwell.py
```diff
@@ -3179,6 +3179,77 @@ def assign_resistive_sheet(
...
+        bool
+            ``True`` when successful, ``False`` when failed.
+
+        References
+        ----------
+        >>> oModule.OrderCoilTerminals
+
+        Examples
+        --------
+        >>> from ansys.aedt.core.generic.constants import Axis
```
gmalinve, line 3208: @jvela018 ok maybe it's a bit too much. ;) you need maximum three coils

tests/system/general/test_28_Maxwell3D.py
```diff
@@ -993,3 +993,52 @@ def test_enable_harmonic_force_layout(self, layout_comp):
             layout_comp.enable_harmonic_force_on_layout_component(
                 comp.name, {nets[0]: layers[1::2], nets[1]: layers[1::2]}
             )
+
+    def test_order_coil_terminals(self, m3d_app):
+        m3d_app.solution_type = "TransientAPhiFormulation"
```
gmalinve, line 998: @jvela018 you are not asserting the raise exception when the solution type is not transient aphi

jvela018: @gmalinve this is addressed in 1042

src/ansys/aedt/core/maxwell.py
jvela018: @gmalinve do you want me to change it to two coils? With three the example is nicer because it's got a connection to swapping current directions as well.

gmalinve: @jvela018 ok, let's keep it :)

tests/system/general/test_28_Maxwell3D.py
```diff
@@ -993,3 +993,52 @@ def test_enable_harmonic_force_layout(self, layout_comp):
...
+            "CoilTerminal2",
+            "CoilTerminal4",
+            "CoilTerminal3",
+            "CoilTerminal5",
+            "CoilTerminal6",
+        ]
+        assert m3d_app.order_coil_terminals(winding_name="Winding1", list_of_terminals=terminal_list_order)
+        m3d_app.solution_type = "Transient"
+        with pytest.raises(AEDTRuntimeError, match="Only available in Transient A-Phi Formulation solution type."):
+            assert m3d_app.order_coil_terminals(winding_name="Winding1", list_of_terminals=terminal_list_order)
```
gmalinve:
```suggestion
        with pytest.raises(AEDTRuntimeError, match="Only available in Transient A-Phi Formulation solution type."):
            m3d_app.order_coil_terminals(winding_name="Winding1", list_of_terminals=terminal_list_order)
```
if you commit from github check that the indentation is correct ;)

PR #6409, opened 2025-07-15
FIX: [order_coil_terminals] Transient naming convention in 2025R2

jvela018:
This method is only available in the APhi formulation. 
The test in 2025R2 fails because the native API's name for the solution type has changed from "TransientAPhiFormulation" to "Transient APhi". This PR should recognizes both names for backward compatibility

Issue #6408


src/ansys/aedt/core/maxwell.py
```diff
@@ -3639,7 +3639,8 @@ def order_coil_terminals(self, winding_name, list_of_terminals):
         ... ]
         >>> app.order_coil_terminals(winding_name=winding, list_of_terminals=updated_connection_order)
         """
-        if self.solution_type != "TransientAPhiFormulation":
+
```
```suggestion
```

jvela018:
Will do in a single PR to address all methods that face the same issue

PR #6411, opened 2025-07-15
FIX: [insert_layout_component] Transient naming change in 2025R2

jvela018:
TransientAPhiFormulation naming has changed in 2025R2 to Transient APhi

Issue #6410


jvela018:
Will do in a single PR to address all methods that face the same issue

PR #6425, opened 2025-07-17
FIX: analyze_from_zero

anur7:
## Description
This PR fixes the bug in analyze_from_zero() method in maxwell.py.
The missing unit test of the will also be added in this PR.

## Issue linked
Issue #6404 

## Checklist

### DIFF
--- a/src/ansys/aedt/core/maxwell.py
+++ b/src/ansys/aedt/core/maxwell.py
@@ -1639,7 +1639,14 @@ class Maxwell(CreateBoundaryMixin):
             raise AEDTRuntimeError("This methods work only with Maxwell Transient Analysis.")
 
         self.oanalysis.ResetSetupToTimeZero(self._setup)
-        self.analyze()
+        if self.boundaries[2].type == "Balloon":
+            self.logger.warning(
+                "With Balloon boundary is not possible to parallelize the simulation "
+                "using the Time Decomposition Method (TDM)."
+            )
+            self.analyze(cores=1)
+        else:
+            self.analyze()
         return True
 
     @pyaedt_function_handler(val="angle")


src/ansys/aedt/core/maxwell.py
```diff
@@ -1640,7 +1640,18 @@ def analyze_from_zero(self):
             raise AEDTRuntimeError("This methods work only with Maxwell Transient Analysis.")
 
         self.oanalysis.ResetSetupToTimeZero(self._setup)
-        self.analyze()
+        bound = []
+        for k in range(len(self.boundaries)):
```
nunesanderson: There would better ways to implement this for loop but I leave to @gmalinve to decide whether to accept it or not.
I would only leave a comment here explaining why you are including this check, since it should be removed after the aedt API is fixed.

nunesanderson (changes requested):
@gmalinve I will leave it to you to approve it.
Thanks @anur7 for fixing the issue.

src/ansys/aedt/core/maxwell.py
gmalinve: @nunesanderson @anur7 I don't personally like how this has been done and I don't think is efficient. Above all if there is a baloon boundary type you don't simply need to set the cores to 1 but also set the auto settings to False. Please look at my suggestion below.

src/ansys/aedt/core/maxwell.py
```diff
@@ -1640,7 +1640,18 @@ def analyze_from_zero(self):
...
+            if self.boundaries[k].type == "Balloon":
+                self.logger.warning(
+                    "With Balloon boundary is not possible to parallelize the simulation "
+                    "using the Time Decomposition Method (TDM)."
+                )
+                bound.append(self.boundaries[k].type)
+        if "Balloon" in bound:
+            self.analyze(cores=1)
```
```suggestion
        if any(boundary.type == "Balloon" for boundary in self.boundaries):
    self.logger.warning(
        "With Balloon boundary, it is not possible to parallelize the simulation "
        "using the Time Decomposition Method (TDM)."
    )
    self.analyze(use_auto_settings=False, cores=1)
```

tests/system/general/test_27_Maxwell2D.py
```diff
@@ -105,6 +105,13 @@ def m2d_setup(add_app):
     app.close_project(app.project_name)
 
 
+@pytest.fixture()
+def m2d_analyze(add_app):
```
gmalinve: @nunesanderson @anur7 please check because there should be already an m2d instance with an empty design. I don't think there is the need to create a new one.

tests/system/general/test_27_Maxwell2D.py
```diff
@@ -834,3 +841,14 @@ def test_export_rl_matrix(self, local_scratch, m2d_export_matrix):
             matrix_type="RL", matrix_name="Matrix1", output_file=export_path_2
         )
         assert os.path.exists(export_path_2)
+
+    def test_analyze_from_zero(self, m2d_analyze):
```
gmalinve: @nunesanderson @anur7 could you please also cover the case when no baloon boundary condition is present in your design?

nunesanderson: @gmalinve Shouldn't it be already covered? That part of the code was not touched.

gmalinve: @anur put an if/else condition and in this test is creating only a baloon type so the code doesn't go into the else (when no baloon is present).

src/ansys/aedt/core/maxwell.py
```diff
@@ -1640,7 +1640,14 @@ def analyze_from_zero(self):
...
         self.oanalysis.ResetSetupToTimeZero(self._setup)
-        self.analyze()
+        if any(boundary.type == "Balloon" for boundary in self.boundaries):
+    self.logger.warning(
+        "With Balloon boundary, it is not possible to parallelize the simulation "
+        "using the Time Decomposition Method (TDM)."
+    )
+    self.analyze(use_auto_settings=False, cores=1)
+        else:
+            self.analyze()
```
gmalinve, line 1660: @nunesanderson @anur7 please check the indentation.

src/ansys/aedt/core/maxwell.py
anur7: @gmalinve I applied your suggestion. Thanks :)

tests/system/general/test_27_Maxwell2D.py
anur7: @gmalinve @nunesanderson added test with non Balloon BC

tests/system/general/test_27_Maxwell2D.py
anur7: @gmalinve Thank you, I checked m2d_add creates an empty design for test

tests/system/general/test_27_Maxwell2D.py
```diff
@@ -834,3 +834,18 @@ def test_export_rl_matrix(self, local_scratch, m2d_export_matrix):
...
+        m2d_app.solution_type = "TransientXY"
+        conductor = m2d_app.modeler.create_circle(origin=[0, 0, 0], radius=10, material="Copper")
+        m2d_app.assign_winding(assignment=conductor.name, is_solid=False, current="5*cos(2*PI*50*time)")
+        region = m2d_app.modeler.create_region(pad_percent=100)
+        m2d_app.assign_balloon(region.edges)
+        setup1 = m2d_app.create_setup()
+        setup1.props["StopTime"] = "2/50s"
+        setup1.props["TimeStep"] = "1/500s"
+        assert m2d_app.analyze_from_zero()
+        m2d_app.modeler.delete(region.name)
```
gmalinve: @nunesanderson @anur7 why are you deleting the region? you simply need to delete the balloon boundary and create vector potential.

anur7: @gmalinve I didn't found the method to delete the BC. I don't think it is implemented for maxwell, that's why I'm deleting the region so the balloon BC is going to be deleted

gmalinve: @anur7 yes it is implemented. You get the boundary object and call delete()

src/ansys/aedt/core/maxwell.py
```diff
@@ -1640,7 +1640,14 @@ def analyze_from_zero(self):
             raise AEDTRuntimeError("This methods work only with Maxwell Transient Analysis.")
 
         self.oanalysis.ResetSetupToTimeZero(self._setup)
-        self.analyze()
+        if any(boundary.type == "Balloon" for boundary in self.boundaries):
+            self.logger.warning(
+                "With Balloon boundary, it is not possible to parallelize the simulation "
+                "using the Time Decomposition Method (TDM)."
+            )
```
SMoraisAnsys, line 1657: Could you add à comment to specify that the simulation will be analyzed without auto settings and with a single core? The current message could make one think that the analyse wont be performed at all.

SMoraisAnsys (commented):
Just à minor comment

src/ansys/aedt/core/maxwell.py
SMoraisAnsys: Sry but would "running the simulation" be a better expression than "the simulation ran" since the solve call is peeformed afterward ?

PR #6457, opened 2025-07-23
Fix monitor documentation

lorenzovecchietti:
## Description
Some monitor classes were not included in the documentation. Also, the `min_aedt_version` decorator seems to be breaking Sphinx ability to retrieve the doc strings to add to the documentation.

### DIFF
--- a/src/ansys/aedt/core/internal/checks.py
+++ b/src/ansys/aedt/core/internal/checks.py
@@ -28,6 +28,7 @@ import os
 import warnings
 
 from ansys.aedt.core.internal.errors import AEDTRuntimeError
+from functools import wraps
 
 
 def install_message(target: str) -> str:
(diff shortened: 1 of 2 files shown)


src/ansys/aedt/core/internal/checks.py
```diff
@@ -85,6 +86,7 @@ def fetch_odesktop_from_desktop_class(item):
                 return desktop_class.odesktop
 
     def aedt_version_decorator(method):
+        @wraps(method)
```
SMoraisAnsys, line 91: Good catch !

src/ansys/aedt/core/internal/checks.py
```diff
@@ -85,6 +86,7 @@ def fetch_odesktop_from_desktop_class(item):
                 return desktop_class.odesktop
 
     def aedt_version_decorator(method):
```
SMoraisAnsys, line 88:
```suggestion
    def aedt_version_decorator(method):
        """Decorator to check AEDT version compatibility for a method."""
```

SMoraisAnsys (dismissed):
LGTM, thanks @lorenzovecchietti
I left a minor comment.

Samuelopez-ansys (approved):

SMoraisAnsys (approved):

PR #6461, opened 2025-07-24
FIX: Via Design Extension version issue

hui-zhou-a:
correct edbversion as input to ViaDesignBackend

### DIFF
--- a/doc/changelog.d/6367.miscellaneous.md
+++ b/doc/changelog.d/6367.miscellaneous.md
@@ -1,0 +1,1 @@
+Pathlib in multiple files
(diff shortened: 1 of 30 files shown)


SMoraisAnsys (approved):

SMoraisAnsys (commented):
Il this fixes the skip we had to do, could you remove it ?

hui-zhou-a:
> Il this fixes the skip we had to do, could you remove it ?

All fixed on pyedb side, but I am blocked by 25.2 which is not installed on pyedb test VM

Issue #6400, opened 2025-07-14
Unable to specify project name when opening an archive


Devin-Crawford:
### Before submitting the issue

- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### Description of the bug

When opening an archive (``"*.aedtz"``) file, a new project is always created with a random name. In many cases, the user may want to specify the name explicitly, and by default it seems more intuitive that the uncompressed project file name would match the archive name but end with the ``"*.aedt"`` extension.

### Steps To Reproduce

In ``"\ansys\aedt\core\applications\design.py"`` the setter of ``Design.oproject`` always creates a new file with a random name. Instead, it should try to create a project with the same name as the archive.

### Which Operating System are you using?

Windows

### Which Python version are you using?

3.10

### Installed packages

All required packages are installed.

PR #6553, opened 2025-08-19
FIX: Save project after recovering archive

Devin-Crawford:
## Description
**Save project with the same name as the archive. Increment the file name suffix if the file already exists.**

## Issue linked
#6400

## Checklist


tests/system/general/test_01_Design.py
```diff
@@ -51,354 +50,360 @@
...
-
-    def test_09_set_objects_temperature(self):
+    def test_03_design_type(self, aedtapp):
+        assert aedtapp.design_type == "HFSS"
+
+    def test_04_projectname(self, aedtapp):
+        assert aedtapp.project_name == test_project_name
+
+    def test_05_lock(self, aedtapp):
+        assert os.path.exists(aedtapp.lock_file)
```
gmalinve, line 114: @Devin-Crawford I guess the missing thing in this file is to refactor all the os.path

Devin-Crawford: @gmalinve This was from the main branch. Let's fix the unit tests in another PR. os.path is still widely used in the unit tests.

Devin-Crawford: See https://github.com/ansys/pyaedt/issues/6555

src/ansys/aedt/core/desktop.py
```diff
@@ -1081,7 +1081,7 @@ def project_path(self, project_name=None):
         else:
             oproject = self.active_project(project_name)
         if oproject:
-            return oproject.GetPath()
+            return Path(oproject.GetPath())
```
Samuelopez-ansys: In Returns docstring, it is still specified "str", I think changing to Path the return is dangerous because this could break user codes. Most of PyAEDT methods, returns strings (and we are first migrating to accept as an argument string or Path)

Devin-Crawford: That's fine. In the course of reviewing I had assumed this was a mistake.
```suggestion
            return oproject.GetPath()
```

Samuelopez-ansys (approved):
LGTM

tests/system/general/test_01_Design.py
```diff
@@ -51,354 +50,360 @@
...
+        aedtapp.set_temporary_directory(tempfile.gettempdir())
+
+    def test_08_objects(self, aedtapp):
+        print(aedtapp.oboundary)
+        print(aedtapp.oanalysis)
+        print(aedtapp.odesktop)
+        print(aedtapp.logger)
+        print(aedtapp.variable_manager)
+        print(aedtapp.materials)
+        print(aedtapp)
```
SMoraisAnsys: Should this LOCs be asserts or removed ?

Devin-Crawford:
@Samuelopez-ansys , @gmalinve : I added one more line in the unit test to hopefully improve codecov, but the lines being flagged by the codecov check and causing the failure for the PR are in fact tested. You can step through the unit test code with the debugger to verify this. Is there any way we can bypass the codecov/patch errors and merge this PR?

Devin-Crawford:
@gmalinve , @Samuelopez-ansys , @maxcapodi78 : Could you please approve? I fixed the unit test.

tests/system/general/test_01_Design.py
Devin-Crawford: I think they should be "assert". This is currently in the main branch.

SMoraisAnsys: Could you change the code to perform the corresponding assert calls ? You are right that this is currently in main but I think we shouldn't have it in the first place. I don't see a reason to keep performing a print in a test.

Devin-Crawford: Done.
```suggestion
        assert aedtapp.oboundary
        assert aedtapp.oanalysis
        assert aedtapp.odesktop
        assert aedtapp.logger
        assert aedtapp.variable_manager
        assert aedtapp.materials
        assert aedtapp
```

Issue #6558, opened 2025-08-21
Wrong calculation in amp2db.m


tadanegishi:
### Before submitting the issue

- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### Description of the bug

<img width="491" height="237" alt="Image" src="https://github.com/user-attachments/assets/54bf8ee1-fe52-4d5c-a334-ccabdfa6aa52" />

### Steps To Reproduce

use the method in MATLAB

### Which Operating System are you using?

Windows

### Which Python version are you using?

3.8

### Installed packages

NA

gmalinve:
Closing as per #6559

PR #6559, opened 2025-08-21
FIX: MATLAB script amp2db.m

tadanegishi:
## Description
Fix amp2db.m to convert dB scale data from linear scale data

## Issue linked
#6558 

## Checklist

### DIFF
--- a/src/pyaedt/sbrplus/matlab/amp2db.m
+++ b/src/pyaedt/sbrplus/matlab/amp2db.m
@@ -10,6 +10,6 @@ function db = amp2db(amp)
 % db = AMP2DB(amp)
 %
 % COPYRIGHT ANSYS, Inc. ALL RIGHTS RESERVED.
-db = 20*log10(real(amp.*conj(amp)));
+db = 10*log10(real(amp.*conj(amp)));
 end  % amp2db
 


gmalinve (changes requested):
Hi @tadanegishi,

You are modifying the wrong file.
PyAEDT source code has been moved to a different folder ansys/aedt/core and the file you want to modify is in the visualization/sbrplus folder.

scanta2:
I am not sure about deleting all that folder unless we check with Bob

Samuelopez-ansys:
@scanta2 @tadanegishi Please check with Bob, because PyAEDT should not host non python files, some of them are already converted to python like the HDM class:

https://aedt.docs.pyansys.com/version/stable/API/visualization/advanced.html#heterogeneous-data-message

If you need to have these files in another public repo, we could think moving them to https://github.com/ansys/example-data

However, this is not ideal, as the code is written in MATLAB and has not been tested in any CI/CD pipeline, so it is not safe to have.

tadanegishi:
@Samuelopez-ansys I have asked Stefano and Bob to suggest good location for those MATLAB scripts. And, I have explained the situation as follow.

Some MATLAB scripts are currently duplicated or only located in pyaedt. For example, we have 3 copies of amp2db.m.
- **AEDT installation**: ANSYS Inc\v252\AnsysEM\Delcross\Savant MATLAB
- **Deprecated folder:** pyaedt\src\pyaedt\sbrplus\matlab
- **Official pyaedt folder**: pyaedt\src\ansys\aedt\core\visualization\advanced\sbrplus\matlab

This PR is to delete MATLAB scripts in the **Deprecated folder** because we already have the same MATLAB scripts in the **Official pyaedt folder**. However, we don’t expect any MATLAB scripts in pyaedt and my suggestion is to migrate the MATLAB scripts to **AEDT installation**. Even if we decide to migrate the MATLAB scripts to **AEDT installation** in R26.1, we need to keep them in **Official pyaedt folder** for a while.

Thank you for suggesting new public repo, but I prefer to locate in AEDT installation. It would be important to locate the MATLAB scripts where users can find them easily.

Samuelopez-ansys:
Thanks @tadanegishi, then in 2026R1, we will remove these Matlab scripts. Thank you for your inputs!

PR #6574, opened 2025-08-26
FIX: 3dlayout component coordinate

hui-zhou-a:
This PR fixes the bug coordinates from 3dlayout component are not imported. 

The name of the coordinates from 3dl component is wrong
currently named as  f"{comp_name}"
Should be named as f"{comp_name}_{pin_name}

### DIFF
--- a/doc/changelog.d/6577.fixed.md
+++ b/doc/changelog.d/6577.fixed.md
@@ -1,0 +1,1 @@
+Configure layout test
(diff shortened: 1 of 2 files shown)


hui-zhou-a:
> Seems like this is not the only place where `components.instances.keys()` is used. Could you also update the coordinate systems for those cases if it makes sense ?
> 
> Also can you extend the description of your PR ? Without an explanation on why those changes are required, it's a bit complicated to follow the need.

@SMoraisAnsys Done

PR #6577, opened 2025-08-27
FIX: configure layout test

hui-zhou-a:
Remove below content from configure file. 
```
        "instances": [
            {
                "name": "Via313",
                "backdrill_parameters": {
                    "from_bottom": {"drill_to_layer": "Inner3", "diameter": "1mm", "stub_length": "0.2mm"}
                },
            },
            {
                "name": "Via314",
                "backdrill_parameters": {
                    "from_bottom": {"drill_to_layer": "Inner3", "diameter": "1mm", "stub_length": "0.2mm"}
                },
            },
        ],
```

The the vias in a fresh EDB don't have names. When it is imported, pyedb will give them names by their IDs. However, the their IDs are not guaranteed the same on different machine. This leads to Via313 not found.

### DIFF
--- a/src/ansys/aedt/core/extensions/project/resources/configure_layout/template.py
+++ b/src/ansys/aedt/core/extensions/project/resources/configure_layout/template.py
@@ -135,21 +135,7 @@ SERDES_CONFIG = {
                 "hole_range": "upper_pad_to_lower_pad",
                 "hole_parameters": {"shape": "circle", "diameter": "0.25mm"},
             },
-        ],
-        "instances": [
-            {
-                "name": "Via313",
-                "backdrill_parameters": {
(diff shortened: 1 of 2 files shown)


SMoraisAnsys (dismissed):

src/ansys/aedt/core/extensions/project/resources/configure_layout/template.py
```diff
@@ -135,21 +135,7 @@
...
-                },
-            },
-            {
-                "name": "Via314",
-                "backdrill_parameters": {
-                    "from_bottom": {"drill_to_layer": "Inner3", "diameter": "1mm", "stub_length": "0.2mm"}
-                },
-            },
-        ],
+        ]
```
SMoraisAnsys:
```suggestion
        ],
```

SMoraisAnsys (approved):

PR #6690, opened 2025-09-26
FEAT: Panels update automatically after update

eblanco-ansys:
## Description
Now panels auto update when updating pyaedt version. Only when updating from releases, not from a branch

## Checklist


src/ansys/aedt/core/extensions/installer/version_manager.py
```diff
@@ -504,16 +514,38 @@ def version_is_leq(version, other_version):
...
+                messagebox.showwarning(
+                    "Warning",
+                    "PyAEDT panels could not be updated in AEDT. You may need to reset them manually.",
+                )
+
             self.clicked_refresh(need_restart=True)
 
-    def reset_pyaedt_buttons_in_aedt(self):
-        response = messagebox.askyesno("Confirm Action", "Are you sure you want to proceed?")
+    def reset_pyaedt_buttons_in_aedt(self, confirm=True):
```
SMoraisAnsys, line 529: Is there a reason to add this confirmation option ? I'm not seeing it being used in your changes.

eblanco-ansys: The confirmation action only pops up when clicking "Reset AEDT Panels". If not, it is applied automatically to avoid breaking the extensions if anything changes inside `PersonalLib` folder

SMoraisAnsys: I don't see the `reset_pyaedt_buttons_in_aedt` ever used with `confirm=True`. Could you pinpoint a location where it is used ?

eblanco-ansys: https://github.com/ansys/pyaedt/blob/0fbf87eaa390af3f316b584d45522780d3dedc83/src/ansys/aedt/core/extensions/installer/version_manager.py#L293-L296

In this lines we don't pass any argument so the default is "True" and the messagebox appears

Issue #5787, opened 2025-02-18
Refactor the code to leverage Mixins


SMoraisAnsys:
I'm pretty sure there are a lot of common methods or snippets that are in multiple classes.
It would be great if we could come up with some Mixin classes to add reusable functionality to multiple classes without creating deep inheritance hierarchies. Typically, such class are small, contain additional methods and should not be instantiated on its own.

The first PR related to this kind of refactoring is https://github.com/ansys/pyaedt/pull/5755 with the boundary creation.

PR #6716, opened 2025-10-01
FEAT: add dir as property

Alberto-DM:
## Description

This PR introduces a **`DirMixin`** and a **central `PyAedtBase` class** to simplify interactive exploration and provide a unified foundation for all project classes.

- `DirMixin` adds a `.dir` property for instances (and optionally class-level access) to quickly inspect available attributes and methods, filtering out private/internal names.  
- `PyAedtBase` inherits from `DirMixin` and serves as a **central base class** for all other classes, making it easy to add additional mixins or shared functionality in the future.

## Example 

Prior to this feature the method exploration required to type back and forth in the command line:
```python
dir(cube)                  
cube.faces                   
dir(cube.faces[0])         
cube.faces[0].edges[0]      
dir(cube.faces[0].edges[0]) 
```

Now we can explore the methods and properties of an object by simply doing this:
```python
cube.dir                  
cube.faces[0].dir                   
cube.faces[0].edges[0].dir         
```

## Benefits

- Faster interactive exploration and autocomplete in REPL/notebooks.
- Single point to add future mixins or shared utilities.
- Cleaner, more maintainable class hierarchy.


ecoussoux-ansys:
After doing some research, I see no objection to your proposal, as I could not find any resource discouraging this type of implementation.

I think using the builtin ``dir`` in code for inspecting attributes and methods is fine, as also discussed [here](https://discuss.python.org/t/use-of-dir-in-code-recommended/5196).
And introducing ``dir`` as a property through a mixin class is I believe a pythonic approach, knowing that such mixins were already considered for ``pyaedt`` in #5787 and implemented in #5755 for a specific functionality.

I do understand the usecase for that new property and the usability improvement. Same goes for the ``PyAedtBase`` class that could gather additional mixins or shared functionality in the future. 
However, can we already think of such additions in the near future that would further populate that ``PyAedtBase`` class? In case that would not yet be very clear, I wonder if the benefits brought by ``DirMixin`` alone justify the pretty overarching refactoring performed here, given the base class gets passed to all (?) others throughout the project?

All in all, the changes do look good to me (nice that tests are provided as well)!
But given how pervasive they are and the lack of a truly relevant assessment on my part regarding the novelty introduced, I would strongly suggest waiting for @SMoraisAnsys's comments on this!

SMoraisAnsys (commented):
Thanks for opening that PR @Alberto-DM 
While I do like the idea of having some mixin that could share behavior over all classes, I'm not really fan of the `dir` property. I might be wrong, but I feel like this property brings:
- risk of confusion for people who are familiar with `dir()`, this "breaks" implicit convention of the language;
- little additional value when the risk of confusion are non negligible.
Also, I think we should focus on improving the type hints provided through the code base. This would really help a lot with autocompletion in modern IDEs. When thinking about notebook, isn't the use of <TAB> already enough to get the attributes ? If it's because the list starts with private / protected attributes and methods, we could just rewrite `__dir__` by sorting the elements and placing the one with `_` at the end of the list.

Other though: renaming into `public_attrs` would avoid confusion.

Open to discussion :)

Alberto-DM:
@SMoraisAnsys I understand your concerns. Regarding the utility of having dir as a property: I personally use `dir()` a lot for exploring properties and methods, and I know I’m not the only one. Renaming the property to something like `public_attrs` (or another name) it is a good compromise.

I also really like the idea of improving autocompletion by reordering the `dir()` output. Overriding `__dir__` is a neat solution, and we can implement it easily now with the new `PyAedtBase` class.

@ecoussoux-ansys yes, implementing `PyAedtBase` was a lot of work, but now it is done 🙂 and I believe it adds real value to PyAEDT, and this kind of refactoring is best done before we reach version 1.0.

Alberto-DM:
Following the suggestions from @SMoraisAnsys I renamed the `dir` property in `public_dir`.
I also defined a `__dir__`method to the DirMixin to sort the `dir()` output.

Alberto-DM:
If the mixin used by the base class is imported from `mixins.py`, a circular import error occurs. This happens because the base class must remain independent and free of any other imports from within the same package. It should always be the rightmost class in the inheritance list of any PyAEDT class, so that the MRO places it just before the built-in `object` class. If `PyAedtBase` imports anything from pyaedt, it will cause a circular import error.

SMoraisAnsys:
> If the mixin used by the base class is imported from `mixins.py`, a circular import error occurs. This happens because the base class must remain independent and free of any other imports from within the same package. It should always be the rightmost class in the inheritance list of any PyAEDT class, so that the MRO places it just before the built-in `object` class. If `PyAedtBase` imports anything from pyaedt, it will cause a circular import error.

For visibility: following the comment above and a direct discussion, we agreed that the base class will be in it's own file and shouldn't have any dependency since it'll be used everywhere in the project. This will ensure that future changes do not conflict and cause circular import.

src/ansys/aedt/core/application/analysis_maxwell_circuit.py
```diff
@@ -23,10 +23,11 @@
 # SOFTWARE.
 
 from ansys.aedt.core.application.analysis import Analysis
+from ansys.aedt.core.base import PyAedtBase
 from ansys.aedt.core.generic.settings import settings
 
 
-class AnalysisMaxwellCircuit(Analysis):
+class AnalysisMaxwellCircuit(Analysis, PyAedtBase):
```
SMoraisAnsys, line 30: Can't we reduce the number of places where `PyAedtBase` is used ? Since it is added to application (which also inherist from Design and so on), adding it to the great great parent class should be enough no ?

Alberto-DM: In theory yes, but I think it is easier just to add it anywhere, it is designed to be used like that. If a single module has a common class I added the base classe only to the common class, but this is something I did just inside a single python file. I did not consider the inheritances across different modules because it gets complex quickly, and we cannot expect that the average contributor will dig into it. So, just add it anywhere and phyton will do the job for you with no overhead.

SMoraisAnsys: I do agree but we are here to review external contributions. @Samuelopez-ansys @maxcapodi78 what are your calls on that ? I'll accept the PR so it can be merged it if that's fine with you (I would suggest not to apply `PyAedtBase` everywhere and to leverage inheritance if possible).

Issue #6724, opened 2025-10-02
create_output_variable doesn't work in Maxwell


jvela018:
### Before submitting the issue

- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### Description of the bug

When calling the create_output_variable method, it calls the native API
`oModule.CreateOutputVariable(variable, expression, solution, self.design_solutions.report_type, context)`

report_type needs to be replaced by solution_type for it to work in Maxwell

### Steps To Reproduce

Run https://examples.aedt.docs.pyansys.com/version/dev/examples/low_frequency/general/resistance.html

### Which Operating System are you using?

Windows

### Which Python version are you using?

3.10

### Installed packages

accessible-pygments==0.0.5
aiohappyeyeballs==2.6.1
aiohttp==3.12.15
aiosignal==1.4.0
alabaster==1.0.0
altgraph==0.17.4
annotated-types==0.7.0
ansys-aedt-toolkits-common==0.14.0
ansys-aedt-toolkits-electronic_transformer @ file:///C:/Users/jvelasco/Documents/ansys-electronic-transformer-toolkit
ansys-aedt-toolkits-template @ file:///C:/Users/jvelasco/Documents/etk/pyaedt-toolkits-template
ansys-api-edb==0.2.1
ansys-edb-core==0.2.1
ansys-pythonnet==3.1.0rc6
ansys-sphinx-theme==1.6.0
ansys-tools-visualization-interface==0.11.0
anyio==4.10.0
argon2-cffi==25.1.0
argon2-cffi-bindings==25.1.0
arrow==1.3.0
asgiref==3.9.1
asttokens==3.0.0
async-lru==2.0.5
async-timeout==5.0.1
attrs==25.1.0
babel==2.17.0
backports.tarfile==1.2.0
beautifulsoup4==4.13.5
bleach==6.2.0
blinker==1.9.0
blosc2==3.8.0
build==1.3.0
certifi==2025.8.3
cffi==2.0.0
cfgv==3.4.0
charset-normalizer==3.4.3
click==8.2.1
clr_loader==0.2.7.post0
colorama==0.4.6
comm==0.2.3
commonmark==0.9.1
contourpy==1.3.2
coverage==7.10.7
cycler==0.12.1
debugpy==1.8.16
decorator==5.2.1
defusedxml==0.7.1
distlib==0.4.0
Django==5.2.6
docutils==0.21.2
et_xmlfile==2.0.0
exceptiongroup==1.3.0
[...]

PR #6726, opened 2025-10-02
FIX: Use solution type instead of report type in CreateOutputVariable

jvela018:
## Description
Use solution_type in create_output_variable method when calling the native API

## Issue linked
#6724

### DIFF
--- a/src/ansys/aedt/core/application/analysis.py
+++ b/src/ansys/aedt/core/application/analysis.py
@@ -1637,11 +1637,13 @@ class Analysis(Design, object):
             solution = self.existing_analysis_sweeps[0]
         if variable in self.output_variables:
             oModule.EditOutputVariable(
-                variable, expression, variable, solution, self.design_solutions.report_type, context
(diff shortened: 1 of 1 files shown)


src/ansys/aedt/core/application/analysis.py
```diff
@@ -1637,11 +1637,13 @@ def create_output_variable(self, variable, expression, solution=None, context=No
             solution = self.existing_analysis_sweeps[0]
         if variable in self.output_variables:
             oModule.EditOutputVariable(
-                variable, expression, variable, solution, self.design_solutions.report_type, context
+                variable, expression, variable, solution, self.design_solutions.solution_type, context
```
gmalinve: @jvela018 this is not the fix but rather we should make available the report_type for DC and AC solutions :)

gmalinve (dismissed):

jvela018:
@gmalinve , 

I've tested this branch in the resistance.py example in 2024R2, 2025R1 and 2025R2. Will you please approve? Thanks

gmalinve (approved):
LGTM

PR #6743, opened 2025-10-08
CI: Add ``ansys/actions/check-actions-security`` action and related fixes

### LINKED ISSUE #6713: [MAINTENANCE] Missing or outdated ``ansys/actions/check-actions-security`` action in ansys/pyaedt
Please enable the ``ansys/actions/check-actions-security`` action in the repository.

See [here](https://actions.docs.ansys.com/version/stable/vulnerability-actions/index.html#check-actions-security-action) an up-to-date documentation to a properly formatted action.

---

This issue was created by a bot. If you have any questions, please ping the PyAnsys Core team for assistance. To ping the team write a comment in this issue with the following mention: `@ansys/pyansys-core`.

If you want this repository to be excluded from this automated maintenance process, please let us know by [filling in the opt-out request form](https://github.com/ansys-internal/pyansys-maintenance-automation/issues/new?labels=&projects=&template=opt-out.yml&title=Request+to+opt-out+from+automated+maintenance+for+ansys%2Fpyaedt).

ecoussoux-ansys:
This PR introduces the ``ansys/actions/check-actions-security`` action in the workflow file ``.github/workflows/ci_cd.yml`` and consequently in the CI of ``pyaedt`` as requested in #6713.

This action is using [``zizmor``](https://docs.zizmor.sh/) to perform an audit of the workflows defined in the ``.github/workflows`` folder. 
More information on the approach for introducing the action is provided [here](https://actions.docs.ansys.com/version/stable/vulnerability-actions/index.html#check-actions-security-action), while instructions for fixing common workflow vulnerabilities and the rationale for addressing them are provided [here](https://dev.docs.pyansys.com/how-to/vulnerabilities.html#addressing-common-vulnerabilities-in-github-actions).

The PR addresses the findings surfaced by the ``zizmor`` audit on the workflow files (performed locally), resulting in the following changes:

* The argument ``persist-credentials: false`` is now systematically used with the action ``actions/checkout``,
* Template expansions ``(${{ ... }})`` are removed from plain run steps inside jobs. Inputs and relevant context variables are expanded in the ``env`` block instead,
* ``permissions`` are now defined on a job by job basis while none are granted at the workflow level. Jobs that do not use secrets are not granted any specific permission,
* An anonymous definition for a job in the ``label.yml`` file is fixed,
* A potentially spoofable bot condition in ``ci_cd.yml`` is fixed (more details on this fix [here](https://docs.zizmor.sh/audits/#bot-conditions)).
    
Note that ``zizmor`` findings of the type _"note[self-hosted-runner]: runs on a self-hosted runner"_ (see related documentation [here](https://docs.zizmor.sh/audits/#self-hosted-runner)) are not addressed by the present changes. 
These findings cannot actually be fixed (since self-hosted runners are required for ``pyaedt``), but they can be silenced: Based on the [implementation](https://github.com/zizmorcore/zizmor/pull/34/files#diff-07e16fbcf99e4ca407d640915f8b74d890e8aec6c34e0ee9b737415475fa8168) of this particular check, it appears that not putting the ``self-hosted`` parameter of the job's ``runs-on`` argument first will result in ``zizmor`` not raising a finding. However, [GitHub's documentation](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#choosing-self-hosted-runners) states that ``self-hosted`` must come first in the array provided for ``runs-on``, hence the decision to not apply any change to get rid of the related findings.
A different approach was adopted in https://github.com/ansys/pyedb/pull/1575 due to the presence - prior to that PR - of array declarations for ``runs-on`` with ``self-hosted`` not being declared as the first item (which does not seem to be causing issues in the workflow).

For further reference, here are a few PRs applying similar changes from other pyansys projects: ansys/pre-commit-hooks#352, ansys/ansys-tools-visualization-interface#366, ansys/pyansys-geometry#2277.
Lastly, the latest release v10.1.4 of ``ansys/actions`` is used here.

Close #6713.

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -23,6 +23,8 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+permissions: {} # Disable default permissions
+
 jobs:
 
   update-changelog:
@@ -43,6 +45,8 @@ jobs:
   vulnerabilities:
     name: "Vulnerabilities"
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     steps:
       - uses: ansys/actions/check-vulnerabilities@c2fa7c93f6883114e0e643599431b33d29f0b13f # v10.1.4
         with:
@@ -87,6 +91,8 @@ jobs:
     name: Check the title of the PR (if needed)
     runs-on: ubuntu-latest
     needs: [block-pyansys-ci-bot]
+    permissions:
+      contents: read
     steps:
       - name: Check the title of the pull request
         if: github.event_name == 'pull_request'
@@ -103,6 +109,8 @@ jobs:
     name: Documentation style check
     runs-on: ubuntu-latest
     needs: [pr-title]
+    permissions:
+      contents: read
     steps:
       - name: Check documentation style
         uses: ansys/actions/doc-style@c2fa7c93f6883114e0e643599431b33d29f0b13f # v10.1.4
@@ -1053,6 +1061,8 @@ jobs:
     if: github.event_name == 'push' && contains(github.ref, 'refs/tags')
     runs-on: ubuntu-latest
     needs: [release]
+    permissions:
+      contents: write
     steps:
       - name: Deploy the stable documentation
         uses: ansys/actions/doc-deploy-stable@c2fa7c93f6883114e0e643599431b33d29f0b13f # v10.1.4
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -157,8 +157,10 @@ jobs:
           use-python-cache: false
       - name: Import python package
         shell: bash
+        env:
+          ACTIVATE_VENV: ${{ steps.build-wheelhouse.outputs.activate-venv }}
         run: |
-          ${{ steps.build-wheelhouse.outputs.activate-venv }}
+          ${ACTIVATE_VENV}
           python -c "import ansys.aedt.core; from ansys.aedt.core import __version__"
 
   unit-tests:
@@ -269,9 +271,10 @@ jobs:
       - name: Run tests marked with 'solvers'
         env:
           PYTHONMALLOC: malloc
+          PYTEST_ARGUMENTS: ${{ env.PYTEST_ARGUMENTS }}
         run: |
           .venv\Scripts\Activate.ps1
-          pytest ${{ env.PYTEST_ARGUMENTS }} --timeout=600 -m solvers
+          pytest ${PYTEST_ARGUMENTS} --timeout=600 -m solvers
 
       - uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7 # v5.5.1
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/ci_cd.yml
```diff
@@ -87,6 +103,8 @@ jobs:
     name: Check the title of the PR (if needed)
     runs-on: ubuntu-latest
     needs: [block-pyansys-ci-bot]
+    permissions:
```
SMoraisAnsys, line 106:
```suggestion
    permissions:
      pull-requests: read
```
Just wondering if that isn't required here ?

.github/workflows/ci_cd.yml
```diff
@@ -23,6 +23,8 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+permissions: {} # Disable default permissions
```
SMoraisAnsys, line 26: Could we available 
```
permissions:
  contents: read
```

This seems to be valuable for many jobs and quite redundant no ?

.github/workflows/label.yml
```diff
@@ -15,13 +15,17 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+permissions: {} # Disable default permissions
```
SMoraisAnsys, line 18: Same comment

.github/workflows/ci_cd.yml
ecoussoux-ansys: According to the ``README`` of the [repository](https://github.com/amannn/action-semantic-pull-request) of the action ``amannn/action-semantic-pull-request`` used under the hood by our ``ansys/actions/check-pr-title``, you are correct.
Actually, the only required permission is ``pull-requests: read`` and the one I had provided is not needed. I updated in af08ee3.

.github/workflows/ci_cd.yml
ecoussoux-ansys: This permission is granted to 5 out of 24 jobs (one less following your first comment!) in ``ci_cd.yml``.
``zizmor`` will not complain about making this available to all jobs in the file, so do you think I should proceed with adding it? 
It does not "feel" that redundant to me though and maybe that's a good habit (although I recognize probably annoying and maybe error-prone too) to ask what job needs what precisely..? But on the other hand I see this global permission being used in pyansys-geometry for instance. 
So, sorry, I'm rather undecided about what to do here..!

.github/workflows/label.yml
ecoussoux-ansys: In ``label.yml`` it is granted to 2/4 jobs.. so it would be the same comment from my side.. 
I'll be fine with applying your suggestion or leaving the permissions as they are now.

.github/workflows/ci_cd.yml
SMoraisAnsys: I assumed content read was required for every job where we clone the repo so I was in favor to adding it be default. I'm surprised that only 5 ou of 24 jobs are requiring that. I assume that pyaedt being public, it can be download with HTTPS and doesn't require to add many things in the default permissions.
Keep things as before 👍

ecoussoux-ansys: To be honest I mainly based my changes regarding the permissions on the examples I could fine in other pyansys projects which had already introduced the ``ansys/actions/check-actions-security`` and fixed the related ``zizmor`` checks. 
Also the documentation says, as rule of thumb, that when a job uses no secret, no permissions should be needed either.. 
All in all I must say that  proceeded more by imitation than by conducting in-depth investigations on the topic..!

PR #6748, opened 2025-10-10
FIX: f-string usage for IronPython compatibility

eblanco-ansys:
Replace f-strings with standard string formatting to ensure compatibility with IronPython.

### DIFF
--- a/doc/source/Resources/pyaedt_installer_from_aedt.py
+++ b/doc/source/Resources/pyaedt_installer_from_aedt.py
@@ -360,7 +360,6 @@ def install_pyaedt():
             subprocess.run(command, check=True, env=env)  # nosec
         else:
             # Try to install uv and use it, fallback to pip if it fails
-            uv_available = True
             try:
                 print("Installing uv in the virtual environment...")
                 subprocess.run([str(pip_exe), "--default-timeout=1000", "install", "uv"], check=True, env=env)  # nosec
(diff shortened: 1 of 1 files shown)


doc/source/Resources/pyaedt_installer_from_aedt.py
```diff
@@ -375,7 +375,7 @@ def install_pyaedt():
                 else:
                     subprocess.run([str(uv_exe), "pip", "install", "pyaedt[all]"], check=True, env=env)  # nosec
             except subprocess.CalledProcessError as e:
-                print(f"uv installation failed with error: {e}")
+                print("uv installation failed")
```
SMoraisAnsys:
```suggestion
                print("uv installation failed with error: {}".format(e))
```
Why not leaving the exception while using another way to retrieve the associated string value ?

doc/source/Resources/pyaedt_installer_from_aedt.py
```diff
@@ -442,7 +442,7 @@ def install_pyaedt():
                 else:
                     subprocess.run([str(uv_exe), "pip", "install", "pyaedt[all]"], check=True, env=env)  # nosec
             except subprocess.CalledProcessError as e:
-                print(f"uv installation failed with error: {e}")
+                print("uv installation failed with")
```
SMoraisAnsys: Same comment

PR #6775, opened 2025-10-16
FIX: Bug located in primitives circuit module

ramin4667:
## Description
**This PR fixes the error due to non-digit circuit element IDs. **

## Issue linked
**The post-processing call for the simulated circuit of this example https://examples.aedt.docs.pyansys.com/version/dev/examples/high_frequency/radiofrequency_mmwave/lumped_element.html raises an error. **

## Checklist

### DIFF
--- a/doc/changelog.d/6822.fixed.md
+++ b/doc/changelog.d/6822.fixed.md
@@ -1,0 +1,1 @@
+Add error message if extension is started with an empty HFSS 3D Layout design
--- a/src/ansys/aedt/core/extensions/hfss3dlayout/cutout.py
+++ b/src/ansys/aedt/core/extensions/hfss3dlayout/cutout.py
@@ -209,6 +209,9 @@ class CutoutExtension(ExtensionHFSS3DLayoutCommon):
     def __load_objects_net(self):
         """Load objects by net from the EDB modeler."""
         res = defaultdict(list)
+        if not self.aedt_application.modeler.edb:
+            self.release_desktop()
+            raise AEDTRuntimeError("Extension cannot be used with an empty HFSS 3D Layout design.")
         for net, net_objs in self.aedt_application.modeler.edb.modeler.primitives_by_net.items():
             res[net].extend(obj.aedt_name for obj in net_objs)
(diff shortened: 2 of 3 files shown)


gmalinve:
Hi @ramin4667 ,

Do you have an issue linked to this PR? if yes please link it.

Thanks,

KR,

Giulia

ramin4667:
> Hi @ramin4667 ,
> 
> Do you have an issue linked to this PR? if yes please link it.
> 
> Thanks,
> 
> KR,
> 
> Giulia

Hi @gmalinve, The PR is updated with a link to an issue.

gmalinve:
Hi @ramin4667, yes. I linked it before.

src/ansys/aedt/core/modeler/circuits/primitives_circuit.py
```diff
@@ -1387,7 +1387,7 @@ def refresh_all_ids(self):
                         o.schematic_id = int(name[1].split(":")[0])
                         objID = int(o.schematic_id)
                     else:
-                        o.id = int(name[1])
+                        o.id = int(name[1]) if name[1].isdigit() else name[1]
```
myoung301: If the id field is permitted to be a string, should we switch to consistently store it as a string? In other words, this change would be:

```suggestion
                        o.id = name[1]
```

I'm not sure if there are other places that rely on this ID being a string, but we probably need to update those. They will also sometimes fail if we conditionally make this a string (the change you are proposing). So, it's probably best to apply any additional changes needed to accept a string ID (if there are any).

myoung301 (changes requested):

src/ansys/aedt/core/modeler/circuits/primitives_circuit.py
ramin4667: String IDs of circuit elements are saved as strings in 5 various locations.

ramin4667:
> @ramin4667 There is at least one change needed for the tests. I saw a failure caused by the test looking for an int where the type is now a string. Please review the failing tests.

@myoung301 Tests are updated and passing.

Issue #6788, opened 2025-10-20
CI: Refactor code related to `test_apply_config_to_edb` or strengthen CI


SMoraisAnsys:
When running the extensions tests on Linux, it happens often that `test_apply_config_to_edb` fails without changes associated to this part of the code.

See https://github.com/ansys/pyaedt/actions/runs/18646749738/job/53156210783

<img width="1328" height="396" alt="Image" src="https://github.com/user-attachments/assets/1c1028c0-e9f2-48db-9ed1-4bb78733ccd4" />

SMoraisAnsys:
Pinging @hui-zhou-a  and @svandenb-dev for visibility

SMoraisAnsys:
Extending this flaky marker to tests/system/extensions/test_export_layout.py::test_export_layout_all_option, see

https://github.com/ansys/pyaedt/actions/runs/18650651620/job/53168385951?pr=6789

SMoraisAnsys:
Extending this flaky marker to  tests/system/extensions/test_export_layout.py::test_export_layout_ipc_only, see

https://github.com/ansys/pyaedt/actions/runs/18681857757/job/53265428535

SMoraisAnsys:
Extending this flaky marker to tests/system/extensions/test_export_layout.py::test_export_layout_config_only, see

https://github.com/ansys/pyaedt/actions/runs/18688682881/job/53289641692

SMoraisAnsys:
Extending this flaky marker to tests/system/extensions/test_post_layout_design.py::test_layout_design_toolkit_microvia, see

https://github.com/ansys/pyaedt/actions/runs/18707742802/job/53349586488

Issue #5524, opened 2024-12-03
ROADMAP: Improve CICD


Samuelopez-ansys:

Samuelopez-ansys:
@SMoraisAnsys Please, could you add the list of vulnerabilities that PyAEDT should tackle?

PR #6789, opened 2025-10-20
CI: Use flaky marker to avoid rerunning all tests

SMoraisAnsys:
## Description
Use flaky marker to avoid rerunning a lot of tests. In the past we had a lot of such tests and, while it seems stable this days, we still have at least one of them https://github.com/ansys/pyaedt/issues/6788. Since we can expect this kind of behavior to happen from time to time due to AEDT release changes, this PR proposes to mark some tests as flaky and rerunning them if needed. This would avoid rerunning a set of N tests every time a flaky test fails during CI.

Note that two markers are introduced since the test failing is only failing for Linux. However, the CI can be easily extended with a job for Windows (I think we have a test that is breaking the CI from time to time, i.e. it reaches the 2 hours time limit on windows even if no test breaks).

## Issue linked
Related to #6788 and #5524 

## Checklist

### DIFF
--- a/tests/system/extensions/test_export_layout.py
+++ b/tests/system/extensions/test_export_layout.py
@@ -25,6 +25,8 @@
 import json
 from pathlib import Path
 
+import pytest
+
 from ansys.aedt.core.edb import Edb
 from ansys.aedt.core.extensions.hfss3dlayout.export_layout import ExportLayoutExtensionData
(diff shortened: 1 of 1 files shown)


eblanco-ansys:
I think this proposal is a great idea, LGTM

MaxJPRey (approved):
LGTM.

PR #6791, opened 2025-10-21
FEAT: Filter deprecated methods from public_dir

### LINKED ISSUE #6769: Improve PyAedtBase.public_dir __repr__ method

It would be very helpful if this method could somehow flag deprecated names.

### Code sample expected

Not sure how to do this in practice, but will look into it.

_No response_

SMoraisAnsys:
## Description
Remove deprecated methods from `public_dir` by parsing docstrings (if any).

## Issue linked
Close #6769

## Checklist


src/ansys/aedt/core/base.py
```diff
@@ -51,7 +54,24 @@ def __dir__(self):
...
+
+            # NOTE: Could be necessary to handle properties that raise exceptions, e.g. NotImplementedError
+            try:
+                obj = getattr(self, name)
+            except (AttributeError, RuntimeError):
+                continue
+
+            doc = inspect.getdoc(obj)
+            if doc and ".. deprecated::" in doc:
+                continue
```
Devin-Crawford, line 70: I think it would be better to display deprecated names and append text indicating the name is deprecated. Something like ``"\*deprecated\*"`` or ``"\<deprecated\>"``

SMoraisAnsys: Currently when you call `public_dir` you get a list of str and each str is the name of a public method/property/attribute. If I take the test as an example, I get `vertex.public_dir == ['position']`, would you rather have `['position', 'position_deprecated <deprecated>']` ?

SMoraisAnsys: Since the methods are deprecated, I would prefer to just remove them from the pool.

Devin-Crawford: I was hoping for something like you suggested which is useful when making an update to deprecate a method.

SMoraisAnsys: I'm not sure to understand what is the value of keeping the deprecated method around since the main goal of `public_dir` is to list the available method/property/attribute. Can you explain it to me so that I can potentially suggest a more suitable solution than the current one?

PR #6799, opened 2025-10-22
FIX: Fixed panels in linux

eblanco-ansys:
## Description
Linux panels need a relative path for the icons in order to work. I added a symlink to the environment icons folder so we don't create a copy of them in PersonalLib folder

## Checklist


src/ansys/aedt/core/extensions/customize_automation_tab.py
```diff
@@ -142,12 +142,32 @@ def add_automation_tab(
...
+
+        # For Linux, create symbolic link and use relative path (if not, AEDT panels break)
+        if is_linux:  # pragma: no cover
+            images_source = Path(ansys.aedt.core.extensions.__file__).parent / "installer" / "images" / "large"
+            images_target = lib_dir / product / "images"
+            if not images_target.exists() and images_source.exists():
+                try:
+                    images_target.symlink_to(images_source)
+                except Exception:
+                    logging.getLogger("Global").warning(
```
SMoraisAnsys, line 154: Just wondering if we couldn't use the Desktop logger here. It would require to pass the `desktop_object` from `add_custom_toolkit` to `add_script_to_menu` and `add_automation_tab` but I think it would help to let the user know that something might have went wrong (and the reason). That would also be valuable for us IMO. What do you think ?

eblanco-ansys: The problem of this method is that is also the one used [after installation](https://aedt.docs.pyansys.com/version/stable/Getting_started/Installation.html#after-installation), so it's not always running from an extension with a desktop instance. We could add the logger as an optional parameter and only use it if received. Let me know what you think about this

eblanco-ansys: Furthermore, the idea is that this method will only be used after installation and will be removed from the version manager, as it will not be necessary to reset the panels as before

eblanco-ansys:
> LGTM, I left a minor comment ?

I created issue #6813 to work on it for next release. I think we can merge this one

PR #6802, opened 2025-10-22
FIX: settings.aedt_version in desktop and design class fix

gmalinve:
## Description
If settings.aedt_version is set before the application init, this is not taken into account.

## Issue linked

## Checklist

### DIFF
--- a/src/ansys/aedt/core/desktop.py
+++ b/src/ansys/aedt/core/desktop.py
@@ -390,7 +390,8 @@ class Desktop(PyAedtBase):
         specified_version = (
             kwargs.get("specified_version")
             or kwargs.get("version")
-            or (settings.aedt_version if (not args or args[0] is None) else args[0])
+            or settings.aedt_version
+            or (None if (not args or args[0] is None) else args[0])
         )
         if "new_desktop_session" in kwargs or "new_desktop" in kwargs:
             new_desktop = kwargs.get("new_desktop_session") or kwargs.get("new_desktop")


src/ansys/aedt/core/desktop.py
```diff
@@ -388,7 +388,9 @@ class Desktop(PyAedtBase):
     def __new__(cls, *args, **kwargs):
         # The following commented lines will be useful when we will need to search among multiple saved desktop.
         specified_version = (
-            kwargs.get("specified_version") or kwargs.get("version") or None if (not args or len(args) < 1) else args[0]
+            kwargs.get("specified_version")
+            or kwargs.get("version")
+            or (settings.aedt_version if (not args or args[0] is None) else args[0])
```
Alberto-DM:
```suggestion
            or settings.aedt_version
            or (None if (not args or args[0] is None) else args[0])
```
if the default value of `settings.aedt_version` is `None` (when it is not set), then my suggestion is functionally equivalent to your code. Nonetheless, I strongly suggest to change the line as suggested to do not rely on a setting default value which might change in the future.

src/ansys/aedt/core/visualization/post/compliance.py
Alberto-DM, line 1: Can you explain the relation between this file modification and the bug? Is it perhaps a merging error?

gmalinve: @Alberto-DM Massimo needed on the fly ;)

src/ansys/aedt/core/application/design.py
```diff
@@ -202,7 +202,7 @@ def __init__(
         self.close_on_exit: bool = close_on_exit
         self._desktop_class = None
         self._desktop_class = self.__init_desktop_from_design(
-            version,
+            settings.aedt_version if settings.aedt_version else version,
```
maxcapodi78, line 205: I suggest 
`version if version else settings.aedt_version`

hui-zhou-a:
@gmalinve thanks for fixing this

PR #6804, opened 2025-10-23
FIX: Allow Object3d to be used in create_current_source_from_objects

SMoraisAnsys:
## Description
Previously the method `create_current_source_from_objects` was failing when one passed a `Object3d` as source or target because the checking of the existence of this object was only performed for `int` and `str`. However, the method should be working with input of type `Object3d` as described in the docstring (and because it makes sense tbh).

## Issue linked
None

## Checklist

### DIFF
--- a/src/ansys/aedt/core/modeler/cad/primitives_3d.py
+++ b/src/ansys/aedt/core/modeler/cad/primitives_3d.py
@@ -35,6 +35,7 @@ from math import sqrt
 from math import tan
 import os
 from pathlib import Path
+from typing import TYPE_CHECKING
 
 from ansys.aedt.core import Edb
 from ansys.aedt.core.base import PyAedtBase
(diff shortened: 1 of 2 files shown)


src/ansys/aedt/core/modeler/cad/primitives_3d.py
```diff
@@ -111,7 +115,7 @@ def __init__(self, application):
         self.multiparts = []
 
     @pyaedt_function_handler(position="origin", dimensions_list="sizes", matname="material")
-    def create_box(self, origin, sizes, name=None, material=None, **kwargs):
+    def create_box(self, origin, sizes, name=None, material=None, **kwargs) -> "Object3d":
```
gmalinve, line 118: @SMoraisAnsys dummy question but why the return type as a string? and why thype hints just for the create_box method?

SMoraisAnsys: Here I'm trying to both improve documentation & IDE auto completion, while avoiding future circular import by adding the import in `if TYPE_CHECKING:`. Since it's not directly imported, it can't be used as a type hint. However, having it in `"..."` makes it explorable through the IDE.

Also, the reason for the changes only on `create_box` is because I was not having good response with my IDE while writing my test. If possible, type hints should be in every function signature (arguments & return).

Issue #6818, opened 2025-10-26
pyaedt material creation data missing


zmladen:
Hello,

I am currently testing PyAnsys/Pyaedt in Maxwell (Release 2024R1). I was wondering what the method set_coreloss_at_frequency actually does. The example code runs successfully, but I can’t find any data at the specified frequencies in the Maxwell GUI.

`
from ansys.aedt.core import Maxwell3d
m3d = Maxwell3d()
box = m3d.modeler.create_box([-10, -10, 0], [20, 20, 20], "box_to_split")
box.material = "magnesium"
m3d.materials["magnesium"].set_coreloss_at_frequency(
                                            points_at_frequency={60 : [[0,0], [1,3.5], [2,7.4]],
                                                                  100 : [[0,0], [1,8], [2,9]],
                                                                  150 : [[0,0], [1,10], [2,19]]}
                                            )
m3d.desktop_class.close_desktop()
`
Frequencies are shown in the "Core Loss vs Frequency" dialog, but the data is missing!

<img width="753" height="495" alt="image" src="https://github.com/user-attachments/assets/f24d02da-5ef1-46af-8d0c-93d97e5fdc6c" />

_Originally posted by @zmladen in https://github.com/ansys/pyaedt/discussions/6781_

PR #6851, opened 2025-11-04
FIX: Move native API Points definition to data_handlers.py

jvela018:
## Description
The method set_coreloss_at_frequency can be used for a single frequency or multiple frequencies. The method was able to set up the frequencies, but was unable to populate the data on the tables. The API argument setup for the Points argument was done in two places - the _dict2arg method and the get_frequency_coefficients method. Unfortunately, the one being handled in _dic2arg, didn't modify the global props under self._props, and it was handled via the get_frequency method. The coefficients were calculated correctly, but the datasets weren't updated.

The issue was fixed by removing the handling of the API Points definition within the get_frequency_coefficient, and exposed it directly through the data_handlers.py.

## Issue linked
#6818 

## Checklist


tests/system/general/test_materials.py
```diff
@@ -64,6 +65,13 @@ def aedt_app(add_app):
     app.close_project(app.project_name, save=False)
 
 
+@pytest.fixture
```
gmalinve: @jvela018 hey, there's no need of this fixture as there is already another one called maxwell_app. You can reuse that one.

jvela018:
> One small thing to improve in the test, otherwise for me it's ok. Thanks @jvela018 :)

@gmalinve Done

PR #6897, opened 2025-11-18
CI: Rework workflow to avoid testing

SMoraisAnsys:
## Description
This PR is aiming at handling the first bullet point of https://github.com/ansys/pyaedt/issues/6896
Here we mitigate resource contention by avoiding to run the test for extensions / emit / solver solutions, unless there is a modification in the folder associated to each topic.
The tests are still run nightly and we'll have to consult them to know if something is going wrong with them.
Tests are still running when the PR is merge to ensure that we have code coverage fully evaluated on the repo. We could investigate if running the test nightly can be leveraged to still get a code coverage for the repo without running the test on main but that should be done in a second PR.

Pinging @jsalant22 @myoung301 @ramin4667 for visibility.

## Issue linked
Associated to #6896 

## Checklist
- [ ] I have tested my changes locally.
- [ ] I have added necessary documentation or updated existing documentation.
- [ ] I have followed the coding style guidelines of this project.
- [ ] I have added appropriate tests (unit, integration, system).
- [ ] I have reviewed my changes before submitting this pull request.
- [ ] I have linked the issue or issues that are solved by the PR if any.
- [ ] I have agreed with the Contributor License Agreement ([CLA](https://developer.ansys.com/form/cla-acceptance)).

PR #6880, opened 2025-11-14
FEAT: Local test configuration cli

eblanco-ansys:
Added cli docs + local test configuration utility + minor improvements

## Checklist
- [ ] I have tested my changes locally.
- [ ] I have added necessary documentation or updated existing documentation.
- [ ] I have followed the coding style guidelines of this project.
- [ ] I have added appropriate tests (unit, integration, system).
- [ ] I have reviewed my changes before submitting this pull request.
- [ ] I have linked the issue or issues that are solved by the PR if any.
- [ ] I have agreed with the Contributor License Agreement ([CLA](https://developer.ansys.com/form/cla-acceptance)).

PR #6926, opened 2025-11-22
CI: Fix uv setup in nightly tests

SMoraisAnsys:
## Description

Adding missing changes from #6897 (related to the setup of `uv` on our Windows self hosted runners)

## Issue linked
None

## Checklist

### DIFF
--- a/doc/changelog.d/6926.maintenance.md
+++ b/doc/changelog.d/6926.maintenance.md
@@ -1,0 +1,1 @@
+Fix uv setup in nightly tests


.github/workflows/nightly-tests.yml
eblanco-ansys, line 282: Why not using `pyaedt stop --all` cli command to remove all processes? With the merge of #6880 `typer` should also be added to test optional dependencies, so we could simply CI readability replacing this block with the cli command

SMoraisAnsys: The thing is that the pyaedt command only handles ansysaedt.exe processes (or ansysaedt on linux). In the past, we also have add cases where specific process of AEDT were still running when something goes wrong. If we want to have the same kind of behavior then the `stop` command needs to be extended to also cover other processes and we would need an exhaustive list of those.

PR #6970, opened 2025-12-01
REFACTOR: run pyaedt script py

hui-zhou-a:
## Description

## Issue linked

## Checklist

### DIFF
--- a/src/ansys/aedt/core/extensions/templates/run_pyaedt_script.py_build
+++ b/src/ansys/aedt/core/extensions/templates/run_pyaedt_script.py_build
@@ -52,7 +52,7 @@ def show_launcher_dialog(initial_dir):
     f = Form()
     f.Text = "Run PyAEDT Script"
     f.Width = 500
-    f.Height = 220
+    f.Height = 180
     f.FormBorderStyle = FormBorderStyle.FixedDialog
     f.MaximizeBox = False
     f.MinimizeBox = False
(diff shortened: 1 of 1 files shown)


eblanco-ansys:
@hui-zhou-a I think it's a good idea to include a textbox like this to allow args to be added easily. However, I don't think `--configure_file` should be there, as it only applies to some scripts and can be confusing for the user. @Samuelopez-ansys what do you think about this?

src/ansys/aedt/core/extensions/templates/run_pyaedt_script.py_build
```diff
@@ -41,60 +41,198 @@ sys.path.append(r"##EXTENSION_TEMPLATES##")
...
+    txt_args = TextBox()
+    txt_args.Left = 110
+    txt_args.Top = top
+    txt_args.Width = 320
+    f.Controls.Add(txt_args)
+
+    top += 30
+    # Configure file
+    lbl_cfg = Label()
+    lbl_cfg.Text = "--configure_file"
```
eblanco-ansys: This needs to be discussed

Samuelopez-ansys: I agree, this is what we discussed. We can only accept "Script Arguments:"

hui-zhou-a: OK. I will create an extension to add configure_file support

Samuelopez-ansys: Sorry @hui-zhou-a but why the configuration file is not just a script argument?

Even an extension is not a solution, we want to avoid adding an extension only for this, we will not allow to merge it either.

please explain why the script argument is not enough

PR #6975, opened 2025-12-02
CI: update Python version to 3.12

eblanco-ansys:
## Description
Updated python version to 3.12 across tests except for EMIT and filters tests

## Checklist


eblanco-ansys:
> I think it's a good idea but as far as I remember the team wanted to keep using the same version as the one delivered with AEDT. Note that it makes sense as it's the python version we leverage to create the extensions venv. If this gets approved none the less, could you perform the same changes in the nightly workflows ?

Sure, I can also implement this in nightly ones. The reason why I thought it was a good idea is that Python 3.12 is significantly faster and should help to lighten the load

PR #6978, opened 2025-12-03
CI: Use python 3.13 in ci

eblanco-ansys:
## Description
As title says

## Checklist

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -1019,9 +1019,7 @@ jobs:
 
       - name: Create virtual environment
         if: steps.changes.outputs.run_tests == 'true'
-        env:
-          UV_PYTHON_VERSION: ${{ env.TESTS_VERSION }}
-        run: uv venv .venv --python $UV_PYTHON_VERSION
+        run: uv venv .venv
 
       - name: Upgrade wheel and setuptools
         if: steps.changes.outputs.run_tests == 'true'
@@ -1136,9 +1134,7 @@ jobs:
 
       - name: Create virtual environment
         if: steps.changes.outputs.run_tests == 'true'
-        env:
-          UV_PYTHON_VERSION: ${{ env.TESTS_VERSION }}
-        run: uv venv .venv --python $UV_PYTHON_VERSION
+        run: uv venv .venv
 
       - name: Upgrade wheel and setuptools
         if: steps.changes.outputs.run_tests == 'true'


Alberto-DM (commented):
Has this change been discussed? Many customers are using Python 3.10 included in AEDT.

eblanco-ansys:
> Has this change been discussed? Many customers are using Python 3.10 included in AEDT.

I won't approve this. I wanted to test if we get a real speed boost in the CI (Python 3.11+ is about 15~20% faster). To test CI speed I couldn't tag it as draft

Alberto-DM (dismissed):
Dummy request just to block accidental merge.

ecoussoux-ansys:
@eblanco-ansys 

In 38a9dd0 I am pushing a tiny change to the ``conf.py`` file which locally on my machine solved the issue we are observing currently on the ``doc-build`` job when trying to use ``numpydoc==1.10.0`` together with python 3.13.

To give a few explanations: I was able to identify the source of the error to be caused by the ``src.ansys.aedt.core.modules.profile.Profiles`` class. In particular, the issue arises because this class inherits from the ``Mapping`` class from the ``collections.abc`` module: ``numpydoc`` seems to be unable to locate the source code of the methods from that class (which it wants to document as well because we ask it to also list the inherited methods / attributes in our API doc) and is consequently crashing.

To solve this, I tried two approaches but only one of them worked:
* As you can see from 38a9dd0, I am adding the files generated by autosummary for the ``Profiles`` class to the exclude list for numpydoc validation. This allowed the doc-build to go through on my machine and I could see no difference between what I obtained for the ``Profiles`` class in the resulting locally-built doc and what is currently available in the online documentation of pyaedt for that class (https://aedt.docs.pyansys.com/version/stable/API/_autosummary/ansys.aedt.core.modules.profile.Profiles.html),
* The other approach that I thought would be more clever, would have been to update that line in the ``intersphinx_mapping`` (still in the ``conf.py``): ``"python": ("https://docs.python.org/3.11", None),``. I thought that by updating it to ``3.13``, sphinx would have been able to grab the source code for the ``Mapping`` class and I wouldn't have had to extend the exclude list for numpydoc validation.. but unfortunately it does not seem to work for me locally. (Therefore I don't really know why it is working fine with ``numpydoc==1.10.0`` and python 3.10.. it could be that the ``Mapping`` class is not accessed the same way in python 3.13 compared to previous versions, but I am not sure..).

Hopefully this enables usage of python 3.13 in conjunction with ``numpydoc==1.10.0`` to build our doc!

eblanco-ansys:
> @eblanco-ansys
> 
> In [38a9dd0](https://github.com/ansys/pyaedt/commit/38a9dd0dbc9b96074f0886bbe565442b1bc2f64e) I am pushing a tiny change to the `conf.py` file which locally on my machine solved the issue we are observing currently on the `doc-build` job when trying to use `numpydoc==1.10.0` together with python 3.13.
> 
> To give a few explanations: I was able to identify the source of the error to be caused by the `src.ansys.aedt.core.modules.profile.Profiles` class. In particular, the issue arises because this class inherits from the `Mapping` class from the `collections.abc` module: `numpydoc` seems to be unable to locate the source code of the methods from that class (which it wants to document as well because we ask it to also list the inherited methods / attributes in our API doc) and is consequently crashing.
> 
> To solve this, I tried two approaches but only one of them worked:
> 
> * As you can see from [38a9dd0](https://github.com/ansys/pyaedt/commit/38a9dd0dbc9b96074f0886bbe565442b1bc2f64e), I am adding the files generated by autosummary for the `Profiles` class to the exclude list for numpydoc validation. This allowed the doc-build to go through on my machine and I could see no difference between what I obtained for the `Profiles` class in the resulting locally-built doc and what is currently available in the online documentation of pyaedt for that class (https://aedt.docs.pyansys.com/version/stable/API/_autosummary/ansys.aedt.core.modules.profile.Profiles.html),
> * The other approach that I thought would be more clever, would have been to update that line in the `intersphinx_mapping` (still in the `conf.py`): `"python": ("https://docs.python.org/3.11", None),`. I thought that by updating it to `3.13`, sphinx would have been able to grab the source code for the `Mapping` class and I wouldn't have had to extend the exclude list for numpydoc validation.. but unfortunately it does not seem to work for me locally. (Therefore I don't really know why it is working fine with `numpydoc==1.10.0` and python 3.10.. it could be that the `Mapping` class is not accessed the same way in python 3.13 compared to previous versions, but I am not sure..).
> 
> Hopefully this enables usage of python 3.13 in conjunction with `numpydoc==1.10.0` to build our doc!

Thank you very much for your contribution and detailed description of what you have changed. In my opinion, it is the ideal solution to the problem. Sphinx is certainly very tricky...

zizmor.yml
```diff
@@ -0,0 +1,5 @@
+rules:
+  bot-conditions:
+    ignore:
+      # ignore line 84 in ci_cd.yml
+      - ci_cd.yml:84
```
ecoussoux-ansys: This file should be deleted.
It was removed in #6991 and is no longer necessary, because the ``bot-conditions`` raised by ``zizmor`` for the ``block-dependabot`` job is now ignored with an in-line comment next to the corresponding line (which actually is not line 84 anymore).

doc/changelog.d/6975.maintenance.md
```diff
@@ -0,0 +1 @@
+Update Python version to 3.12
```
ecoussoux-ansys: Why do we have 2 changelog fragments in the PR? 
This one should probably be removed (does not have the proper PR number), right?

.github/workflows/ci_cd.yml
```diff
@@ -1015,7 +1019,9 @@ jobs:
 
       - name: Create virtual environment
         if: steps.changes.outputs.run_tests == 'true'
-        run: uv venv .venv
+        env:
+          UV_PYTHON_VERSION: ${{ env.TESTS_VERSION }}
```
ecoussoux-ansys: I don't get why you made this change in only 2 of the 11 jobs that implement exactly that same "Create virtual environment" step.
Is there a reason why it had to be updated inside ``system-tests-filter-windows`` and ``system-tests-emit-windows`` but not in the others?

doc/changelog.d/6975.maintenance.md
eblanco-ansys: I previously named the PR with that title, so I'll remove it

.github/workflows/ci_cd.yml
eblanco-ansys: I previously implemented the CI to run all tests on python 3.13 to test the performance, but filter and emit tests only run on python 3.10. That's the reason why I performed the change. After checking that there isn't a substantial improvement I went back to 3.10 for all the tests. I forgot to remove this argument that was only used for filters and emit.

Thanks for noticing😉

ecoussoux-ansys: Oh I see, that makes sense, thanks for the explanations and the follow-up commit! 👍

SMoraisAnsys:
Was that PR merged by mistake ? I though the team wanted to keep working with 3.10 to align with the version in AEDT. Did i miss something @Samuelopez-ansys ?

Samuelopez-ansys:
> Was that PR merged by mistake ? I though the team wanted to keep working with 3.10 to align with the version in AEDT. Did i miss something @Samuelopez-ansys ?

If I am not wrong, for the tests is still using 3.10 right?

eblanco-ansys:
> > Was that PR merged by mistake ? I though the team wanted to keep working with 3.10 to align with the version in AEDT. Did i miss something @Samuelopez-ansys ?
> 
> If I am not wrong, for the tests is still using 3.10 right?

That's correct, we are aligned with AEDT version for the tests

Issue #6955, opened 2025-11-27
Add library ansys-edb-core in version maanger


hui-zhou-a:
### Description of the feature

PyEDB gRPC is dependent on ansys-edb-core, which needs to be updated regularly together with PyEDB

### Steps for implementing the feature

NA

### Useful links and references

all

hui-zhou-a:
@eblanco-ansys could you help on this?

PR #6979, opened 2025-12-03
FIX: Update pip install command to use eager upgrade strategy for pyaedt and pyedb

eblanco-ansys:
## Description
Now version manager updates all packages, even if they comply pyproject.toml requirements

## Issue linked
#6955

## Checklist


eblanco-ansys:
@hui-zhou-a this should fix #6955. Can you check?

src/ansys/aedt/core/extensions/installer/version_manager.py
```diff
@@ -405,7 +405,7 @@ def update_pyaedt(self):
             if self.pyaedt_version > latest_version:
                 pip_args = ["install", f"pyaedt[all]=={latest_version}"]
             else:
-                pip_args = ["install", "-U", "pyaedt[all]"]
+                pip_args = ["install", "-U", "--upgrade-strategy", "eager", "pyaedt[all]"]
```
SMoraisAnsys: should eager be used for both cases if it's getting added ?

eblanco-ansys: `self.pyaedt_version > latest_version` usually means we have dev version installed and we want to install the stable one. The recommended way to do it is not using eager because it can break the environment, but it's true that if we don't use it we cannot guarantee latest versions will be installed for all packages

Samuelopez-ansys: What is the status of this? Should we merge?

eblanco-ansys: @hui-zhou-a was testing the changes last week. I think we can merge if there are no issues

PR #6999, opened 2025-12-11
FIX: Allow EMIT table string inputs

mcoleman-ansys:
## Description
All users to input string with unit into tables (eg. "100 kW", "10 MHz").

## Issue linked

## Checklist


src/ansys/aedt/core/emit_core/nodes/emit_node.py
```diff
@@ -551,6 +551,154 @@ def _is_column_data_table(self):
...
+        data : list of tuples
+            User inputted table data.
+
+        Returns
+        -------
+        list of tuples
+            Data converted to SI units.
+        """
+        exceptions = {
+            "TxHarmonicNode": {"Absolute": [None, "PowerUnit"], "Relative": [None, "Power (dBc)"]},
```
jsalant22, line 578:
```suggestion
            "TxHarmonicNode": {
                "Absolute": [None, "PowerUnit"], 
                "Relative": [None, "Power (dBc)"]},
```

src/ansys/aedt/core/emit_core/nodes/emit_node.py
```diff
@@ -551,6 +551,154 @@ def _is_column_data_table(self):
...
+                "Absolute": ["FrequencyUnit", "PowerUnit"],
+                "RelativeBandwidth": ["FrequencyUnit", "Attenuation (dB)"],
+            },
+            "TxBbEmissionNode": {
+                "Absolute": ["FrequencyUnit", "Amplitude (dBm/Hz)"],
+                "RelativeBandwidth": ["FrequencyUnit", "Amplitude (dBm/Hz)"],
+                "RelativeOffset": ["FrequencyUnit", "Amplitude (dBm/Hz)"],
+                "BroadbandEquation": ["FrequencyUnit (MHz)", "Amplitude (dBm/Hz)"],
+            },
+            "RxMixerProductNode": {"Absolute": [None, None, "PowerUnit"], "Relative": [None, None, "Power (dBc)"]},
```
jsalant22, line 589:
```suggestion
            "RxMixerProductNode": {
                "Absolute": [None, None, "PowerUnit"], 
                "Relative": [None, None, "Power (dBc)"]},
```

tests/system/emit/test_emit.py
```diff
@@ -2008,6 +2008,164 @@ def test_fm_fsk_freq_deviation(self, emit_app):
...
+
+        tx_harmonics = cast(TxHarmonicNode, tx_harmonics)
+        tx_narrowband_emissions_mask = cast(TxNbEmissionNode, tx_narrowband_emissions_mask)
+        tx_spurious_emissions = cast(TxSpurNode, tx_spurious_emissions)
+        tx_broadband_noise_profile = cast(TxBbEmissionNode, tx_broadband_noise_profile)
+
+        rx_mixer = cast(RxMixerProductNode, rx_mixer)
+        rx_saturation = cast(RxSaturationNode, rx_saturation)
+        rx_selectivity = cast(RxSelectivityNode, rx_selectivity)
+        rx_spurious_responses = cast(RxSpurNode, rx_spurious_responses)
```
jsalant22: You can leave as is, but convenient shorthand that Carlos showed me a while back is: 

`tx_harmonics : TxHarmonicNode = tx_spectral_profile.add_custom_tx_harmonics()`

src/ansys/aedt/core/emit_core/nodes/emit_node.py
```diff
@@ -551,6 +551,154 @@ def _is_column_data_table(self):
             return False
         return True
 
+    def _check_column_table_data(self, data):
+        """Converts user inputted int or string table data to SI units.
```
jsalant22, line 560: Can we add a list of the impacted nodes here? It'll make it easier if we need to debug/modify this in the future.

src/ansys/aedt/core/emit_core/nodes/emit_node.py
```diff
@@ -551,6 +551,154 @@ def _is_column_data_table(self):
...
+                        for unit_type, valid_units_list in EMIT_VALID_UNITS.items():
+                            if unit in valid_units_list:
+                                valid_unit = True
+                                unit_system = unit_type
+                                break
+                        if valid_unit:
+                            row_list[i] = self._convert_to_internal_units(value, unit_system)
+                        else:
+                            raise ValueError(f"{unit} are not valid units for this property.")
+                else:
```
jsalant22, line 643: Just to be extra safe, we should verify this is a float/int and throw a ValueError if not.

src/ansys/aedt/core/emit_core/nodes/emit_node.py
```diff
@@ -551,6 +551,154 @@ def _is_column_data_table(self):
...
+                    # Handle numeric inputs
+                    row_list[i] = value
+            data_return.append(tuple(row_list))
+        return data_return
+
+    def _check_node_prop_table_data(self, data):
+        """Converts user inputted int or string table data to SI units.
```
jsalant22, line 672: As above, can we add a list of the impacted nodes here? It'll make it easier if we need to debug/modify this in the future.

src/ansys/aedt/core/emit_core/nodes/emit_node.py
```diff
@@ -551,6 +551,154 @@ def _is_column_data_table(self):
...
+                # Extract column unit if present in column header
+                col_unit = None
+                if "(" in cols[i]:
+                    col_unit = cols[i][cols[i].find("(") + 1 : cols[i].find(")")]
+
+                    # Update unit type based on column unit
+                    if col_unit in EMIT_VALID_UNITS["Frequency"]:
+                        units[i] = "FrequencyUnit"
+                    elif col_unit in EMIT_VALID_UNITS["Power"]:
+                        units[i] = "PowerUnit"
```
jsalant22, line 706: It _shouldn't_ happen since it's coming from internal code, but you should add an `else` to handle any error cases. It'll help catch issues if someone changes `EMIT_VALID_UNITS` or more likely if we add a new table type in EMIT that uses something other than Freq/Power.

src/ansys/aedt/core/emit_core/nodes/emit_node.py
```diff
@@ -551,6 +551,154 @@ def _is_column_data_table(self):
...
+
+                if "(" in cols[i] and isinstance(val, str):
+                    # Check for function input strings
+                    if "rf" in val.lower():
+                        row_list[i] = val
+                        continue
```
jsalant22: Can we add a comment specifying which nodes these apply to? I think it's just Spurious Emissions and Spurious Responses

tests/system/emit/test_emit.py
```diff
@@ -2008,6 +2008,164 @@ def test_fm_fsk_freq_deviation(self, emit_app):
...
+
+        # Tx Spurious Emissions Table Data
+        tx_spurious_emissions.spur_table_units = tx_spurious_emissions.SpurTableUnitsOption.ABSOLUTE
+        spur_data = [("30 kHz", "50 MHz", "100 kW"), ("4000000 Hz", 5e6, 0.05), (70e6, 10e5, "1000 W")]
+        tx_spurious_emissions.table_data = spur_data
+        assert tx_spurious_emissions.table_data == [(30e3, 50e6, 80), (4e6, 5e6, 0.05), (70e6, 1e6, 60)]
+        tx_spurious_emissions.spur_table_units = tx_spurious_emissions.SpurTableUnitsOption.RELATIVE
+        spur_data = [("20 MHz", "50 MHz", "5 dBc"), ("4000000 Hz", 5e6, "5 dBc"), (70e6, 10e5, "6 dBc")]
+        tx_spurious_emissions.table_data = spur_data
+        assert tx_spurious_emissions.table_data == [(20e6, 50e6, 5), (4e6, 5e6, 5), (70e6, 1e6, 6)]
```
jsalant22: We should add a case with "RF"

src/ansys/aedt/core/emit_core/nodes/emit_node.py
```diff
@@ -551,6 +551,154 @@ def _is_column_data_table(self):
...
+                if "(" in cols[i]:
+                    col_unit = cols[i][cols[i].find("(") + 1 : cols[i].find(")")]
+
+                    # Update unit type based on column unit
+                    if col_unit in EMIT_VALID_UNITS["Frequency"]:
+                        units[i] = "FrequencyUnit"
+                    elif col_unit in EMIT_VALID_UNITS["Power"]:
+                        units[i] = "PowerUnit"
+
+                if "(" in cols[i] and isinstance(val, str):
```
jsalant22, line 708: Why check for "("? Isn't the data formatted like this: `("4000000 Hz", 5e6, 0.05)`, so the values in cols shouldn't have "(" unless I need more coffee...

src/ansys/aedt/core/emit_core/nodes/emit_node.py
```diff
@@ -551,6 +551,154 @@ def _is_column_data_table(self):
...
+                        row_list[i] = val
+                        continue
+
+                    # Process input values and units
+                    s = val.strip().replace(" ", "")
+                    unit_index = s.find(next(filter(str.isalpha, s)))
+                    value = float(s[:unit_index])
+                    input_unit = s[unit_index:]
+
+                    # Exception for dBc and dBm/Hz units
```
jsalant22:
```suggestion
                    # Handle dBc and dBm/Hz units
```

My fault for calling them "exceptions", "special cases" is probably better so people viewing the code don't expect "Exceptions" to be thrown when the cases are encountered.

src/ansys/aedt/core/emit_core/nodes/emit_node.py
```diff
@@ -551,6 +551,154 @@ def _is_column_data_table(self):
...
+                        row_list[i] = consts.unit_converter(value, units[i][:-4], input_unit, col_unit)
+                else:
+                    # Process values that are stored in SI units
+                    if isinstance(val, str):
+                        value, unit = self._string_to_value_units(val)
+                        if unit == "dBc":
+                            row_list[i] = value
+                        else:
+                            row_list[i] = self._convert_to_internal_units(val, units[i][:-4])
+                    elif units[i] == "none":
```
jsalant22:
```suggestion
                    elif units[i] is None:
```

src/ansys/aedt/core/emit_core/nodes/emit_node.py
jsalant22: Actually, looks like you handle these errors down below so the else statement can just be `units[i] = None`

src/ansys/aedt/core/emit_core/nodes/emit_node.py
```diff
@@ -551,6 +551,154 @@ def _is_column_data_table(self):
...
+                    else:
+                        row_list[i] = consts.unit_converter(value, units[i][:-4], input_unit, col_unit)
+                else:
+                    # Process values that are stored in SI units
+                    if isinstance(val, str):
+                        value, unit = self._string_to_value_units(val)
+                        if unit == "dBc":
+                            row_list[i] = value
+                        else:
+                            row_list[i] = self._convert_to_internal_units(val, units[i][:-4])
```
jsalant22, line 745: Is the `[:-4]` to remove "Unit" from "FrequencyUnit" or "PowerUnit"? If so, just change those to `Frequency` and `Power` above since we're specifying the strings anyway (unless they can also come from somewhere else outside this function).

src/ansys/aedt/core/emit_core/nodes/emit_node.py
```diff
@@ -551,6 +551,154 @@ def _is_column_data_table(self):
...
+                    elif units[i] == "none":
+                        row_list[i] = val
+                    else:
+                        row_list[i] = self._convert_to_internal_units(val, units[i][:-4])
```
jsalant22: Same as above re: `[:-4]`

tests/system/emit/test_emit.py
jsalant22: There's also a bug in 25.2 related to the NodeID so we might need to either restructure this code. Basically, if you get one node (e.g. tx_harmonics) and then get a subsequent node (e.g. rx_mixer), aedt/iemit lose track of the NodeID for tx_harmonics. 

So you basically need to make any changes to tx_harmonics before you get a different node. We can see if the tests fail before restructuring though.

tests/system/emit/test_emit.py
```diff
@@ -2008,6 +2008,164 @@ def test_fm_fsk_freq_deviation(self, emit_app):
...
+            tx_broadband_noise_profile.noise_behavior = tx_broadband_noise_profile.NoiseBehaviorOption.EQUATION
+            broadband_data = [("RF+10.0", 10), ("10 MHz", "5 dBm/Hz"), (100e6, "30 dBm/Hz")]
+            tx_broadband_noise_profile.table_data = broadband_data
+            assert tx_broadband_noise_profile.table_data == [("RF+10.0", 10), (10e6, 5), (100e6, 30)]
+            tx_broadband_noise_profile.noise_behavior = (
+                tx_broadband_noise_profile.NoiseBehaviorOption.RELATIVE_BANDWIDTH
+            )
+            broadband_data = [(1e6, 10), ("10 MHz", "5 dBm/Hz"), (100e6, "30 dBm/Hz")]
+            tx_broadband_noise_profile.table_data = broadband_data
+            assert tx_broadband_noise_profile.table_data == [(1e6, 10), (10e6, 5), (100e6, 30)]
```
jsalant22, line 2176: Testing looks good overall, looks like a good mix of unitless and values with units tested. Once CodeCov completes, we should make sure all branches of if/else statements are tested and add some error testing for bad inputs too.

jsalant22 (commented):
@mcoleman-ansys  PR looks great overall. Mostly nitpicking and some minor suggestions.

jsalant22:
@mcoleman-ansys Just saw EmitTest failed. Hard to tell from that error if it's because of the NodeID error or not. Are they passing locally? If so, what version are you running? The current official release is 25.2, so that's what is run by these tests (makes it fun when we're developing on 26.2!!)

src/ansys/aedt/core/emit_core/nodes/emit_node.py
mcoleman-ansys: Yes, the `[:-4]`  is to remove the "Unit" portion. The units for each column come from the table properties with `units = self._get_property("TableUnitTypes", True)` which returns 'FrequencyUnit', 'PowerUnit' or 'none'

src/ansys/aedt/core/emit_core/nodes/emit_node.py
mcoleman-ansys: I wish this were the case, the `_get_properties` function returns a string of "none" instead of a NoneType object

jsalant22 (approved):

PR #7016, opened 2025-12-16
FEAT: Update Icepak mesh settings for 25R2 and mesh fusion support

lorenzovecchietti:
Update mesh settings to support mesh fusion in icepak


gmalinve:
KUDOS!!!

Issue #7017, opened 2025-12-16
Bug located in variable_manager.set_variable


Ivanz7:
### Before submitting the issue

- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### Description of the bug

It reports errors with PyAEDT version 0.23.0.

<img width="2329" height="1756" alt="Image" src="https://github.com/user-attachments/assets/286d617a-e903-494c-854b-09a7def1261e" />

But there is no such errors with PyAEDT version 0.21.0

<img width="2701" height="1573" alt="Image" src="https://github.com/user-attachments/assets/e2a34095-2947-4c9f-830a-354bc7d86e51" />

### Steps To Reproduce

# 
from pyaedt import Circuit

app = Circuit(version="252")
app.variable_manager.set_variable(name = 'Frequency', expression = '0GHz', circuit_parameter=False)

### Which Operating System are you using?

Windows

### Which Python version are you using?

3.10

### Installed packages

PyAEDT version 0.23.0

gmalinve:
Hi @Ivanz7, could you try if this merged PR: #7020 fix your problem? is merged into main already so you can try with it.
If this solved your problem could you please close this issue?

Ivanz7:
Hi @gmalinve, Thanks for your information. I have tried with PR: 7020 which has fixed this problem. I will Close this issue.

PR #7020, opened 2025-12-16
FIX: create variable array

hui-zhou-a:
grpc method doesn't recognize ' 

The fix is to replace ' with "


gmalinve:
@hui-zhou-a I think this other issue: https://github.com/ansys/pyaedt/issues/7017 can be fixed by yours, right? If yes, can you please link it and I'll approve the PR.

hui-zhou-a:
@gmalinve #7017 doesn't look like a related problem. Better to debug and create a new PR.

PR #7041, opened 2025-12-24
FEAT: Improve assign_em_loss method

lorenzovecchietti:


src/ansys/aedt/core/icepak.py
```diff
@@ -1100,73 +1100,87 @@ def edit_design_settings(
...
         """
         if surface_objects is None:
             surface_objects = []
-        if assignment is None:
-            assignment = []
-
-        self.logger.info("Mapping EM losses.")
+        if not isinstance(assignment, list):
+            assignment = [assignment]
+        if isinstance(assignment[0], Object3d):
```
Samuelopez-ansys, line 1179: What if the list of assignments is not only Object3d? I think you should check that all elements are Object3d

lorenzovecchietti: I stuck to `Union[str, Object3d, List[str], List[Object3d]]` because I wanted to avoid the overhead of mixed-type lists.

If we support `List[Union[...]]`, we lose the ability to verify inputs quickly; we'd have to iterate through the entire list every time. Given the amount of input cleanup pyaedt already does, I worry this will degrade performance for complex projects compared to native APIs.

That said, I'm open to changing this if you think supporting mixed inputs is necessary. Just let me know!

src/ansys/aedt/core/icepak.py
```diff
@@ -1100,73 +1106,87 @@ def edit_design_settings(
...
+        sweep : str
+            Name of the source setup sweep from which to map losses.
+        source_project_name : str, optional
+            Name of the source project from which to map losses. The default is ``None``, in which case the
+            source from the same project is used.
+        name : str, optional
+            Name to assign to the boundary condition.
+            The default is ``None`` in which case it will be automatically generated.
+        map_frequency : str, float, list of float, or list of str, optional
+            Frequency or frequencies to map. The default is ``None``.
```
SMoraisAnsys:
```suggestion
        map_frequency : str, float, list of float, or list of str, optional
            Frequency or frequencies to map. The default is ``None``.
        source_project_name : str, optional
            Name of the source project from which to map losses. The default is ``None``, in which case the
            source from the same project is used.
        name : str, optional
            Name to assign to the boundary condition.
            The default is ``None`` in which case it will be automatically generated.
```

src/ansys/aedt/core/icepak.py
```diff
@@ -1100,73 +1106,87 @@ def edit_design_settings(
...
         self,
-        design="HFSSDesign1",
-        setup="Setup1",
-        sweep="LastAdaptive",
-        map_frequency=None,
-        surface_objects=None,
-        source_project_name=None,
-        parameters=None,
-        assignment=None,
+        assignment: Union[str, Object3d, List[str], List[Object3d]],
```
SMoraisAnsys:
```suggestion
        assignment: str | Object3d | list[str] | list[Object3d],
```
I think we can use a more moderne syntaxe since we are CPython 3.10+ and remove some of the typing imports. I'm only proposing a change for the first line  because I can't propose changes for all the lines at once from GH interface...

src/ansys/aedt/core/icepak.py
```diff
@@ -28,6 +28,11 @@
 import os
 from pathlib import Path
 import re
+from typing import Dict
```
SMoraisAnsys: We might consider removing those as suggested in a coming comment

src/ansys/aedt/core/icepak.py
```diff
@@ -2023,27 +2057,6 @@ def globalMeshSettings(
         )
         return True
 
-    @pyaedt_function_handler()
-    def delete_em_losses(self, bound_name):
```
SMoraisAnsys, line 2027: You are removing this method because the other method returns an object that can be directly deleted, right ?

src/ansys/aedt/core/icepak.py
Samuelopez-ansys: I agree, I am only asking for a check, that all elements of the list are Object3d, and if they are not, we could raise an Exception.

src/ansys/aedt/core/icepak.py
lorenzovecchietti: Yes, exactly. `assign_em_losses` returns a boundary object that can be deleted. I removed this because it doesn't make sense to have a delete method just for this type of boundary object.

Issue #7038, opened 2025-12-22
Improve function get_coupling_in_range() from touchstone_parser.py


amichel0205:
### Description of the feature

Improve function get_coupling_in_range() from touchstone_parser.py:

1. Allow to specify frequency range in which looking for coupling
2. Allow user to exclude S parameter from different pin of the same component
3. Allow user the possibility of excluding S parameter from a list of component

### Steps for implementing the feature

1.

        # identify frequency index at which to start and stop the check
        k_start = bisect.bisect_left(self.frequency.f, start_frequency)
        k_stop = bisect.bisect_right(self.frequency.f, stop_frequency)

replace nb_freq with k_stop for k in range(k_start, k_stop, frequency_sample):

2.

                    refdes1 = self.port_names[i].split("_")[0]
                    refdes2 = self.port_names[j].split("_")[0]
                    if refdes1 == refdes2:
                        continue

3.

in investigation

PR #7061, opened 2026-01-07
FEAT: Update_get_coupling_in_range

amichel0205:
## Description
Improve function get_coupling_in_range() from touchstone_parser.py:

Allow to specify frequency range in which looking for coupling
Allow user to exclude S parameter from different pin of the same component
Allow user the possibility of excluding S parameter from a list of component

## Issue linked
Improve function get_coupling_in_range() from touchstone_parser.py #7038

## Checklist

### DIFF
--- a/.github/workflows/ci_cd.yml
+++ b/.github/workflows/ci_cd.yml
@@ -275,7 +275,7 @@ jobs:
           python-version: ${{ env.TESTS_VERSION }}
 
       - name: Set up uv
-        uses: astral-sh/setup-uv@681c641aba71e4a1c380be3ab5e12ad51f415867 # v7.1.6
+        uses: astral-sh/setup-uv@61cb8a9741eeb8a550a1b8544337180c0fc8476b # v7.2.0
         with:
           enable-cache: false
           prune-cache: true
@@ -353,7 +353,7 @@ jobs:
           python-version: ${{ env.TESTS_VERSION }}
 
       - name: Set up uv
-        uses: astral-sh/setup-uv@681c641aba71e4a1c380be3ab5e12ad51f415867 # v7.1.6
+        uses: astral-sh/setup-uv@61cb8a9741eeb8a550a1b8544337180c0fc8476b # v7.2.0
         with:
           enable-cache: false
           prune-cache: true
@@ -428,7 +428,7 @@ jobs:
           python-version: ${{ env.TESTS_VERSION }}
 
       - name: Set up uv
-        uses: astral-sh/setup-uv@681c641aba71e4a1c380be3ab5e12ad51f415867 # v7.1.6
+        uses: astral-sh/setup-uv@61cb8a9741eeb8a550a1b8544337180c0fc8476b # v7.2.0
         with:
           enable-cache: false
           prune-cache: true
@@ -514,7 +514,7 @@ jobs:
           python-version: ${{ env.TESTS_VERSION }}
 
       - name: Set up uv
-        uses: astral-sh/setup-uv@681c641aba71e4a1c380be3ab5e12ad51f415867 # v7.1.6
+        uses: astral-sh/setup-uv@61cb8a9741eeb8a550a1b8544337180c0fc8476b # v7.2.0
         with:
           enable-cache: false
           prune-cache: true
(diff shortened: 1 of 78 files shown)


amichel0205:
Hello Samuel,

Yes, it is not needed. The function works with an S parameter. The aedb was used to identify port location. I remove this part as it was not working as expected. 
I intend to create a script if no such function exist that will identified (and this maybe out of standard PyAEDT function) that will work with aedb and log file created from this function such we can extract more than port location like net connected to pin and so on.

amichel0205:
get_coupling_in_range is working on S parameter file and export a log file of this format:
S(MA9_2,MA60_O) Loss= -42.86dB Freq= 2.000GH, MA9_, MA60_O
S(MA9_3,MA60_O) Loss= -59.31dB Freq= 2.000GH, MA9_, MA60_O
S(MA9_4,MA60_O) Loss= -59.94dB Freq= 5.000GH, MA9_, MA60_O

Where the port name are RefDes_Pin. I think it would be better to have another function that will browse this log file such we can extract different properties from the component/ pin (like net name) if needed. 
This function maybe not enough generic to have it as standard function so I balance to either create a new function or share a script via "How to" issue.

src/ansys/aedt/core/visualization/advanced/touchstone_parser.py
```diff
@@ -180,104 +181,128 @@ def reduce(
...
-        high_loss=-60,
-        frequency_sample=5,
-        output_file=None,
-        aedb_path=None,
-        design_name=None,
-    ):
+        start_frequency: float = 1e9,
+        stop_frequency: float = 10e9,
+        max_loss: float = -40.0,
+        min_loss: float = -60,
```
SMoraisAnsys:
```suggestion
        min_loss: float = -60.0,
```
Just to align with the type hint. Could you also reorder the parameters to have min defined before max ?

tests/system/visualization/test_touchstone_parser.py
```diff
@@ -87,38 +87,50 @@ def test_check_touchstone_file(test_tmp_dir):
 
 
 def test_get_coupling_in_range(test_tmp_dir):
-    touchstone_file_path_o = TEST_DIR / "HSD_PCIE_UT_main_cutout.s16p"
-    touchstone_file_path = shutil.copy2(touchstone_file_path_o, test_tmp_dir / "HSD_PCIE_UT_main_cutout.s16p")
+    touchstone_file_path_o = TEST_DIR / "si_verse_cutout.s16p"
+    touchstone_file_path = shutil.copy2(touchstone_file_path_o, test_tmp_dir / "si_verse_cutout.s16p")
```
SMoraisAnsys, line 91: If you have to use new data, could you remove `HSD_PCIE_UT_main_cutout.s16p` and `HSD_PCIE_UT.aedt` ? I think these data are not used by any other test so we could remove it.

tests/system/visualization/test_touchstone_parser.py
```diff
@@ -87,38 +87,50 @@ def test_check_touchstone_file(test_tmp_dir):
...
-        start_frequency=1e9, high_loss=-60, low_loss=-40, frequency_sample=5, output_file=str(output_file)
+        start_frequency=2e9,
+        stop_frequency=5e9,
+        include_same_component=True,
+        component_filter=[],
+        include_filter=True,
+        frequency_sample=5,
+        output_file=str(output_file),
+    )
+    assert isinstance(res, list)
```
SMoraisAnsys, line 105: COuld this test be extend to check the content of res and not only the type ? Currently we are not really testing correctly and only know that the call worked + the return type is a list. If an empty list is returned for every call or `["whatever", "I", "want"]` we wouldn't see it :/

SMoraisAnsys (dismissed):
Thanks for the PR @amichel0205, I left a few comments.
Note that you are currently breaking the PyAEDT API. Could you extend the description of this PR to explain the reason for that ?

amichel0205:
We need to discuss about the change with loss definition.
maxloss should be -60db is define more losses than minloss of -40db
loss means loss :-) which mean -60db is higher loss than -40db.

amichel0205:
I've reverted change for loss parameters.
I reverted parameter max_loss and min_loss to high_loss and low_loss as it makes more sens for RF/HF users. With regard to loss definition -60dB is higher loss than -40dB.

SMoraisAnsys (approved):

PR #7079, opened 2026-01-12
FIX: make sure rename fcn is in pyaedt

jsalant22:
## Description

## Issue linked

## Checklist


Samuelopez-ansys:
Hi @jsalant22 , could it make sense to have all the duplicated methods (rename, duplicate...) in EmitNode class? If I am not wrong, the other classes (Amplifier, Cable...) could inherit it, and the coverage could increase because with this PR, we would need to test all the rename methods of all components.

jsalant22:
> Hi @jsalant22 , could it make sense to have all the duplicated methods (rename, duplicate...) in EmitNode class? If I am not wrong, the other classes (Amplifier, Cable...) could inherit it, and the coverage could increase because with this PR, we would need to test all the rename methods of all components.

All of the classes in the _generated_ folder inherit from EmitNode, but not all of them allow renaming. Duplicate/Delete are also blocked in some instances (I'm not sure if it's always the same cases) So we'd either need to:

1. Add another class that inherits from EmitNode that we can add rename/duplicate/delete to and then have the necessary _generated_ classes inherit from that. OR
2. Add logic to `rename()` that gives a warning if a class isn't renamable. This also means that classes that can't be renamed/duplicated/deleted would have these functions show up for auto-complete even though they can't be used which is something we'd like to avoid.

The `test_all_generated_emit_node_properties` class has the ability to test all the renames, but it looks like it's currently being skipped. I can try and implement that.

jsalant22:
Actually, taking a 2nd look at EmitNode, we already have EmitNode.name which is inherited by all generated classes and we were intending to deprecate EmitNode._rename() in the future (presumably with PyAEDT 1.0). 

If it makes sense, I can just remove all the rename() functions for the generated classes now. Thoughts? @Samuelopez-ansys

Samuelopez-ansys:
It is Ok for me, in PYAEDT objects we do not have a rename method, because it is the job of the name property setter.

In addition, what about all the duplicate methods? Could not be part of EmitNode?

jsalant22:
> It is Ok for me, in PYAEDT objects we do not have a rename method, because it is the job of the name property setter.
> 
> In addition, what about all the duplicate methods? Could not be part of EmitNode?

I'll create a story for us to look at moving all the duplicate/delete methods to EmitNode. Don't want to handle that with this PR though.

jsalant22:
@Samuelopez-ansys Can we merge this?

Samuelopez-ansys:
@jsalant22 One test is failing: test_emitters_radios

jsalant22:
@Samuelopez-ansys Sorry, page stopped auto-refreshing and I missed that.

PR #7106, opened 2026-01-16
CI: Skip tests docstrings

eblanco-ansys:
## Description
Introduce an `analyze-changes` script to assess code changes in CI, enhancing the workflow by determining whether to skip tests based on the analysis results. Leveraged it in the CICD for skipping heavy tests where they are not affected by code logic changes

## Checklist

### DIFF
--- a/doc/changelog.d/7143.fixed.md
+++ b/doc/changelog.d/7143.fixed.md
@@ -1,0 +1,1 @@
+7142 - error in polar plotting.
--- a/src/ansys/aedt/core/visualization/advanced/farfield_visualization.py
+++ b/src/ansys/aedt/core/visualization/advanced/farfield_visualization.py
@@ -821,8 +821,8 @@ class FfdSolutionData(PyAedtBase):
             Generate the plot in polar coordinates. The default is ``True``. If ``False``, the plot
             generated is rectangular.
         max_theta : float or int, optional
-            Maximum theta angle for plotting. The default is ``180``, which plots the far-field data for
-            all angles. Setting ``max_theta`` to 90 limits the displayed data to the upper
+            Maximum theta angle for plotting far-field data. The default value is ``180``, which
+            plots all angles. Setting ``max_theta`` to 90 limits the displayed data to the upper
             hemisphere, that is (0 < theta < 90).
 
         Returns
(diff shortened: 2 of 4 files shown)


SMoraisAnsys (dismissed):
Thanks for the contribution @eblanco-ansys !

I think part of the proposed changes could greatly enhance the CI but I'm a bit conflicted with the docstring check. First, the checking should be removed from the CLI and, if needed, we can create a .ci folder at the root of the repo to contain CI related stuff. Second, I think the implementation has to cover a lot of edge cases (this can be seen from all the checks currently in the proposed changes) and we might still be missing some of them, so discarding testing could really bite us back :/ 

Checking the kind of changes performed can be valuable though. For example having a look into the folder impacted by the changes would easily help us to distinguish between non code related PRs and the others. Note that anything affecting the CI should trigger the whole workflow though. Otherwise, we might also miss something.

eblanco-ansys:
> Thanks for the contribution @eblanco-ansys !
> 
> I think part of the proposed changes could greatly enhance the CI but I'm a bit conflicted with the docstring check. First, the checking should be removed from the CLI and, if needed, we can create a .ci folder at the root of the repo to contain CI related stuff. Second, I think the implementation has to cover a lot of edge cases (this can be seen from all the checks currently in the proposed changes) and we might still be missing some of them, so discarding testing could really bite us back :/
> 
> Checking the kind of changes performed can be valuable though. For example having a look into the folder impacted by the changes would easily help us to distinguish between non code related PRs and the others. Note that anything affecting the CI should trigger the whole workflow though. Otherwise, we might also miss something.

I forgot to add it in the PR comments, but the fact that I implemented it in the cli initially was just for easy local testing. I agree with you that all the stuff should be moved to another folder. If not, cli can become too complex.
My initial idea was:
- Skip all tests if we just changed documentation
- Skip tests if changes affect docstrings or comments 
- If code changes affect to the logic, then run everything

If I'm not wrong CI changes trigger the whole workflow with this implementation.

I'm sure this can be widely improved to cover all the edge cases, but I think having something like this would be very beneficial to us. IMO, it would push us to improve more little details of the documentation because we don't need to wait for the full pipeline, that currently is one of the main limitations.

Thanks a lot for the feedback! @SMoraisAnsys

PR #7131, opened 2026-01-22
CI: Add uv lock file for testmon and reproducibility

SMoraisAnsys:
## Description
Following the testmon required changes identified by @eblanco-ansys  and the work started in https://github.com/ansys/actions/pull/991, this PR adds a uv lock file that should be used for testing. This will be a first step to see how it should be integrated into a PyAnsys repo at large.

## Issue linked
Associated to https://github.com/ansys/actions/issues/985

## Checklist

### DIFF
--- a/doc/changelog.d/7138.maintenance.md
+++ b/doc/changelog.d/7138.maintenance.md
@@ -1,0 +1,1 @@
+Temporary remove jupyter target from all
--- a/doc/source/Resources/pyaedt_installer_from_aedt.py
+++ b/doc/source/Resources/pyaedt_installer_from_aedt.py
@@ -402,7 +402,6 @@ def install_pyaedt():
                 subprocess.run([str(uv_exe), "pip", "install", "wheel"], check=True, env=env)  # nosec
                 if args.version <= "231":
                     subprocess.run([str(uv_exe), "pip", "install", "pyaedt[all]=='0.9.0'"], check=True, env=env)  # nosec
-                    subprocess.run([str(uv_exe), "pip", "install", "jupyterlab"], check=True, env=env)  # nosec
                     subprocess.run([str(uv_exe), "pip", "install", "ipython", "-U"], check=True, env=env)  # nosec
                     subprocess.run([str(uv_exe), "pip", "install", "ipyvtklink"], check=True, env=env)  # nosec
                 else:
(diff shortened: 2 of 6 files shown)


eblanco-ansys:
I changed some things in the cicd:
- Removed setup-python job. Not needed anymore, this is now handled directly by uv
- Removed retry action. We shouldn't need it anymore, tests are consistent enough
- Deleted "Upgrade wheel and setuptools" job
- Unit and integration tests don't run through `ansys/actions/tests-pytest` anymore

Probably some small things can be improved, but I think it's ready for merge
@SMoraisAnsys @Samuelopez-ansys @ecoussoux-ansys @gmalinve

.github/workflows/update-testmondata-cache.yml
```diff
@@ -259,15 +279,28 @@ jobs:
...
+
+      - name: Remove Ansys processes (if any)
+        shell: bash
         run: |
-          source .venv/bin/activate
-          uv pip install .[tests]
+          for pid in $(ps -eo pid,comm,args | grep -iE "ansys.inc|ansysem" | grep -v grep | awk '{print $1}'); do
+            echo "Killing PID $pid"
+            kill -9 "$pid"
+          done
```
ecoussoux-ansys, line 327: Here we have the "Remove Ansys processes (if any)" step twice now.

eblanco-ansys:
> I reviewed the changes and just found the minimal comment below.
> 
> I would also have one question: Should we also align the `nightly-tests.yml` and `weekly_tests.yml` workflows with the changes introduced in this PR?

Yes, we should align them too

.github/workflows/update-testmondata-cache.yml
eblanco-ansys: > Here we have the "Remove Ansys processes (if any)" step twice now.

Thanks for the catch!! @ecoussoux-ansys

.github/workflows/ci_cd.yml
```diff
@@ -220,6 +220,27 @@ jobs:
         with:
           persist-credentials: false
 
+      - name: Install Xvfb
+        shell: bash
+        run: sudo apt-get install -y xvfb
+
+      - name: Set up uv
```
SMoraisAnsys, line 232: @eblanco-ansys I was keeping the setup-python to leverage the base image of the containers, see https://github.com/astral-sh/setup-uv?tab=readme-ov-file#do-i-still-need-actionssetup-python-alongside-setup-uv. But we can remove it if you prefer. Maybe it would be worth to setup the python version input then ?

eblanco-ansys: Now python version is introduced inside the `uv sync --frozen --python "$TESTS_VERSION" --extra unit-tests` command, but knowing it's faster using `setup-version` I'll revert to previous implementation

Issue #7136, opened 2026-01-23
Add method for PyAEDT example test


anur7:
Currently, PyAEDT examples must be manually executed and verified whenever a new PyAEDT version is released. This process is time‑consuming .

It would be beneficial to have methods (or a dedicated workflow) that automatically test all PyAEDT examples during each new release cycle.

SMoraisAnsys:
I don't think adding a method in pyaedt for this purpose should be performed. Given the content of the check, a simple call to `is_close` inside of the example should be enough. If the condition is not valid, then an exception can be raised.

gmalinve:
@anur7 closing this one as implmented in examples repo.

PR #7137, opened 2026-01-23
FEAT: implemented a method to test PyAEDT examples

anur7:
## Description
**This method is intended to automatically test PyAEDT examples for new PyAEDT releases, eliminating the need to manually run the examples each time a new version is published.**

## Issue linked
#7136

## Checklist

### DIFF
--- a/doc/changelog.d/7137.added.md
+++ b/doc/changelog.d/7137.added.md
@@ -1,0 +1,1 @@
+Implemented a method to test PyAEDT examples


gmalinve:
@SMoraisAnsys thanks for the feedback :) from my point of view it's better that this stays in PyAEDT because it can be later reused by other examples. 
I personally don't like the idea that the customer downloads the examples with the reference check method implementation in it if this is what you mean. 
Ideally this method will be used by all examples to check whether something has changed in results from one version to another (or for whichever other reason) and have it in all of them I don't think is nice.
Let me know if you intended something else ;)

@Samuelopez-ansys

tests/unit/test_variable.py
```diff
@@ -438,3 +440,13 @@ def test_validator_float_type_invalidate_zeros(validation_float_input):
...
+
+
+def test_reference_check():
+    a = 15
+    b = 20
+    error = 0.1
+    file = Path(__file__)
+    with pytest.raises(ValueError) as excinfo:
+        reference_check(a, b, error, file)
+    assert Path(file).name in str(excinfo.value)
```
gmalinve, line 452: @anur7 there's actually no need of this line as it does not affect code cov.

tests/unit/test_variable.py
```diff
@@ -438,3 +440,13 @@ def test_validator_float_type_invalidate_zeros(validation_float_input):
     validation_errors = generate_validation_errors(property_names, expected_settings, actual_settings)
 
     assert len(validation_errors) == 1
+
+
+def test_reference_check():
+    a = 15
+    b = 20
+    error = 0.1
+    file = Path(__file__)
```
gmalinve, line 449: @anur7 I think is safer if you use the pytest fixture test_tmp_dir and build a dummy path instead of retrieving the path of the test file.
Something like:
file = test_tmp_dir / "my_file.py"

@Samuelopez-ansys

SMoraisAnsys:
> @SMoraisAnsys thanks for the feedback :) from my point of view it's better that this stays in PyAEDT because it can be later reused by other examples. I personally don't like the idea that the customer downloads the examples with the reference check method implementation in it if this is what you mean. Ideally this method will be used by all examples to check whether something has changed in results from one version to another (or for whichever other reason) and have it in all of them I don't think is nice. Let me know if you intended something else ;)
> 
> @Samuelopez-ansys

I dont see the need to create a method in pyaedt when the code is an if statement and raison of an exczption. If is_close didnt existe, I would have understood but it is already part of the API.

gmalinve:
@SMoraisAnsys but the is_close does not reaise an exception which is what we want to catch in the CI/CD of the examples if any results is outside a specific range. 

@Samuelopez-ansys

src/ansys/aedt/core/generic/numbers_utils.py
```diff
@@ -496,3 +498,18 @@ def _find_units_in_dependent_variables(variable_value, full_variables=None):
...
+
+def reference_check(var_test, reference, error, file):
+    """Check the reference value in PyAEDT examples throwing error in case of mismatch.
+
+    Parameters
+    ----------
+    var_test: scalar value of the quantity to test
+    reference: reference value of the testing quantity
+    error: percentage error
+    file: path of the texting example
```
gmalinve, line 511: @anur7 here you are missing types.

SMoraisAnsys:
> @SMoraisAnsys but the is_close does not reaise an exception which is what we want to catch in the CI/CD of the examples if any results is outside a specific range.
> 
> @Samuelopez-ansys

I agree but that only means that we have to add the following snippet into the examples
```
if not is_close(...):
    raise RuntimeError("....")
```
instead of 
```
reference_check(...)
```

Still thinking this shouldn't be added, sry :)

nunesanderson:
@gmalinve @SMoraisAnsys @Samuelopez-ansys 

If we will include it in many examples, I would suggest creating the external function as Abdun did. Otherwise, in case we need to change that behavior in the future, we would need to touch all the examples. In the proposed way, we would touch only one function.
That's why I'm more in favor of calling the reference_check(...) in the examples. This way, we can do whatever we want within it.

SMoraisAnsys:
> @gmalinve @SMoraisAnsys @Samuelopez-ansys
> 
> If we will include it in many examples, I would suggest creating the external function as Abdun did. Otherwise, in case we need to change that behavior in the future, we would need to touch all the examples. In the proposed way, we would touch only one function. That's why I'm more in favor of calling the reference_check(...) in the examples. This way, we can do whatever we want within it.

If the method was larger I wouldn't oppose to it but here it's only 2 lines of code :/

Samuelopez-ansys:
I think this PR can be closed right?

anur7:
yes @Samuelopez-ansys. This PR can be closed since we decided to implement the tests directly in the examples

Issue #7142, opened 2026-01-24
Bug located in farfield_visualization


Devin-Crawford:
### Before submitting the issue

- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### Description of the bug

Polar plots in the ``contour_plot`` method cut off half the plot. This could be
seen in the component array example.

``max_theta`` is passed as an argument to ``new.plot_contour()`` on line 873 in "farfield_visualization.py". The ``max_theta`` in the high level method has a diferent meaning than ``max_theta`` in ``plot_contour()`` however, which causes the plot to be cut in half.

### Steps To Reproduce

See the [component array example](https://examples.aedt.docs.pyansys.com/version/dev/examples/high_frequency/antenna/array.html). Only half of the polar contour plot is displayed.

### Which Operating System are you using?

Windows

### Which Python version are you using?

3.12

### Installed packages

All dependencies are installed.

Devin-Crawford:
Resolved in PR #7142

PR #7143, opened 2026-01-24
FIX: 7142 - error in polar plotting.

Devin-Crawford:
max_theta was incorrectly interpreted in a refactor of the far-field plotting, resulting in half the polar plot being cut off. This has been corrected.

## Description
Fix error in polar plotting.

``max_theta`` was incorrectly interpreted passed from ``FarfieldVisualization.plot_contour()`` to
``matplotlib.plot_contour()`` with a value of 180. This caused only half of the polar plot to be
rendered.

## Issue linked
#7142 
#7146

### DIFF
--- a/.github/workflows/update-testmondata-cache.yml
+++ b/.github/workflows/update-testmondata-cache.yml
@@ -1,9 +1,7 @@
 name: Update Testmon Cache on Merge
 
 on:
-  pull_request:
-    types:
-      - closed
+  push:
     branches:
       - main
   workflow_dispatch:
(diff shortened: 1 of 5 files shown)


src/ansys/aedt/core/visualization/advanced/farfield_visualization.py
```diff
@@ -821,8 +821,8 @@ def plot_contour(
             Generate the plot in polar coordinates. The default is ``True``. If ``False``, the plot
             generated is rectangular.
         max_theta : float or int, optional
-            Maximum theta angle for plotting. The default is ``180``, which plots the far-field data for
-            all angles. Setting ``max_theta`` to 90 limits the displayed data to the upper
+            Maximum theta angle for plotting far-ield data. The default value is ``180``, which
```
Samuelopez-ansys:
```suggestion
            Maximum theta angle for plotting far-field data. The default value is ``180``, which
```

src/ansys/aedt/core/visualization/advanced/farfield_visualization.py
```diff
@@ -874,7 +874,6 @@ def plot_contour(
             trace=0,
             polar=polar,
             levels=levels,
-            max_theta=max_theta,
```
Samuelopez-ansys, line 877: Instead of changing the plot_contour max_theta default value to 360, I think it is better to hardcode max_theta=360 here in the farfield plot_contour, and keep plot_contour=180 asthe default in matplotlib file.

Plot_contour is used by other plotters, not only for far field, and I am not sure what the impact is.

Devin-Crawford: The max_theta in plot_contour has a completely different meaning here. It basically cuts the rotational angle off at 180 degrees on a plot that goes from 0 to 360 or from -180 to 180. In other words, the power plot should ALWAYS rotate by 360 degrees.

Devin-Crawford: All polar plots will by default require full-circle 360 degree plotting, so this was a bug.

PR #7167, opened 2026-01-29
CHORE: Check extension changes

Samuelopez-ansys:
## Description
This pull request enhances the CI/CD workflow to enable more granular and efficient testing by detecting changes at the module level and conditionally running system tests only for affected modules. The main updates include improvements to the change analysis script, the addition of module-specific outputs, and modifications to the workflow logic to trigger tests only when relevant modules are changed.

**Module-level change detection and conditional testing:**

* The `.ci/scripts/analyze_changes.py` script now detects changes in specific modules (`emit`, `extensions`, `filter_solutions`) and sets corresponding outputs (`emit_changed`, `extensions_changed`, `filter_solutions_changed`). These outputs are written to GitHub Actions and echoed for visibility.

* The `.github/workflows/ci_cd.yml` workflow consumes these new outputs to conditionally run system tests for each module only when relevant files have changed. This reduces unnecessary test runs and speeds up the CI/CD process.

**Workflow and output improvements:**

* The workflow now prints a summary of the change analysis results as GitHub Actions notices, improving transparency and debugging.

## Issue linked

## Checklist

### DIFF
--- a/.github/workflows/update-testmondata-cache.yml
+++ b/.github/workflows/update-testmondata-cache.yml
@@ -50,11 +50,53 @@ jobs:
           exit 1
 
   # ------------------------------------------------------------------
+  # ANALYZE CHANGES
+  # ------------------------------------------------------------------
+  analyze-changes:
+    name: Analyze code changes
+    runs-on: ubuntu-latest
+    needs: check-reference
+    outputs:
+      skip_tests: ${{ steps.analyze.outputs.skip_tests }}
+      doc_only: ${{ steps.analyze.outputs.doc_only }}
+      run_all: ${{ steps.analyze.outputs.run_all }}
+    permissions:
+      contents: read # Required to read repository content and git history
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
+        with:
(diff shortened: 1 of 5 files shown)


Samuelopez-ansys:
> Just wondering, isn't testmon already enough to handle the fact of running / not running some specific tests ?

Indeed is able to do it. But with this, we do not waste the time of the VM. Testmon needs at least to take the self hosted runner, checkout the repo, install all the deps and then check if something has to be tested. For EMIT, Filter Solutions and Extensions, we are able to know even without testmon if this is needed, because they are well separated in the code. 

Extensions are a bit more tricky, because a change in PyAEDT code could impact a particular extension, but the plan is moving the extensions, so at some point this approach is similar, I am open to discussions.

Samuelopez-ansys:
> IF we have nightly tests of the extensions and can catch that something went wrong then I don't mind. But someone will have to check the nightly run ;)
> 
> Note that I'm still not really reassured with the automatic discard of testing add with the CI script. I think it adds brittleness to our CI when it has already been quitte impacted by testmon.

I understand your concerns. Nightly jobs are not skipping any test, so if something goes wrong, we will see it. I would merge this PR, and see the impact, @ansys/pyaedt-maintainers will need to make a more exhaustive review to verify all works. If we see any problem, we will revert it.

PR #7195, opened 2026-02-05
FEAT: Add pyedb grpc support to pyaedt

maxcapodi78:
## Description
Introduced a new settings environment variable use_pyedb_grpc which is False by default but can be set to True and enable users to run pyedb through ansys.edb.core package. This is gonna work from AEDT 2025R2

## Issue linked

## Checklist


doc/source/Getting_started/Contributing.rst
```diff
@@ -511,6 +512,7 @@ Parameter descriptions:
 - ``use_local_example_data``: When ``true``, uses local example data for tests.
 - ``local_example_folder``: Path to the local example data folder.
 - ``skip_modelithics``: When ``true``, skips Modelithics-related tests.
+- ``use_pyedb_grpc``: When ``true``, use use PyEDB with gRPC for database access.
```
Samuelopez-ansys:
```suggestion
- ``use_pyedb_grpc``: When ``true``, use PyEDB with gRPC for database access.
```

src/ansys/aedt/core/generic/settings.py
```diff
@@ -238,6 +239,7 @@ def __init__(self):
         self.__grpc_secure_mode = DEFAULT_GRPC_SECURE_MODE
         self.__grpc_local = DEFAULT_GRPC_LOCAL
         self.__grpc_listen_all = DEFAULT_GRPC_LISTEN_ALL
+        self.__pyedb_use_grpc = None
```
SMoraisAnsys:
```suggestion
        self.__pyedb_use_grpc: bool = False
```

src/ansys/aedt/core/generic/settings.py
```diff
@@ -252,6 +254,18 @@ def _update_settings(self):
 
     # ########################## gRPC properties ##########################
 
+    @property
+    def pyedb_use_grpc(self):
```
SMoraisAnsys:
```suggestion
    def pyedb_use_grpc(self) -> bool:
```

src/ansys/aedt/core/generic/settings.py
```diff
@@ -252,6 +254,18 @@ def _update_settings(self):
...
+        """Flag for whether to use Pyedb with grpc or not.
+        Grpc is supported from AEDT 25R2.
+        The default is ``False``.
+        """
+        return self.__pyedb_use_grpc
+
+    @pyedb_use_grpc.setter
+    def pyedb_use_grpc(self, value):
```
SMoraisAnsys:
```suggestion
    def pyedb_use_grpc(self, value: bool):
```

tests/conftest.py
```diff
@@ -96,13 +97,16 @@
...
+USE_PYEDB_GRPC = config.get("use_pyedb_grpc", DEFAULT_CONFIG.get("use_pyedb_grpc"))
+
 os.environ["PYAEDT_SCRIPT_VERSION"] = DESKTOP_VERSION
 
 # ================================
 # PyAEDT settings
 # ================================
 
 settings.use_grpc_api = USE_GRPC
+settings.pyedb_use_grpc = USE_PYEDB_GRPC
```
SMoraisAnsys: Can't we use the pyaedt_settings.yaml file in tests/ instead ?

SMoraisAnsys (commented):
Thankfs for the changes, I left a few comments.
I would also update the documentation associated to pyaedt_settings.yaml file:
- doc/source/User_guide/settings.rst
- doc/source/Resources/pyaedt_settings.yaml
- tests/pyaedt_settings.yaml  (mentionned in a comment bu open to discussion)

tests/conftest.py
Samuelopez-ansys: I was thinking that it is not even needed in the conftest, because by default pyedb_use_grpc is None, and then PyAEDT checks if the AEDT version, if it is upper 2025R2 then uses grpc, I think it is better this approach.

Samuelopez-ansys: So , I deleted it from conftest. It is like the use_grpc_api, this setting is not part of the conftest, because PyAEDT internally decides which one to use.

doc/source/Resources/pyaedt_settings.yaml
```diff
@@ -128,6 +128,8 @@ general:
   block_figure_plot: false
   # Skip license check during Desktop initialization
   skip_license_check: true
+  # Enable or disable gRPC API for the EDB module
+  pyedb_use_grpc: null
```
SMoraisAnsys, line 133: Shouldn't that be false ?

doc/source/User_guide/settings.rst
```diff
@@ -157,3 +157,5 @@ Below is the content that can be updated through the YAML file.
         pyd_libraries_path: null
         # Path to the user defined PyD libraries. If set, it overrides the default pyd_libraries_path
         pyd_libraries_user_path: null
+        # Enable or disable gRPC API for the EDB module.
+        pyedb_use_grpc: null
```
SMoraisAnsys, line 162: Same comment

src/ansys/aedt/core/edb.py
```diff
@@ -107,19 +107,26 @@ def Edb(
     >>> app = Edb("/path/to/file/myfile.gds")
 
     """
-    # Use EDB legacy (default choice)
     from pyedb import Edb
 
+    if settings.pyedb_use_grpc is None and settings.aedt_version > "2025.2":
```
SMoraisAnsys: Why are you using None ?

Samuelopez-ansys: It is None, because the gRPC API is used if AEDT version is 2026 R1 or later, otherwise the legacy COM object is used.

So, interanlly PyAEDT decides which one to use by default:


    if settings.pyedb_use_grpc is None and settings.aedt_version > "2025.2":  # pragma: no cover
        settings.pyedb_use_grpc = True

    use_grpc = True if settings.pyedb_use_grpc and settings.aedt_version > "2025.2" else False 

If the user enables it explictly then it will bypass it (if the AEDT version is compatible).

src/ansys/aedt/core/generic/settings.py
```diff
@@ -252,6 +254,19 @@ def _update_settings(self):
...
+        """Flag for whether to use PyEDB with gRPC or not.
+        GRPC is supported from AEDT 25R2.
+        The default is ``False``, if AEDT version is lower than 2026 R1.
+        .
+        """
+        return self.__pyedb_use_grpc
+
+    @pyedb_use_grpc.setter
+    def pyedb_use_grpc(self, value: bool):
+        self.__pyedb_use_grpc = value
```
SMoraisAnsys:
```suggestion
    def pyedb_use_grpc(self) -> Optional[bool]:
        """Flag for whether to use PyEDB with gRPC or not.
        The default is ``None``.
        """
        return self.__pyedb_use_grpc

    @pyedb_use_grpc.setter
    def pyedb_use_grpc(self, value: bool):
        self.__pyedb_use_grpc = value
```

src/ansys/aedt/core/edb.py
```diff
@@ -107,19 +107,26 @@ def Edb(
     >>> app = Edb("/path/to/file/myfile.gds")
 
     """
-    # Use EDB legacy (default choice)
     from pyedb import Edb
 
+    if settings.pyedb_use_grpc is None and settings.aedt_version > "2025.2":  # pragma: no cover
+        settings.pyedb_use_grpc = True
```
SMoraisAnsys, line 129: Could we log / warn here so that users know about this change of behavior ? I mean the incoming log is fine but it would be great to let people know that this settings is defined automagically because no setting was provided and AEDT version is 2026.1 or higher.

src/ansys/aedt/core/generic/settings.py
```diff
@@ -111,6 +111,7 @@
     "use_local_example_data",
     "pyd_libraries_path",
     "pyd_libraries_user_path",
+    "pyedb_use_grpc",
 ]
 ALLOWED_GRPC_SETTINGS = ["grpc_secure_mode", "grpc_local", "grpc_listen_all"]
```
SMoraisAnsys: Shouldn't this setting join the list ALLOWED_GRPC_SETTINGS instead ?

SMoraisAnsys (approved):
LGTM, I left minor comments

PR #7209, opened 2026-02-06
CI: Remove codecov logic for unwanted workflows

SMoraisAnsys:
## Description
Removing codecov settings for PR as we no longer can leverage code coverage when working with testmon.

Here is an issue mentionning that [problem](https://github.com/tarpas/pytest-testmon/issues/86) and here is the reason / limit that testmon is facing [concurrent usage of settrace doesn't work in Python ecosystem · Issue #97 · tarpas/pytest-testmon](https://github.com/tarpas/pytest-testmon/issues/97).

The code coverage is kept for the nightly build though. That way we would should be able to keep the codecov badge in the README.

## Issue linked
None

## Checklist

### DIFF
--- a/.ci/pyproject.toml
+++ b/.ci/pyproject.toml
@@ -79,7 +79,7 @@ dotnet = [
     "pywin32>=303; platform_system=='Windows'",
 ]
 doc = [
-    "ansys-sphinx-theme>=1.0.0,<1.7",
+    "ansys-sphinx-theme>=1.0.0,<1.8",
     "jupyter",
     "numpydoc==1.10.0",
     "recommonmark",
--- a/.ci/uv.lock
+++ b/.ci/uv.lock
@@ -200,8 +200,7 @@ source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "ansys-api-edb" },
     { name = "ansys-tools-common" },
-    { name = "django", version = "5.2.10", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.12'" },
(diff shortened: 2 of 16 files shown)


Samuelopez-ansys (dismissed):

MaxJPRey:
Could this lead to a decrease of the quality as the PR don't have the indicator anymore?
It is also true that sometimes the data shared were not accurate -false positive-.

SMoraisAnsys:
> Could this lead to a decrease of the quality as the PR don't have the indicator anymore? It is also true that sometimes the data shared were not accurate -false positive-.

Indeed but we can't have both. We'll have to be careful in evaluating that the modifications and added tests. Note that we could teach people to run their changes locally by checking which set of tests were run by testmon and double checking coverage.
Also, we have the nightly runs that will let us know if things go wrong.

eblanco-ansys:
We should probably migrate this part into a separate section, now it's a bit hidden:
https://aedt.docs.pyansys.com/version/stable/Getting_started/Contributing.html#run-tests-in-vscode-and-pycharm
We can even keep it there and create a new section for Ansys PyAEDT devs. We could add some recommendations for testing coverage locally before submitting a PR in the new section

SMoraisAnsys:
> We should probably migrate this part into a separate section, now it's a bit hidden: https://aedt.docs.pyansys.com/version/stable/Getting_started/Contributing.html#run-tests-in-vscode-and-pycharm We can even keep it there and create a new section for Ansys PyAEDT devs. We could add some recommendations for testing coverage locally before submitting a PR in the new section

We discussed about that previously and having such a document in the doc (internal) would be great :)
Some existing part of the documentation might be migrated into it too

PR #7248, opened 2026-02-12
FIX: Fixed excitations to support the A-Phi solvers

### LINKED ISSUE #7121: fix assign_current for APhi
There are new args for APhi:

oModule.AssignCurrent(
	[
		"NAME:Current1",
		"Faces:="		, [7],
		"Phase:="		, "2deg",
		"Current:="		, "1mA",
		"CurrentExcitationModel:=", "Single Potential",
		"Point out of terminal:=", False
	])

CurrentExcitationModel is not implemented.

@nunesanderson @jvela018

nunesanderson:
## Description
As the title says.

## Issue linked
close #7121 

## Checklist

### DIFF
--- a/src/ansys/aedt/core/visualization/post/field_data.py
+++ b/src/ansys/aedt/core/visualization/post/field_data.py
@@ -22,6 +22,8 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
+from __future__ import annotations
+
 from abc import abstractmethod
 from collections import defaultdict
 import csv
@@ -1268,7 +1270,7 @@ class FieldPlot(PyAedtBase):
     @pyaedt_function_handler()
     def get_points_value(
         self, points: list | dict, filename: str | None = None, visibility: bool = False
-    ) -> dict | pd.DataFrame:  # pragma: no cover
+    ) -> "dict | pd.DataFrame":  # pragma: no cover
         """
         Get points data from field plot.
 


tests/system/general/test_maxwell_3d.py
```diff
@@ -78,6 +79,13 @@ def layout_comp(add_app_example):
     app.close_project(app.project_name, save=False)
 
 
+@pytest.fixture
+def m3d_app_maxwell_bars(add_app_example):
```
gmalinve: @nunesanderson do you think we can create the model from scratch in each test? I checked and it's a very simple design with two boxes and a region. 
It would be better to do it as it's done in the other tests.

tests/system/general/test_maxwell_3d.py
```diff
@@ -1221,3 +1229,59 @@ def test_assign_sink(m3d_app, maxwell_versioned) -> None:
...
+        assignment=m3d_app_maxwell_bars.modeler.object_list[1].faces[0], amplitude=2, name="voltage1"
+    )
+    voltage2 = m3d_app_maxwell_bars.assign_voltage(
+        assignment=m3d_app_maxwell_bars.modeler.object_list[1].faces[4],
+        amplitude=3,
+        name="voltage2",
+        excitation_model="Double Potentials with Ground",
+        swap_direction=True,
+    )
+    assert m3d_app_maxwell_bars.boundaries[0].name == "current1"
```
gmalinve: @nunesanderson minor comment: instead of looking for the object in boundaries, you could simply rely on the returned objects current1, voltage1 and voltage2 and assert like: current1.name == "current1" etc. (as you are doing already in the props asserts)

src/ansys/aedt/core/maxwell.py
gmalinve, line 775: @nunesanderson I just noticed that this method has 2 arguments in the docstring that are not used in the logic.
One being "initial_current" and the other "has_initial_current". Also there is a comment in the middle: "# TO CHECK FOR AC MAGNETIC A-PHI". Could you please remove them? thanks :)

src/ansys/aedt/core/maxwell.py
```diff
@@ -1081,7 +1099,17 @@ def assign_rotate_motion(
...
+    def assign_voltage(
+        self,
+        assignment: list,
+        amplitude: float | None = 1,
+        name: str | None = None,
+        excitation_model: str = "Single Potential",
+        initial_current: float | None = 0,
+        swap_direction: bool | None = False,
+        phase: float | None = 0,
+        has_initial_current: bool | None = False,
```
gmalinve, line 1111: @nunesanderson these arguments are missing from the docstring. Coul you please add them? thanks! :)

tests/system/general/test_maxwell_3d.py
nunesanderson: Yes, we could. But I thought it would make more sense to reuse it.

tests/system/general/test_maxwell_3d.py
nunesanderson: As far as I understood, the current1... access the data set by the user and the  m3d_app_maxwell_bars.boundaries[1] gets it from aedt, so I decided to test both ways to make sure.

tests/system/general/test_maxwell_3d.py
Samuelopez-ansys: I agree with Giulia. Opening a project introduces an extra delay and also introduces random read/write random failures

tests/system/general/test_maxwell_3d.py
Samuelopez-ansys: Both ways is ok

gmalinve: @nunesanderson both ways are correct, I just said it was a minor comment :)

src/ansys/aedt/core/maxwell.py
```diff
@@ -798,6 +799,15 @@ def assign_current(
         name : str, optional
             Name of the current excitation.
             The default is ``None`` in which case a generic name will be given.
+        excitation_model : str, optional
```
Samuelopez-ansys, line 806: @nunesanderson In the docstring swap_direction and phase are duplicated

Samuelopez-ansys:
Moving to draft until issues are not fixed

PR #7264, opened 2026-02-15
FEAT: Improve consistency

### LINKED ISSUE #7257: Improve consistency of get_object...

The following methods seem redundant and probably could be combined:
``` python
GeometryModeler.get_objects_w_string()
GeometryModeler.get_object_from_name()
```
I propose a single method:
``` python
GeometryModeler.get_objects_by_name()
```
using "by" instead of "from" is consistent with ``get_objects_by_material`` and the redundant methods would be consolidated.

### Code sample expected

``` python
modeler.get_objects_by_name("Mag")
```
returns a list of all objects whose name contains the string "Mag".

If only one specific object is needed this is very easy and fast using list comprehension.

_No response_

Devin-Crawford:
## Description
Add method ``modeler.get_objects_by_name`` to be consistent with ``modeler.get_objects_by_material``

## Issue linked
Close #7257 

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ exclude: |
 repos:
 
 - repo: https://github.com/astral-sh/ruff-pre-commit
-  rev: 4924b0e01e032fea073ad04a1c5cfa7e4add0afb  # frozen: v0.15.6
+  rev: e05c5c0818279e5ac248ac9e954431ba58865e61  # frozen: v0.15.7
   hooks:
     - id: ruff-check
     - id: ruff-format
@@ -53,7 +53,7 @@ repos:
     - id: check-pre-commit-ci-config
 
 - repo: https://github.com/ansys/pre-commit-hooks
-  rev: fd1d6b0b4baeb08e32d4de0909999576695c0433  # frozen: v0.5.2
+  rev: 680b297da7584a07eb7c92b6b4e9baf88ece3e0e  # frozen: v0.6.0
   hooks:
     - id: add-license-headers
       exclude: |
--- a/doc/changelog.d/7433.maintenance.md
+++ b/doc/changelog.d/7433.maintenance.md
@@ -1,0 +1,1 @@
+Pre-commit automatic update


src/ansys/aedt/core/modeler/cad/primitives.py
```diff
@@ -7463,23 +7463,28 @@ def get_obj_id(self, assignment):
...
-    def get_object_from_name(self, assignment):
-        """Return the object from an object name.
+    def get_objects_by_name(self, assignment, case_sensitive: bool = True):
+        """Return the objects given a search string.
 
         Parameters
         ----------
         assignment : str
-            Name of the object.
+            String used to filter by object names..
```
MaxJPRey:
```suggestion
            String used to filter by object names.
```

src/ansys/aedt/core/modeler/cad/primitives.py
```diff
@@ -7463,23 +7463,28 @@ def get_obj_id(self, assignment):
...
+        case_sensitive : bool, optional
+            Whether the string is case-sensitive. The default is ``True``.
 
         Returns
         -------
-        :class:`ansys.aedt.core.modeler.cad.object_3d.Object3d`
-            3D object returned.
+        list of class:`ansys.aedt.core.modeler.cad.object_3d.Object3d`
+            Returns a list of objects whose names contain the
+            search string..
```
MaxJPRey:
```suggestion
            search string.
```

src/ansys/aedt/core/modeler/cad/primitives.py
```diff
@@ -7462,6 +7462,30 @@ def get_obj_id(self, assignment):
...
+        list of class:`ansys.aedt.core.modeler.cad.object_3d.Object3d`
+            Returns a list of objects whose names contain the
+            search string.
+
+        """
+        if case_sensitive:
+            return [o for name, o in self.objects_by_name.items() if assignment in name]
+        else:
+            return [o for name, o in self.objects_by_name.items() if assignment.lower() in name.lower()]
+        return None
```
SMoraisAnsys:
```suggestion
```
This should never happen given the use of if: ... else: ...

SMoraisAnsys:
Any thoughts on my comment @Devin-Crawford @Samuelopez-ansys @MaxJPRey (you reviewed, blame yourself for the ping :p)

src/ansys/aedt/core/modules/material.py
```diff
@@ -452,6 +454,24 @@ def value(self, val) -> None:
...
+                with open(val) as f:
+                    reader = csv.reader(f, delimiter="\t")
+                    next(reader)
+                    for row in reader:
+                        if len(row) != 2:
+                            raise ValueError(f"Line {lineno}: expected 2 columns, got {len(parts)}")
+                        if not (row[0].isnumeric() and row[1].isnumeric()):
+                            raise ValueError(f"Line {lineno}: expected numeric values, got {row[0]} and {row[1]}")
+                        datalist.append([float(row[0]), float(row[1])])
+                    self.value = datalist
```
Samuelopez-ansys: I would check if datalist is an empty list, and in that case, raise an Exception.

We could also use: from ansys.aedt.core.generic.file_utils import read_csv

And why are you assuming the file has a header? I prefer assume that the file is just 2 columns, and we could add a new parameter in read_csv to specify the parameter (,  or /t ....)

Samuelopez-ansys: @Devin-Crawford

src/ansys/aedt/core/application/design.py
```diff
@@ -2828,37 +2828,41 @@ def import_dataset1d(
...
+
+            if name[0] == "$":
+                name = name[1:]
+                is_project_dataset = True
+
+            return self.create_dataset(
+                name, xlist, ylist, is_project_dataset=is_project_dataset, x_unit=units[0], y_unit=units[1], sort=sort
+            )
+        else:
+            self.logger.error("Input file does not exist.")
```
Samuelopez-ansys: I think it should raise an Exception

Samuelopez-ansys:
> Should we rework the API to work differently ? Like having a single method to do both approaches ? We could allow different kind of matching mechanism, e.g. (exact, contains, start/endswith, regex), whille keeping the pattern argument (meaning varying depending on the matching mechanism) and en/disabling case sensitive.
> 
> Open to discussion

What do you mean @SMoraisAnsys ?

get_objects_by_name and get_object_from_name?

I would keep get_object_from_name which already exist, and I would add a new argument to specify if you want a perfect match (like now) or partial (with Sebastien's ideas).

src/ansys/aedt/core/application/design.py
Devin-Crawford: I agree.

```suggestion
            error_message = f"Input file '{in_file}' does not exist."
            self.logger.error(error_message)
            raise FileNotFoundError(error_message)
```

src/ansys/aedt/core/modules/material.py
Devin-Crawford: Throwing away the header matches the default AEDT behavior when you import the B-H curve from a file. It is very specific to material property assignment when the material is defined as x,y values in a text file.

Samuelopez-ansys: But what if the user has its own file, and this file does not contain headers? Or if the separatos are commas?

I think it is better to create a method like "import_value_from_file"  that could have as argument the number of headers that it has, the delimeter type... and then internally just do: self.value = ....

Devin-Crawford:
> > Should we rework the API to work differently ? Like having a single method to do both approaches ? We could allow different kind of matching mechanism, e.g. (exact, contains, start/endswith, regex), whille keeping the pattern argument (meaning varying depending on the matching mechanism) and en/disabling case sensitive.
> > Open to discussion
> 
> What do you mean @SMoraisAnsys ?
> 
> get_objects_by_name and get_object_from_name?
> 
> I would keep get_object_from_name which already exist, and I would add a new argument to specify if you want a perfect match (like now) or partial (with Sebastien's ideas).

- ``get_objects_by_material()`` returns a list of objects.
- ``get_object_from_name()`` returns a single object and the name must be specified exactly.
- ``get_objects_by_name()`` is new and is consistent with the first method above. There 
  are two inconsistencies: "from" vs "by". It doesn't matter which preposition 
  is used but they should
  be consistent. ``get_object_from_name()`` wouldn't be necessary if we 
  have ``get_objects_by_name()``. We could add a named argument ``exact_match`` to return only
  a single object and then we would have ``get_object_from_name()`` functionality.

Samuelopez-ansys:
Hi @Devin-Crawford , could you please add a test for the new material case, where you pass a tab file?

Samuelopez-ansys:
@SMoraisAnsys @Devin-Crawford I added with fnmatch what you suggested, now get_objects_by_name is very complete and accept the wildcard pattern *, like other Ansys tools:

        # Exact match
        >>> objs = modeler.get_objects_by_name("Patch_1")

        # All objects whose name starts with "Substrate"
        >>> objs = modeler.get_objects_by_name("Substrate*")

        # All objects whose name ends with "_gnd"
        >>> objs = modeler.get_objects_by_name("*_gnd")

        # All objects whose name contains "patch"
        >>> objs = modeler.get_objects_by_name("*patch*", case_sensitive=False)

        # Mid-string wildcard: names like "Sub_1", "Sub_gnd_1".
        >>> objs = modeler.get_objects_by_name("Sub*_1")

Samuelopez-ansys:
Before merging, @SMoraisAnsys could you please make a last review?

src/ansys/aedt/core/application/design.py
```diff
@@ -2854,37 +2854,41 @@ def import_dataset1d(
...
+            if name[0] == "$":
+                name = name[1:]
+                is_project_dataset = True
+
+            return self.create_dataset(
+                name, xlist, ylist, is_project_dataset=is_project_dataset, x_unit=units[0], y_unit=units[1], sort=sort
+            )
+        else:
+            error_message = f"Input file '{in_file}' does not exist."
+            raise FileNotFoundError(error_message)
```
SMoraisAnsys:
```suggestion
            raise FileNotFoundError(f"Input file '{in_file}' does not exist.")
```

src/ansys/aedt/core/modeler/cad/primitives.py
```diff
@@ -4496,7 +4497,7 @@ def create_faceted_bondwire_from_true_surface(
         str
             Name of the bondwire created.
         """
-        old_bondwire = self.get_object_from_name(assignment)
+        old_bondwire = self.get_objects_by_name(assignment)[0]
         if not old_bondwire:
             return False
```
SMoraisAnsys:
```suggestion
        bojects = self.get_objects_by_name(assignment)
        if objects:
            old_bondwire = objects[0]
        else:
            raise ValueError(f"Assignment {assignment} does not match an object.")
```
If the returned list is empty, the code will return an exception.

src/ansys/aedt/core/modeler/cad/primitives.py
```diff
@@ -7601,7 +7602,57 @@ def get_obj_id(self, assignment: str) -> int | None:
...
-    def get_object_from_name(self, assignment: str) -> Object3d:
+    def get_objects_by_name(
+        self,
+        assignment: str,
+        case_sensitive: bool = True,
+    ) -> list[Object3d]:
+        """Return the objects whose names match a wildcard pattern.
+
+        The ``*`` character acts as a wildcard that matches any sequence of
+        characters (including none).  The matching mode is inferred
```
SMoraisAnsys:
```suggestion
        characters (including none). The matching mode is inferred
```

src/ansys/aedt/core/modeler/cad/primitives.py
```diff
@@ -7616,8 +7667,8 @@ def get_object_from_name(self, assignment: str) -> Object3d:
 
         """
         if assignment in self.object_names:
-            # object_id = self.get_obj_id(objname)
             return self.objects[assignment]
+        return None
```
SMoraisAnsys: Not sure if this is required as, by default, every python function return None

src/ansys/aedt/core/modules/material.py
```diff
@@ -452,6 +454,22 @@ def value(self, val: list[float] | float) -> None:
             if len(val) == 4:
                 self._property_value[0].value = val
                 self._material._update_props(self.name, val, update_aedt=self._material._material_update)
+        elif isinstance(val, Path):
```
SMoraisAnsys: Maybe adding a check on the kind of file here ? Just to ensure that we call `csv.reader()` on a CSV file ?

Samuelopez-ansys:
@SMoraisAnsys @eblanco-ansys Thank you for your reviews, please this PR is ready for the final approval

PR #7272, opened 2026-02-16
FEAT: Panel button groups

eblanco-ansys:
## Description
- Added button groups
- Moved jupyter inside the group
- Added class for easily manage xml buttons/groups

image

## Checklist

### DIFF
--- a/src/ansys/aedt/core/extensions/installer/cli_setup.py
+++ b/src/ansys/aedt/core/extensions/installer/cli_setup.py
@@ -38,8 +38,7 @@ version = os.environ.get("PYAEDT_DESKTOP_VERSION", None)
 print("Loading the PyAEDT CLI.")
 
 try:
-    from ansys.aedt.core import * # noqa: F401
-    import ansys.aedt.core  # noqa: F401
+    from ansys.aedt.core import *  # noqa: F401
     from ansys.aedt.core import Desktop
     from ansys.aedt.core.generic.general_methods import active_sessions
     from ansys.aedt.core.generic.general_methods import is_windows
@@ -53,7 +52,6 @@ except ImportError:
         sys.path.append(str(console_setup_dir / ".." / ".." / ".."))
 
     from ansys.aedt.core import *  # noqa: F401
-    import ansys.aedt.core  # noqa: F401
     from ansys.aedt.core import Desktop
     from ansys.aedt.core.generic.general_methods import active_sessions
     from ansys.aedt.core.generic.general_methods import is_windows


src/ansys/aedt/core/extensions/installer/cli_setup.py
```diff
@@ -0,0 +1,198 @@
...
+
+"""
+Launches an interactive shell with an instance of HFSS.
+
+Omitting the .py in the name of this file hides it from the Electronics Desktop Toolkit menu.
+It should be hidden from this menu because the scripts in that menu are meant to be executed using IronPython
+cpython_console.py should be run instead of this script.
+
+This file can also serve as a template to modify PyAEDT scripts to take advantage of the command line arguments
+provided by the launcher
```
SMoraisAnsys: I'm not sure to understand the description here. Why are you mentionning the removal of ".py" and the fact that it should be hidden ?

src/ansys/aedt/core/extensions/installer/cli_setup.py
```diff
@@ -0,0 +1,198 @@
...
+"""
+
+import os
+import atexit
+from pathlib import Path
+import sys
+from IPython import get_ipython
+import tempfile
+
+aedt_process_id = int(os.environ.get("PYAEDT_PROCESS_ID", None)) # pragma: no cover
```
SMoraisAnsys: Instead of adding `pragma: no cover` everywhere, can you ignore the file from the coverage ? Also, why isn't it tested ?

src/ansys/aedt/core/extensions/installer/cli_setup.py
eblanco-ansys: To be honest, I'm not sure of what it means. I copy pasted `console_setup.py` file and modified it. It has the same header

src/ansys/aedt/core/extensions/installer/cli_setup.py
eblanco-ansys: I tried to cover them some months ago, but it wasn't easy to test by the way the script is made. If you want I can work on that or exclude it directly as you suggested

SMoraisAnsys: Ideally I would like for it to be tested but I know it's though to do something with how we interact with some files. Open to opinions

src/ansys/aedt/core/extensions/installer/cli_setup.py
Samuelopez-ansys: Please @eblanco-ansys delete this description.

src/ansys/aedt/core/extensions/installer/cli_setup.py
eblanco-ansys: From now I excluded them from coverage

src/ansys/aedt/core/cli/panels.py
```diff
@@ -51,6 +52,12 @@ def add_panels(
         "--skip-version-manager",
         help="Skip installing the Version Manager tab",
     ),
+    reboot: bool = typer.Option(
+        False,
+        "--reboot",
```
eblanco-ansys: LGTM, but IMO we should use `--reset` rather than `--reboot`. It sounds better to me. When restoring the panels, we always refer to it as "resetting" them

Samuelopez-ansys (approved):

PR #7298, opened 2026-02-21
FIX: dxf fixes on mcad & ecad

gkorompi:
1. get_dxf_layers updated to support older versions of .dxf
2. empty xml path caused anstranslator to faile in RedHat 9 when calling ImportCad
3. no need to have a layers list as an input requirement for import_dxf when we call get_dxf_layers inside of it

2 different customers face issues with these functions.

### DIFF
--- a/src/ansys/aedt/core/application/analysis_3d.py
+++ b/src/ansys/aedt/core/application/analysis_3d.py
@@ -1299,7 +1299,7 @@ class FieldAnalysis3D(Analysis, PyAedtBase):
         else:
             for layer in layers:
                 if layer not in dxf_layers:
-                    self.logger.error(f"{layer} does not exist in specified dxf.")
+                    self.logger.error(f"{layer} does not exist in specified DXF file.")
                     return False
 
         if hasattr(self, "is3d") and self.is3d:


src/ansys/aedt/core/generic/file_utils.py
```diff
@@ -821,6 +821,11 @@ def find_indices(list_to_check, item_to_find):
         for idx in indices:
             if "2" in lines[idx + index_offset]:
                 layer_names.append(lines[idx + index_offset + 1].replace("\n", ""))
+        if not layer_names:
```
eblanco-ansys, line 824: Please, add some tests to cover the new lines you implemented

src/ansys/aedt/core/application/analysis_3d.py
```diff
@@ -1247,9 +1247,9 @@ def import_dxf(
         ----------
         input_file : str or :class:`pathlib.Path`
             Path to the DXF file.
-        layers : list
-            List of layer names to import. To get the dxf_layers in the DXF file,
-            you can call the ``get_dxf_layers`` method.
+        layers : list, optional
+            List of layer names to import, if empty all layers will be imported.
```
SMoraisAnsys:
```suggestion
            List of layer names to import, if ``None`` or empty list all layers will be imported.
```

src/ansys/aedt/core/application/analysis_3d.py
```diff
@@ -1247,9 +1247,9 @@ def import_dxf(
...
+            To get the dxf_layers in the DXF file, you can call the ``get_dxf_layers`` method.
```
SMoraisAnsys:
```suggestion
            To get the layers in the DXF file, you can call the ``get_dxf_layers`` method.
```

src/ansys/aedt/core/application/analysis_3d.py
```diff
@@ -1294,10 +1294,13 @@ def import_dxf(
...
-        for layer in layers:
-            if layer not in dxf_layers:
-                self.logger.error(f"{layer} does not exist in specified dxf.")
-                return False
+        if not layers:
+            layers = dxf_layers
+        else:
+            for layer in layers:
+                if layer not in dxf_layers:
+                    self.logger.error(f"{layer} does not exist in specified dxf.")
```
SMoraisAnsys:
```suggestion
                    self.logger.error(f"{layer} does not exist in specified DXF file.")
```

src/ansys/aedt/core/generic/file_utils.py
```diff
@@ -821,6 +821,11 @@ def find_indices(list_to_check, item_to_find):
...
+            index_offset = 1
+            for idx in indices:
+                if "2" in lines[idx + index_offset]:
+                    layer_names.append(lines[idx + index_offset + 1].replace("\n", ""))
```
SMoraisAnsys, line 828: Could you explain the logic behind this code ?

src/ansys/aedt/core/hfss3dlayout.py
```diff
@@ -1433,7 +1433,7 @@ def _import_cad(
             aedb_path = aedb_path.replace(old_name, project_name)
             self.logger.warning("aedb_exists. Renaming it to %s", project_name)
         if xml_path is None:
-            xml_path = Path("").name
+            xml_path = str(Path(cad_path).with_suffix(".xml"))
```
SMoraisAnsys, line 1436: Why are you casting into `str` when the previous code was using `Path` ? That's also the case for the `elif` a line below

SMoraisAnsys (changes requested):
Thanks for the contribution @gkorompi Can you have a look into my comments ?

Samuelopez-ansys:
Hi @gkorompi what is the status of this PR? I will move it to draft until you fix the suggestions

PR #7303, opened 2026-02-23
FIX: add EmitNode parent

jsalant22:
Add a couple convenience functions to EmitNode to get the node's fullname and it's parent node's name.

## Description

## Issue linked

## Checklist

### DIFF
--- a/src/ansys/aedt/core/emit_core/nodes/emit_node.py
+++ b/src/ansys/aedt/core/emit_core/nodes/emit_node.py
@@ -110,7 +110,7 @@ class EmitNode:
 
     @name.setter
     @min_aedt_version("2025.2")
-    def name(self, requested_name: str):
+    def name(self, requested_name: str) -> None:
         """Renames the node/component.
 
         Parameters
--- a/doc/changelog.d/7326.miscellaneous.md
+++ b/doc/changelog.d/7326.miscellaneous.md
@@ -1,0 +1,1 @@
+Continue PyAEDT Typing
--- a/doc/source/Getting_started/Contributing.rst
+++ b/doc/source/Getting_started/Contributing.rst
@@ -273,7 +273,7 @@ and contain at least two classes:
 
 
     class MyExtension(ExtensionCommon):
-        def __init__(self, *args, **kwargs):
+        def __init__(self, *args, **kwargs) -> None:
             super().__init__(*args, **kwargs)
             self.add_extension_content()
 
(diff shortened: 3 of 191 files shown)


src/ansys/aedt/core/emit_core/nodes/emit_node.py
```diff
@@ -150,6 +150,29 @@ def _parent(self):
...
+
+        Returns
+        -------
+        Str
+            Full node name of this node's parent node.
+        """
+        return self._get_property("Parent", True)
+
+    @property
+    def _full_node_name(self):
```
SMoraisAnsys, line 178: are those related to a specific version of AEDT ? If so, please use the decorator for minimal version.

jsalant22: @SMoraisAnsys The entire EmitNode class (and subclasses) requires 252 or newer. I don't think we have any decorators for the classes specifically, but the functions in emit_core/results/Revision that are used to retrieve these nodes do version checking.

Samuelopez-ansys: @jsalant22 I think you can use the decorator min_aedt_version

It is used in some places in pyaedt, and in this way we can avoid that a user trying a previous EMIT version

jsalant22: @Samuelopez-ansys @SMoraisAnsys Added decorators

jsalant22:
@Samuelopez-ansys @SMoraisAnsys Anything else for this PR?

SMoraisAnsys:
> @Samuelopez-ansys @SMoraisAnsys Anything else for this PR?

There is a huge conflict it seems

Samuelopez-ansys:
@jsalant22 I have fixed all conflict, the root cuase was that you added -> None to all setters, and the setter should not return anything.

@SMoraisAnsys Do you agree adding the version decorator to all properties? Or maybe we could add one single call at __init__ class level?

Samuelopez-ansys:
Any comment @SMoraisAnsys ?

jsalant22:
@SMoraisAnsys @Samuelopez-ansys FWIW, I had Copilot analyze the version decorator and he (she?) didn't think the implementation would handle it at a class level and recommended adding them for each method.

Samuelopez-ansys:
@jsalant22 Sometimes copilot does not give the most efficient answer. I am waiting @SMoraisAnsys to see his feedback regarding all these decorators

PR #7366, opened 2026-03-10
FEAT: add new feature to coil extension

gmalinve:
## Description
In this PR I added a new functionality for flat coils in order to extend the end connection.
This is quite useful when assigning excitations tha must have a start and stop assignment on the region.

## Issue linked

## Checklist


src/ansys/aedt/core/modeler/advanced_cad/coil.py
```diff
@@ -374,7 +374,29 @@ def create_vertical_path(self) -> list:
         return polyline
 
     @pyaedt_function_handler()
-    def create_sweep_profile(self, start_point: list[float], polyline: "Object3d") -> str:
-        profile = self._app.modeler.create_circle("YZ", start_point, "wire_radius", num_sides="section_segmentation")
+    def create_sweep_profile(self, polyline: "Object3d") -> str:
```
SMoraisAnsys, line 377: Can you add docstrings for that method ?

SMoraisAnsys, line 377: Can you add docstrings for that method ?

gmalinve:
@eblanco-ansys 
image

Samuelopez-ansys:
@gmalinve which commands did you use to get the coverage? I think it is useful for the community

src/ansys/aedt/core/extensions/maxwell3d/create_coil.py
```diff
@@ -82,7 +84,8 @@
 class CoilExtensionData(ExtensionCommonData):
     """Data class containing user input and computed data."""
 
-    is_vertical: bool = EXTENSION_DEFAULT_ARGUMENTS["is_vertical"]
+    is_vertical: bool = field(default_factory=lambda: EXTENSION_DEFAULT_ARGUMENTS["is_vertical"])
+    create_3d_comp: bool = field(default_factory=lambda: EXTENSION_DEFAULT_ARGUMENTS["create_3d_comp"])
```
SMoraisAnsys, line 88: I'm not sure about the need to use a field here. The idea is to specificly leverage a static value defined in EXTENSION_DEFAULT_ARGUMENTS. No one should be updating that dict so no reason to use a field imo.

src/ansys/aedt/core/extensions/maxwell3d/create_coil.py
```diff
@@ -250,46 +269,24 @@ def main(data: CoilExtensionData) -> bool:
...
-        name=data.name,
-        variables_to_include=list(params),
-        assignment=[profile_name],
-    )
+    profile_name = coil.create_sweep_profile(polyline)
+
+    if data.create_3d_comp:
+        # Create and replace 3D Component
+        comp_path = Path(aedtapp.working_directory, data.name + ".a3dcomp")
+        params = [f.name for f in fields(data)] & aedtapp.variable_manager.design_variables.keys()
```
SMoraisAnsys:
```suggestion
        params = {f.name for f in fields(data)} & aedtapp.variable_manager.design_variables.keys()
```
I think this is incorrect. Operator `&` should be working with set and dict keys but not with list and dict keys.

src/ansys/aedt/core/extensions/maxwell3d/create_coil.py
```diff
@@ -250,46 +269,24 @@ def main(data: CoilExtensionData) -> bool:
...
+        aedtapp.modeler.create_3dcomponent(
+            input_file=str(comp_path),
+            variables_to_include=list(params),
+            # name=data.name,
```
SMoraisAnsys:
```suggestion
```

src/ansys/aedt/core/extensions/maxwell3d/create_coil.py
gmalinve: @SMoraisAnsys PyCharm flags as errors these bool default values. By reading some posts online I discovered that many type checkers treat dictionary indexing as a potentially mutable default and that the correct fix is to use a default_factory.

SMoraisAnsys: That's right, the fact that we are using CAPS for the name is to let people know that this shouldn't be modified (but we do not garantee it). That's making me considering the fact of using `MappingProxyType` recursively to ensure that we have a dict like structure that is frozen (or we could also created a frozen data class).

Anyway, thanks for the feedback :) Pinging @ecoussoux-ansys for future discussions.

PR #7378, opened 2026-03-12
FIX:  Incorrect field summary behavior with dependent variables

lorenzovecchietti:
The automatic generation of the variation string in the field summary creation was failing to include all dependent variables. As a result, the csv export was failing.


Samuelopez-ansys:
Is it possible to have in the example a dependent variable? Just to avoid this issue in the future. Let me know if it makes sense.

we are trying to fix ‘mistakes’ of the past, and if we find a better way to test a method or something missed, we are trying to add it

lorenzovecchietti:
> Is it possible to have in the example a dependent variable? Just to avoid this issue in the future. Let me know if it makes sense.
> 
> we are trying to fix ‘mistakes’ of the past, and if we find a better way to test a method or something missed, we are trying to add it

@Samuelopez-ansys Fair ask :)   Done.

PR #7411, opened 2026-03-17
REFACTOR: lazy load libraries

Samuelopez-ansys:
## Description
This pull request refactors how the `pandas` library is imported and used across several modules in the codebase. Instead of importing `pandas` at the top level and issuing warnings on import failure, the code now uses a lazy import pattern: `pandas` is only imported when needed, and clear error messages are raised if it is not installed. This improves startup performance, avoids unnecessary warnings, and makes the codebase cleaner and more robust. Several function signatures and internal checks have also been updated to accommodate this change.

Key changes include:

**Lazy Import Pattern for pandas**

* Replaced top-level `pandas` imports and warnings with a `_import_pandas()` helper function in all affected modules. This function only imports `pandas` when needed and raises a clear `ImportError` if not installed. (`src/ansys/aedt/core/modules/profile.py`, `src/ansys/aedt/core/visualization/post/field_data.py`, `src/ansys/aedt/core/visualization/post/field_summary.py`, `src/ansys/aedt/core/visualization/post/post_circuit.py`, `src/ansys/aedt/core/visualization/post/solution_data.py`)

* Updated all usages of `pd.DataFrame`, `pd.Series`, and similar to use the locally imported `pd_module` from `_import_pandas()`.

**API and Type Hint Updates**

* Changed function signatures to use string type annotations for pandas types (e.g., `"pd.DataFrame"`, `"list | Series"`) to avoid direct references to `pandas` at import time

* Improved handling of pandas-enabled outputs: if pandas is not available, the code now raises clear errors or falls back to returning lists/dicts, with logging as appropriate

**Configuration and Initialization**

* Updated logic for enabling pandas output in `SolutionData` to check for pandas availability dynamically, and to log a warning if enabling is attempted without pandas installed.

These changes make the codebase more robust to missing optional dependencies and improve user experience by providing clear, actionable error messages.

## Issue linked

## Checklist

### DIFF
--- a/.github/workflows/ci-release.yml
+++ b/.github/workflows/ci-release.yml
@@ -133,6 +133,63 @@ jobs:
           ${ACTIVATE_VENV}
           python -c "import ansys.aedt.core; from ansys.aedt.core import __version__"
 
+  build-wheelhouse-rocky:
+    name: Build wheelhouse (Rocky Linux)
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ['3.10', '3.11', '3.12', '3.13']
+    permissions:
+      contents: read   # Required to read repository content for building
+    container:
+      image: rockylinux/rockylinux:8@sha256:f5529992e67440c1a4ae7788244d4381c6909159a88eacd95b7523ae47ced82e
+    steps:
+      - name: Install build deps
+        shell: bash
+        run: |
+          set -eux
+          dnf -y update
+          dnf groupinstall -y "Development Tools"
+          dnf -y install \
+            gcc make wget tar xz bzip2 \
+            bzip2-devel zlib-devel xz-devel \
+            openssl-devel libffi-devel sqlite-devel \
+            readline-devel tk-devel \
+            libuuid-devel expat-devel \
+            pkgconf-pkg-config ca-certificates
+          update-ca-trust
+
+      - name: Set up uv
+        uses: astral-sh/setup-uv@eac588ad8def6316056a12d4907a9d4d84ff7a3b # v7.3.0
+        with:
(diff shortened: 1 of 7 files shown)


Samuelopez-ansys:
All the warnings could be misleading, and pandas should not be needed at Desktop level, so this PR is lazy loading pandas.

image

src/ansys/aedt/core/visualization/post/field_data.py
```diff
@@ -45,17 +45,30 @@
...
-    warnings.warn(
-        "The Pandas module is required to run functionalities of ansys.aedt.core.visualization.post.field_data.\n"
-        "Install with \n"
-        """>> pip install pandas"""
-    )
+else:
     pd = None
 
 
+def _import_pandas():
```
SMoraisAnsys: I'm not really a fan of having this function around. Can't we simply add the import in the method and not adding the warning at all ? This would simplify things and shouldn't be something that users can't understand.

If adding this kind of message is a wanted behavior, could we create a decorator on the methods afflicted by this kind of behavior (not only for `pandas`) and apply it everywhere ? Something like `@requires(package_name)`. Note that this could have some side effects as we could end up catching any ImportError and not only those directly associated to the package imported.

Samuelopez-ansys: Hi @SMoraisAnsys ,

It is not impacting performance (not much), but as you can see in my previous comment, all the warnings were misleading, and some users were thinking that PYAEDT was not working.

Let's then delete the warning, as you said

Samuelopez-ansys: Hi @SMoraisAnsys , please could you review it again? Now, it is just importing pandas wherever is needed.

PR #996, opened 2022-03-25
Fix examples

maxcapodi78:

PR #7453, opened 2026-03-26
FIX: Minor updates to comply with latest ``pyvista`` versions

### LINKED ISSUE #7451: Bug located in ``populate_pyvista_object()`` method of ``vizualization/plot/pyvista.py``

### Description of the bug

The bug was encountered while working on https://github.com/ansys/pyaedt-examples/pull/548.

Using the latest ``pyvista`` release (currently v0.47.1), an error occurs in the ``low_frequency/general/resistance.py`` example when calling ([at this line](https://github.com/ansys/pyaedt-examples/blob/d77eba1af9ed588cc5614ccb8e34af8e1aeb7d46/examples/low_frequency/general/resistance.py#L222)) the ``plot()`` method of the class ``ModelPlotter`` defined in ``vizualization/plot/pyvista.py`` ([here](https://github.com/ansys/pyaedt/blob/6c3dd6bd8707dcd17fff66374ab769054b02d144/src/ansys/aedt/core/visualization/plot/pyvista.py#L1362)).
This ``pyaedt`` method is calling the ``populate_pyvista_object()`` method (same file) where [at this line](https://github.com/ansys/pyaedt/blob/6c3dd6bd8707dcd17fff66374ab769054b02d144/src/ansys/aedt/core/visualization/plot/pyvista.py#L1344) we are trying to set a new attribute (``viewup``) to an object of class ``Camera`` (a ``pyvista`` class).

However, ``pyvista`` introduced a breaking change in v0.46.0 (pyvista/pyvista#7716) preventing new attributes from being set on PyVista classes.
This exact line is not covered by our tests. 
A fix for this could be to use the ``pyvista.set_new_attribute()`` method as described [here](https://docs.pyvista.org/api/utilities/_autosummary/pyvista.set_new_attribute).
The only consequence of this issue for now it that the latest version of ``pyvista`` can't be used for the ``pyaedt-examples``.

@SMoraisAnsys, @Samuelopez-ansys, FYI. 

### Steps To Reproduce

In a python virtual environment, install the requirements for the ``pyaedt-examples`` listed in the ``requirements/requirements.txt`` file on that repo.
Replace the install of ``pyvista`` with the latest release available for that package.
Then run the ``low_frequency/general/resistance.py`` example.

Windows

3.10

### Installed packages

From ``requirements/requirements_doc.txt`` in ``pyaedt-examples`` with the latest ``pyvista`` release available.

ecoussoux-ansys:
This PR is fixing the issue observed in the ``ModelPlotter`` class when using ``pyvista`` >= 0.46.0, as described in #7451.

On top of that two minor modifications are introduced here : 
*  In the ``_read_case()`` method (still of the ``ModelPlotter`` class), the ``algorithm`` input is explicitly set to ``None`` in the call to the ``extract_surface()`` method. This is following the note (and silencing the associated warning) added to this method in recent versions (see source code [here](https://github.com/pyvista/pyvista/blob/release/0.47/pyvista/core/filters/data_object.py#L3343-L3531)),
* In the ``pyproject.toml``, the ``[jupyter]`` optional dependency group is added to ``pyvista`` inside the ``[examples]`` optional dependency for ``pyaedt``. This is necessary to properly render 3D ``pyvista`` plots in Jupyter when using latest ``pyvista`` releases as discussed in https://github.com/ansys/ansys-aedt-toolkits-radar-explorer/pull/34. This will be useful for the ``pyaedt-examples`` repo which depends on ``pyaedt[examples]`` and contains 3D ``pyvista`` plots to be rendered inside of Jupyter notebooks.

I tested these changes locally and the affected example (``low_frequency/general/resistance.py`` from the ``pyaedt-examples``) can now be executed without error (as script and notebook) using the latest ``pyvista`` version.

Closes #7451


src/ansys/aedt/core/visualization/plot/pyvista.py
```diff
@@ -1341,7 +1341,7 @@ def populate_pyvista_object(self) -> None:
             if isinstance(self.camera_position, (tuple, list)):
                 self.pv.camera.position = self.camera_position
                 self.pv.camera.focal_point = self.focal_point
-                self.pv.camera.viewup = self.view_up
+                pv.set_new_attribute(self.pv.camera, "viewup", self.view_up)
```
Samuelopez-ansys: @ecoussoux-ansys Why is only needed "viewup"? And position and focal_point?

ecoussoux-ansys:
Thanks @Samuelopez-ansys and @SMoraisAnsys for the feedback.
Let me try to answer on the two topics.

First, regarding the attributes of the object of class ``pyvista.plotting.camera.Camera`` (for us the ``self.pv.camera``). All the attributes you listed are defined at init by the class implementation as can be seen from [the docs](https://docs.pyvista.org/api/core/_autosummary/pyvista.camera). Therefore no specific syntax is required here to set new values for those.
It's different from ``viewup`` as this one is _not_ an attribute of the class ``pyvista.plotting.camera.Camera``.

However, on the second point raised regarding this ``viewup`` attribute: After looking into it more (my first approach was just trying to "fix the pyvista error" and I admit I was not thinking of why this attribute was there..), I believe that setting this attribute is not useful.
This particular line was introduced by this pretty old PR : #996
And already there, we were using the value in ``self.view_up`` to update a property of the ``self.pv.camera`` object in two different ways in two methods of the class ``ModelPlotter`` : 
* In the method ``populate_pyvista_object()`` we were doing ``self.pv.camera.viewup = self.view_up``,
* In the method ``animate()`` we are doing ``self.pv.camera.up = self.view_up``.

Could it be that it was a mistake from the beginning? 
In cb272c2 I aligned the syntax in ``populate_pyvista_object()`` on the one in ``animate()`` as I believe it makes much more sense : It updates an attribute that actually exists (with a meaningful value for it too!) and I could not find any resources on what is supposed to be achieved by defining a new attribute ``viewup`` as was done before..!

I tested again this new change on the example (resistance) from the ``pyaedt-examples`` and everything works fine. 
Let me know what you think ! Thanks!

Samuelopez-ansys:
Great description @ecoussoux-ansys , I agree with your solution

PR #7475, opened 2026-04-01
FIX: Implemented modification for 26R1 SP

maxcapodi78:
## Description
**Please provide a brief description of the changes made in this pull request.**

## Issue linked
**Please mention the issue number or describe the problem this pull request addresses.**

## Checklist
- [ ] I have tested my changes locally.
- [ ] I have added necessary documentation or updated existing documentation.
- [ ] I have followed the coding style guidelines of this project.
- [ ] I have added appropriate tests (unit, integration, system).
- [ ] I have reviewed my changes before submitting this pull request.
- [ ] I have linked the issue or issues that are solved by the PR if any.
- [ ] I have agreed with the Contributor License Agreement ([CLA](https://developer.ansys.com/form/cla-acceptance)).

PR #7479, opened 2026-04-01
FEAT: Enhance extension manager with optional extensions support

eblanco-ansys:
## Description
This PR enhances the PyAEDT extension tooling with support for optional extensions and a broader UI/theme refresh across the extension manager and version manager.

### What changed
- Added optional-extension metadata support in toolkit catalogs via `optional = true` entries for selected extensions.
- Implemented optional-extension behavior in the Extension Manager:
  - Added logic to identify optional extensions by category/toolkit metadata.
  - Added global optional-extension enable/disable state handling.
  - Disabled interaction (launch/pin/open web) for optional extensions when unavailable.
  - Added user-facing tooltip/log messaging when actions are blocked due to missing optional dependencies.
- Improved extension category resolution so mapped categories (for example Common/Product name mappings) resolve correctly to toolkit metadata.
- Updated extension cards/layout and UX:
  - Wider minimum window, fixed-width card grid behavior, improved scroll/canvas sizing.
  - Better card title handling (truncation with tooltip for long labels).
  - Improved pin icon placement and disabled visual state for optional entries.
- Refined dialog/log window theming and canvas styling consistency for light/dark modes.
- Refactored and modernized theme infrastructure in shared extension UI code:
  - Expanded light/dark palettes, including disabled button colors.
  - Updated default fonts/styles and action-button variants.
  - Improved tooltip theming to adapt to active theme.
  - Improved scrollbar, entry, combobox, radiobutton, and frame styling consistency.
- Updated Version Manager UI:
  - New min/max window bounds and more compact layout.
  - Theme application improvements across non-ttk widgets.
  - Refined basic/advanced layout alignment and update notification sizing.

image
image

When optional extensions are disabled, they get displayed as unactive:
image

## Issue linked
No issue linked in this PR yet.

### DIFF
--- a/.github/workflows/ci-release.yml
+++ b/.github/workflows/ci-release.yml
@@ -31,7 +31,7 @@ jobs:
       contents: write        # Required to commit and push changelog updates to the repository
       pull-requests: write   # Required to create pull requests with changelog updates
     steps:
-      - uses: ansys/actions/doc-deploy-changelog@de753211965a233968fb2cda376da99376b1b9e5 # v10.2.11
+      - uses: ansys/actions/doc-deploy-changelog@7d55ff0b9f67c77dae50a286d711ee1ae27b0a25 # v10.2.12
         with:
           bot-email: ${{ secrets.PYANSYS_CI_BOT_EMAIL }}
           bot-user: ${{ secrets.PYANSYS_CI_BOT_USERNAME }}
@@ -45,7 +45,7 @@ jobs:
     permissions:
       contents: read   # Required to read repository content for vulnerability scanning
     steps:
-      - uses: ansys/actions/check-vulnerabilities@de753211965a233968fb2cda376da99376b1b9e5 # v10.2.11
+      - uses: ansys/actions/check-vulnerabilities@7d55ff0b9f67c77dae50a286d711ee1ae27b0a25 # v10.2.12
         with:
           python-version: ${{ env.MAIN_PYTHON_VERSION }}
           python-package-name: ${{ env.PACKAGE_NAME }}
@@ -59,7 +59,7 @@ jobs:
     permissions:
       contents: read   # Required to read workflow files and check for security issues
     steps:
-      - uses: ansys/actions/check-actions-security@de753211965a233968fb2cda376da99376b1b9e5 # v10.2.11
+      - uses: ansys/actions/check-actions-security@7d55ff0b9f67c77dae50a286d711ee1ae27b0a25 # v10.2.12
         with:
           generate-summary: true
           token: ${{ secrets.GITHUB_TOKEN }}
@@ -72,7 +72,7 @@ jobs:
       contents: read   # Required to read documentation files for style checking
     steps:
       - name: Check documentation style
-        uses: ansys/actions/doc-style@de753211965a233968fb2cda376da99376b1b9e5 # v10.2.11
+        uses: ansys/actions/doc-style@7d55ff0b9f67c77dae50a286d711ee1ae27b0a25 # v10.2.12
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           fail-level: 'warning'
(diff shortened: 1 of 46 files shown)


src/ansys/aedt/core/extensions/installer/extension_manager.py
```diff
@@ -1414,6 +1507,36 @@ def launch_web_url(self, category: str, option: str) -> bool:
...
+    def _get_extension_info(self, category: str, option: str) -> dict:
+        """Return toolkit metadata for a built-in extension entry."""
+        if not self.toolkits or option.lower() == "custom":
+            return {}
+        return self._get_category_toolkits(category).get(option, {})
+
+    @property
+    def optional_extensions(self) -> bool:
+        """Whether optional extensions are enabled for interaction."""
+        return getattr(self, "_optional_extensions", False)
```
SMoraisAnsys: Why are you not simply calling self._optional_extensions ? It is set at the begining, no ?

SMoraisAnsys (commented):
I'm not sure to understand how the "optional" is activated / deactivated from the UI. Is that possible with the current implementation ?

tests/unit/extensions/test_extension_manager.py
```diff
@@ -110,6 +112,93 @@ def test_extension_manager_load_extensions(mock_toolkits, mock_desktop, mock_aed
...
+                "name": "Generate report",
+                "script": "create_report.py",
+                "optional": True,
+                "url": "https://example.com",
+            }
+        }
+    }
+    mock_toolkits.return_value = toolkit_data
+
+    extension = ExtensionManager.__new__(ExtensionManager)
```
SMoraisAnsys, line 194: Why are you using `__new__` here ? Oo
Seems like it's used everywhere but I don't know why not just creating the object through the class call only

tests/unit/extensions/test_version_manager.py
```diff
@@ -160,8 +160,10 @@ def get(self):
 
     desktop.release_desktop = MagicMock()
 
-    # Instantiate
-    manager = vm.VersionManager(ui, desktop)
+    # Instantiate without live PyPI lookups or background update threads.
+    with patch("ansys.aedt.core.extensions.installer.version_manager.get_latest_version", return_value="Unknown"):
+        with patch.object(vm.VersionManager, "check_for_pyaedt_update_on_startup", return_value=None):
```
SMoraisAnsys, line 165: Just a warning as this patch are only valid for the `manager = vm.VersionManager(ui, desktop)` call and get deactivated afterward. If it's needed afterward, this won't be applicable anymore.
Please ignore this comment if it's under control :)

tests/unit/extensions/test_extension_manager.py
eblanco-ansys: Initializing with `__new__` avoids a full initialization. Since we are not testing tkinter here, it's more secure to initialize the extension this way. I will perform a full refactoring of this file in the extension migration, because we could probably use this technique for other tests and avoid tcl issues

src/ansys/aedt/core/extensions/installer/extension_manager.py
eblanco-ansys: Correct, but I wanted to handle the logic for setting up the parameter in a property because we will use here `is_pyaedt_in_edt` method from #7475, which is not already merged

tests/unit/extensions/test_version_manager.py
eblanco-ansys: This is just needed at the beginning of the test. I was having some issues with the background update thread in the tests so now is not triggered

src/ansys/aedt/core/extensions/installer/extension_manager.py
eblanco-ansys: We can wait for #7475 merge and I will apply the modifications :)

Samuelopez-ansys (approved):

PR #7572, opened 2026-04-23
FEAT: RaptorX setup

gkorompi:
Adding raptorX setup in pyaedt.


gkorompi:
@Samuelopez-ansys as discussed in today's call

tests/system/layout/test_3dlayout_modeler.py
```diff
@@ -1351,3 +1351,11 @@ def test_create_coordinate_system(aedt_app) -> None:
 def test_create_scattering(aedt_app) -> None:
     aedt_app.create_setup()
     assert aedt_app.create_scattering()
+
+
+def test_create_raptorx_setup(aedt_app) -> None:
```
SMoraisAnsys, line 1357: Can you add a test for the case where the `ic_mode` is false ?

src/ansys/aedt/core/application/analysis_3d_layout.py
```diff
@@ -318,6 +318,9 @@ def create_setup(self, name: str = "MySetupAuto", setup_type: str | None = None,
         elif setup_type in SetupKeys.SetupNames:
             setup_type = SetupKeys.SetupNames.index(setup_type)
         name = self.generate_unique_setup_name(name)
+        if setup_type == 62 and not self.ic_mode:
+            self.logger.error("RaptorX is supported only in 'IC' mode")
+            return None
```
Samuelopez-ansys: @gkorompi  why not raising an exeption?

It can be AEDTRuntimeError

Samuelopez-ansys (changes requested):

tests/system/layout/test_3dlayout_modeler.py
```diff
@@ -1351,3 +1351,13 @@ def test_create_coordinate_system(aedt_app) -> None:
...
+    aedt_app.ic_mode = False
+    assert not aedt_app.create_setup(name="raptor_setup", setup_type="RaptorX")
```
Samuelopez-ansys: Now that it raises an exception, you need to test here, with a "with pytest.raise...."

Samuelopez-ansys (changes requested):

Samuelopez-ansys (approved):

PR #7604, opened 2026-05-04
FIX: compute_power_by_layer

### LINKED ISSUE #7605: Bug located in compute_power_by_layer

### Description of the bug

AEDT is raises an error when computing power loss if the net has no field data on a specific layer. When it happens, PyAEDT will raise an error and release the desktop.

This is correct behavior but not wanted in compute_power_by_layer

### Steps To Reproduce

execute compute_power_by_layer

Windows

3.12

### Installed packages

all

hui-zhou-a:
## Description

Add try-except to handle the situation when a net has no field data on the layer. 
- prevent the code from breaking
- set loss to 0

## Issue linked
Close #7605 

## Checklist


hui-zhou-a:
close #7605

hui-zhou-a:
To handle this situation, the user need to turn off release_on_exception
```
from ansys.aedt.core import Hfss3dLayout

aedt = r"D:\solved_project.aedt"


h3d = Hfss3dLayout(aedt, version="2026.1", non_graphical=True)

settings.release_on_exception = False
loss = h3d.post.compute_power_by_layer()
settings.release_on_exception = True

h3d.release_desktop(False, False)
```

src/ansys/aedt/core/visualization/post/post_3dlayout.py
```diff
@@ -157,9 +158,27 @@ def _compute_power_loss(self, net_filter=None, layer_filter=None, solution: str
...
+                        try:
+                            self._app.post.fields_calculator.add_expression(my_expression, "")
+                            loss = self._app.post.fields_calculator.evaluate(
+                                my_expression["name"], solution, intrinsics={}
+                            )
+                        except Exception:
+                            self.logger.add_warning_message(
+                                f"Net `{net_name}` has no field data on layer `{layer_name}`."
+                            )
+                            loss = 0
```
SMoraisAnsys: I would agree with this pattern (same as `import_from_openstreet_map`) if the second `try:...` snippet was modifying the evaluation method. But that's not the case. Why do you trigger the same code a second time ? Also, what the logic with assigning `loss = 0` ? Multiple issues could arise and lead to this situation.

Please update the PR description.

SMoraisAnsys (changes requested):

src/ansys/aedt/core/visualization/post/post_3dlayout.py
hui-zhou-a: The second try is unnecessary and it is removed now. The fix is to workaround AEDT error when there is no field data. In this case, loss is set to 0 instead of breaking. The expression in this method has a fixed pattern, so it is not likely other issue could lead to the exception.

Samuelopez-ansys: @SMoraisAnsys if the layer has several polygons and of them has not power information, AEDT raises an exception, so the fallback is setting these valeus to 0.

Please @hui-zhou-a add a description to the PR.

src/ansys/aedt/core/visualization/post/post_3dlayout.py
```diff
@@ -157,9 +158,21 @@ def _compute_power_loss(self, net_filter=None, layer_filter=None, solution: str
...
 
-                loss = self._app.post.fields_calculator.evaluate(my_expression["name"], solution, intrinsics={})
+                try:
+                    self._app.post.fields_calculator.add_expression(my_expression, "")
+                    loss = self._app.post.fields_calculator.evaluate(my_expression["name"], solution, intrinsics={})
+                except Exception as e:  # pragma: no cover
+                    if settings.release_on_exception:
+                        raise GrpcApiError(
+                            f"Failed to compute power loss for Net `{net_name}` on layer `{layer_name}`."
+                            f"Consider disabling `settings.release_on_exception` for fallback compute. "
```
SMoraisAnsys:
```suggestion
                            f"Failed to compute power loss for Net `{net_name}` on layer `{layer_name}`. "
                            "Set `settings.release_on_exception = False` to fall back to a zero loss value instead of raising."
```

SMoraisAnsys (dismissed):
The changes make more sense now, thanks @hui-zhou-a.
I left a minor comment on the error message.

Samuelopez-ansys:
@hui-zhou-a  Is it possible to add a test?

hui-zhou-a:
> @hui-zhou-a Is it possible to add a test?

Sorry, no. I couldn't make up such a test case. The problem was reported on a customer design.

Samuelopez-ansys:
Hi @hui-zhou-a , this is not a suggestion; we can not add new code that is not tested. You can take the project used for the tests: def test_dcir(dcir_example_project)
And modify it to have a polygons without power.

SMoraisAnsys:
I agree with @Samuelopez-ansys, worst case scenario we can mock the situation and only add a unit test (not system test)

hui-zhou-a:
> Hi @hui-zhou-a , this is not a suggestion; we can not add new code that is not tested. You can take the project used for the tests: def test_dcir(dcir_example_project) And modify it to have a polygons without power.



> Hi @hui-zhou-a , this is not a suggestion; we can not add new code that is not tested. You can take the project used for the tests: def test_dcir(dcir_example_project) And modify it to have a polygons without power.

Tried, but I cannot reproduce the problem on this example board.

hui-zhou-a:
Added test on mock

tests/system/layout/test_3dlayout_dcir.py
```diff
@@ -62,3 +64,15 @@ def test_dcir(dcir_example_project) -> None:
...
+@pytest.mark.skipif(is_linux, reason="Not Supported on Linux.")
+@pytest.mark.skipif(DESKTOP_VERSION == "2025.2", reason="WAITING BUG FIX")
+@patch(
+    "ansys.aedt.core.visualization.post.fields_calculator.FieldsCalculator.add_expression",
+    side_effect=RuntimeError("mock add_expression failure"),
+)
+def test_dcir2(mock_add_expression, dcir_example_project) -> None:
+    assert dcir_example_project.post.compute_power_by_layer()
+    assert mock_add_expression.called
+    assert any(call.args and call.args[-1] == "" for call in mock_add_expression.call_args_list)
```
SMoraisAnsys:
```suggestion
@pytest.mark.skipif(is_linux, reason="Not Supported on Linux.")
@pytest.mark.skipif(DESKTOP_VERSION == "2025.2", reason="WAITING BUG FIX")
@patch(
    "ansys.aedt.core.visualization.post.fields_calculator.FieldsCalculator.add_expression",
    side_effect=RuntimeError("Dummy exception"),
)
def test_dcir_power_loss_fallback_to_zero(mock_add_expression, dcir_example_project) -> None:
    """Test fallback to zero."""
    result = dcir_example_project.post.compute_power_by_layer()

    assert result is not None
    assert all(entry == 0.0 for entry in result.values())
```
The proposed test does not bring much sense in terms of validation. This one should ensure that the fall back is used and the result is the one expected.

SMoraisAnsys (dismissed):

hui-zhou-a:
#7605

hui-zhou-a:
@SMoraisAnsys Could you help me on Labeler check? I referenced this PR in an issue, but it just cannot pass this test.

SMoraisAnsys:
> @SMoraisAnsys Could you help me on Labeler check? I referenced this PR in an issue, but it just cannot pass this test.

Your PR is not up to date so it can't run a job that it doesn't know.

Samuelopez-ansys:
You need to link an issue in the description, if not issue exists, just comment that no issue linked.

hui-zhou-a:
I linked this PR to the issue manually. It should be fine now.

Samuelopez-ansys (approved):

PR #7645, opened 2026-05-12
FIX: enhance Unix socket port discovery for AEDT processes

Samuelopez-ansys:
## Description

This pull request improves the reliability and documentation of Unix socket port discovery for AEDT processes on Linux. The main focus is on making session detection more robust, especially in environments where the `ss` utility is missing or restricted, and on providing clearer installation and troubleshooting instructions for users.

**Enhancements to Unix socket port discovery:**

* Added a new fallback mechanism in `general_methods.py` that parses `/proc/net/unix` and `/proc/<pid>/fd` to discover AEDT Unix-socket ports when the `ss` utility is unavailable or lacks sufficient permissions. This ensures session detection works across a wider range of Linux distributions and user configurations.

* Improved the `_run_ss_xlp` function to first attempt discovery using `ss -xlp`, and if that fails (or is incomplete), automatically fall back to the new `/proc`-based method. The function now also logs debug information for troubleshooting.

* Updated imports in `general_methods.py` to include `Path` and `shutil` for the new fallback logic.

**Documentation improvements:**

* Expanded the Linux installation instructions to clarify environment variable setup, explain the fallback mechanism for session discovery, and provide guidance for users whose `$PATH` does not include `/usr/sbin`. Troubleshooting steps are now included for cases where PyAEDT opens a new AEDT session instead of connecting to an existing one.
* Added a changelog entry summarizing the enhancement to Unix socket port discovery for AEDT processes.

## Issue linked
No issue linked.

## Checklist


maxcapodi78 (dismissed):
Great catch

src/ansys/aedt/core/generic/general_methods.py
```diff
@@ -710,51 +712,138 @@ def is_digit(c):
...
+    # Format: Num RefCount Protocol Flags Type St Inode Path
+    inode_to_port: dict[str, int] = {}
+    try:
+        for line in proc_net_unix.read_text().splitlines():
+            port_match = re.search(r"-(\d+)\.sock", line)
+            if not port_match:
+                continue
+            parts = line.split()
+            if len(parts) >= 7:
+                inode_to_port[parts[6]] = int(port_match.group(1))
```
SMoraisAnsys: Just wondering if we should check for `==` instead or keep `>=` and access parts with `-1` instead of 6 (assuming that the last value is the one we want). Just an open question here, feel free to discard.

tests/unit/test_general_methods.py
```diff
@@ -132,3 +135,76 @@ def test_valid_versions(self, version):
...
+            sock.listen(1)
+
+            current_pid = os.getpid()
+
+            # Inject the current process as the only "AEDT" target so that
+            # _run_ss_xlp inspects our own /proc/<pid>/fd entries.
+            with patch(
+                "ansys.aedt.core.generic.general_methods._get_target_processes",
+                return_value=[(current_pid, ["/ansysedt"])],
+            ):
```
SMoraisAnsys: Can you decorate the test instead ?

SMoraisAnsys (dismissed):
I left minor comments, LGTM otherwise :)

src/ansys/aedt/core/generic/general_methods.py
Samuelopez-ansys: Good catch, but I'd keep it as-is. 

The /proc/net/unix format is Num RefCount Protocol Flags Type St Inode Path, so the inode is always at index 6 (fixed columns), while -1 would point to the Path string and break the inode→port mapping used later against socket:[<inode>] symlinks. >= is also safer than == since unnamed sockets have 7 tokens and named ones 8+.

If anything, the only tightening worth doing is >= 8 to require the Path column (already implied by the .sock regex match above).

doc/source/Getting_started/Installation.rst
```diff
@@ -225,12 +225,34 @@ Linux support
...
+    export ANSYSEM_ROOT261=/path/to/AnsysEM/v261/AnsysEM
+    export LD_LIBRARY_PATH=$ANSYSEM_ROOT261:$LD_LIBRARY_PATH
+
+The version suffix in ``ANSYSEM_ROOT<XYZ>`` must match your installed AEDT release
+(for example, ``251`` for 2025 R1, ``252`` for 2025 R2, ``261`` for 2026 R1).
+The default installation path on Linux is typically
+``/opt/AnsysEM/v<XYZ>/AnsysEM`` or ``/usr/ansys_inc/v<XYZ>/AnsysEM``.
+
+.. note::
+    On some Linux distributions (such as RHEL/CentOS 8), the ``ss`` utility used
```
SMoraisAnsys, line 243:
```suggestion
.. note::

    On some Linux distributions (such as RHEL/CentOS 8), the ``ss`` utility used
```
I think a separator is needed

SMoraisAnsys (dismissed):

maxcapodi78 (dismissed):
Great fix... Hope it covers now all use cases and OS

SMoraisAnsys (approved):

PR #7666, opened 2026-05-19
CI: Refactor label logic to use inner logic

SMoraisAnsys:
## Description
Rework the label logic to avoid using external action and add trailer logic. This can be adapted to only parse the PR logic, atm both the PR description & commits are checked.

## Issue linked
None

## Checklist


.github/actions/label-from-trailers/action.yml
```diff
@@ -0,0 +1,136 @@
...
+      shell: bash
+      env:
+        GH_TOKEN: ${{ inputs.github-token }}
+        PR_NUMBER: ${{ inputs.pr-number }}
+      run: |
+        set -uo pipefail
+
+        # ---------- Fetch PR ----------
+
+        PR_BODY=$(gh pr view "$PR_NUMBER" --repo ansys/pyaedt --json body \
```
MaxJPRey, line 36: Instead of hardcoding ansys/pyaedt, we could use ${{ github.repository }} passed via env, or accept repository as an input.

This is mainly if we want to make the action more generic later in our ansys-action repo.

.github/actions/label-from-trailers/action.yml
```diff
@@ -0,0 +1,136 @@
...
+runs:
+  using: composite
+  steps:
+    - name: Parse trailers and apply labels
+      shell: bash
+      env:
+        GH_TOKEN: ${{ inputs.github-token }}
+        PR_NUMBER: ${{ inputs.pr-number }}
+      run: |
+        set -uo pipefail
```
MaxJPRey, line 32: To prevent failures to be silently ignored unless explicitly handled, I would suggest:
```suggestion
        set -euo pipefail
```

SMoraisAnsys: That is on design. The current code leverage some calls that can return exit code 1 without being an issue.

Samuelopez-ansys (approved):
LGTM, is this PR ready?

PR #7702, opened 2026-05-27
DOCS: Update main docs page

eblanco-ansys:
## Description
Provide `whatsnew` banner

## Issue linked
No issue

## Checklist


SMoraisAnsys (changes requested):
If possible I would avoid to add a banner and instead leverage the `"whatsnew"` option of the ansys-sphinx-theme, see for example [this](https://github.com/ansys/pymechanical/blob/main/doc/source/conf.py#L244-L248) in pymechanical.

The main reason I would prefer to avoid anything related to banner is that we have automatic stuff running through actions / automation to add some already and I would prefer to avoid creating conflicts. In fact, I discovered an existing issue and was working on that this morning (banner related to secure options), see [this PR](https://github.com/ansys/actions/pull/1339).

eblanco-ansys:
> If possible I would avoid to add a banner and instead leverage the `"whatsnew"` option of the ansys-sphinx-theme, see for example [this](https://github.com/ansys/pymechanical/blob/main/doc/source/conf.py#L244-L248) in pymechanical.
> 
> The main reason I would prefer to avoid anything related to banner is that we have automatic stuff running through actions / automation to add some already and I would prefer to avoid creating conflicts. In fact, I discovered an existing issue and was working on that this morning (banner related to secure options), see [this PR](https://github.com/ansys/actions/pull/1339).

I see, it makes sense👍

eblanco-ansys:
> If possible I would avoid to add a banner and instead leverage the `"whatsnew"` option of the ansys-sphinx-theme, see for example [this](https://github.com/ansys/pymechanical/blob/main/doc/source/conf.py#L244-L248) in pymechanical.
> 
> The main reason I would prefer to avoid anything related to banner is that we have automatic stuff running through actions / automation to add some already and I would prefer to avoid creating conflicts. In fact, I discovered an existing issue and was working on that this morning (banner related to secure options), see [this PR](https://github.com/ansys/actions/pull/1339).

This is a good solution from the main screen, but if the user opens any other section from the docs won't notice it. How could we address this then?

doc/changelog.d/whatsnew.yml
```diff
@@ -0,0 +1,27 @@
+fragments:
+  - title: PyAEDT 1.0 release highlights
+    version: 1.0.0
+    content: |
```
SMoraisAnsys: I think this is too big, we should have a look into how it renders in the HTML files and how others are handling that. Providing references to other pages might be a good idea here (like a migration page)

doc/source/_static/custom.css
```diff
@@ -4,30 +4,70 @@
...
+
 .col-xl-7 {
     flex: 0 0 53.33333%;
     max-width: 53.33333%;
-  }
+}
 
 .bd-toc {
     padding-top: 5em;
 }
```
SMoraisAnsys, line 178: I think we can discard those changes right ?

SMoraisAnsys (changes requested):

doc/changelog.d/whatsnew.yml
eblanco-ansys: It's not very big and doesn't grab the attention of the visitor:
image
We should definitely add more links to this section, like this one: https://aedt.docs.pyansys.com/version/stable/release_1_0.html#release-1-0
What I don't like is that in other sections this new left pane it's not visible:
image

SMoraisAnsys: I assumed that the whole content would render on the left side, but that's not the case. Great !

Also, I assumed that the previous banner proposition was modifying the logic behind the already existing banners as I couldn't see the secure grpc one. However that's not the case and this issue resulted from another thing (it is now handled). Therefore, aren't the current online documentation banners already enough ? For example, if I look into 0.28.0 (landing page, API, ...) I get the following
image

eblanco-ansys: For that case I agree, if you are checking an old version of the docs it's clear. The problem is if you are visualizing directly the latest documentation (default scenario). IMO 1.0 version is something important we should announce in the whole page, as pandas does here: https://pandas.pydata.org/docs/index.html

SMoraisAnsys: If there is a link to the breaking changes / blog like in pandas then I'm fine with it. If it's just about notifying people that there is a release 1.0 when they are already watching it, the interest is none.

PR #7711, opened 2026-06-01
CHORE: Add workflow for syncing main branch to release/2.0

Samuelopez-ansys:
## Description

Create workflow for validate PRs in release 2.0 branch, and check if the branch is diverging.
 

## Issue linked
None

## Checklist


.github/workflows/sync-release.yml
```diff
@@ -0,0 +1,1209 @@
...
+      contents: read
+    outputs:
+      has_conflicts: ${{ steps.check.outputs.has_conflicts }}
+    steps:
+      - name: Checkout release/2.0
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          ref: release/2.0
+          fetch-depth: 0
+          persist-credentials: true
```
SMoraisAnsys:
```suggestion
          persist-credentials: false
```
Since this step only checks for merge, I don't think there is a need to persist the credentials.

.github/workflows/sync-release.yml
```diff
@@ -0,0 +1,1209 @@
...
+        run: git fetch origin main
+
+      - name: Try merge (dry-run)
+        id: check
+        run: |
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git config user.name  "github-actions[bot]"
+          # Keep release/2.0 version of __init__.py to avoid version conflicts
+          git config merge.ours.driver true
+          echo 'src/ansys/aedt/core/__init__.py merge=ours' >> .gitattributes
```
SMoraisAnsys: Is that to ensure that we don't update the version of the project ? If yes, is there a reason not to do it as we could simply change the version number when releasing ?

.github/workflows/sync-release.yml
```diff
@@ -0,0 +1,1209 @@
...
+          if git merge --no-commit --no-ff origin/main; then
+            echo "has_conflicts=false" >> "$GITHUB_OUTPUT"
+            git merge --abort 2>/dev/null || true
+          else
+            echo "has_conflicts=true" >> "$GITHUB_OUTPUT"
+            git merge --abort 2>/dev/null || true
+          fi
+          # Clean up temporary .gitattributes change
+          git checkout -- .gitattributes 2>/dev/null || true
```
SMoraisAnsys:
```suggestion
          if git merge --no-commit --no-ff origin/main; then
            echo "has_conflicts=false" >> "$GITHUB_OUTPUT"
          else
            echo "has_conflicts=true" >> "$GITHUB_OUTPUT"
          fi
```
If we get rid of the gitattributes update (previous comment), maybe we could also remove the `git merge --abort` since the following jobs will clone the repo again. Could you give it a try ?

.github/workflows/sync-release.yml
```diff
@@ -0,0 +1,1209 @@
...
+          fetch-depth: 0
+          persist-credentials: true
+          token: ${{ secrets.PYANSYS_CI_BOT_TOKEN }}
+
+      - name: Create or update sync branch
+        id: sync
+        run: |
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git config user.name  "github-actions[bot]"
+          # Keep release/2.0 version of __init__.py to avoid version conflicts
```
SMoraisAnsys: Same comment, we problably can keep the other version until we do a manual release any way

SMoraisAnsys (changes requested):
Thanks for this sync workflow @Samuelopez-ansys, I've left a few comments.
Also, aren't the rocky related jobs missing ?

.github/workflows/sync-release.yml
Samuelopez-ansys: Indeed this is the reason. Then you suggest that branch release/2.0 __init__.py should be always syncronize with main one right?

ecoussoux-ansys (commented):
Nothing to add to Sebastien's review, thanks !

Samuelopez-ansys:
Hi @SMoraisAnsys, I think I applied your suggestions. I tested it in the fork, and it works, the only condition is having in the __init__ of releaase/2.0 the same version number than the main to avoid the conflict

SMoraisAnsys (commented):
Open question, should we also add a job to ensure that every PR, except for sync ones (i.e. non schedule and non worfklow dispatch), contain breaking changes information (like `!` in the PR title, breaking change label, ...) ? When we'll push / merge something manually, we won't go through PR workflow so that shouldn't be a problem. What do you think ?

.github/workflows/sync-release.yml
```diff
@@ -0,0 +1,1239 @@
...
+          # Skip for automated sync PRs
+          if echo "$PR_TITLE" | grep -qi "sync main into release"; then
+            echo "Automated sync PR detected, skipping breaking change check."
+            exit 0
+          fi
+          # Check for breaking change indicators
+          if echo "$PR_TITLE" | grep -q '!:' || echo "$PR_LABELS" | grep -qi 'breaking'; then
+            echo "Breaking change information found."
+          else
+            echo "::error::PRs targeting release/2.0 must indicate breaking changes. Add '!' to the PR title (e.g., 'FEAT!: ...') or apply a 'breaking-change' label."
```
SMoraisAnsys:
```suggestion
            echo "::error::PRs targeting release/2.0 must indicate breaking changes. Add '!' to the PR title (e.g., 'FEAT!: ...') or the 'Breaking-Change:' trailer, or apply a 'breaking-change' label."
```

SMoraisAnsys (dismissed):
LGTM, I left a last minor comment :)

.github/workflows/sync-release.yml
```diff
@@ -0,0 +1,1239 @@
...
+            exit
```
SMoraisAnsys: Shouldn't we exit 1 here ?

Samuelopez-ansys:
@SMoraisAnsys Thanks for the review, performed your suggestions

eblanco-ansys (approved):

SMoraisAnsys (approved):

PR #7797, opened 2026-06-19
FIX: improved documentation on close_on_exit desktop argument and enhanced the behavior. 

Alberto-DM:
## Description

The `close_on_exit` docstring was reporting "This option is used only when Desktop is used in a context manager (``with`` statement)." This is not correct because the  `close_on_exit` argument also determines whether aedt closes after closing the interpreter session. 
In the current behavior `close_on_exit` is set to `True` as default, which is correct for the context manager, but it is wrong if Desktop is invoked directly (we don't want to close an existing session which pyaedt just connects to). Moreover it has an opposite behavior with respect to the application classes like `Hfss` where the default is `close_on_exit=False`.

This PR updates Desktop shutdown behavior to align close_on_exit with runtime session behavior and documented context-manager semantics.

New close_on_exit behavior:
- When close_on_exit is not explicitly provided:
  - inside a context manager (with Desktop(...)), same behavior as before: 
    - always treated as True (close on exit)
  - outside a context manager: 
    - True if PyAEDT opens a new AEDT session
    - False if PyAEDT connects to an existing AEDT session
- When `close_on_exit` is explicitly set (True/False), same behavior as before: user choice is always respected (inside or outside a context manager).

The docstring has been updated accordingly.

## Tests
Added/updated tests for `close_on_exit` logic.

## Issue linked
None

## Checklist


src/ansys/aedt/core/desktop.py
```diff
@@ -793,7 +809,9 @@ def __init__(
         _desktop_sessions[self.aedt_process_id] = self
         # Register the desktop closure to be called at exit unless asked not to.
         atexit.register(
-            lambda: self.__release_and_close_desktop(close_projects=close_on_exit, close_aedt_app=close_on_exit)
+            lambda: self.__release_and_close_desktop(
+                close_projects=self.close_on_exit, close_aedt_app=self.close_on_exit
```
SMoraisAnsys:
```suggestion
                close_projects=self.__close_on_exit, close_aedt_app=self.__close_on_exit
```
Avoids indirection through property

tests/unit/test_desktop.py
```diff
@@ -302,3 +302,105 @@ def test_grpc_server_args_repr_with_insecure_all_raise_error(mock_settings, monk
 
     with pytest.raises(AEDTRuntimeError):
         str(_ServerArgs(mode=TransportMode.INSECURE))
+
+
+def _patch_init_dependencies(monkeypatch, *, new_desktop_after_grpc):
```
SMoraisAnsys, line 307:
```suggestion
def _patch_init_dependencies(monkeypatch, new_desktop_after_grpc):
```
Should this star be removed ?

SMoraisAnsys (dismissed):
LGTM, I left minor comments. Thanks for handling this side effect.

tests/unit/test_desktop.py
Alberto-DM: I prefer not, I did it to force ``new_desktop_after_grpc`` as keyword-only argument. 
It is a delicate patch, and I wanted to make clear its behavior, and I wanted to avoid future use as ``_patch_init_dependencies(monkeypatch, True)`` without understanding what it was really patched.

src/ansys/aedt/core/desktop.py
Alberto-DM: done. Thank you

Samuelopez-ansys (approved):

PR #7801, opened 2026-06-19
FIX: Fix aedt_version_id and aedt_version_string

Alberto-DM:
## Description
When connecting to an existing session, in case the user specifies the port or the pid and not the version, `desktop.aedt_version_id` and `desktop.aedt_version_string` were showing the last stable version installed on system, even if the desktop connected to and older version. 

Now the correct version is shown. 

## Issue linked
none

## Checklist


src/ansys/aedt/core/desktop.py
```diff
@@ -2396,7 +2399,6 @@ def __check_version(self, specified_version, student_version):
 
         # Save the version information in the instance and global settings for later use
         self.aedt_version_id = specified_version
-        settings.aedt_version = specified_version
```
Samuelopez-ansys, line 2399: Why settings.aedt_version is not set?

Alberto-DM: self.aedt_version_id setter already sets settings.aedt_version, so repeating it here is useless.

src/ansys/aedt/core/desktop.py
```diff
@@ -2571,15 +2573,6 @@ def __initialize(
             )
             self.port = self.grpc_plugin.port
             self.aedt_process_id = self.odesktop.GetProcessID()
-            # NOTE: This is particularly necessary for rpyc connections where the version information is not available
```
Samuelopez-ansys, line 2574: @Alberto-DM Why this is not needed?

Alberto-DM: This is likely old code leftover. At this point `aedt_version_id` is never `None`. 
`aedt_version_id` is set by `__check_version` at the beginning of `__init__` and later refreshed by `check_starting_mode`. 
By the time we reach this point in `__initialize` aedt is already started, so it would never go inside this branch.

It could happen though that a version is required and set in `aedt_version_id`, but PyAEDT attached to a different version. In this case the version was not updated, which is what I fixed in the `__init_desktop`. 
Also, `__init_desktop`, is a failsafe since it is called at the very end of `Desktop.__init__`.

If the explanation is not enough, I discussed the modification with Massimo 😄

Samuelopez-ansys: THis is more than enough :), I was trying to figure out, if something was missed.

Great job!

Samuelopez-ansys (approved):

PR #7814, opened 2026-06-21
CI: Cleanup analyze changes logic

SMoraisAnsys:
## Description
Follow up of multiple PRs performed to change our CI. A cleanup was required.
In particular, this removes the analyze change logic that is no longer required in some cases (emit, extensions and filter solutions).

Open question: should we discuss about reverting part of https://github.com/ansys/pyaedt/pull/7167 ?

## Issue linked
None

## Checklist


Samuelopez-ansys (approved):

Samuelopez-ansys:
@SMoraisAnsys Which one should we revert?

SMoraisAnsys:
> @SMoraisAnsys Which one should we revert?

Given that activity on the repo has lessened, I was wondering if we could just use pytest + testmon and remove the analyze logic. This would avoid having to double check every day that running the tests passed in the nightly run. Currently, we might have situations where some tests are not run but should be and our only way to see that is through the nightly runs.

Samuelopez-ansys:
> > @SMoraisAnsys Which one should we revert?
> 
> Given that activity on the repo has lessened, I was wondering if we could just use pytest + testmon and remove the analyze logic. This would avoid having to double check every day that running the tests passed in the nightly run. Currently, we might have situations where some tests are not run but should be and our only way to see that is through the nightly runs.

Ok, let's remove it, and let' see, my only concern is extensions, which were not working well with testmon and they are still failing randomly.

SMoraisAnsys:
> > > @SMoraisAnsys Which one should we revert?
> > 
> > 
> > Given that activity on the repo has lessened, I was wondering if we could just use pytest + testmon and remove the analyze logic. This would avoid having to double check every day that running the tests passed in the nightly run. Currently, we might have situations where some tests are not run but should be and our only way to see that is through the nightly runs.
> 
> Ok, let's remove it, and let' see, my only concern is extensions, which were not working well with testmon and they are still failing randomly.

Then let us wait for the extensions to be extracted ?

PR #7828, opened 2026-06-24
FEAT: add option to plot in volume - cloud generator

### LINKED ISSUE #7829: Add functionality to plot within the volume of an object - point cloud generator

gmalinve:
## Description
Right now the extension only plot on object's surfaces but for some applications it's important to be able to generate a point cloud within the volume of selected objects.

## Issue linked
Linked to #7829 

## Checklist


tests/unit/extensions/test_points_cloud.py
```diff
@@ -80,6 +80,7 @@ def test_point_cloud_extension_generate_button_multiple_objects(mock_hfss_app_wi
     assert [long_name] == data.choice
     assert 1000 == data.points
     assert "" == data.output_file
+    assert data.in_volume
```
SMoraisAnsys:
```suggestion
    assert not data.in_volume
```
From the code it should be False by default, right ?

SMoraisAnsys (commented):
Can you clarify the documentation with the behavior of "in_volume" w.r.t. the number of points specified ?

My understanding of the code is that when you specify that you want in volume you end up with two files that aim at providing you the same amount of points. Would it make sense to allow providing numbers for volume and surface points ? If yes, we can default to 0 for volume and keep the same behavior as before :)

gmalinve:
@SMoraisAnsys as discussed, I moved the logic of the point cloud in volume inside the point_cloud method in pyvista file :)

src/ansys/aedt/core/extensions/common/points_cloud.py
```diff
@@ -52,7 +52,7 @@
 IS_STUDENT = is_student()
 
 # Extension batch arguments
-EXTENSION_DEFAULT_ARGUMENTS = {"choice": "", "points": 1000, "output_file": ""}
+EXTENSION_DEFAULT_ARGUMENTS = {"choice": "", "points": 1000, "output_file": "", "in_volume": True}
```
SMoraisAnsys: Should we keep the previous behavior instead and set it to False ?

src/ansys/aedt/core/extensions/common/points_cloud.py
```diff
@@ -222,15 +223,28 @@ def browse_output_location() -> None:
...
+        # In volume checkbox
+        ttk.Label(input_frame, text="Generate in volume:", style="PyAEDT.TLabel").grid(
+            row=3, column=0, pady=10, sticky="w"
+        )
+        self._widgets["in_volume_entry"] = tkinter.IntVar()
+        self.in_volume_checkbox = ttk.Checkbutton(
+            input_frame, variable=self._widgets["in_volume_entry"], style="PyAEDT.TCheckbutton"
+        )
+        self.in_volume_checkbox.grid(row=3, column=1, pady=10, padx=5)
+        self._widgets["in_volume_entry"].set(1)
```
SMoraisAnsys: Cf my previous point

src/ansys/aedt/core/extensions/common/points_cloud.py
```diff
@@ -362,9 +380,19 @@ def main(data: PointsCloudExtensionData):
     return str(point_cloud[list(point_cloud.keys())[0]][0])
 
 
-def generate_point_cloud(aedtapp: Desktop, selected_objects: list[str], num_points: int, output_file: str = None):
-    """Generate point cloud from selected objects"""
+def generate_point_cloud(
+    aedtapp: Desktop,
+    selected_objects: list[str],
+    num_points: int,
+    output_file: str = None,
```
SMoraisAnsys:
```suggestion
    output_file: str | None = None,
```

src/ansys/aedt/core/visualization/plot/pyvista.py
```diff
@@ -1661,70 +1661,121 @@ def generate_geometry_mesh(self) -> "Mesh":
         return self.meshes
 
     @pyaedt_function_handler()
-    def point_cloud(self, points: int = 10) -> dict:
+    def point_cloud(self, points: int = 10, in_volume=False) -> dict:
```
SMoraisAnsys, line 1664: Could you simplify the code by extracting the inner function as private / protected methods ?

tests/system/extensions/test_points_cloud.py
```diff
@@ -37,8 +37,8 @@
 def test_point_cloud_extension_logic(add_app_example, test_tmp_dir) -> None:
```
SMoraisAnsys: We need another test so that we have both volume and surface. Not only the volume part.

SMoraisAnsys (dismissed):
LGTM but I left some comments that should be addressed first.
Thanks for the refactoring @gmalinve  !!

gmalinve:
@SMoraisAnsys everything should have been  addressed completely now.

Samuelopez-ansys (approved):

PR #7847, opened 2026-06-26
FIX: aedt_process_id connecting to correct session

Samuelopez-ansys:
## Description
This pull request updates the user guide examples and refines the logic for determining the AEDT starting mode and version selection when connecting to existing sessions. The changes improve clarity in documentation, ensure the correct startup mode is chosen based on session and environment, and enhance version detection for session attachment.

**Documentation improvements:**

* Updated code examples in `desktop_sessions.rst` to use direct imports (`from ansys.aedt.core import Desktop, Hfss, Maxwell3d`) and context manager variable assignment (`with Desktop(...) as d:`), making the usage clearer and more Pythonic. [[1]](diffhunk://#diff-ebf19fdb040c26b7c2ba29d053bffc8bb93cf211c52c06d9e7c8015ee9d6e0d9L49-R57) [[2]](diffhunk://#diff-ebf19fdb040c26b7c2ba29d053bffc8bb93cf211c52c06d9e7c8015ee9d6e0d9L68-R77)

**Logic changes for AEDT startup and session handling:**

* Changed the order and conditions for selecting the AEDT starting mode in `desktop.py`: now prioritizes `settings.remote_rpc_session` before checking for platform-specific requirements, ensuring remote session settings take precedence. 
* Enhanced version selection in `__check_version`: when attaching to an existing session (`self.aedt_process_id` is not `None`), the code now tries to infer the AEDT version from active sessions, improving robustness when connecting to running instances.

## Issue linked
None

## Checklist
- [ ] I have tested my changes locally.
- [ ] I have added necessary documentation or updated existing documentation.
- [ ] I have followed the coding style guidelines of this project.
- [ ] I have added appropriate tests (unit, integration, system).
- [ ] I have reviewed my changes before submitting this pull request.
- [ ] I have linked the issue(s) that are solved by the PR if any.
- [ ] I have assigned this PR to myself.
- [ ] I have added the minimum version decorator to any new backend method implemented.
- [ ] I have agreed with the Contributor License Agreement ([CLA](https://developer.ansys.com/form/cla-acceptance)).

PR #7861, opened 2026-06-30
FIX: test_export_to_aedt_test_FAIL_New

ramin4667:
## Description
**New folder for test.**

## Issue linked
**COM connection issue**

## Checklist


Samuelopez-ansys:
Hi @ramin4667 , I think you should plan to move to gRPC in the short term. dotnet is going to be removed from PyAEDT this year because PyEDB does not need it anymore. And pythonnet is not maintained anymore, and PyAnsys team had to fork it to keep it alive. 

@maxcapodi78 @SMoraisAnsys What do you think?

.github/workflows/manual_draft.yml
```diff
@@ -712,6 +712,7 @@ jobs:
         run: |
           .venv\Scripts\Activate.ps1
           pip install . --group tests
+          pip install "pywin32>=303" "ansys-pythonnet>=3.1.0rc3"
```
SMoraisAnsys:
```suggestion
         pip install .[all-dotnet] --group tests
```
Would that work for you ?

doc/changelog.d/7860.fixed.md
```diff
@@ -0,0 +1 @@
+Pyaedt test export to aedt test fail
```
SMoraisAnsys: This file should be removed.

src/ansys/aedt/core/desktop.py
```diff
@@ -1197,7 +1197,7 @@ def odesktop(self) -> object:
         >>> d = Desktop()
         >>> d.odesktop
         """
-        if settings.use_grpc_api:
+        if self.is_grpc_api:
```
SMoraisAnsys, line 1438: @Samuelopez-ansys Is there a reason why this wasn't aligned with recent changes or can it be modified as proposed ?

pyproject.toml
```diff
@@ -131,6 +131,8 @@ tests = [
     "pytest-xdist>=3.5.0",
     "pytest-testmon>=2.0.0",
     "cryptography",
+    "ansys-pythonnet>=3.1.0rc3; platform_system=='Windows'",
+    "pywin32>=303; platform_system=='Windows'",
```
SMoraisAnsys:
```suggestion
```
Dependency groups should only define dev requirements. If you need extra things then you must add the associated install target / extra when installing pyaedt for testing, e.g. `uv install  pyaedt[all] --group tests`

pyproject.toml
```diff
@@ -365,3 +367,9 @@ directory = "test"
 name = "Test"
 showcontent = true
 
+[tool.pyright]
+include = ["src", "tests"]
+venvPath = "."
+venv = ".venv"
+pythonVersion = "3.10"
+extraPaths = ["src"]
```
SMoraisAnsys:
```suggestion
```
I'm working on adding `ty` which is equivalent to `pyright` but faster and maintained by astral. Can you remove that .

uv.lock
```diff
@@ -3728,6 +3728,8 @@ tests = [
     { name = "pytest-testmon", specifier = ">=2.0.0" },
     { name = "pytest-timeout", specifier = ">=2.3.0" },
     { name = "pytest-xdist", specifier = ">=3.5.0" },
+    { name = "ansys-pythonnet", marker = "sys_platform == 'win32'", specifier = ">=3.1.0rc3" },
+    { name = "pywin32", marker = "sys_platform == 'win32'", specifier = ">=303" },
```
SMoraisAnsys:
```suggestion
```

SMoraisAnsys (changes requested):
Here are some minor comments, thanks for the contribution !

src/ansys/aedt/core/desktop.py
ramin4667: In PR #7847 lines 948-962 the connection mode depend on the actual session not the global preference. After init, self.is_grpc_api reflects the real mode; settings.use_grpc_api is only what the user asked for when starting a session. When on main, odesktop checks settings.use_grpc_api it returns settings.use_grpc_api = True that is default config for tests and also returns self.is_grpc_api = False that is COM config for FilterSolutions.    
Then odesktop takes the gRPC path, calls grpc_plugin, and fails with AttributeError: 'Desktop' object has no attribute 'grpc_plugin'.
odesktop should follow the session this desktop instance is using, not the global preference. That’s the same idea as #7847.
Using self.is_grpc_api is intentional and aligned with #7847: when Filter Solutions (or any flow) attaches via COM while settings.use_grpc_api remains True, the instance is COM but odesktop was still entering the gRPC retry path and failing on missing grpc_plugin.

doc/changelog.d/7860.fixed.md
ramin4667: Removed

.github/workflows/manual_draft.yml
ramin4667: Updated

pyproject.toml
ramin4667: Removed

pyproject.toml
ramin4667: Removed

src/ansys/aedt/core/desktop.py
Samuelopez-ansys: This is correct.

Samuelopez-ansys:
Hi @ramin4667 , tests are failing locally, I do not understand how they will work if use_grpc is set to True.

Samuelopez-ansys:
How can I run these tests locally?

ramin4667:
> Hi @ramin4667 , tests are failing locally, I do not understand how they will work if use_grpc is set to True.

@Samuelopez-ansys , you are right, I set it to False.

PR #7918, opened 2026-07-15
FIX: Connect to existing student version

### LINKED ISSUE #7884: Attaching to a remote gRPC session validates the port on localhost: `_validate_port()` called without `machine`
### Description

When attaching to an already-running **remote** AEDT gRPC session (`new_desktop=False`, `machine=<remote-host>`, `port=<port>`), the initial port validation in `Desktop.__init_grpc` calls `self._validate_port(self.port)` **without forwarding the machine argument**. `_validate_port` falls back to `machine=None`, so `is_grpc_session_active` scans **localhost** instead of the remote host. Since no session exists locally on that port, PyAEDT either flips `new_desktop = True` and silently starts a *local* instance, or the attach fails with `Failed to connect to Desktop Session`.

The helpers already accept `machine` since PR #7495 (`_validate_port(self, port, machine=None)`, `is_grpc_session_active(port, machine)`), but the two call sites in `__init_grpc` were never wired up.

### Steps to reproduce

1. On a remote Windows machine, start AEDT with the classic insecure gRPC server: `ansysedt.exe -grpcsrv 50051 -ng`
2. On the client (pyaedt 1.2.0), use the documented legacy-transport workaround:
   ```python
   import os
   os.environ["PYAEDT_USE_PRE_GRPC_ARGS"] = "True"
   from ansys.aedt.core import Hfss
   settings.grpc_secure_mode = False
   hfss = Hfss(machine="<remote-host>", port=50051, new_desktop=False, version="2026.1")
   ```
3. The attach fails (or a local instance is spawned) even though the remote session is alive, because the port was validated on localhost.

### Root cause / suggested fix

`desktop.py` (v1.2.0), `__init_grpc`, two call sites around lines 3266-3279:

```python
# current
self.__port = self._validate_port(self.port)
...
self.__port = self._validate_port(self.port)

# fixed (2-line change)
self.__port = self._validate_port(self.port, self.machine)
...
self.__port = self._validate_port(self.port, self.machine)
```

With this patch applied locally, attach to a manually started `-grpcsrv` server through the legacy transport works again — verified on AEDT 2026 R1 / Windows / pyaedt 1.2.0, with the connected `odesktop.GetProcessID()` matching the pre-existing server PID (i.e. a genuine attach, not a silently spawned new instance), and repeatable re-attach after `release_desktop`.

### Environment

- pyaedt 1.2.0, AEDT 2026 R1, Windows 11
- Related: #7871 (same error-message family, open), #7360 (user diagnosed the same localhost validation and shorted `is_grpc_session_active` to connect), #7495 (added the `machine` parameter to the helpers but not to these call sites)

### LINKED ISSUE #7891: AEDT 2025 R2 Student version fails to launch for PyAEDT 1.2.0.

### Description of the bug

I understand that this issue had been attempted to be patched in 1.2.0 but running PYAEDT_USE_PRE_GRPC_ARGS=True + grpc_secure_mode=False does not solve the issue.

A way to make it work is patching `general_methods.py`'s `active_sessions()` to set the default of `student_version` from False to True (This is not a permanent solution and the real fix should be passing arguments into `is_grpc_session_active`). This is caused by `is_grpc_session_active` calling `active_sessions` without arguments, resulting in the grpc check failing since it would be checking for a process with a different name.

Note that in this version, the GRPC_ARGS=True and secure_mode=False patch does not need to be applied for it to work.

### Steps To Reproduce

Same reproduction steps as #7842.

Windows

3.10

### Installed packages

annotated-doc==0.0.4
annotated-types==0.7.0
ansys-api-edb==0.3.2
ansys-edb-core==0.3.3
ansys-tools-common==0.5.1
attrs==26.1.0
certifi==2026.6.17
charset-normalizer==3.4.9
click==8.4.2
colorama==0.4.6
defusedxml==0.7.1
grpcio==1.82.1
idna==3.18
jsonschema==4.26.0
jsonschema-specifications==2025.9.1
markdown-it-py==4.2.0
mdurl==0.1.2
numpy==2.2.6
platformdirs==4.10.0
protobuf==6.33.6
psutil==7.2.2
pyaedt==1.2.0
pydantic==2.13.4
pydantic_core==2.46.4
pyedb==0.80.1
Pygments==2.20.0
PyYAML==6.0.3
referencing==0.37.0
requests==2.34.2
rich==15.0.0
rpds-py==0.30.0
scipy==1.15.3
scooby==0.11.2
shellingham==1.5.4
toml==0.10.2
tomli==2.4.1
tomli_w==1.2.0
typer==0.26.8
typing-inspection==0.4.2
typing_extensions==4.16.0
urllib3==2.7.0
xmltodict==1.0.4

Samuelopez-ansys:
## Description
Student version was connecting to existing session.

Platform: windows, linux

## Issue linked
Close #7891 
Close #7884 

## Checklist


maxcapodi78 (dismissed):

eblanco-ansys (dismissed):

SMoraisAnsys (dismissed):

tzhou-yyds:
Verified on Windows 11, AEDT 2026 R1, this PR branch (1.3.dev0):

✅ **Attach to existing session works** — started `ansysedt.exe -grpcsrv 50700 -ng` manually, then `Desktop(version="2026.1", port=50700, new_desktop=False, non_graphical=True)` connects to the same process (verified `odesktop.GetProcessID()` == manual instance PID, ruling out a silent new instance). `release_desktop(close_projects=False, close_on_exit=False)` leaves the instance alive as expected. This resolves our #7884 local-attach scenario.

⚠️ Two observations from the same test bench:
1. With the legacy workaround env var `PYAEDT_USE_PRE_GRPC_ARGS=True` set, the same attach **silently launches a new instance** instead (PID mismatch, no error raised). Users who adopted that workaround for #7884 will hit silent failures after upgrading — a code warning or docs note might be worth it.
2. After an abnormally interrupted session, re-attaching to the same port can also silently spawn a new instance until stale processes exit.

Remote-machine attach (the original #7884 remote case) was not tested — no routable gRPC path between our two machines.

Samuelopez-ansys:
Hi @tzhou-yyds , thanks for the feedback, I applied a small fix that could impact the issue you mentioned, I tested 2026R1:

app_1 = Hfss(version="2026.1", port=50700, new_desktop=False, non_graphical=True, machine="127.0.0.1")

With an open AEDT in non graphical mode with the env. variable PYAEDT_USE_PRE_GRPC_ARGS, and it not open a new AEDT instance.

tzhou-yyds:
Re-tested with the latest `fix/issue_7891` (41658c8) on 2026R1, Windows 11 — **the small fix works**:

- **Case B** (`PYAEDT_USE_PRE_GRPC_ARGS=True` + `Desktop(version='2026.1', port=50700, new_desktop=False, non_graphical=True)`): attaches to the manually launched `ansysedt -grpcsrv 50700 -ng` instance, `GetProcessID()` matches the manual PID — **no new instance spawned** (previously this env var caused a silent new instance).
- **Case C** (your exact form: `Hfss(version='2026.1', port=50700, new_desktop=False, non_graphical=True, machine='127.0.0.1')` with the env var set): attach OK, PID matches as well.
- `release_desktop(close_projects=False, close_desktop=False)` leaves the manual instance alive as expected in both cases.

Thanks for the quick turnaround!

src/ansys/aedt/core/desktop.py
```diff
@@ -3085,39 +3081,75 @@ def _check_machine(self) -> None:
...
+
+        version_neg = self.aedt_version_id[2:4] + self.aedt_version_id[5]
+        version_neg += "_nongraphical" if not self.non_graphical else "_graphical"
+        version_neg += "_student" if self.student_version else ""
+
+        if self.new_desktop:
+            for el in all_sessions.values():
+                if self.port in el.values():
+                    self.logger.warning(f"Port {self.port} is already in use. Finding a new free port.")
+                    self.port = _find_free_port()
```
SMoraisAnsys: I think we shoudl directly play with `__port` instead of the `port` property. That way we are sure that any change in the property (like this PR) wouldn't affect the inner logic. Open to discussion though :)

src/ansys/aedt/core/generic/general_methods.py
```diff
@@ -1492,6 +1508,109 @@ def active_sessions(
...
+    for version, sessions in return_dict_filtered.items():
+        if any(port == -1 for port in sessions.values()):
+            for pid in [i for i, v in sessions.items() if v == -1]:
+                version_number = version.replace("_student", "").replace("_nongraphical", "").replace("_graphical", "")
+                sessions[pid] = _check_connection_grpc_port(
+                    connections,
+                    pid,
+                    version_number,
+                    True if "nongraphical" in version else False,
+                )
```
SMoraisAnsys: I'm not sure how this step interact with the returned value, isn't there something missing here ?

src/ansys/aedt/core/generic/general_methods.py
```diff
@@ -1492,6 +1508,109 @@ def active_sessions(
     return return_dict_filtered
 
 
+@pyaedt_function_handler()
+def all_active_sessions() -> dict[str, dict]:
```
SMoraisAnsys:
```suggestion
def all_active_sessions() -> dict[str, dict[int, int]]:
```

src/ansys/aedt/core/generic/general_methods.py
```diff
@@ -1492,6 +1508,109 @@ def active_sessions(
...
+        4. **TCP Connection Analysis**: Falls back to checking active TCP connections via psutil.
+
+    Port Detection Results:
+        - Positive integer, gRPC session on that port.
+        - ``-1``: COM session (no gRPC server running).
+
+
+    Returns
+    -------
+    dict[str, dict]
```
SMoraisAnsys:
```suggestion
    dict[str, dict[int, int]]
```

src/ansys/aedt/core/generic/general_methods.py
```diff
@@ -1492,6 +1508,109 @@ def active_sessions(
...
+        Dictionary mapping AEDT process IDs to their corresponding ports.
```
SMoraisAnsys:
```suggestion
        Dictionary mapping AEDT version to the associated process IDs and their corresponding ports.
```

src/ansys/aedt/core/desktop.py
```diff
@@ -3085,39 +3081,75 @@ def _check_machine(self) -> None:
...
+                    return self.port
+            # No active sessions found, open a new AEDT session
             self.new_desktop = True
-        return port
+            return self.port
 
     @pyaedt_function_handler()
     def _assign_port(self):
-        self.__port = 0
+        self.port = 0
```
SMoraisAnsys: Same remark as above, I would keep the `__port` value

src/ansys/aedt/core/generic/general_methods.py
```diff
@@ -1492,6 +1508,109 @@ def active_sessions(
...
+        Port is set to ``-1`` if the session is using COM instead of gRPC.
+
+    Examples
+    --------
+    Get all active AEDT sessions (any version, any mode):
+
+
+    """
+    # Step 1: Determine target process names based on version type and operating system
```
SMoraisAnsys: Steps do not align with the docstring description and misses the second step :p

tests/unit/test_validate_port_sequence.py
```diff
@@ -0,0 +1,294 @@
...
+        assert d1._validate_port() == base_port
+
+    # Start 2026.1 non-graphical session (no active sessions yet), new_desktop is False, and PyAEDT will flip it
+    d2 = _make_desktop(port=base_port, version="2026.1", student_version=False, non_graphical=False, new_desktop=False)
+    with patch("ansys.aedt.core.desktop.all_active_sessions", return_value={}):
+        assert d2._validate_port() == base_port
+        assert d2.new_desktop
+
+
+def test_new_session_port_0(mock_settings):
```
SMoraisAnsys: Please split that into multiple tests isntead of one gathering everything. Same comment for the following ones

SMoraisAnsys (dismissed):
Overall LGTM but there is something bothering me with "Step 6". Otherwise, I left minor comments.

src/ansys/aedt/core/generic/general_methods.py
Samuelopez-ansys: it is working because sessions[pid] is updating the value of the return_dict_filtered, due to the mutable python property, but I will update it to make it more clear

Samuelopez-ansys:
Hi @SMoraisAnsys , I applied your suggestions, please could you review it again? 

I reverted the self.__port set instead of using the property setter, but I also modified it in other places of desktop.py to make it consistent, please take a look. I do not understand why this is better than using the property setter, in my opinion in was more clean, but I also understand that if the property setter changes, it could be dangerous, and actually it opens the door to the question if the self.port property should have a setter, or not, because it is a very sensitive variable.

eblanco-ansys (dismissed):

maxcapodi78 (dismissed):
LGTM

doc/source/User_guide/desktop_sessions.rst
```diff
@@ -108,3 +108,35 @@ Recommendations
...
+  requested port is already in use by any active session, PyAEDT chooses a different
+  free port (``_find_free_port``) so the new session can be started.
+- If a remote RPyC RPC connection is configured (``settings.remote_rpc_session``): PyAEDT
+  uses the remote session and short-circuits further local port checks (``new_desktop`` is
+  set to ``False`` and the requested port is used).
+- If there is an active session for the same AEDT version and the requested port is used by
+  that session, PyAEDT connects to it (reuse).
+- If there is an active session for the same AEDT version but the opposite display mode
+  (graphical vs non-graphical) using the requested port, PyAEDT flips the ``non_graphical``
+  flag and connect to that session.
```
SMoraisAnsys:
```suggestion
  flag and connect to that session (reuse).
```

src/ansys/aedt/core/generic/general_methods.py
```diff
@@ -1338,7 +1349,12 @@ def is_grpc_session_active(port: int, machine: str | None = None) -> bool:
     if machine and machine not in ["localhost", "127.0.0.1", "::ffff:127.0.0.1", socket.gethostname()]:
         return _is_port_occupied(port, machine)
 
-    return True if port in active_sessions().values() else False
+    return (
+        True
+        if port
+        in active_sessions(version=version, student_version=student_version, non_graphical=non_graphical).values()
+        else False
+    )
```
SMoraisAnsys, line 1354:
```suggestion
    return (
        port
        in active_sessions(version=version, student_version=student_version, non_graphical=non_graphical).values()
    )
```
The check itself is enough

src/ansys/aedt/core/generic/general_methods.py
```diff
@@ -1492,6 +1508,136 @@ def active_sessions(
...
+
+    Returns
+    -------
+    dict[str, dict[int, int]]
+        Nested dictionary structure:
+
+        - **Outer key** (str): AEDT version identifier with format:
+          ``v<version>_<mode>[_student]`` (e.g., ``v241_graphical``, ``v251_nongraphical_student``)
+        - **Inner key** (int): Process ID (PID) of the AEDT session
+        - **Inner value** (int): gRPC port number, or ``-1`` for COM sessions
```
SMoraisAnsys, line 1554: Very good idea, really helpful

src/ansys/aedt/core/generic/general_methods.py
```diff
@@ -1492,6 +1508,136 @@ def active_sessions(
...
+
+    Examples
+    --------
+    Get all active AEDT sessions (any version, any mode):
+
+    >>> from ansys.aedt.core.generic.general_methods import all_active_sessions
+    >>> sessions = all_active_sessions()
+
+    """
+    # Step 1: Determine target process names based on version type and operating system
```
SMoraisAnsys: Feels like step 1 and 2 should be merged, I don't see the point to split the definition of variable `targets` and the docstring states that step 1 searches for AEDT processes.

src/ansys/aedt/core/generic/general_methods.py
```diff
@@ -1492,6 +1508,136 @@ def active_sessions(
...
+
+    targets = ["ansysedtsv.exe", "ansysedt.exe"]
+
+    # Step 2: Get all matching AEDT processes from the system
+    # Returns list of tuples: [(pid, command_line_args), ...]
+
+    target_processes = _get_target_processes(targets)
+
+    # AEDT processes launched
+    return_dict = {pid: -1 for pid, _ in target_processes}
```
SMoraisAnsys: For consistency with other functions, should we name the return variable `res` instead ? I know this has been used in multiple places and could ease our future review process if applied in every place.

src/ansys/aedt/core/desktop.py
```diff
@@ -801,7 +799,7 @@ def __init__(
         )
         self.__close_on_exit_arg = close_on_exit
         self.__machine = machine if machine else None
-        self.__port = port
+        self.__port = port if port is not None else 0
```
SMoraisAnsys, line 802: This might have impact on rpc and/or older version of AEDT but I can't predict what atm :/

src/ansys/aedt/core/desktop.py
```diff
@@ -3094,23 +3090,59 @@ def _check_machine(self) -> None:
         self.machine = "127.0.0.1"
 
     @pyaedt_function_handler()
-    def _validate_port(self, port, machine=None):
+    def _validate_port(
```
SMoraisAnsys, line 3093: The removal of `machine` could also have impact on rpc and/or older version of AEDT. Could you double check that ?

src/ansys/aedt/core/desktop.py
```diff
@@ -3094,23 +3090,59 @@ def _check_machine(self) -> None:
...
+            self._assign_port()
+            return self.port
+        all_sessions = all_active_sessions()
+        version = self.aedt_version_id[2:4] + self.aedt_version_id[5]
+        version += "_nongraphical" if self.non_graphical else "_graphical"
+        version += "_student" if self.student_version else ""
+
+        version_neg = self.aedt_version_id[2:4] + self.aedt_version_id[5]
+        version_neg += "_nongraphical" if not self.non_graphical else "_graphical"
+        version_neg += "_student" if self.student_version else ""
```
SMoraisAnsys:
```suggestion
        base = self.aedt_version_id[2:4] + self.aedt_version_id[5]
        student = "_student" if self.student_version else ""
        mode, mode_neg = (
            ("_nongraphical", "_graphical") if self.non_graphical else ("_graphical", "_nongraphical")
        )
        version = f"{base}{mode}{student}"
        version_neg = f"{base}{mode_neg}{student}"
```
While reading I didn't catch (at first) that only gaphical was negated. This could help (open remark, feel free to discard)

src/ansys/aedt/core/desktop.py
```diff
@@ -3094,23 +3090,59 @@ def _check_machine(self) -> None:
...
+
+        if self.new_desktop:
+            for el in all_sessions.values():
+                if self.port in el.values():
+                    self.logger.warning(f"Port {self.port} is already in use. Finding a new free port.")
+                    self.__port = _find_free_port()
+                    break
+            return self.port
+        elif settings.remote_rpc_session:  # remote session -> no port check
+            self.logger.warning(f"Remote session found on port {self.port}. Using it.")
```
SMoraisAnsys:
```suggestion
            self.logger.warning(f"Remote session activated, no port checking.")
```
I don't think that we are checking that here so I would remove that log or modify it as the comment above says.

src/ansys/aedt/core/desktop.py
```diff
@@ -3094,23 +3090,59 @@ def _check_machine(self) -> None:
...
+            self.new_desktop = False
+            return self.port
+        elif version in all_sessions and self.port in all_sessions[version].values():
+            self.logger.info(f"Port {self.port} session has been found.")
+            return self.port
+        elif version_neg in all_sessions and self.port in all_sessions[version_neg].values():
+            mode = "graphical" if self.non_graphical else "non_graphical"
```
SMoraisAnsys, line 3128: If my previous refactoring is used, this `mode` could probably be modified to use a variable already defined.

tests/unit/test_validate_port_sequence.py
```diff
@@ -0,0 +1,309 @@
...
+    return m
+
+
+def test_new_session_no_active_sessions(mock_settings):
+    """New AEDT session on port 50051, no active sessions."""
+    # Start 2026.1 graphical session (no active sessions yet)
+    d1 = _make_desktop(
+        port=DEFAULT_PORT, version="2026.1", student_version=False, non_graphical=False, new_desktop=True
+    )
+    with patch("ansys.aedt.core.desktop.all_active_sessions", return_value={}):
```
SMoraisAnsys: Can't that patch being added at the test level (same for other tests) ?

SMoraisAnsys (commented):
Thanks for the changes @Samuelopez-ansys Here are a few comments.
Given how this interacts with the desktop session handling, could you double check that this doesn't break anything with older versions of AEDT ?

src/ansys/aedt/core/desktop.py
Samuelopez-ansys: If passed port is None, it is safer to convert it here to 0 because None is not accepted by AEDT API. So I think this is better and I did not find any case that it could fail.

src/ansys/aedt/core/desktop.py
Samuelopez-ansys: No call sites pass a machine argument to _validate_port. All internal calls use self._validate_port() with no args (e.g., in __init_grpc), and unit tests call it without arguments.

_validate_port already handles remote RPC via settings.remote_rpc_session and uses self.machine elsewhere (e.g., launch logic and older-version mode decisions). Removing a machine parameter from _validate_port therefore does not break RPC handling or the older-AEDT-version decision tree because those rely on settings.remote_rpc_session or self.machine attributes, not on a separate machine parameter.

tests/unit/test_validate_port_sequence.py
```diff
@@ -0,0 +1,334 @@
...
+@patch("ansys.aedt.core.desktop.all_active_sessions")
+@patch("ansys.aedt.core.desktop._find_free_port")
+def test_new_session_port_0_with_active_sessions_different_version(find_free_port, all_active_sessions, mock_settings):
+    """New AEDT session on port 0 with active sessions in different versions."""
+    # Start 2026.1 (new_desktop=True)
+    d4 = _make_desktop(port=BASE_PORT, version="2026.1", student_version=False, non_graphical=False, new_desktop=True)
+    all_active_sessions.return_value = ACTIVE_SESSIONS
+    find_free_port.return_value = RANDOM_PORT
+    assert d4._validate_port() == RANDOM_PORT
+    d4._Desktop__logger.info.assert_called_with(f"New AEDT session is starting on gRPC port {RANDOM_PORT}.")
```
SMoraisAnsys: I would define the return value of `all_active_sessions` and `find_free_port` in the decorator. Or at least define the return value before declaring `d4`. That way, we know that the setup is finished when we define the object  (`d4`) to evaluate.

SMoraisAnsys (dismissed):
LGTM, if I can ask for a last change it would be to write the test in such a way : setup, exercice, assert, teardown (if needed). Here the fixture and patch do a nice job but it feels like some of the setup is performed after the object to evaluate is created. If possible I would first define everything required, then define the object to evaluate and finally perform the asserts. I try to give more information in my comment.

SMoraisAnsys (approved):

Issue #7943, opened 2026-07-28
Bug located in change_stackup_type


maxcapodi78:
### Before submitting the issue

- [x] I have searched among the existing issues
- [x] I am using a Python virtual environment

### Description of the bug

the actual method in pyaedt is applying dielectric thickness ignoring completely the fill_material of signal layers.

<img width="1027" height="315" alt="Image" src="https://github.com/user-attachments/assets/74fdb848-19bd-41c7-b6f3-e3c79d0cc3a6" />

<img width="1037" height="433" alt="Image" src="https://github.com/user-attachments/assets/57043a50-186f-4a49-9620-a968c4dd5e09" />

### Steps To Reproduce

open a laminate stackup
run
h3d.modeler.stackup.change_stackup_type("Overlap")

### Which Operating System are you using?

Windows

### Which Python version are you using?

3.10

### Installed packages

-

PR #7944, opened 2026-07-28
FIX: Improve the method change_stackup_type 

maxcapodi78:
## Description
Improve the method change_stackup_type to automatically merge dielectric layers as expected in case of switch from laminate to overlap or viceversa.
When switching from laminate to overlapping there is a dielectric simplification carried out in the UI which was not performed into the pyaedt method. the new alghoritm mimic the same behaviour trying to merge the dielectrics during the stackup type change.

  Application:  hfss3dlayout
  AEDT-Version: 24.1, 24.2, 25.1, 25.2, 26.1

## Issue linked
#7943 

## Checklist

### DIFF
--- a/src/ansys/aedt/core/modules/layer_stackup.py
+++ b/src/ansys/aedt/core/modules/layer_stackup.py
@@ -27,6 +27,8 @@
 The module provides all layer stackup functionalities for the Circuit and HFSS 3D Layout tools.
 """
 
+from copy import deepcopy
+
 from ansys.aedt.core.base import PyAedtBase
 from ansys.aedt.core.generic.constants import unit_converter
 from ansys.aedt.core.generic.data_handlers import str_to_bool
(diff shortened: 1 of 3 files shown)


src/ansys/aedt/core/modules/layer_stackup.py
```diff
@@ -1844,12 +2095,28 @@ def change_stackup_type(self, mode: str = "MultiZone", number_zones: int = 3) ->
         else:
             self.logger.error("Stackup mode has to be Multizone, Overlap or Laminate.")
             return False
-        for v in list(self.layers.values()):
+        len_ll = len(list(self.layers.values())) + 1
```
SMoraisAnsys: Would it make sense to rename that as "layer_iterator" ? I was wondering why the +1 was added and understood it below

tests/unit/test_layer_stackup.py
```diff
@@ -0,0 +1,148 @@
...
+            "NAME:Sublayer",
+            "Thickness:=",
+            thickness,
+            "LowerElevation:=",
+            lower_elevation,
+            "Material:=",
+            material,
+        ],
+    ]
+    if layer_id is not None:
```
SMoraisAnsys, line 53: Should this "layer_id" handling be removed? I don't see any code where the value is `None`.

tests/unit/test_layer_stackup.py
```diff
@@ -0,0 +1,148 @@
...
+            thickness,
+            "LowerElevation:=",
+            lower_elevation,
+            "Material:=",
+            "copper",
+            "FillMaterial:=",
+            fill_material,
+        ],
+    ]
+    if layer_id is not None:
```
SMoraisAnsys, line 77: Same comment as before

src/ansys/aedt/core/modules/layer_stackup.py
```diff
@@ -1809,6 +1811,255 @@ def add_layer(
...
+                and cls._get_argument_value(layer_args, "Type:=") == "signal"
+            ):
+                sublayer = cls._get_stackup_sublayer(layer_args)
+                interval = cls._get_stackup_sublayer_interval(sublayer, default_unit) if sublayer else None
+                if interval:
+                    signal_layers.append((layer_args, sublayer, interval))
+
+        split_args = []
+        split_index = 1
+        tolerance = 1e-12
```
SMoraisAnsys, line 1950: Should that be an input ?

SMoraisAnsys (dismissed):
Given the newly added code, I'm wondering if it wouldn't be beneficial to create a separate class (even if only defined in the scope of the `Layers` class) to handle the "argument" logic ? Just an open question, the code seems fine otherwise :)

src/ansys/aedt/core/modules/layer_stackup.py
Alberto-DM: no, because this tolerance is necessary to deal with a known AEDT issue. Sometimes some floats like 0.2 became 0.19999999999999999. They are even show like that in the UI.

tests/unit/test_layer_stackup.py
Alberto-DM: True, but it might be `None` in future tests for layer stackup. This is a new test file and more tests are expected.

tests/unit/test_layer_stackup.py
Alberto-DM: same as before

src/ansys/aedt/core/modules/layer_stackup.py
Alberto-DM: changed in len_layers_list

Alberto-DM (approved):
LGTM

PR #8022, opened 2026-08-24
DOCS: Add note on documentation's landing page about doc artifacts for older versions

ecoussoux-ansys:
## Description
This PR is a follow-up to #8003.
Unfortunately, the changes introduced in that PR did not fix the `pages build and deployment` workflow, which was until now still failing on the `deploy` job.

This issue could ultimately be resolved by removing the documentation related to earlier versions of the project from the `gh-pages` branch (see latest run of the `pages build and deployment` workflow [here](https://github.com/ansys/pyaedt/actions/runs/32735832844)): Currently the documentation available at https://aedt.docs.pyansys.com/ covers up to the version 0.27 of pyaedt.

This PR adds a brief note to the documentation's landing page, stating that doc artifacts for all previous versions of pyaedt can be downloaded directly from the GitHub releases page.

## Issue linked
N/A

## Checklist


ecoussoux-ansys:
Many aspects are open for discussion of course :
1. The wording of the note I added obviously but also I recognize that it may not deserve to be on such a prominent location on the landing page, so it can be moved as well !
2. My manual commit on the `gh-pages` branch (see https://github.com/ansys/pyaedt/commits/gh-pages/) has removed the documentation of versions 0.14 to 0.26 included. This can be easily adjusted/reverted !
3. ~~I'd suggest keeping the changes introduced in #8003. This was suggested by Revathy as a better practice (pushing smaller commits to that branch).~~ After discussing with @SMoraisAnsys, changes performed in #8003 are reverted: The history on the `gh-pages` branch is growing too fast (especially through the use of the `doc-deploy-pr` action), which is not sustainable in the long run.
4. The solution suggested by Eduardo (archiving the cutout versions on a dedicated branch and displaying them under a different doc page) is not being pursued, because it appears that there are no mechanisms available to effectively implement it (see https://github.com/orgs/community/discussions/47867).

eblanco-ansys (dismissed):

doc/source/conf.py
```diff
@@ -323,6 +323,14 @@ def setup(app):
         "limit": 10,
         "ignoreLocation": True,
     },
+    "announcement_banner": [
+        {
+            "message": "Documentation artifacts for versions prior to 0.27 are available for download "
```
SMoraisAnsys:
```suggestion
            "message": "Documentation artifacts for older versions are available for download "
```

SMoraisAnsys (changes requested):
Marking as request changes to ensure discussions before merging.
Can you bring that topic to our team's meeting because I think that the banner is not a good solution for this kind of situation. If this banner was only appearing on the "Older version" page, I wouldn't mind. However, I don't think we have that ability and I would prefer to avoid having stacked announcement if possible. My main reason being that I'm not sure if users would check that banner for this kind of information. Other reasons include people just closing the banner because it takes place on the screen which leads to no information available :/

ecoussoux-ansys:
> Marking as request changes to ensure discussions before merging. Can you bring that topic to our team's meeting because I think that the banner is not a good solution for this kind of situation. If this banner was only appearing on the "Older version" page, I wouldn't mind. However, I don't think we have that ability and I would prefer to avoid having stacked announcement if possible. My main reason being that I'm not sure if users would check that banner for this kind of information. Other reasons include people just closing the banner because it takes place on the screen which leads to no information available :/

Thanks for your feedback @SMoraisAnsys !
Sure, I'll gather input from the rest of our team - certainly during next week's meeting.
I agree with your comment that the announcement banner is probably not that appropriate. And that info likely does not belong directly on the landing page (adding a third note there is making the top of that page too cluttered).
For reference, here are screenshots for these two options : 

image

---

image

---

Your suggestion of a message or banner on the "version page" matches what @eblanco-ansys suggested me but indeed, the `ansys-sphix-theme` does not provide a control to insert such content in that automatically generated page.
Other projects within the Pyansys ecosystem are currently dealing with the issue of removing content related to older versions from the online doc; let's see if a common approach to informing users about this can be found.

Samuelopez-ansys:
Hi @ecoussoux-ansys , this PR is on hold right? Just to move it to draft

ecoussoux-ansys:
I am eventually closing this PR:
*   Following discussions within the PyAnsys Core Team, it was agreed to aim at a generic solution to the `gh-pages` branch overflow issue, as PyAEDT is not the only project facing this problem. See: https://github.com/ansys/actions/issues/1530
*   Once a mechanism is in place to address this issue, it will be implemented for PyAEDT as well. Waiting for that unified approach is acceptable, since in the meantime, the `pages build and deployment` workflow remains functional for PyAEDT (after having manually cut out some content from the `gh-pages` branch as described above).
*   The ecosystem-wide solution should also introduce an appropriate way for informing users on how to access documentation for older project versions (the original topic of this PR). The two options suggested here (a note on the landing page or an announcement banner) were rejected as unsatisfactory during the review. It is absolutely possible that a note / banner could be added to the page listing the documentation versions available online (https://aedt.docs.pyansys.com/version/index.html); note that this page is generated by the `ansys/action` `_doc-geng-versions-page` (internally called by the `doc-deploy-dev` and `doc-deploy-stable` actions).
*   Reverting the changes made in #8003 (`force-orphan: false` for the `gh-pages` branch)—which is necessary for PyAEDT, as discussed in my previous comment—is being handled in #8063.

PR #8025, opened 2026-08-24
Fix AttributeError masking dropped AEDT connection in solution_type fallback

### LINKED ISSUE #8020: AttributeError: 'HfssConstants' has no attribute 'default_solution' masks the real connection error (constants define solution_default)
## Description

`design_solutions.py` falls back to `self._design_type.default_solution` when `GetSolutionType()` raises, but every constant class in `aedt_constants.py` defines the attribute as `solution_default`. The fallback therefore raises `AttributeError` itself, and that error replaces the original one.

This matters because of *when* the fallback runs: it is reached exactly when the gRPC connection to AEDT has broken. The user then sees an error naming a design-type constant, which points investigation at design types — where nothing is wrong — instead of at the lost connection, which is the actual failure. It cost us most of a debugging session before we read the source.

The same file uses both spellings, a few lines apart, which is what makes this easy to miss (line numbers from 1.3.0):

```python
# DesignSolution.solution_type getter, line 74 - correct
elif self._odesign:
    try:
        self._solution_type = self._odesign.GetSolutionType()
    except Exception:
        self._solution_type = self._design_type.solution_default   # OK

# DesignSolution.solution_type setter, line 94 - raises
elif self._odesign:
    try:
        self._solution_type = self._odesign.GetSolutionType()
    except Exception:
        self._solution_type = self._design_type.default_solution   # AttributeError

# HFSSDesignSolution.solution_type getter, line 218 - raises
    except Exception:
        self._solution_type = self._design_type.default_solution   # AttributeError
```

## Steps to reproduce

No AEDT installation or licence required:

```bash
python -c "from ansys.aedt.core.generic.aedt_constants import HfssConstants as H; print(hasattr(H, 'solution_default'), hasattr(H, 'default_solution'))"
```

Output on 1.3.0:

```
True False
```

Every constant class is affected, not only `HfssConstants` — `Q3dConstants`, `IcepakConstants`, `Maxwell3dConstants`, `CircuitConstants` and the rest all define `solution_default`.

To observe the masking end to end: connect to AEDT over gRPC, cause the session to drop, then read any design property. The `AttributeError` above is raised instead of a connection error.

## Occurrence count

Counting occurrences in `src/ansys/aedt/core/application/design_solutions.py`:

| version | `solution_default` (correct) | `default_solution` (raises) |
|---|---:|---:|
| 1.3.0 | 3 | 9 |
| 1.4.0 | 1 | 3 |
| main (2026-08-24) | 0 | 2 |

The counts fall because the file has been refactored, not because the bug was addressed: the wrong spelling survives in all three, including in `HFSSDesignSolution`.

## Suggested fix

Rename the remaining `default_solution` accesses to `solution_default`. The constants themselves need no change.

Separately, chaining the original exception would stop this class of masking regardless of any attribute name:

```python
except Exception as exc:
    raise AedtRuntimeError("lost connection to AEDT") from exc
```

## Which version are you using?

1.3.0 (source of 1.4.0 and main also inspected)

## Which operating system are you using?

Windows

## What is your Python version?

3.11

## Are you using any additional Ansys products?

AEDT 2026 R1, connected over gRPC.

agu2347:
### Bug

Reported in #8020. `design_solutions.py` falls back to `self._design_type.default_solution` when `GetSolutionType()` raises, but every constant class in `generic/aedt_constants.py` (`HfssConstants`, `Q3dConstants`, `IcepakConstants`, `Maxwell3dConstants`, `CircuitConstants`, etc.) only defines `solution_default`:

```
python -c "from ansys.aedt.core.generic.aedt_constants import HfssConstants as H; print(hasattr(H, 'solution_default'), hasattr(H, 'default_solution'))"
True False
```

The fallback therefore raises `AttributeError` itself. Because this fallback runs exactly when the gRPC connection to AEDT has broken, the user sees `'HfssConstants' object has no attribute 'default_solution'` instead of a connection error, which points debugging at design types instead of the actual failure.

`DesignSolution.solution_type`'s getter already uses the correct `solution_default` spelling; the setter (`value is None` branch) and both of `HFSSDesignSolution.solution_type`'s getter and setter still used `default_solution`. On current `main` this affects 9 call sites in `src/ansys/aedt/core/application/design_solutions.py`, all identical in shape (`self._design_type.default_solution`).

### Fix

Rename all 9 remaining `default_solution` accesses to `solution_default`, matching every other call site in the file and the actual attribute the constant classes define. No changes needed to the constants themselves.

### Testing

No AEDT installation or license is required to reproduce or verify this — the bug is a pure attribute-name mismatch between two Python modules.

Added `tests/unit/test_design_solutions.py`:
- `test_constants_only_define_solution_default` — confirms the constant classes define `solution_default` and not `default_solution`.
- `test_design_solution_getter_falls_back_on_connection_error` — control, exercises the getter's already-correct branch.
- `test_design_solution_setter_falls_back_on_connection_error` — regression test for the setter's `value is None` branch, using a fake `ODesign` whose `GetSolutionType()` raises to simulate a dropped connection.
- `test_design_solution_setter_no_odesign_falls_back` — regression test for the no-`_odesign` fallback branch.
- `test_hfss_design_solution_getter_falls_back_on_connection_error` — regression test for `HFSSDesignSolution`'s getter.

Verified all 3 regression tests reproduce the exact reported `AttributeError` against the unfixed code (reverted `design_solutions.py` to the current `main` version, reran — 3 failed with `AttributeError: type object 'HfssConstants'/'Q3dConstants' has no attribute 'default_solution'`, 2 passed as controls), then reapplied the fix and reran — all 5 pass.

Fixes #8020

### DIFF
--- a/src/ansys/aedt/core/application/design_solutions.py
+++ b/src/ansys/aedt/core/application/design_solutions.py
@@ -86,14 +86,14 @@ class DesignSolution(PyAedtBase):
                 DesignType.EMIT,
                 DesignType.Q3D,
             ]:
-                self._solution_type = self._design_type.default_solution
+                self._solution_type = self._design_type.solution_default
             elif self._odesign:
                 try:
                     self._solution_type = self._odesign.GetSolutionType()
                 except Exception:
-                    self._solution_type = self._design_type.default_solution
+                    self._solution_type = self._design_type.solution_default
             else:
-                self._solution_type = self._design_type.default_solution
+                self._solution_type = self._design_type.solution_default
         elif value and value in self._solution_options:
             self._solution_type = value
             if self._solution_options[value]["name"]:
@@ -215,9 +215,9 @@ class HFSSDesignSolution(DesignSolution, PyAedtBase):
                 elif "Terminal" in self._solution_type:
                     self._solution_type = "Terminal"
             except Exception:
-                self._solution_type = self._design_type.default_solution
+                self._solution_type = self._design_type.solution_default
         elif self._solution_type is None:
-            self._solution_type = self._design_type.default_solution
+            self._solution_type = self._design_type.solution_default
         return self._solution_type
 
     @solution_type.setter
@@ -246,9 +246,9 @@ class HFSSDesignSolution(DesignSolution, PyAedtBase):
                     elif "Terminal" in self._solution_type:
                         self._solution_type = "Terminal"
                 except Exception:
-                    self._solution_type = self._design_type.default_solution
+                    self._solution_type = self._design_type.solution_default
             else:
-                self._solution_type = self._design_type.default_solution
+                self._solution_type = self._design_type.solution_default
         elif value and value in self._solution_options and self._solution_options[value]["name"]:
             if value == "Transient" or value == "Transient Network":
                 value = "Transient Network"
@@ -415,7 +415,7 @@ class Maxwell2DDesignSolution(DesignSolution, PyAedtBase):
             try:
                 self._solution_type = self._odesign.GetSolutionType()
             except Exception:
-                self._solution_type = self._design_type.default_solution
+                self._solution_type = self._design_type.solution_default
         return self._solution_type
 
     @solution_type.setter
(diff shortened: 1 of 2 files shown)


Samuelopez-ansys:
Hi @agu2347, please sign the CLA, and then I will review your code. Thanks for your contribution!

Samuelopez-ansys:
Hi @agu2347 , please accept the CLA: https://cla.developer.ansys.com/?next=https://github.com/ansys/pyaedt/pull/8025

Samuelopez-ansys:
Hi @agu2347 ,

Thank you for your contribution. I cherry picked it and I created a new PR:

https://github.com/ansys/pyaedt/pull/8072

PR #8070, opened 2026-09-07
FEAT: Set variables in one call

Samuelopez-ansys:
## Description
As title says. PyAEDT was not allowing set variables in batch.

Application: hfss, maxwell, circuit, q3d
AEDT-Version: 25.2, 26.1, 27.1
Platform: windows, linux

## Issue linked
None

## Checklist


src/ansys/aedt/core/application/variables.py
```diff
@@ -1190,28 +1196,126 @@ def set_variable(
...
+            if nm in self.variables:
+                circuit_parameter[i] = self.variables[nm].is_circuit_parameter
+
+        # Separate project variables from design variables
+        project_indices = [i for i, nm in enumerate(names) if nm.startswith("$")]
+        design_indices = [i for i, nm in enumerate(names) if not nm.startswith("$")]
+
+        # Further separate design variables by circuit_parameter (only matters for circuit designs)
+        design_circuit_indices = [i for i in design_indices if circuit_parameter[i]]
+        design_local_indices = [i for i in design_indices if not circuit_parameter[i]]
```
SMoraisAnsys, line 1244: For efficiency, should this be refactored with something like
```python
for i, (nm, is_circuit) in enumerate(zip(names, circuit_parameter)):
    if nm.startswith("$"):
        project_indices.append(i)
        continue
.... some code needs to be added ...
```

This would make the code more verbose but would avoid the multi loops. If those loops can't be too time consuming then ignore my comment :)

src/ansys/aedt/core/application/variables.py
```diff
@@ -1190,28 +1196,126 @@ def set_variable(
...
+                read_only=[read_only[i] for i in design_circuit_indices],
+                hidden=[hidden[i] for i in design_circuit_indices],
+                description=[description[i] for i in design_circuit_indices],
+                sweep=[sweep[i] for i in design_circuit_indices],
+                is_post_processing=[is_post_processing[i] for i in design_circuit_indices],
+                overwrite=overwrite,
+                circuit_parameter=True,
+                is_project=False,
+            )
+            success = success and result
```
SMoraisAnsys, line 1262: Should we interupt the code once a failure is detected instead of going through the whole function ?

src/ansys/aedt/core/application/variables.py
```diff
@@ -1227,131 +1331,128 @@ def set_variable(
...
+            prop_type = "PostProcessingVariableProp"
+
+        if isinstance(expression, str):
+            variable = expression
+        elif isinstance(expression, Variable):
+            variable = expression.evaluated_value
+        elif isinstance(expression, Quantity):
+            variable = str(expression)
+        elif is_number(expression):
+            variable = str(expression)
```
SMoraisAnsys, line 1429:
```suggestion
        elif isinstance(expression, Quantity) or is_number(expression):
            variable = str(expression)
```

SMoraisAnsys (dismissed):
LGTM, I left some minor comments

src/ansys/aedt/core/application/variables.py
```diff
@@ -1259,7 +1259,8 @@ def set_variable(
                 circuit_parameter=True,
                 is_project=False,
             )
-            success = success and result
+            if not result:
+                return False
```
SMoraisAnsys, line 1271: Should we log and raise in that case (and the following ones) ?

src/ansys/aedt/core/application/variables.py
```diff
@@ -1190,28 +1196,133 @@ def set_variable(
...
+            if nm in self.variables:
+                circuit_parameter[i] = self.variables[nm].is_circuit_parameter
+            if nm.startswith("$"):
+                project_indices.append(i)
+            else:
+                design_indices.append(i)
+
+        # Further separate design variables by circuit_parameter (only matters for circuit designs)
+        design_circuit_indices = [i for i in design_indices if circuit_parameter[i]]
+        design_local_indices = [i for i in design_indices if not circuit_parameter[i]]
```
SMoraisAnsys: Sry I should have finished the code, I was thinking about something like

```python
project_indices = []
design_indices = []
design_circuit_indices = []
design_local_indices = []

for i, (nm, is_circuit) in enumerate(zip(names, circuit_parameter)):
    if nm.startswith("$"):
        project_indices.append(i)
        continue

    design_indices.append(i)
    if is_circuit:
        design_circuit_indices.append(i)
    else:
        design_local_indices.append(i)
```

src/ansys/aedt/core/application/variables.py
Samuelopez-ansys: In the internal metho, we already log, so it is not needed and also avoids log duplication (_set_variables_single_call). About the raise, this would introduce a regression, because right now it is returning False when the variable is not correctly created, in the future we could consider chaning to raising exception por PyAEDT 2.0. @SMoraisAnsys

SMoraisAnsys (dismissed):

SMoraisAnsys (approved):

PR #8087, opened 2026-09-10
FIX: Python 3.13 compatibility issues

Samuelopez-ansys:
## Description
Some issues were found with python 3.13 unit tests.

1) Desktop destructor issues.

In Python 3.13, pytest is stricter about unraisable exceptions raised during object finalization. The Desktop destructor was assuming the object had completed initialization and could safely access close_on_exit. In the failing mTLS startup test, the Desktop instance can be partially initialized when teardown runs, so the destructor hit a missing private attribute and raised AttributeError from del.
This produced an unraisable exception during garbage collection, which pytest then reported as an ExceptionGroup and obscured the actual test intent. The fix makes the destructor defensive:
• it reads close_on_exit through getattr with a safe fallback
•it catches any exception raised during cleanup
•it avoids propagating errors from del entirely

This is necessary because destructors must never raise during interpreter shutdown or partial object lifetime, especially under Python 3.13, where pytest surfaces those situations more aggressively.
The change is intentionally small and scoped to cleanup safety; it does not modify the underlying AEDT startup behavior or the test assertion itself. It only prevents destructor-time exceptions from turning a valid negative test into a misleading pytest failure.

2) Remove get_solution_data from report test, it is duplicated and also it is failing randomly (both 3.10 and 3.13).
Random failure in this test. Not needed because get_solution_data is tested in test_get_solution_data.

3) EMIT from 2027R1 is compatible with any python version, so we needed to refactor a bit the load of the modules in the test.

4) A regression in AEDT lists has been fixed, so I reverted the test back.

## Issue linked

None

## Checklist


tests/system/solvers/sequential/test_desktop_grpc_transport.py
```diff
@@ -544,9 +544,17 @@ def test_desktop_default_mtls_failure_due_to_bad_certificate(monkeypatch, tmp_pa
         Desktop(version=AEDT_VERSION, non_graphical=NON_GRAPHICAL, new_desktop=True, port=port)
     from ansys.aedt.core.generic.general_methods import active_sessions
 
-    sessions = active_sessions()
-    for j, k in sessions.items():
-        os.kill(j, signal.SIGTERM)
+    sessions = active_sessions() or {}
```
SMoraisAnsys:
```suggestion
    sessions = active_sessions()
```
Mainly wondering why do we need to add `or {}` ?

tests/system/visualization/test_post_processing_field_report.py
```diff
@@ -299,14 +299,9 @@ def test_reports_by_category_fields(h3d_potter_horn) -> None:
 
 
 def test_reports_by_category_modal_solution(h3d_potter_horn) -> None:
-    variations2 = h3d_potter_horn.available_variations.nominal_variation(dependent_params=False)
```
SMoraisAnsys, line 302: Can you also explain those changes in your PR description ?

eblanco-ansys (dismissed):

tests/system/emit/test_emit.py
```diff
@@ -49,10 +49,14 @@
...
+# AEDT 2027R1 removes the EmitApiPython dependency which needed to be
+# specifically compiled for each supported Python version.
 # Prior to 2025R1, the Emit API supported Python 3.8,3.9,3.10,3.11
 # Starting with 2025R1, the Emit API supports Python 3.10,3.11,3.12
-if ((3, 8) <= sys.version_info[0:2] <= (3, 11) and DESKTOP_VERSION < "2025.1") or (
-    (3, 10) <= sys.version_info[0:2] <= (3, 12) and DESKTOP_VERSION > "2024.2"
+if (
+    DESKTOP_VERSION >= "2027.1"
+    or ((3, 8) <= sys.version_info[0:2] <= (3, 11) and DESKTOP_VERSION < "2025.1")
+    or ((3, 10) <= sys.version_info[0:2] <= (3, 12) and DESKTOP_VERSION > "2024.2")
```
SMoraisAnsys, line 59: Should we remove (in another PR for example) anything related to 24R2 ?

SMoraisAnsys (approved):

